PR #13, opened 2016-09-22
Mcol 290

dhall-InfiniDB:


dbcon/execplan/treenode.h
```diff
@@ -39,8 +39,7 @@
 #include "calpontsystemcatalog.h"
 #include "exceptclasses.h"
 #include "dataconvert.h"
-
-namespace messageqcpp {
+namespace messageqecpp {
```
LinuxJedi: Why the namespace change? Won't that break things?

PR #15, opened 2016-09-23
MCOL-97

LinuxJedi:
This branch does the following:

1. Removes boost_idb
2. Removes some unused binaries
3. Removes the built-in net-snmp
4. Searches for and uses the OS net-snmp (and errors if it can't find one)


LinuxJedi:
Closing this, it currently breaks snmpd

PR #30, opened 2016-10-03
MCOL-328 Remove extra '+' in float->string in func

LinuxJedi:
Functions such as reverse() that do a float/double to string conversion
use printf's %g to do it. Unfortunately this adds a '+' symbol before
the exponent symbol. MariaDB does not do this.

This patch just strips out that symbol.

### DIFF
--- a/utils/funcexp/functor_str.h
+++ b/utils/funcexp/functor_str.h
@@ -90,6 +90,7 @@ protected:
 		// Bug3788, use the shorter of fixed or scientific notation for floating point values.
 		// [ the default format in treenode.h is fixed-point notation ]
 		char buf[20];
+        char *dest, *src;
 		switch (fp->data()->resultType().colDataType)
 		{
 			case execplan::CalpontSystemCatalog::DOUBLE:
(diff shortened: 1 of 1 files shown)


utils/funcexp/functor_str.h
```diff
@@ -104,6 +105,18 @@ class Func_Str : public Func
 				return fp->data()->getStrVal(row, isNull);
 			break;
 		}
+        // Strip off the '+' that %g puts before the exponent symbol
+        dest = src = buf;
+        while(*src != '\0')
+        {
```
davidwbt: does it make sense to scan the whole string - couldn't you just check the first character and then adjust the pointer if it's plus?

dhall-InfiniDB: The problem is that the '+' is on the exponent and appears in the middle of the string.

davidwbt: Ah, ok missed that. So yes this is necessary.

LinuxJedi:
I'm going to try a different way first to see if I can build it using two sprintf calls instead. As long as it doesn't break anything I'll go with that method. Please don't merge this yet.

LinuxJedi:
Other way is better, abandoning this for a new branch coming shortly

PR #37, opened 2016-10-10
MCOL-98

benthompson15:


LinuxJedi:
With func_div.cpp did you meant to do more than just NULL returns? If not then I don't think the extra includes are needed.

PR #60, opened 2016-11-17
remove generated file releasenum and add it to gitignore file

benthompson15:


LinuxJedi:
Exactly as discussed. Tested locally and works fine.

PR #67, opened 2016-12-06
MCOL-406 Add schema to table_usage()

LinuxJedi:
Schema name should be included in results


greenlion:
Maybe the command line should be table_usage(schema, table) or the WHERE clause on the I_S should include:
... AND (table_schema = DATABASE() or DATABASE() IS NULL) ...

LinuxJedi:
I agree and have discussed it internally. We will make a change along those lines shortly.

PR #96, opened 2017-01-14
MCOL-505 Performance improvements to ExeMgr

LinuxJedi:
This fix improves the performance of ExeMgr by doing the following:

* Significantly reduces the amount of time the xml configuration is
scanned
* Uses a much faster way to determine the CPU core count
* Reduces the amount of times certain allocations are executed
* Rowgroup pre-allocates vectors for 1024 rows

This improves performance for the first query of a connection and the
performance for smaller result sets. It may well improve performance in
other areas too.


LinuxJedi:
Merging as already approved with 'develop' branch

PR #118, opened 2017-02-08
MCOL-533 Increase max file limit

LinuxJedi:
It is possible for a query with many extents to exceed the original
65536.


LinuxJedi:
Closing for now as I don't think this is really needed

PR #119, opened 2017-02-08
MCOL-533 Increase max file limit

LinuxJedi:
It is possible for a query with many extents to exceed the original
65536.

### DIFF
--- a/exemgr/main.cpp
+++ b/exemgr/main.cpp
@@ -1256,7 +1256,7 @@ int setupResources()
         return -1;
     }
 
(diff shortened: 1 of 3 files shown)


dleeyh:
While large table may have many extents, Each PrimProc thread would still process them one batch at a time.  Would each thread destroy the FD after processing?

LinuxJedi:
yes, but you could have enough threads running in parallel that all of the extents are in use (and therefore open) at any one time.

LinuxJedi:
Closing for now as I don't think this is really needed

PR #121, opened 2017-02-11
Improve Debian package versioning

jbfavre:
Similar to https://github.com/mariadb-corporation/mariadb-columnstore-server/pull/32
Might be added to https://jira.mariadb.org/browse/MCOL-565

Same warning about CI compatibility


LinuxJedi:
Added to MCOL-565 for Ben to review

LinuxJedi:
We are completely refactoring the way this code is compiled and built. We may need to make similar changes, but it would be done very differently.

PR #123, opened 2017-02-13
Make packages management optional

jbfavre:
Currently postConfigure requires deb or rpm packages to be copied on disk in /root
It also connect to each nodes using ssh to uninstall packages, push its own ones and install its own through expect scripts.
This makes MariaDB Columnstore very unfriendly with configuration management systems, and it's even worse with containers.

This patch aims to:
- add a new command line option to postConfigure: `-m` to tell postConfigure *not* to manage packages (default behaviour is still to manage them)
- pass a new option to expect scripts to reflect package management choice
- update expect scripts to respect this option

I have still to implement the same behaviour in one place at least: `processmanager.cpp` which check for package in `ProcessManager::addModule`.
I'll have a look at it asap.

Cheers

### DIFF
--- a/oam/install_scripts/binary_installer.sh
+++ b/oam/install_scripts/binary_installer.sh
@@ -37,6 +37,7 @@ set UNM [lindex $argv 10]
 if { $UNM != "" } {
 	set USERNAME $UNM
 }
+set DEPLOY_PACKAGES [lindex $argv 11]
 
 set BASH "/bin/bash "
 if { $DEBUG == "1" } {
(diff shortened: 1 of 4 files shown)


LinuxJedi:
This covers part of MCOL-522 (https://jira.mariadb.org/browse/MCOL-522) which David Hill is working on, it is probably best that he reviews this.

davidhilldallas:
You will also want to double check the removeModule logic in ProcMon, it will remove packages also.

PR #135, opened 2017-03-08
MCOL-613 - add check for pm1 upgrade

davidhilldallas:


davidhilldallas:
updates made to check for all Ip addesses, please review

PR #145, opened 2017-04-06
when setting systemName also modify my.cnf to update server_audit_sys…

benthompson15:
…log_info value.


davidhilldallas:
reviewed, need to check the JIRA to see if there is good enough info for Daniel to test with

PR #162, opened 2017-05-03
MCOL-689 Remove 64KB VARCHAR response limit

LinuxJedi:
This was a hard coded limit due to StringStore not being able to handle
more than this. It restricts hex() unnecessarily and is now redundant.


LinuxJedi:
Wrong MCOL

PR #257, opened 2017-09-15
MCOL-859 DDL System catalog mutex deadlock

LinuxJedi:
A race between a DDL change and a system catalog version cache update
can cause a deadlock in DDLProc. This change makes the mutex recursive.


LinuxJedi:
Not to be merged until after 1.1.0

PR #355, opened 2017-12-20
MCOL-497 securing mysql connection with TLS in crossengine and querystats code.

drrtuy:
Please consider the PR.


LinuxJedi:
Great work! One very small additional commit please:
Can you edit oam/etc/Columnstore.xml and add empty XML nodes for the additional items in the CrossEngineSupport section? The user then has a place to fill in as needed that we can document (as with the other settings).

PR #428, opened 2018-03-15
MCOL-1222 - improved the waitforsystemactive function

davidhilldallas:


dhall-MariaDB:
Using a 10 second delay in a loop can cause an inordinate wait for active. The previous delay of 3 seconds was chosen as a compromise between hitting it too hard and having the user wait too long.

PR #440, opened 2018-04-09
MCOL-1182 Fix cross engine filters

LinuxJedi:
Don't delete a join filter that hasn't actually been used yet. It
appears that tree pruning is happening in the wrong place.


LinuxJedi:
Doesn't fully work yet

PR #493, opened 2018-06-07
Mcol 1201 b

dhall-MariaDB:


dbcon/joblist/joblistfactory.cpp
```diff
@@ -891,21 +896,55 @@ const JobStepVector doAggProject(const CalpontSelectExecutionPlan* csep, JobInfo
 
             continue;
         }
-
+#if 0
+        // MCOL-1201 Add support for multi-parameter UDAnF
```
LinuxJedi: Is this code intentionally disabled?

PR #495, opened 2018-06-14
MCOL-1474 Add error handling to PriorityThreadPool

LinuxJedi:
When an error occurs PriorityThreadPool could either cause PrimProc to crash or just completely ignore the problem and lose the job. This patch adds better error handling.


LinuxJedi:
Please do not merge until after 1.1.5 has been released

PR #502, opened 2018-06-21
MCOL-1456 Restore the pre-1.2 processing behavior for queries in autoswitch mode

drrtuy:


dhall-MariaDB:
This fixes the mode two switch over, but it sill doesn't work in mode 0.

PR #554, opened 2018-09-07
MCOL-1694 & MCOL-1505 Improved exception handling

LinuxJedi:
This patch catches exceptions in DDLProc, DMLProc and ExeMgr which could
potentially happen during startup. Logging them instead of silently
ignoring them (or crashing in ExeMgr).


LinuxJedi:
Please do not merge until 1.1.6 is released

PR #576, opened 2018-09-27
MCOL-1750 Stop threadpool memory leak

LinuxJedi:
Threadpool could leak thread stack memory if the thread returns without
being joined. This is the case when the 10 minute cleanup timer is
triggered. We now detach threadpool threads so they clean themselves up.


LinuxJedi:
Closed because it is causing cpimport to crash with a mutex issue

PR #584, opened 2018-10-07
MCOL-266 Add support for BOOLEAN/BOOL data type

tntnatbry:
Since BOOLEAN is simply an alias for TINYINT(1), only the parser files are updated to add support for it in ColumnStore.

Am I missing something here?

### DIFF
--- a/dbcon/ddlpackage/ddl.l
+++ b/dbcon/ddlpackage/ddl.l
@@ -184,6 +184,8 @@ TINYTEXT {return TINYTEXT;}
 TEXT {return TEXT;}
 MEDIUMTEXT {return MEDIUMTEXT;}
 LONGTEXT {return LONGTEXT;}
+BOOL {return BOOL;}
+BOOLEAN {return BOOLEAN;}
 
 \n { lineno++;}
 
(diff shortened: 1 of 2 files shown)


drrtuy:
Firstly many thanks for this feature pull request. We appreciate you contribution.
From my perspective and [this doc](https://mariadb.com/kb/en/library/boolean/)it looks OK. So we will test it with DQL, DML and come back with the results.

LinuxJedi:
Unfortunately yes you are. This does not add any kind of limit so it would be possible for an INSERT/UPDATE query to insert > 1 using functions, MariaDB server won't catch this since they are executed in the engine. In addition cpimport and mcsapi would also need to be made aware of BOOL/BOOLEAN and how to handle the words "true"/"false" and limits.

To quote Futurama:

Fry: Bender, what is it?
Bender: Whoa, what an awful dream. Ones and zeros everywhere. And I thought I saw a two.
Fry: It was just a dream, Bender. There's no such thing as two.

LinuxJedi:
OK, ignore my previous comment, mcsapi and cpimport support is the only thing missing since MariaDB itself doesn't cap these like I thought it did.

davidwbt:
Hi, @tntnatbry, can you confirm that you are contributing this whole pull request, including one or
several files that are either new files or modified ones, under the BSD-new license. https://en.wikipedia.org/wiki/BSD_licenses.  We will provide you credit for the fix in our release notes but this protects both you and mariadb in terms of any possible licensing issues.

tntnatbry:
Hi @davidwbt, I am submitting this patch under the BSD-new license.

LinuxJedi:
Buildbot failures were due to another problem which has now been resolved.

PR #616, opened 2018-11-08
Merge develop-1.1 into develop

LinuxJedi:


utils/funcexp/functor_str.h
```diff
@@ -120,7 +120,8 @@ class Func_Str : public Func
                 break;
 
             default:
-                return fp->data()->getStrVal(row, isNull);
+				fFloatStr = fp->data()->getStrVal(row, isNull);
+                return;
                 break;
```
pleblanc1976: can remove the break

utils/funcexp/func_concat_ws.cpp
```diff
@@ -94,10 +94,11 @@ string Func_concat_ws::getStrVal(Row& row,
 
 #else
     string str;
-
+    string tmp;
     for ( uint32_t i = 1 ; i < parm.size() ; i++)
     {
-        str += string(stringValue(parm[i], row, isNull).c_str());
+		string(stringValue(parm[i], row, isNull).c_str(), tmp);
```
pleblanc1976: What was there before looks odd, turns a string to a c-string, then copies it.  In the new version since stringValue() returns nothing, can remove the string ctor around it.

utils/funcexp/functor_str.h
LinuxJedi: That generally isn't a good idea. If a later date we remove the return (may not be likely, but hypothetical) we may missing adding the break back. Better to let the compiler realise that.

utils/funcexp/func_concat_ws.cpp
LinuxJedi: That is a good point, but since this is a merge from 1.1 to develop we should probably fix it there. I'll file a bug. In the mean time this isn't doing any harm, just a pain.

PR #618, opened 2018-11-12
MCOL-1855 RENAME TABLE now supports tablenames with non-alphanumerical chars.

drrtuy:
RENAME doesn`t work with tables whose names contains special or non-ASCII chars. 
The commit alters ha_calpont_impl_rename_table_() to decode paths received from the server.

### DIFF
--- a/dbcon/mysql/ha_calpont_ddl.cpp
+++ b/dbcon/mysql/ha_calpont_ddl.cpp
@@ -1887,6 +1887,7 @@ int ProcessDDLStatement(string& ddlStatement, string& schema, const string& tabl
     return rc;
 }
 
+// Fails with `/` sign in table name
 pair<string, string> parseTableName(const string& tn)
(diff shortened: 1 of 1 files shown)


dbcon/mysql/ha_calpont_ddl.cpp
```diff
@@ -2342,6 +2343,58 @@ int ha_calpont_impl_delete_table_(const char* db, const char* name, cal_connecti
...
+  @brief
+  Find and return a pointer to the last slash in the name.
+
+  @details
+  This f() finds and returns position of the last slash sign found in
+  the path or NULL.
+
+  Called from ha_calpont_ddl.cpp by decode_table_name().
+*/
+const char* last_slash_pos(const char *name, size_t name_len)
```
LinuxJedi: could you not just use strrchr()?

PR #619, opened 2018-11-13
MCOL-1885 Fix libmysql error handling

LinuxJedi:
If an error occurs in mysql_real_query() it fell through and tried to
get a result set anyway. This caused the MariaDB error message to be
forgotten and a generic one returned instead.

This fix returns the query error message and also uses the int error
code for generic errors instead of casting it to an unsigned int.

### DIFF
--- a/utils/libmysql_client/libmysql_client.cpp
+++ b/utils/libmysql_client/libmysql_client.cpp
@@ -104,6 +104,7 @@ int LibMySQL::run(const char* query)
     {
         fErrStr = "fatal error runing mysql_real_query() in libmysql_client lib";
         ret = -1;
+        return ret;
     }
 
     fRes = mysql_use_result(fCon);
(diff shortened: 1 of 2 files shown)


LinuxJedi:
DO NOT MERGE UNTIL 1.2.1 IS RELEASED

utils/libmysql_client/libmysql_client.cpp
```diff
@@ -104,6 +104,7 @@ int LibMySQL::run(const char* query)
     {
         fErrStr = "fatal error runing mysql_real_query() in libmysql_client lib";
         ret = -1;
+        return ret;
```
pleblanc1976: I don't know this code.  This has the side effect of not setting fRes; it will be left with its old value.  If that is important, then this will need to be fixed.  If it's not important, then this will work.  I'll leave it to your judgement.

LinuxJedi: Yes, that is fine, the only possibility in that case is fRes = NULL unless we are using the class incorrectly. It is only designed to make a connection, execute a single query and then end. The destructor cleans up. If run returns a -1 then fRes won't be used because an error occurred. Everything is aborted. Exceptions would have probably been better here but this is basically a merge of two sections of code that were doing the same thing.
At a later date we should maybe consider connection pooling here for performance.

pleblanc1976: Looks good to me then, will merge once 1.2.1 is finished.

PR #627, opened 2018-11-16
MCOL-1875 - fixed issue with calulating module cpu usage

davidhilldallas:


oamapps/serverMonitor/cpuMonitor.cpp
```diff
@@ -556,7 +556,7 @@ void ServerMonitor::getCPUdata()
 
     string tmpsystemCpu = tmpDir + "/processCpu";
     
-    cmd = "top -b -n 6 -d 1 | awk '{print $5}' | grep %id  > " + tmpsystemCpu;
+    cmd = "top -b -n 6 -d 1 | grep '%Cpu' | awk '{print $8}'  > " + tmpsystemCpu;
```
LinuxJedi: Instead of an executable calling an external tool like this (which will not work for everyone) why not read /proc/stat twice and do a calculation on the jiffies?

PR #651, opened 2018-12-06
MCOL-1347 ALTER TABLE ADD COLUMN creates a column with width 4 for a varchar(1) column. 

drrtuy:


LinuxJedi (changes requested):
This appears to fix the case for varchar(1). But if, for example, varchar(2) is used then we end up with an 8 byte column.

PR #656, opened 2018-12-11
MCOL-1676 Window Functions with FLOAT type gives incorrect answer

davidhall586:


pleblanc1976 (approved):
I think this is OK, but this class could sure use better variable names and comments.  ie: this is the in type, this is the out type, etc.

PR #658, opened 2018-12-17
MCOL-1662 Use version buffer for UPDATEs

LinuxJedi:
For some reason version buffer is not turned up for update queries which
means that the version number for dictionary blocks is not changed.
This can lead to dirty cache reads resulting in _CpNoTf_ in the results.

This patch turns on version buffer for updates.

### DIFF
--- a/writeengine/server/we_dmlcommandproc.cpp
+++ b/writeengine/server/we_dmlcommandproc.cpp
@@ -2515,6 +2515,7 @@ uint8_t WE_DMLCommandProc::processUpdate(messageqcpp::ByteStream& bs,
 	boost::scoped_array<int> preBlkNums(new int[columnsUpdated.size()]);
 	boost::scoped_array<OID> oids(new OID[columnsUpdated.size()]);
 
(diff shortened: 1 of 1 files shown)


writeengine/server/we_dmlcommandproc.cpp
```diff
@@ -2515,6 +2515,7 @@ uint8_t WE_DMLCommandProc::processUpdate(messageqcpp::ByteStream& bs,
 	boost::scoped_array<int> preBlkNums(new int[columnsUpdated.size()]);
 	boost::scoped_array<OID> oids(new OID[columnsUpdated.size()]);
 
+    BRMWrapper::setUseVb(true);
```
pleblanc1976, line 2518: I'm not familiar with the changes between 1.0 and 1.1.  If I understand the problem though, doesn't this need to be added to the various insert and delete fcns as well?  Those need to version the blocks being changed as well don't they?

pleblanc1976 (approved):

PR #660, opened 2018-12-20
MCOL-498 Optionally disable disk space preallocation. 

drrtuy:
Plz look into JIRA's issue for details.


utils/idbdatafile/CMakeLists.txt
```diff
@@ -16,7 +16,7 @@ set(idbdatafile_LIB_SRCS
 
 add_library(idbdatafile SHARED ${idbdatafile_LIB_SRCS})
 
-target_link_libraries(idbdatafile ${NETSNMP_LIBRARIES})
+target_link_libraries(idbdatafile ${NETSNMP_LIBRARIES} ${ENGINE_OAM_LIBS})
```
pleblanc1976: note to self.  I don't see a change yet that requires the new dependency.

utils/idbdatafile/IDBPolicy.cpp
```diff
@@ -224,9 +228,73 @@ void IDBPolicy::configIDBPolicy()
...
+    {
+        pt++;
+        continue;
+    }
+    
+    // CS could return here b/c it initialised this singleton and
+    // there is no DBRootX sections in XML.
+    if (pt == moduledbrootlist.end())
+    {
+        return;
```
pleblanc1976, line 263: Would this indicate no dbroots are assigned to this PM?

writeengine/shared/we_fileop.cpp
```diff
@@ -1096,69 +1100,100 @@ int FileOp::initColumnExtent(
             Stats::stopParseEvent(WE_STATS_WAIT_TO_EXPAND_COL_EXTENT);
         else
             Stats::stopParseEvent(WE_STATS_WAIT_TO_CREATE_COL_EXTENT);
+#endif
+        // MCOL-498 Skip the huge preallocations if the option is set
+        // for the dbroot
+        if ( bOptExtension )
+        {
+            bOptExtension = (idbdatafile::IDBPolicy::PreallocSpace(dbRoot))
```
pleblanc1976, line 1115: since bOptExtension is true here, this reduces to bOptExtension = ..PreallocSpace(dbRoot)

writeengine/bulk/we_colbuf.cpp
```diff
@@ -101,15 +102,33 @@ void ColumnBuffer::resizeAndCopy(int newSize, int startOffset, int endOffset)
...
 #endif
-    size_t nitems = fFile->write(fBuffer + startOffset, writeSize) / writeSize;
+    size_t nitems;
+    if ( fillUpWNulls )
+        nitems = fFile->write(newBuf, BYTE_PER_BLOCK) / BYTE_PER_BLOCK;
+    else
+        nitems = fFile->write(fBuffer + startOffset, writeSize) / writeSize;
 
     if (nitems != 1)
         return ERR_FILE_WRITE;
```
pleblanc1976, line 137: need to dealloc newBuf before returning

writeengine/bulk/we_colbuf.cpp
```diff
@@ -101,15 +102,33 @@ void ColumnBuffer::resizeAndCopy(int newSize, int startOffset, int endOffset)
 //------------------------------------------------------------------------------
 // Write data stored up in the output buffer to the segment column file.
 //------------------------------------------------------------------------------
-int ColumnBuffer::writeToFile(int startOffset, int writeSize)
+int ColumnBuffer::writeToFile(int startOffset, int writeSize, bool fillUpWNulls)
```
pleblanc1976: Optional.  fillUpWNulls should probably be called fillUpWEmpty instead.

writeengine/bulk/we_colbufcompressed.cpp
```diff
@@ -242,12 +247,14 @@ int ColumnBufferCompressed::writeToFile(int startOffset, int writeSize)
                 "; part-"     << fColInfo->curCol.dataFile.fPartition <<
                 "; seg-"      << fColInfo->curCol.dataFile.fSegment   <<
                 "; addBytes-" << writeSize <<
+                "; extraBytes-" << fillUpWNullsWriteSize <<
                 "; totBytes-" << (fNumBytes + writeSize);
             fLog->logMsg( oss.str(), MSGLVL_INFO2 );
         }
 
         memcpy(bufOffset, (fBuffer + startOffset), writeSize);
```
pleblanc1976: I'm a little unclear how this is supposed to work.  It might make more sense later.  With my current understanding, the empty values should be copied to bufOffset here...?

pleblanc1976: A little more reading makes me think the buffer is initialized with empty values, so that would be unnecessary.  Please confirm that behavior

writeengine/wrapper/we_colop.cpp
```diff
@@ -1533,12 +1542,16 @@ int ColumnOp::writeRow(Column& curCol, uint64_t totalRow, const RID* rowIdArray,
     char     charTmpBuf[8];
     uint64_t  emptyVal;
     int rc = NO_ERROR;
+    bool fillUpWEmptyVals = false;
+    bool fistRowInBlock = false;
```
pleblanc1976: a typo worth fixing

pleblanc1976 (changes requested):
Some of this was pretty hard to understand without reading through all of the writeengine code.  The harder it is to read the code, the more we have to rely on testing.  A couple comments...
1) Replace all of the variable names & comments that suggest we're padding with NULL values rather than empty values.
2) Add a component test or two.  Ex: a script to import a column, big & small, with preallocation on and off, then verify the expected result from dumpcol or coldump (whatever it's called).

utils/idbdatafile/IDBPolicy.cpp
drrtuy: Yes. It would. Is this relevant in this case?

writeengine/bulk/we_colbuf.cpp
drrtuy: Strongly agree on that.

writeengine/bulk/we_colbuf.cpp
drrtuy: Good catch.

writeengine/wrapper/we_colop.cpp
drrtuy: Fixed.

writeengine/shared/we_fileop.cpp
drrtuy: bOptExtension is set when CS increases an extent only and not for abbreviated extent so I prefered conditional jump to a function call.

writeengine/wrapper/we_colop.cpp
pleblanc1976: Sorry, I meant 'fistRowInBlock' -> 'firstRowInBlock'.  :)

utils/idbdatafile/IDBPolicy.cpp
pleblanc1976: No, it's just for my own understanding of the code.

writeengine/wrapper/we_colop.cpp
drrtuy: I didn't miss the typo for the second time :)

writeengine/bulk/we_colbufcompressed.cpp
drrtuy: It is not only about the buffer initialization but about how many bytes was written.
This additional bytes must make singular INSERTs happy but I am looking into it in details now.

drrtuy:
As you suggested I will add a test for this and comment most of the changes to make it more readable.

writeengine/bulk/we_colbufcompressed.cpp
drrtuy: This change is unnecessary indeed. Nice catch.

pleblanc1976: Groovy

PR #661, opened 2018-12-20
MCOL-1726 Make cleartablelock remove txn from BRM

LinuxJedi:
With an API txn (and likely other scenarios) cleartablelock will roll
back the txn and lock but leave a stale txn pointer in the BRM. This
causes restart/shutdown to warn that transactions are still open.

This patch makes sure BRM removes the txn from the list.


pleblanc1976 (approved):
Was a brm rollback never happening before this??

This acts as a catch-all for cases where we're not doing that for whatever reason, and it looks OK.

PR #662, opened 2018-12-20
Mcol 1607 - postconfig must support writing hostnames to config file

pleblanc1976:
Took a whack at this.  After some discussion in the ticket, we believe the scope for dyn dns support should be at the initial cluster config (the postconfig run).  I also added support in the addModule command in mcsadmin as the boundary was unclear at the time.


pleblanc1976:
Will rebase on develop-1.2 instead of develop.

PR #669, opened 2018-12-31
MCOL-270 Add support for MEDIUMINT data type

tntnatbry:
This PR adds support for the MEDIUMINT data type in ColumnStore.

Since it is stored as an int32, the allowed range for signed and unsigned MEDIUMINT types is [-8388608, 8388607] and [0, 16777215] respectively (unlike the other numeric data types where the supported range is 2 less).

Existing values of the signed and unsigned INT data type are used for the null and empty row representation for MEDIUMINT.

Support already existed in the function expression, udfsdk, and window function modules.

I am submitting this patch under the BSD-new license.


drrtuy:
We appreciate the contribution and will proceed with testing.
Happy New Year!

LinuxJedi (approved):
Nice work! You appear to have all the bases covered I can think of. Only minor concern is the possibility of an aggregation overflow which we will need to test.

Many thanks for your contribution

PR #677, opened 2019-01-16
MCOL-2009 MCOL-1702 : join() inside of joblist::abort caused threads …

benthompson15:
…to be blocked on an abort


pleblanc1976 (approved):

pleblanc1976:
Will merge once buildbot is happy.

PR #680, opened 2019-01-23
MCOL-2001 Make the parameters for redistribute case insensitive

davidhall586:


drrtuy:
Despite the fact that mcsadmin doesn't call for speed or performance I would avoid using iequals b/c it creates complex objects underneath, e.g. locale.

davidhall586:
Changed the iequals() to:
#if _MSC_VER
			if (_strnicmp(arguments[1].c_str(), "start", 5) == 0))
#else
			if (strncasecmp(arguments[1].c_str(), "start", 5) == 0)
#endif
etc.
We could spend a bit more time and make it work with short versions, such as "sta" and "sto", but, this will do.

benthompson15 (approved):

PR #687, opened 2019-02-08
MCOL-2149 Fix cpimport decimal saturation

LinuxJedi:
If we saturate int64_t during string -> decimal conversion then end
processing there instead of continuing. This preserves a good saturation
value.


pleblanc1976 (approved):
Looks good, can't believe I had it not returning on error!

PR #704, opened 2019-03-05
Mcol 1822 c

davidhall586:


utils/funcexp/func_substring_index.cpp
```diff
@@ -61,7 +62,7 @@ std::string Func_substring_index::getStrVal(rowgroup::Row& row,
     if (isNull)
         return "";
 
-    int64_t count = fp[2]->data()->getIntVal(row, isNull);
+    size_t count = fp[2]->data()->getIntVal(row, isNull);
```
LinuxJedi, line 65: Is this not a signed to unsigned conversion? Not sure if this would be a problem, just pointing it out.

utils/funcexp/func_round.cpp
```diff
@@ -200,6 +203,67 @@ double Func_round::getDoubleVal(Row& row,
...
+    if (x.scale > 0)
+    {
+        while (x.scale-- > 0)
+            d /= 10.0;
+    }
+    else
+    {
+        while (x.scale++ < 0)
+            d *= 10.0;
+    }
```
LinuxJedi, line 263: I would have thought using exp10() (or similar) would be more optimal to the above.

utils/dataconvert/dataconvert.cpp
```diff
@@ -871,7 +872,7 @@ bool mysql_str_to_datetime( const string& input, DateTime& output, bool& isDate
 
 bool mysql_str_to_time( const string& input, Time& output, long decimals )
 {
-    int32_t datesepct = 0;
+//    int32_t datesepct = 0;
```
LinuxJedi, line 875: There is no need to comment this out, could just be deleted.

LinuxJedi (approved):
Approved with comments

utils/funcexp/func_substring_index.cpp
davidhall586: Upon further reflection, this is probably not a good idea. We should revert to int64_t.

utils/funcexp/func_round.cpp
davidhall586: Probably so. This code was already there in the copied code and I didn't see the improvement possibility.

PR #705, opened 2019-03-06
MCOL-2160 MCOL-1883 RENAME now supports both '/' symbols in table names and crash described in MCOL-1883 has been fixed.

drrtuy:
MCOL-2160 MCOL-1883 RENAME now supports both '/' symbols in table 
    names and doesn't crash on multi-RENAME statements when 
    non-alphanumeric chars are presented in table names.

    Replaced INFINIDB_DEBUG with MCS_DEBUG in ha_calpont_ddl.cpp

    Removed unused debug outputs.


pleblanc1976 (approved):
This will let '/' be part of a tablename.

PR #714, opened 2019-03-13
MCOL-1883 This commit fixes RENAME behavior that doesn't work properly with table     names that contains '/' sign in the middle.

drrtuy:
MCOL-1883 This commit fixes RENAME behavior that doesn't work properly with table     names that contains '/' sign in the middle. Given the name `some/table/` in to or  from. RENAME uses a prefix to the left of the first '/', namely `some` and doesn't  consider the trailer `/table/`.

### DIFF
--- a/dbcon/mysql/ha_calpont_ddl.cpp
+++ b/dbcon/mysql/ha_calpont_ddl.cpp
@@ -2373,33 +2373,36 @@ void decode_table_name(char *buf, const char *path, size_t pathLen)
 
   @details
   Parses the path to extract both database
-  and table names, e.g path ./test/d$ produces
-  a pair of strings 'test' and 'd$'. This f() looks for a '/'
-  token only twice to allow '/' symbol in table names.
+  and table names. This f() assumes the path format
+  ./test/d$
+  and f() produces a pair of strings 'test' and 'd$'.
+  This f() looks for a '/' symbols only twice to allow '/'
+  symbol in table names. The f() supports international
+  glyphs in db or table names.
 
   Called from ha_calpont_ddl.cpp by ha_calpont_impl_rename_table_().
 */
-pair<string, string> parseTableName(const string& tn)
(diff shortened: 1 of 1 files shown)


dbcon/mysql/ha_calpont_ddl.cpp
```diff
@@ -2373,33 +2373,36 @@ void decode_table_name(char *buf, const char *path, size_t pathLen)
...
+    db_pnt = path_cursor;
+    while (*path_cursor != '/')
+    {
+        path_cursor++;
+    }
+    path_cursor++;
+    tbl_pnt = path_cursor;
+    db.assign(db_pnt, tbl_pnt - 1 - db_pnt);
+    tb.assign(tbl_pnt, path + strlen(path) - tbl_pnt);
+
```
pleblanc1976, line 2403: This looks ok given well-defined input, but I'm not sure why it was necessary.  Is there an implicit conversion in the boost tokenizer, which is mangling the international symbols?

pleblanc1976: Another question I guess, since the loops that scan for '/' aren't doing any interpretation, can we say for sure that value doesn't appear in any uft8* chars?  (Not something I know about)

drrtuy: Tokenizer returns the list of the tokens that are b/w the delimiter. So I have to manually glue tokens with '/' signs. This adds a complexity adding extra conditions into the function.

drrtuy: The '/' itself isn't used in international symbols encoding. The symbol's code looks like @00aa and even '@' sign is encoded.

pleblanc1976: ok, as long as the / ascii value isn't ever in those encodings then this will work.

pleblanc1976 (approved):

PR #721, opened 2019-03-26
MCOL-498 The knob to disable disk space preallocation for segment files.

drrtuy:


pleblanc1976 (approved):
Reviving ancient unit tests, ambitious!  The new test looks good.  Verifies the expected file size with and without the new feature for both numeric and dict extents.  Between that and the regression test we have good coverage.

pleblanc1976 (changes requested):
Several of the buildbot minions failed for whatever reason.  Need to look into those first...

sneJ-:
> Several of the buildbot minions failed for whatever reason. Need to look into those first...

The tests fail because MCOL-1995 hasn't been upmerged from develop-1.2 to develop in mariadb-columnstore-api. Nothing to worry about on you side.

PR #730, opened 2019-04-10
Mcol 1559

davidhall586:


LinuxJedi (approved):
This would be easier if we stored length with string data throughout the code. Many thanks for this!

PR #745, opened 2019-04-30
MCOL-2051: CS uses 'NULL' literal as NULL when UPDATEs varchar field

ssedov:
Added proxying of NULL values.
Note: empty strings are still considered NULLs later on.


LinuxJedi:
Builldbot is currently under maintenance, hence the failures.

drrtuy (approved):

drrtuy:
Tested the failed builds manually.

PR #746, opened 2019-05-01
unneeded  k for loop

gscteam:
Hotsport identified in perf top, It seems that the k-loop is not needed there.


drrtuy:
Agree, this loop could be avoided. Could you post license-mantra?

gscteam:
Submitted  under the BSD license.

LinuxJedi (approved):

gscteam:
There is one condition that might be missing, if  columnsUpdated.size() is 0, then nothing get done. It is unclear for me when would that be the case, since the j loop for (unsigned int j = 0; j < columnsUpdated.size(); j++) is also based on that. So if that size is 0 it would not even reach that far.

drrtuy:
Thanks for commenting this. We will look into this.

чт, 9 мая 2019 г., 1:15 gscteam <notifications@github.com>:

PR #751, opened 2019-05-07
MCOL-537 Address existing compiler warnings.

drrtuy:


ddlproc/ddlproc.cpp
```diff
@@ -103,7 +104,10 @@ int main(int argc, char* argv[])
     // This is unset due to the way we start it
     program_invocation_short_name = const_cast<char*>("DDLProc");
 
-    setupCwd();
+    if ( setupCwd() < 0 )
+    {
+        std::cerr << "Could not set working directory" << std::endl;
```
pleblanc1976: Since we're handling the error now, and it's likely important that this succeed, this should exit after failure, IMO.

oamapps/alarmmanager/alarmmanager.cpp
```diff
@@ -484,11 +484,10 @@ void ALARMManager::sendAlarmReport (const char* componentID, int alarmID, int st
...
         processName = repProcessName;
 
-    int returnStatus = API_SUCCESS;           //default
-    ByteStream msg1;
+ByteStream msg1;
 
-    // setup message
-    msg1 << (ByteStream::byte) alarmID;
+// setup message
+msg1 << (ByteStream::byte) alarmID;
```
pleblanc1976: looks like we lost indentation here

primitives/linux-port/column.cpp
```diff
@@ -1421,7 +1421,8 @@ namespace primitives
 void PrimitiveProcessor::p_Col(NewColRequestHeader* in, NewColResultHeader* out,
                                unsigned outSize, unsigned* written)
 {
-    memcpy(out, in, sizeof(ISMPacketHeader) + sizeof(PrimitiveHeader));
```
pleblanc1976, line 1424: I'm surprised casting is necessary in all of these memcpy calls.  Do you know which version of gcc is complaining about this?

tools/dbloadxml/colxml.cpp
```diff
@@ -46,7 +46,10 @@ int main(int argc, char** argv)
 #ifdef _MSC_VER
     //FIXME
 #else
+#pragma GCC diagnostic ignored "-Wunused-result"
     setuid( 0 ); // set effective ID to root; ignore return status
+    // Why should we raise privileges if we don't care?
```
pleblanc1976: My guess is that failure there isn't always a real problem.  Failures later would tell the user whether it was a real problem or not.  But also this must be ancient code written under different specs.

pleblanc1976 (approved):

pleblanc1976 (commented):
It's probably fine, but we should make sure it still compiles on our supported platforms before merging.

pleblanc1976:
In Centos 6, we're getting this error:  "error: #pragma GCC diagnostic not allowed inside functions"
so, would need to wrap whole functions apparently, or test for compiler version and uglify the code.

Centos 7 and Deb 8 apparently don't like the format-truncation suppressions, may be a syntax thing though:  "warning: unknown option after ‘#pragma GCC diagnostic’ kind [-Wpragmas]
 #pragma GCC diagnostic ignored "-Wformat-truncation="

Yeah, we'll need to iterate on this a little to get it to compile.

pleblanc1976 (changes requested):
Need to make compilers happy.

primitives/linux-port/column.cpp
drrtuy: All starting from gcc 7.

tools/dbloadxml/colxml.cpp
drrtuy: I agree with that guess. I will add extra message that notifies about the failure to simplify troubleshooting.

primitives/linux-port/column.cpp
drrtuy: pragma is another option how to disable warnings. But I prefer to have less pragmas disables the warnings throughout the code.

utils/threadpool/threadpool.h
```diff
@@ -75,7 +77,7 @@ class ThreadPoolGroup
     boost::thread* create_thread(F threadfunc)
     {
         boost::lock_guard<boost::shared_mutex> guard(m);
-        std::auto_ptr<boost::thread> new_thread(new boost::thread(threadfunc));
+        std::unique_ptr<boost::thread> new_thread(new boost::thread(threadfunc));
```
drrtuy, line 81: This could potentially break SUSE12 build.

drrtuy: As I expected there is no std::unique_ptr in centos 6/suse 12 stdlib. I think I have to put preprocessor conditional expression to check for GCC version higher then 5.

pleblanc1976:
Odd, looks like BB only ran the windows check.  Will check with Ben.

pleblanc1976:
Ben says there are a lot of build jobs queued atm, and the ones for this PR are scheduled.  Will wait for results.

utils/threadpool/threadpool.h
drrtuy: Couldn't do much about the warning produced in U16. g++ uses 97 standard by default but std::unique_ptr is supported since c++11.

pleblanc1976: the Boost smart ptrs will work.  std::unique_ptr == boost::scoped_ptr

pleblanc1976 (approved):

pleblanc1976:
BB is still failing in U16 & C7, but it looks like it has nothing to do with this.

I'll leave it to you to decide whether to switch ptr types to eliminate the warning in U16.

PR #756, opened 2019-05-10
Mcol 3296

davidhall586:


dmlproc/dmlproc.cpp
```diff
@@ -551,16 +553,16 @@ int main(int argc, char* argv[])
...
 	temp = toInt(cf->getConfig(DMLProc, "ServerThreads"));
 	if (temp > 0)
 			serverThreads = temp;
 
-	temp = toInt(cf->getConfig(DMLProc, "ServerQueueSize"));
-	if (temp > 0)
-			serverQueueSize = temp;
+//	temp = toInt(cf->getConfig(DMLProc, "ServerQueueSize"));
+//	if (temp > 0)
+//			serverQueueSize = temp;
```
drrtuy, line 565: Could we just remove but comment this?

dmlproc/dmlprocessor.cpp
```diff
@@ -374,9 +374,27 @@ int PackageHandler::synchTableAccess()
...
+            LoggingID logid(21, fSessionID, fTxnid);
+            logging::Message::Args args1;
+            logging::Message msg(1);
+            ostringstream oss;
+            oss << "Txn is waiting for" << tableOidQueue.front() << " " << dmlPackage->get_SQLStatement() << "; |" << dmlPackage->get_SchemaName() <<"|";
+            args1.add(oss.str());
+            args1.add((uint64_t)fTableOid);
+            msg.format(args1);
+            logging::Logger logger(logid.fSubsysID);
+            logger.logMessage(LOG_TYPE_DEBUG, msg, logid);
```
drrtuy, line 387: Why should CS log this w/o checking whether debug is turned on or not?

drrtuy (approved):

dmlproc/dmlprocessor.cpp
drrtuy: The reason behind this block:
What happened is that, if a query has to wait in this queue, the wait happens before the "start sql" log. So we never saw in the log that the query had been submitted until it was released from the queue. And when it was we had no idea it had been queued. With this log, we have a better post-mortem diagnostic. If it gets where this log line spews, then something is most likely broken and that spewing is a clue.

PR #763, opened 2019-05-16
Branches/davidjmott/mcol 1758 1.0

ghost:
reference counted stl allocator


cpackEngineRPM.cmake
```diff
@@ -118,239 +118,11 @@ SET(ignored
...
-"/usr/local/mariadb/columnstore/mysql/calsetuserpriority.sql"
-"/usr/local/mariadb/columnstore/mysql/calremoveuserpriority.sql"
-"/usr/local/mariadb/columnstore/mysql/calshowprocesslist.sql"
-"/usr/local/mariadb/columnstore/mysql/columnstore_info.sql"
-${ignored})
+SET(CPACK_RPM_platform_USER_FILELIST ${ignored})
 
+SET(CPACK_RPM_libs_USER_FILELIST ${ignored})
+
+SET(CPACK_RPM_storage-engine_USER_FILELIST ${ignored})
```
pleblanc1976, line 126: for the changes above, is it safe to say that ignored == the large list of files deleted by the changes?  If not, does this still work?

utils/common/shared_pool_allocator.h
```diff
@@ -0,0 +1,195 @@
...
+      }
+
+      void push(void* pBuff)
+      {
+        node_type * pNode = reinterpret_cast<node_type*>(pBuff);
+        for (;;){
+          pNode->_next = _root.load();
+          if (_root.compare_exchange_strong(pNode->_next, pNode)) break;
+        }
+      }
```
pleblanc1976, line 133: On the surface, push & pop appear to be doing more work than what they replace.  How does performance look when you do a big join or big group by, etc?

IIRC, the stlpoolallocator is used single-threaded b/c stl containers aren't thread-safe, so at least in that application the atomic ops wouldn't be necessary.  Curious if and how much overhead they introduce.

utils/common/shared_pool_allocator.h
```diff
@@ -0,0 +1,195 @@
...
+
+      shared_pool_allocator<T>& operator=(const shared_pool_allocator<T> & src){
+        _stack = src._stack;
+        return *this;
+      }
+
+      pointer allocate(size_type, const void *hint = 0) {
+        void * pRet = _stack->pop();
+        return reinterpret_cast<pointer>(pRet);
+      }
```
pleblanc1976, line 205: I suspect this won't work with hash tables, because the same allocator used to allocate nodes is also used to allocate the table (unless that has changed in C++11 or 17, idk).  Might need to add another path to allocation for cases like that.  Ex, for node-sized allocations use the pool, for others use something else.

utils/common/shared_pool_allocator.h
```diff
@@ -0,0 +1,195 @@
...
+      void deallocate(pointer p, size_type n) {
+        _stack->push(p);
+        bool bBreak = false;
+      }
+      size_type max_size() const throw() { return (sizeof(void*) * std::numeric_limits<uint8_t>::max()); }
```
pleblanc1976: 256 pointers?

utils/common/shared_pool_allocator.h
```diff
@@ -0,0 +1,195 @@
...
+        return oRet;
+
+      }
+
+      void * pop(){
+        for (;;) {                  
+          node_type * pRet = _root.load();
+          if (!pRet) {
+            void * pVoid = NULL;
+            if (posix_memalign(&pVoid, sizeof(void*), alloc_node_size) || !pVoid) throw std::bad_alloc();
```
pleblanc1976, line 116: This seems to be allocating every node individually.  Curious to see how this affects the loading time of a struct with a lot of nodes.

pleblanc1976: Hm, the destruction time as well.  Since these are all individually allocated nodes, they'd also have to be deallocated one by one when the alloc_stack goes away.

pleblanc1976 (changes requested):
A couple things got my attention.  Chatted with Dave Mott for clarification/confirmation of what I found.

1) the table portion of the hash table might not get allocated right b/c the allocator assumes it's always allocating the same node size
2) it rounds node size up to the next power of 2 bytes (worst case is a significant waste issue)
3) in the busy-system use case, mem might never be deallocated and stay at its highest allocation point
   - also need to think about when mem limit is hit.  I think this will result in mem being overallocated until all queries using it have been cancelled.
4) in the opposite use case, one query at a time, or sparse activity, all nodes will be created and destroyed individually for each query which would definitely be a performance regression
5) in the typical use case it will be somewhere in between, which might be OK, but not sure
6) needs to be benchmarked & real mem usage patterns understood / approved of before we merge it in.

davidjmott:
[allocator_performance_comparison.xlsx](https://github.com/mariadb-corporation/mariadb-columnstore-engine/files/3210640/allocator_performance_comparison.xlsx)
Here's some profiling results of the supplied tests

ghost:
Here's some profiling results of the supplied tests

[allocator_performance_comparison.xlsx](https://github.com/mariadb-corporation/mariadb-columnstore-engine/files/3210677/allocator_performance_comparison.xlsx)

drrtuy:
I'm going to compare performance effect with QAs. Plz don't merge this yet.

PR #772, opened 2019-05-20
MCOL-537 Regression test doesn't tolerate 'failed' in stderr, stdout.

drrtuy:
I reformulate the messages.

    Changed version in preprocessor conditions to avoid compilation
    warnings in Debian 9.

    Disabled sign-compare check for generated files in DML/DDL.

### DIFF
--- a/dbcon/ddlpackage/CMakeLists.txt
+++ b/dbcon/ddlpackage/CMakeLists.txt
@@ -9,6 +9,8 @@ ADD_CUSTOM_COMMAND(
     DEPENDS ddl.y ddl.l
 )
 
+set_source_files_properties(ddl-scan.cpp PROPERTIES COMPILE_FLAGS -Wno-sign-compare)
+
 # Parser puts extra info to stderr.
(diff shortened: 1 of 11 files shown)


pleblanc1976 (changes requested):
There are still a couple distros it's failing to compile on.  I would suggest separating the issues.  This ticket only requires changing the 'failed' text, which would be easy & quick.  I'd open another ticket for the remaining compiler warnings and get to that when you have spare time (HA!).

drrtuy:
Centos 6 fails b/c we stoped to support it when enabled C++11, U16 fails b/c of the failed API test - Mcol1408. Andrew promiced to look into this. Windows build complains about non-absolute paths for mcsimport.

LinuxJedi:
We should remove develop from the CentOS 6 builder. I will indeed look into the MCOL-1408 regression soon.

pleblanc1976 (approved):
Ok will approve & merge & etc.

PR #778, opened 2019-06-03
fix to possible memory corruption

ghost:
b >> reinterpret_cast< ByteStream::doublebyte&>(fAsc); casts a byte to a short so two bytes get written to fAsc.

### DIFF
--- a/dbcon/execplan/arithmeticcolumn.cpp
+++ b/dbcon/execplan/arithmeticcolumn.cpp
@@ -326,7 +326,8 @@ void ArithmeticColumn::serialize(messageqcpp::ByteStream& b) const
     ObjectReader::writeParseTree(fExpression, b);
     b << fTableAlias;
     b << fData;
-    b << static_cast<const ByteStream::doublebyte>(fAsc);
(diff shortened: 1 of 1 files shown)


davidhall586:
I think it would be a better solution to add bool as a supported type to bytestream, rather than make a copy like this. Using sizeof(bool), we could determine the proper number of bytes to move into the stream.

drrtuy:
ByteStream divides supported data types in floats and ints future dividing ints by width so I would just use byte + cast instead in ser/des functions.

ghost:
> I think it would be a better solution to add bool as a supported type to bytestream, rather than make a copy like this. Using sizeof(bool), we could determine the proper number of bytes to move into the stream.

I began removing all the overloads and creating a handful of specializations. One implementation handled POD while a few other handled containers. This exposed a couple bigger problems with the bytestream implementation that need to be addressed. I don't want to add to the pile right now.

davidjmott:
This is a fix for a memory corruption bug.

davidhall586 (approved):

PR #796, opened 2019-06-26
Branches/davidjmott/mcol 1503

ghost:
a lot of work that needs to be in place for flexible builds and installers


ghost:
this introduces a way to get reliable build dependencies from a legacy sort of automake setup
since there are circular dependencies, theres targets called <original target>_h which only carry include definitions
the <project name>_h projects are defined first and able to resolve at generate time
its a sort of cmake 'forward declaration'
it decouples the required include directories from the link libraries and compile flags
so can still have a spaghetti include structure but cmake will resolve link dependencies and perfrom minimial rebuilds
you can checkout the repo from within server/storage and it will build as part of the server build

can also build them separately by passing -DMySQL_BINARY_DIR=<location of server build dir>
in both cases it will use the server's CMAKE_BUILD_TYPE to prevent conflicts
most of the changes are cmake related. i shuffled around a few forward declarations and fully qualified a bunch of types, reworked include hierarchy here and there sorta build things

ghost:
going to clean up with squash commit

PR #805, opened 2019-07-12
Brm io cleanup (MCOL-3242)

pleblanc1976:
This is the follow up to the PR for MCOL-3242.  The initial changes to force the BRM structs to use their IDBDataFile paths (vs fstream path) seems to be working.  This PR removes the old fstream paths.

### DIFF
--- a/versioning/BRM/extentmap.cpp
+++ b/versioning/BRM/extentmap.cpp
@@ -1133,85 +1133,6 @@ void ExtentMap::reserveLBIDRange(LBID_t start, uint8_t size)
 	    ...   (* numFL)
 */
 
-void ExtentMap::loadVersion4(ifstream& in)
-{
-    int emNumElements, flNumElements;
-
-    in.read((char*) &emNumElements, sizeof(int));
-    in.read((char*) &flNumElements, sizeof(int));
-    idbassert(emNumElements > 0);
-
-    void *fExtentMapPtr = static_cast<void*>(fExtentMap);
-    memset(fExtentMapPtr, 0, fEMShminfo->allocdSize);
-    fEMShminfo->currentSize = 0;
-
-    // init the free list
-    memset(fFreeList, 0, fFLShminfo->allocdSize);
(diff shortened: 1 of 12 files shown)


versioning/BRM/sessionmanagerserver.h
```diff
@@ -123,10 +123,7 @@ class SessionManagerServer
      * It does not destroy the semaphores.  Those persist until the system
      * is shut down.
      */
-    virtual ~SessionManagerServer()
-    {
-        if (txnidfd >= 0 ) close(txnidfd);
-    }
+    virtual ~SessionManagerServer();
```
drrtuy, line 126: We probably don't need virtual dtor at all.

versioning/BRM/extentmap.cpp
```diff
@@ -1557,86 +1412,23 @@ void ExtentMap::save(const string& filename)
...
     }
-    else   // this is the fstream version to be expired
-    {
-        ofstream out;
 
-        // Make em writes to disk use a buffer size of StrmBufSize bytes (instead of the default 8K)
-        const unsigned StrmBufSize = 1 * 1024 * 1024;
-        scoped_array<char> buf(new char[StrmBufSize]);
-        out.rdbuf()->pubsetbuf(buf.get(), StrmBufSize);
+    //allocdSize = fFLShminfo->allocdSize / sizeof(InlineLBIDRange);
```
drrtuy, line 1416: Just a quick cosmetic question? Do we really need to preserve the comments?

drrtuy (approved):
I have no objections on the patch merely a cosmetic notes. 
Great job indeed.

ghost (approved):

versioning/BRM/sessionmanagerserver.h
pleblanc1976: Now that it does nothing that's true.  We can do a second round of cleanup at some point.

versioning/BRM/extentmap.cpp
pleblanc1976: Probably not.  :)

PR #809, opened 2019-07-31
Mcol 174

davidhall586:
Allows numerical functions on strings, so long as the string can be interpreted as a number. Ex: '255' or '0x0b1123df'
Also replaced custom power function with standard to allow for more than nine decimal places.


LinuxJedi (changes requested):
This PR doesn't compile

LinuxJedi (approved):

PR #811, opened 2019-08-01
MCOL-2219 Fix space handling in DDL parser

LinuxJedi:
Allow non-alphanumeric character after space in column names.

### DIFF
--- a/dbcon/ddlpackage/ddl.l
+++ b/dbcon/ddlpackage/ddl.l
@@ -75,10 +75,8 @@ ident_cont [A-Za-z\200-\377_0-9\$]
 identifier {ident_start}{ident_cont}*
 extended_identifier {ident_start}{extended_ident_cont}*
 /* fully qualified names regexes */
-ident_w_spaces {identifier}\x20*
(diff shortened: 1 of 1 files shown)


dbcon/ddlpackage/ddl.l
```diff
@@ -65,7 +65,7 @@ double_quote \"
 grave_accent `
 
 comment ("--"{non_newline}*)
-extended_ident_cont [A-Za-z\200-\377_0-9\$#,()\[\].;\:\+\-\*\/\%\^\<\>\=!&|@\\]
+extended_ident_cont [ A-Za-z\200-\377_0-9\$#,()\[\].;\:\+\-\*\/\%\^\<\>\=!&|@\\]
```
drrtuy, line 68: We better clean up lexer removing 'column_ident_quoted' and 'ident_w_spaces' patterns given this change. JFYI CS works as expected after I removed both patterns and checked regression 023 and MCOL-2219 case.

drrtuy (changes requested):
Nice and simplistic change. Could you clean up removing unused patterns.

drrtuy (approved):

PR #812, opened 2019-08-05
MCOL-2178: Run CS with vanilla 10.4

tntnatbry:
1. Roman's work on the addition of infrastructure for select_handler, derived_handler, and group_by_handler in CS.
2. Several bug/crash fixes.
3. Cleanup of MIGR:: singleton from the plugin code.
4. Addition of session variables for Enabling (default value)/Disabling the use of smart handlers.

To be compiled against the Server (https://github.com/MariaDB/server) branch bb-10.4-MDEV-19831


tntnatbry:
@drrtuy FYI.

dbcon/mysql/CMakeLists.txt
```diff
@@ -22,6 +22,8 @@ SET ( libcalmysql_SRCS
     ha_pseudocolumn.cpp)
 
 add_definitions(-DMYSQL_DYNAMIC_PLUGIN)
+add_definitions(-DEBUG_WALK_COND)
+add_definitions(-DINFINIDB_DEBUG)
```
LinuxJedi, line 26: I think these two will need to be turned off before release.

dbcon/mysql/ha_calpont.cpp
```diff
@@ -159,6 +174,10 @@ static int infinidb_init_func(void* p)
     calpont_hton->commit = calpont_commit;
     calpont_hton->rollback = calpont_rollback;
     calpont_hton->close_connection = calpont_close_connection;
+    calpont_hton->create_group_by = create_calpont_group_by_handler;
```
LinuxJedi, line 177: Can you think of a good reason to keep the "InfiniDB" engine now?

dbcon/mysql/ha_calpont_impl.cpp
```diff
@@ -779,28 +764,14 @@ int fetchNextRow(uchar* buf, cal_table_info& ti, cal_connection_info* ci, bool h
...
-                    (*f)->field_length = 310;
-
-                    //double double_val = *(double*)(&value);
-                    //f2->store(double_val);
-                    if ((f2->decimals() == DECIMAL_NOT_SPECIFIED && row.getScale(s) > 0)
-                            || f2->decimals() < row.getScale(s))
-                    {
-                        f2->dec = row.getScale(s);
-                    }
+                    (*f)->field_length = 40;
```
LinuxJedi, line 767: Why are we truncating the length of double to the length of float?

dbcon/mysql/ha_calpont_impl.cpp
```diff
@@ -3183,8 +2892,7 @@ int ha_calpont_impl_create(const char* name, TABLE* table_arg, HA_CREATE_INFO* c
     cal_connection_info* ci = reinterpret_cast<cal_connection_info*>(get_fe_conn_info_ptr());
 
     // @bug1940 Do nothing for select query. Support of set default engine to IDB.
-    if (thd->infinidb_vtable.vtable_state == THD::INFINIDB_CREATE_VTABLE ||
-            string(name).find("@0024vtable") != string::npos)
+    if (string(name).find("@0024vtable") != string::npos)
```
LinuxJedi, line 2895: I suspect we don't need this now.

dbcon/mysql/ha_calpont_impl.cpp
```diff
@@ -5275,4 +4878,475 @@ int ha_calpont_impl_group_by_end(ha_calpont_group_by_handler* group_hand, TABLE*
...
+
+    if (!getenv("CALPONT_HOME"))
+    {
+        string calpontHome(home + "/Calpont/etc/");
+        setenv("CALPONT_HOME", calpontHome.c_str(), 1);
+    }
+
+    if (!getenv("CALPONT_CONFIG_FILE"))
+    {
+        string calpontConfigFile(home + "/Calpont/etc/Columnstore.xml");
```
LinuxJedi, line 4908: Whatever this is, please tell me it is wrong. We shouldn't have 'Calpont' in new paths and it isn't a valid old path.

dbcon/mysql/ha_mcs_pushdown.cpp
```diff
@@ -0,0 +1,620 @@
...
+
+void mutate_optimizer_flags(THD *thd_)
+{
+   // MCOL-2178 Disable all optimizer flags as it was in the fork.
+    // CS restores it later in SH::scan_end() and in case of an error
+    // in SH::scan_init()
+    set_original_optimizer_flags(thd_->variables.optimizer_switch, thd_);
+    thd_->variables.optimizer_switch = OPTIMIZER_SWITCH_IN_TO_EXISTS |
+        OPTIMIZER_SWITCH_EXISTS_TO_IN |
+        OPTIMIZER_SWITCH_COND_PUSHDOWN_FOR_DERIVED;
```
LinuxJedi, line 28: We need to rethink this. We actually only had one or two flags that didn't work but ended up switching most of them off in a massive hammer fashion. We should probably fix this.

LinuxJedi (approved):

dbcon/mysql/CMakeLists.txt
drrtuy: These two will be removed for sure.

dbcon/mysql/ha_calpont.cpp
drrtuy: No, 1.4 is a good milestone to drop Infinidb support.

dbcon/mysql/ha_calpont_impl.cpp
drrtuy: We previously set field_length to 310 that was the reason behind MCOL-2243.

dbcon/mysql/ha_calpont_impl.cpp
drrtuy: For sure.

dbcon/mysql/ha_calpont_impl.cpp
drrtuy: This is an outdated way of changing module role for debugging purposes in CSC catalog connections. We don't use this mechanism so I will remove them.

dbcon/mysql/ha_mcs_pushdown.cpp
drrtuy: I would better disable all the flags and later enable them one by one. But this work is big enough for a separate issue.

dbcon/mysql/ha_calpont_impl.cpp
LinuxJedi: Possibly, I'm still not convinced that was the actual cause. Changing it caused regressions which is why we aren't touching this for now. Please revert it until we can fix it properly.

dbcon/mysql/ha_mcs_pushdown.cpp
LinuxJedi: Completely agree

dbcon/mysql/ha_calpont_impl.cpp
drrtuy: The client claims it was.

PR #814, opened 2019-08-07
Corporaton -> Corporation

pleblanc1976:
Noticed this typo in all/most of our license headers.  Did a mass sed substitution to fix.


pleblanc1976:
wait a min, this is including the s3 commits as well.

PR #817, opened 2019-08-09
Follow up to PR #812

tntnatbry:


LinuxJedi (commented):
DEBUG_WALK_COND is useful for debugging, it appears that this patch removes it completely rather than turning it off. Any reason why?

drrtuy:
No, there is no reason to remove this macro so we will revise the PR.

tntnatbry:
I didn't realise this was useful. I have reverted this change and created a new PR #818. Closing this.

PR #827, opened 2019-08-16
MCOL-3317 Moved fill-next-block from writeRow() into allocRowId.

drrtuy:
Intro* INSERT statements could face a non-existant block when MCOL-498 feature
    is enabled. writeRow() guard blocks was supposed to proactively create empty
    blocks. The pre-patch logic failed when first value in the block has been
    removed by DELETE and this overwrites the whole valid block with empty magics.
    This patch moves proactive creation logic into allocRowId().

### DIFF
--- a/dbcon/ddlpackageproc/droptableprocessor.cpp
+++ b/dbcon/ddlpackageproc/droptableprocessor.cpp
@@ -289,9 +289,9 @@ DropTableProcessor::DDLResult DropTableProcessor::processPackage(ddlpackage::Dro
 
         //get a unique number
         VERBOSE_INFO("Removing the SYSTABLE meta data");
-//#ifdef IDB_DDL_DEBUG
+#ifdef IDB_DDL_DEBUG
         cout << fTxnid.id << " Removing the SYSTABLEs meta data" << endl;
-//#endif
+#endif
         bytestream << (ByteStream::byte)WE_SVR_DELETE_SYSTABLE;
         bytestream << uniqueId;
         bytestream << (uint32_t) dropTableStmt.fSessionID;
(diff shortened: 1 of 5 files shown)


dbcon/ddlpackageproc/droptableprocessor.cpp
```diff
@@ -686,9 +686,9 @@ DropTableProcessor::DDLResult DropTableProcessor::processPackage(ddlpackage::Dro
     //Flush primProc cache
     rc = cacheutils::flushOIDsFromCache( oidList );
     //Delete extents from extent map
-//#ifdef IDB_DDL_DEBUG
+#ifdef IDB_DDL_DEBUG
     cout << fTxnid.id << " Drop table deleteOIDs" << endl;
-//#endif
+#endif
```
pleblanc1976, line 691: Did you mean to disable these printouts?

writeengine/wrapper/we_colop.cpp
```diff
@@ -507,14 +517,27 @@ int ColumnOp::allocRowId(const TxnID& txnid, bool useStartingExtent,
...
+                    // MCOL-498 add a block.
+                    // DRRTUY TODO Given there is no more hwm pre-allocated we
+                    // could extend the file once to accomodate all records.
+                    if (rc != NO_ERROR)
+                    {
+                        if (rc == ERR_FILE_EOF)
+                        {
+                            uint64_t emptyVal = getEmptyRowValue(newCol.colDataType, newCol.colWidth);
+                            setEmptyBuf(buf, BYTE_PER_BLOCK, emptyVal, newCol.colWidth);
+                            RETURN_ON_ERROR(saveBlock(newCol.dataFile.pFile, buf, newHwm));
```
pleblanc1976, line 529: Am I correct in believing that these calls to saveBlock() replace the parts of write_row() deleted below?

pleblanc1976 (approved):
It's pretty hard to understand changes to write engine, but given your description of the problem in slack, and some of my own assumptions, I think this code is doing the right thing.  To a large degree though, I have to trust your test results.  :)

dbcon/ddlpackageproc/droptableprocessor.cpp
drrtuy: I mean to guard them with macros instead of printing them to null in production.

writeengine/wrapper/we_colop.cpp
drrtuy: Precisely.

PR #857, opened 2019-09-09
MCOL-3488.  Writeengine getting IDB-FS an odd way

pleblanc1976:
Found that writeengine was getting an IDBFilesystem instance in a way different than other components.  Errors would only show up during test112, when a compression chunk needs to expand (lots of updates in the same region of the table).

Made those instances get an SMFilesystem (the cloud type) instance when appropriate.


LinuxJedi (approved):
We should probably change this stuff to a switch() statement at some point...

PR #863, opened 2019-09-14
Fix for IF function in the order by clause and handle hexadecimal literals.

tntnatbry:
Fixes working_tpch1/misc/CCBillEU.sql and working_tpch1_compareLogOnly/view/mts_view.40.sql.

### DIFF
--- a/dbcon/mysql/ha_calpont_execplan.cpp
+++ b/dbcon/mysql/ha_calpont_execplan.cpp
@@ -298,6 +298,34 @@ bool sortItemIsInGroupRec(Item* sort_item, Item* group_item)
     return found;
 }
 
+/*@brief check_sum_func_item - This traverses Item       */
+/**********************************************************
+* DESCRIPTION:
+* This f() walks Item looking for the existence of
+* a Item::REF_ITEM, which references an item of
+* type Item::SUM_FUNC_ITEM
+* PARAMETERS:
+*    Item * Item to traverse
+* RETURN:
+*********************************************************/
+void check_sum_func_item(const Item* item, void* arg)
+{
+    bool* found = reinterpret_cast<bool*>(arg);
+
+    if (*found)
+        return;
+
+    if (item->type() == Item::REF_ITEM)
+    {
+        const Item_ref* ref_item = reinterpret_cast<const Item_ref*>(item);
+        Item* ref_item_item = (Item*) *ref_item->ref;
+        if (ref_item_item->type() == Item::SUM_FUNC_ITEM)
+        {
+            *found = true;
+        }
+    }
+}
+
 /*@brief sortItemIsInGrouping- seeks for an item in grouping*/
 /***********************************************************
  * DESCRIPTION:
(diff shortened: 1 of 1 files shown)


dbcon/mysql/ha_calpont_execplan.cpp
```diff
@@ -5150,6 +5190,16 @@ void gp_walk(const Item* item, void* arg)
 
                 case STRING_RESULT:
                 {
+                    // Special handling for 0xHHHH literals
+                    const Type_handler *tph = item->type_handler();
+                    if (dynamic_cast<const Type_handler_hex_hybrid*>(tph) != NULL)
```
drrtuy: That is cheaper to use typeid(*tph)==typeid(Type_handler_hex_hybrid) b/c dynamic_cast recurses through the class hierarchy. However tph could be of a different class that inherits from Type_handler_hex_hybrid. If the suggestion is applicable then use it.

dbcon/mysql/ha_calpont_execplan.cpp
```diff
@@ -5150,6 +5190,16 @@ void gp_walk(const Item* item, void* arg)
...
+                    {
+                        Item_hex_hybrid *hip = (Item_hex_hybrid*) item;
+                        gwip->rcWorkStack.push(new ConstantColumn((int64_t)hip->val_int(), ConstantColumn::NUM));
+                        (dynamic_cast<ConstantColumn*>(gwip->rcWorkStack.top()))->timeZone(gwip->thd->variables.time_zone->get_name()->ptr());
```
drrtuy: Heavy to read. Kindly separate operations to make it readable.

dbcon/mysql/ha_calpont_execplan.cpp
```diff
@@ -5150,6 +5190,16 @@ void gp_walk(const Item* item, void* arg)
 
                 case STRING_RESULT:
                 {
+                    // Special handling for 0xHHHH literals
+                    const Type_handler *tph = item->type_handler();
+                    if (dynamic_cast<const Type_handler_hex_hybrid*>(tph) != NULL)
+                    {
+                        Item_hex_hybrid *hip = (Item_hex_hybrid*) item;
```
drrtuy: I know that MDB code uses c-type cast but we should stick with cpp-style reinterpret_cast. Feel free to tell me if I use c-style cast accidently.

dbcon/mysql/ha_calpont_execplan.cpp
tntnatbry: This is a good idea. Done.

dbcon/mysql/ha_calpont_execplan.cpp
tntnatbry: Done.

dbcon/mysql/ha_calpont_execplan.cpp
tntnatbry: Done. I could not directly use reinterpret_cast as the compiler complained about the constness of item (as it is passed as const Item* to gp_walk()). So I had to first cast away the constness using const_cast.

tntnatbry:
PR is updated to incorporate the feedback. Please review.

drrtuy (approved):

PR #864, opened 2019-09-15
MCOL-2178 CS now sets optimizer flags like it did in the fork-era.

drrtuy:
This happens in external_lock() whilst locking the table.

Fixes LIMIT=1 optimization for EXISTS_SUBS subqueries.

external_lock() contains if condition that gives false positive
for SH + pushed conditions.

external_lock() now resets in_subquery_conversion_threshold
variable that governs IN_INTO_SUBQUERY optimization for
queries run in table mode.

external_lock() now purges dynamicall allocated condInfo for
SH and DH execution path.

Commented out UNION check b/c if condition gives false positives
and silently enables table mode execution for queries w/o
UNION.

### DIFF
--- a/dbcon/mysql/ha_mcs_pushdown.cpp
+++ b/dbcon/mysql/ha_mcs_pushdown.cpp
@@ -286,8 +286,8 @@ void check_user_var_func(const Item* item, void* arg)
     if (item->type() == Item::FUNC_ITEM)
     {
         const Item_func* ifp = reinterpret_cast<const Item_func*>(item);
-        std::string funcName = ifp->func_name();
-        if (funcName == "set_user_var" || funcName == "get_user_var")
+        std::string funcname = ifp->func_name();
+        if (funcname == "set_user_var" || funcname == "get_user_var")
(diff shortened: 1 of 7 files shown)


dbcon/mysql/ha_mcs_pushdown.cpp
```diff
@@ -718,15 +718,17 @@ create_columnstore_select_handler(THD* thd, SELECT_LEX* select_lex)
         return handler;
     }
 
+    // This block gives false positive in
+    // partitionOptimization/vTpch11.
     // Disable SH processing for UNION in a subquery.
     // This way, the server falls back to using the DH for executing
     // the UNION, if enabled.
-    TABLE_LIST *tbl_ptr = select_lex->get_table_list();
+    /*TABLE_LIST *tbl_ptr = select_lex->get_table_list();
```
drrtuy, line 726: 4 Gagan. Could you look why this block causes false positive in the mentioned test?

tntnatbry (approved):

PR #869, opened 2019-09-20
MCOL-3436: Modify signal handling to not shutdown on non term or core signals.

benthompson15:


pleblanc1976 (approved):
If you're sure those are the sigs we should shutdown on and there aren't any missing, this looks OK.  I'll let you double-check the list and merge.

LinuxJedi:
I recommend a follow-up in the future which uses utils/common/crashtrace.*

LinuxJedi (approved):

PR #871, opened 2019-09-24
MCOL-3514 Add support for S3 to cpimport

LinuxJedi:
cpimport now has the ability to use libmarias3 to read an object from an
S3 bucket instead of a file on local disk.

This also moves libmarias3 to utils/libmarias3.


LinuxJedi:
Out of tree build broken but this, please hold for a follow-up

benthompson15 (approved):

PR #877, opened 2019-09-25
Handle approximation of double values with a scientific notation properly in FORMAT function.

tntnatbry:
Conditional fix for working_tpch1_compareLogOnly/misc/bug3488.sql.

The results for FORMAT function on a double field do not currently match the reference db as the server is truncating the output length to 70. This is due to ltem::max_length set to 22 for a double field. This needs to be fixed upstream.


drrtuy:
Don't merge this.

LinuxJedi (changes requested):
There are likely much better (less expensive) ways of doing this.

tntnatbry:
We will revisit this at a later point, closing this for now.

PR #885, opened 2019-10-03
Allow a window function in the ORDER BY clause.

tntnatbry:
Fixes 5 tests in working_tpch1/windowFunctions:

working_tpch1/windowFunctions/q0027.sql                                                 
working_tpch1/windowFunctions/q0032.sql
working_tpch1/windowFunctions/q0032b.sql                                                
working_tpch1/windowFunctions/q0033.sql                                                 
working_tpch1/windowFunctions/q0335.sql


drrtuy:
Great job. But are you sure that WF happens before sorting during anex step?

tntnatbry:
WFS executes before TNS. When an order by exists in the presence of a window function in the query, WFS first executes the function, then performs sorting on the resulting rows.

drrtuy (approved):

PR #887, opened 2019-10-04
MCOL-3498: only update bashrc if DataFileEnvFile is set.

benthompson15:


oam/install_scripts/module_installer.sh
```diff
@@ -127,12 +127,14 @@ if [ -n "$plugin" ]; then
```
pleblanc1976, line 127: This seems to mean that this code has something to do with local queries.  Odd.

pleblanc1976 (approved):

LinuxJedi (approved):

PR #892, opened 2019-10-07
MCOL-3541: add modtime to smls outputs

benthompson15:


storage-manager/src/smls.cpp
```diff
@@ -80,7 +82,11 @@ void lsOffline(const char *path)
             }
             else
                 cout.width(15);
-            cout << right << _stat.st_size << left << " " << entry << endl;
+
+            struct tm *my_tm = localtime(&_stat.st_mtim.tv_sec);
+            char date[32];
```
pleblanc1976: localization possibilities may need more space...?  I'd say pad this to 100 bytes to eliminate the question b/c who cares.

pleblanc1976 (changes requested):

pleblanc1976 (approved):

storage-manager/src/smls.cpp
```diff
@@ -80,7 +82,11 @@ void lsOffline(const char *path)
             }
             else
                 cout.width(15);
-            cout << right << _stat.st_size << left << " " << entry << endl;
+
+            struct tm *my_tm = localtime(&_stat.st_mtim.tv_sec);
```
LinuxJedi, line 86: Not sure if it matters, but this isn't thread safe

pleblanc1976: it's a single-threaded tool

PR #894, opened 2019-10-07
MCOL-3357

benthompson15:


storage-manager/src/Cache.cpp
```diff
@@ -56,20 +56,32 @@ Cache::Cache()
...
-        logger->log(LOG_CRIT, "Cache/cache_size is not set");
-        throw runtime_error("Please set Cache/cache_size in the storagemanager.cnf file");
+        logger->log(LOG_CRIT, "Cache/cache_size is not set. Using default value = %zi",defaultCacheValue);
+        maxCacheSize = defaultCacheValue;
+        //throw runtime_error("Please set Cache/cache_size in the storagemanager.cnf file");
     }
     try
     {
-        maxCacheSize = stoul(stmp);
+        size_t configCacheValue = stoul(stmp);
```
pleblanc1976, line 69: stoul() only handles unsigned 64-bit, might want to make that match the size_t type.

pleblanc1976 (changes requested):
Doesn't implement the listener interface.  Should be mainly a cut & paste job to make it do that.  Listener pattern -> better encapsulation & more maintainable.

storage-manager/src/Cache.cpp
LinuxJedi: Did you mean unsigned 32-bit? stoull() is the 64-bit one.

pleblanc1976: Sry, typo, yes, stoull() is what you're after Ben.

PR #898, opened 2019-10-09
Add status variables to show ColumnStore version and commit hash.

tntnatbry:
2 new status variables are added:
  "Columnstore_commit_hash": Displays the abbreviated commit hash of the HEAD of the version tree.
  "Columnstore_version": Displays the CS version based on parsing of the VERSION file in cmake.

### DIFF
--- a/dbcon/mysql/ha_calpont.cpp
+++ b/dbcon/mysql/ha_calpont.cpp
@@ -33,6 +33,8 @@ static int calpont_commit(handlerton* hton, THD* thd, bool all);
 static int calpont_rollback(handlerton* hton, THD* thd, bool all);
 static int calpont_close_connection ( handlerton* hton, THD* thd );
 handlerton* mcs_hton;
+char cs_version[25];
+char cs_commit_hash[41]; // a commit hash is 40 characters
 
 // handlers creation function for hton.
 // Look into ha_mcs_pushdown.* for more details.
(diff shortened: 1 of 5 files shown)


dbcon/mysql/ha_calpont.cpp
```diff
@@ -33,6 +33,8 @@ static int calpont_commit(handlerton* hton, THD* thd, bool all);
 static int calpont_rollback(handlerton* hton, THD* thd, bool all);
 static int calpont_close_connection ( handlerton* hton, THD* thd );
 handlerton* mcs_hton;
+char cs_version[256];
+char cs_commit_hash[256];
```
LinuxJedi: IIRC a git hash is 40 characters, so is this overkill?

dbcon/mysql/ha_calpont.cpp
```diff
@@ -128,6 +130,9 @@ static int columnstore_init_func(void* p)
 
     fprintf(stderr, "Columnstore: Started; Version: %s-%s\n", columnstore_version.c_str(), columnstore_release.c_str());
 
+    memcpy(cs_version, columnstore_version.c_str(), columnstore_version.size());
+    memcpy(cs_commit_hash, columnstore_commit_hash.c_str(), columnstore_commit_hash.size());
+
```
LinuxJedi, line 138: I suspect these should be strncpy()?

dbcon/mysql/ha_calpont.cpp
drrtuy: I wonder if we can get away including columnstoreversion.h and using its variables as the source of the status vars?

tntnatbry: I arbitrarily chose the length to 256 earlier. Updated this part to set the length of cs_commit_hash to 41 and cs_version to 25 (hoping 25 is enough length for the version).

The only available status variable types are C-style strings (SHOW_CHAR). Using the variables of columnstoreversion.h directly is therefore not possible as they are defined as C++ string objects.

dbcon/mysql/ha_calpont.cpp
tntnatbry: Makes sense. Updated this to strncpy instead of memcpy.

drrtuy (approved):

PR #899, opened 2019-10-09
MCOL-3357: SM config change listener

benthompson15:


storage-manager/src/Cache.h
```diff
@@ -94,6 +101,7 @@ class Cache : public boost::noncopyable
         void reset();
         void validateCacheSize();
         
+        virtual void configListener() override;
     private:
         Cache();
```
pleblanc1976, line 107: I suggest adding a dtor where the config-listener is deregistered.  Future modders won't think of it.

pleblanc1976 (changes requested):

storage-manager/src/Cache.h
pleblanc1976: nvm, just noticed it has one and it's doing the right thing.

pleblanc1976 (approved):

PR #901, opened 2019-10-09
Replace ha_calpont with ha_mcs in the source code and filenames in the plugin code.

tntnatbry:


LinuxJedi (approved):
Will definitely cause conflicts with the branches I'm working on. But looks good to me :)

drrtuy (approved):

PR #902, opened 2019-10-10
Create separate functions for saving and restoring group by list columns.

tntnatbry:


dbcon/mysql/ha_mcs_pushdown.cpp
```diff
@@ -43,6 +43,36 @@ void restore_optimizer_flags(THD *thd_)
     }
 }
 
+int save_group_list(THD* thd, SELECT_LEX* select_lex, Group_list_ptrs *&group_list_ptrs)
```
drrtuy: Reference to a pointer works but look a bit overcomplicated though. Could you replace it with Group_list_ptrs** and deref it inside?

tntnatbry: Done. PR is updated with this change.

drrtuy (approved):

PR #903, opened 2019-10-11
MCOL-3551 Use generic MariaDB Server paths

LinuxJedi:
This branch enforces the use of generic MariaDB server paths for their
binaries and data rather than custom paths.

/usr/local/mariadb/columnstore is now only for columnstore with this
patch.

It should be noted that this removes the auto-mounting of external
MariaDB UM data storage for AWS.

This is also a fix for MCOL-3510 when buildbot is reconfigured.


LinuxJedi:
Some missing things are going into another branch. Closing this one

PR #909, opened 2019-10-14
Mcol 3536 1.2

davidhall586:


drrtuy (approved):
IdbSortSpec::fLocale has been added for a later usage when CS will extract locale from a table settings.

utils/windowfunction/idborderby.cpp
```diff
@@ -130,13 +130,20 @@ int StringCompare::operator()(IdbCompare* l, Row::Pointer r1, Row::Pointer r2)
...
+        const char* s1 = (const char*)l->row1().getStringPointer(fSpec.fIndex);
+        const char* s2 = (const char*)l->row2().getStringPointer(fSpec.fIndex);
+        // For Japanese, coll.compare() may not be as correct as strncmp
+        if (JPcodePoint)
+        {
+            ret = fSpec.fAsc * strncmp(s1, s2, max(len1,len2));
+        }
+        else
+        {
+            const std::collate<char>& coll = std::use_facet<std::collate<char> >(loc);
```
LinuxJedi, line 144: We should maybe consider setting this up once somewhere (maybe per thread, can't remember thread safety) so we don't have to do it on every compare.

PR #916, opened 2019-10-22
MCOL-894 Parallel sorting.

drrtuy:
MCOL-894 Finished with parallel workers invocation.
     The implementation lacks final aggregation step.

MCOL-894 TupleAnnexStep's init and destructor are now parallel execution aware.

    Implemented final merging step for parallel execution finalizeParallelOrderBy().

    Templated unit test to use it with arbitrary number of rows, threads.

Reuse LimitedOrderBy in the final step

MCOL-894 Cleaned up finalizeParallelOrderBy.

MCOL-894 Add and propagate thread variable that controls a number of threads.

    Optimized comparators used for sorting and add corresponding UTs.

    Refactored TupleAnnexStep::finalizeParallelOrderByDistinct.

    Parallel sorting methods now preallocates memory in batches.


utils/windowfunction/idborderby.cpp
```diff
@@ -6,8 +6,8 @@
    as published by the Free Software Foundation; version 2 of
    the License.
 
-   This program is distributed in the hope that it will be useful,
-   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   his program is distributed in the hope that it will be useful,
+   ut WITHOUT ANY WARRANTY; without even the implied warranty of
    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
```
LinuxJedi: Please fix the license edit

LinuxJedi (changes requested):
Unfortunately the typo in the license is important. Once that is fixed it looks good to me :)

utils/windowfunction/idborderby.cpp
drrtuy: Sure thing but I will wait for Patrick's review first.

dbcon/joblist/jlf_tuplejoblist.cpp
```diff
@@ -494,12 +496,13 @@ void adjustLastStep(JobStepVector& querySteps, DeliveredTableMap& deliverySteps,
...
 
-//       if (jobInfo.limitCount != (uint64_t) - 1)
-//        {
     if (jobInfo.orderByColVec.size() > 0)
+    {
         tas->addOrderBy(new LimitedOrderBy());
-
-//        }
+        if (jobInfo.orderByThreads > 1)
+            tas->setParallelOp();
```
pleblanc1976, line 503: tested both ways i assume (single & parallel)?

dbcon/joblist/jlf_tuplejoblist.cpp
```diff
@@ -518,7 +521,11 @@ void adjustLastStep(JobStepVector& querySteps, DeliveredTableMap& deliverySteps,
         if (jobInfo.trace) cout << "Output RowGroup 2: " << rg2.toString() << endl;
 
         AnyDataListSPtr spdlIn(new AnyDataList());
-        RowGroupDL* dlIn = new RowGroupDL(1, jobInfo.fifoSize);
+        RowGroupDL* dlIn;
+        if (jobInfo.orderByColVec.size() > 0) 
+            dlIn = new RowGroupDL(jobInfo.orderByThreads, jobInfo.fifoSize);
+        else
+            dlIn = new RowGroupDL(1, jobInfo.fifoSize);
```
pleblanc1976, line 528: this might be a mismatch.  Checking the # of cols to order by as a way to decide how many threads will consume the data, doesn't seem right.  Also, seems like the # of threads consuming the data is always orderByThreads...?  Could you double-check this?

dbcon/joblist/joblistfactory.cpp
```diff
@@ -1722,7 +1722,7 @@ void makeVtableModeSteps(CalpontSelectExecutionPlan* csep, JobInfo& jobInfo,
                          JobStepVector& querySteps, JobStepVector& projectSteps, DeliveredTableMap& deliverySteps)
 {
     // @bug4848, enhance and unify limit handling.
-    if (csep->limitNum() != (uint64_t) - 1)
+    //if (csep->limitNum() != (uint64_t) - 1)
```
pleblanc1976: the code block below looks like it depends on there being a limit, is commenting this line intentional?

dbcon/joblist/tupleannexstep.cpp
```diff
@@ -149,25 +170,41 @@ void TupleAnnexStep::setOutputRowGroup(const rowgroup::RowGroup& rg)
...
+    uint64_t id = 1;
     fRowGroupIn = rgIn;
     fRowGroupIn.initRow(&fRowIn);
-
-    if (fOrderBy)
+    if(fParallelOp && fOrderBy)
     {
-        fOrderBy->distinct(fDistinct);
-        fOrderBy->initialize(rgIn, jobInfo);
+        fOrderByList.resize(fMaxThreads+1);
```
pleblanc1976, line 179: what is the +1 for?  This looks like we have maxthreads + 1 LimitedOrderBy instances...?

dbcon/joblist/tupleannexstep.cpp
```diff
@@ -129,6 +134,22 @@ TupleAnnexStep::TupleAnnexStep(const JobInfo& jobInfo) :
 
 TupleAnnexStep::~TupleAnnexStep()
 {
+    if(fParallelOp)
+    {
+        if(fOrderByList.size() > 0)
+        {
+            for(uint64_t id = 0; id <= fMaxThreads; id++)
```
pleblanc1976: if this is supposed to deallocate everything in fOrderByList, better to use size() as the iteration limit for clarity.  That is, unless you read other code, you can't tell that this cleans up fOrderByList completely.

dbcon/joblist/tupleannexstep.cpp
```diff
@@ -213,17 +248,58 @@ void TupleAnnexStep::run()
...
+    if(fParallelOp)
+    {
+        // Indexing begins with 1
+        fRunnersList.resize(fMaxThreads);
+        fInputIteratorsList.resize(fMaxThreads+1);
+
+        // Activate stats collecting before CS spawns threads.
+        if (traceOn()) dlTimes.setFirstReadTime();
+
+        // *DRRTUY Make this block conditional
```
pleblanc1976, line 260: is telemetry conditional anywhere else?  I could vote that we should not make it conditional.  it's not that expensive and potentially very useful in the field.

dbcon/joblist/tupleannexstep.cpp
```diff
@@ -213,17 +248,58 @@ void TupleAnnexStep::run()
...
+        StepTeleStats sts;
+        sts.query_uuid = fQueryUuid;
+        sts.step_uuid = fStepUuid;
+        sts.msg_type = StepTeleStats::ST_START;
+        sts.total_units_of_work = 1;
+        postStepStartTele(sts);
+
+        for(uint32_t id = 1; id <= fMaxThreads; id++)
+        {
+            fInputIteratorsList[id] = fInputDL->getIterator();
```
pleblanc1976, line 270: what does the 0 element signify in this list?

dbcon/joblist/tupleannexstep.cpp
```diff
@@ -459,7 +540,7 @@ void TupleAnnexStep::executeNoOrderByWithDistinct()
                         fJobList->abortOnLimit((JobStep*) this);
                     }
 
-                    if (UNLIKELY(fRowGroupOut.getRowCount() >= 8192))
+                    if (UNLIKELY(fRowGroupOut.getRowCount() >= rowgroup::rgCommonSize))
```
pleblanc1976, line 543: I like this mod.

dbcon/joblist/tupleannexstep.cpp
```diff
@@ -596,6 +677,579 @@ void TupleAnnexStep::executeWithOrderBy()
...
+            else
+                fConstant->fillInConstants(fRowIn, fRowOut);
+            rowSize = fRowOut.getSize();
+            fRowGroupOut.incRowCount();
+            fRowOut.nextRow(rowSize);
+            finalPQ.pop();
+            count++;
+        }
+    }
+    catch (const std::exception& ex)
```
pleblanc1976: does anything in the try block throw?  if not we can safely get rid of the try-catch stuff.

pleblanc1976: actually, there are a lot of try-catch blocks in this patch where nothing gets thrown (afaik), which could be eliminated.

dbcon/joblist/tupleannexstep.cpp
```diff
@@ -596,6 +677,579 @@ void TupleAnnexStep::executeWithOrderBy()
...
+                    break;
+
+                if (currentPQ.capacity()-currentPQ.size() < doubleRGSize)
+                {
+                    currentPQ.reserve(QUEUE_RESERVE_SIZE);
+                }
+
+                rg.setData(&rgDataIn);
+                rg.getRow(0, &r);
+                rowSize = r.getSize();
```
pleblanc1976: does this need to be fetched for every RG?  I think the data coming from a single datalist has a single format.  Did you find that not to be the case?

dbcon/joblist/tupleannexstep.cpp
```diff
@@ -596,6 +677,579 @@ void TupleAnnexStep::executeWithOrderBy()
...
+                rowCount = rg.getRowCount();
+
+                for (uint64_t i = 0; i < rowCount; ++i)
+                {
+                    limOrderBy->processRow(r);
+                    r.nextRow(rowSize);
+               }
+            }
+            
+            // *DRRTUY Implement a method to skip elements in FIFO
```
pleblanc1976, line 1145: what's the use case for skipping elements?

dbcon/joblist/tupleannexstep.h
```diff
@@ -21,8 +22,11 @@
...
 
-#include "jobstep.h"
+#include <queue>
+#include <boost/thread/thread.hpp>
 
+#include "jobstep.h"
+#include "limitedorderby.h"
 
 // forward reference
 namespace fucexp
```
pleblanc1976: that's a typo isn't it?  Supposed to be 'funcexp'?

dbcon/mysql/ha_calpont_execplan.cpp
```diff
@@ -7872,6 +7867,8 @@ int getSelectPlan(gp_walk_info& gwi, SELECT_LEX& select_lex, SCSEP& csep, bool i
                 {
                     Item_int* select = (Item_int*)select_lex.master_unit()->global_parameters()->select_limit;
                     csep->limitNum(select->val_int());
+                    // MCOL-894 Activate parallel ORDER BY
+                    csep->orderByThreads(gwi.thd->variables.infinidb_orderby_threads);
```
pleblanc1976, line 7871: This may be something we can reasonably hide from the user to not create more vars or configurables.  Depending on the characteristics of the alg, a good default would be 1-2 * num_cores.

utils/windowfunction/idborderby.cpp
```diff
@@ -78,29 +131,109 @@ int IntCompare::operator()(IdbCompare* l, Row::Pointer r1, Row::Pointer r2)
...
+            ret = -fSpec.fNf;
+    }
+    else
+    {
+        if (v1 > v2)
+            ret = fSpec.fAsc;
+        else if (v1 < v2)
+            ret = -fSpec.fAsc;
+    }
+
```
pleblanc1976, line 186: random thought.  for conciseness & modularity, we could template the comparison code, or since the vars all have the same names, it could be a macro.

pleblanc1976 (changes requested):
I only found a couple potential problems.  Would like them addressed/clarified before approving.  For the next PR I would suggest more in-line code comments so we can more easily understand the code.

This patch is large, so correctness will have to rely on your testing.  If you haven't already done so, run your unit tests with valgrind to verify memory accesses esp down error paths, and use gdb to make sure no threads are leaked.

utils/windowfunction/idborderby.cpp
drrtuy: Fixed

dbcon/joblist/jlf_tuplejoblist.cpp
drrtuy: Yes, tested both paths in UT and functional tests in the regression.

dbcon/joblist/jlf_tuplejoblist.cpp
drrtuy: The original code explicitly assigns only 1 consumer to this DL while this if block asigns orderByThreads only if CS has sorting in the query. Sorting is the only operation in TNS that could potentially have > 1 consumer.

dbcon/joblist/tupleannexstep.h
LinuxJedi: It is an old typo and appears in two different files. Whilst not relevant to this PR we should probably kill it later.

dbcon/joblist/tupleannexstep.cpp
drrtuy: Good point. The change will make reading the code easier.

dbcon/joblist/tupleannexstep.cpp
drrtuy: Yes, to avoid vector idx-es calculation later having indices as natural number starting with 1.

dbcon/joblist/tupleannexstep.cpp
drrtuy: Yes, it is. 
I would say we want to introduce changes into our runtime analyze framework. The "must have" will be execution plan w/o actually running a query. The change you suggest may be a part of this frameworks change.

dbcon/joblist/tupleannexstep.cpp
drrtuy: Nothing. I don't like to calculate idx-es prefering to have them as natural numbers.

dbcon/joblist/tupleannexstep.cpp
drrtuy: Yeah, it hangs in the air for a long time.

dbcon/joblist/tupleannexstep.cpp
drrtuy: Good point rowSize can be calculated only once.

dbcon/joblist/tupleannexstep.cpp
drrtuy: Very good question. Here is an example 1st phase sorting threads only takes x-th RGData from the DL and for this I compare received RGData% number of threads with thread id and skip all that produces non zero in this expression. With suggested change I only need to subscribe for the x-th RGData initially and later use more() without additional checks.
A sidethought. We also need a DL that can dynamicly change a number of consumers/producers to allow us to add processing threads on the fly. What do you think about this?

dbcon/joblist/tupleannexstep.h
drrtuy: Good catch.

dbcon/mysql/ha_calpont_execplan.cpp
drrtuy: When a user sets it to one CS uses original single-thread mode. This mode will be more efficient when sorting huge results sets w/o reasonable LIMIT(<1000000). One has to set the session variable to 1 to use original single-thread sorting.
We can eliminate this knob using three approaces:
1) Make the 2nd sorting phase parallel. Then parallel sorting will be faster under real life workloads.
2) When CS will have a statistics that allow us to predict result set to deduce that and automaticaly pick the proper sorting mode
3) If we have RowgroupDLs that can add subscribers on the fly we can add threads as need them thus not using parallel sorting for small datasets.

utils/windowfunction/idborderby.cpp
drrtuy: Good point. This would reduce the size of the code. I will add this suggestion into the roadmap of the parallel order by.

dbcon/joblist/tupleannexstep.h
drrtuy: I think we don't need it here. I'll check it.

dbcon/joblist/joblistfactory.cpp
drrtuy: We might need it in 1.2 to preserve the behavior. In 1.4 I removed this check though.

dbcon/joblist/tupleannexstep.cpp
drrtuy: fOrderByList is a vector of pointers thus we loose one uint64_t here.

dbcon/joblist/jlf_tuplejoblist.cpp
drrtuy: [Here](https://github.com/mariadb-corporation/mariadb-columnstore-regression-test/pull/156/files) is the PR into regression.

pleblanc1976: :+1:

dbcon/joblist/joblistfactory.cpp
pleblanc1976: Is removing it the correct thing to do given that the code in that block seems to depend on there being a limit?

dbcon/joblist/tupleannexstep.cpp
pleblanc1976: I don't quite follow, but that's OK.  As long as you've looked at it again and think it's correct, that's good enough for me.  Could you add a code comment to clarify?

dbcon/joblist/tupleannexstep.cpp
pleblanc1976: The # of consumers in the fifo might be better understood as the # of reads an element should have before fifo gets rid of it.  If any thread can process any element (which is usually the case), then you can safely set the # of consumers to 1, as each element only needs to be read once.  If for some reason you have an alg that requires X threads to see every element (I can't think of any), then you would set the # of consumers to X.

Short version, I think you want to set numConsumers to 1, and you likely don't need to write something to skip elements or dynamically change numConsumers.

drrtuy:
I've just pushed commit that follows up the review.

dbcon/joblist/tupleannexstep.h
pleblanc1976: yeah we must not need it at all, b/c right now it references nothing.  Whether to remove it here or not, eh, who cares.  :)

dbcon/mysql/ha_calpont_execplan.cpp
pleblanc1976: We can debate whether this should be exposed or automatic 'offline'.  There is no problem with exposing it for now, just something to think about.

utils/windowfunction/idborderby.cpp
pleblanc1976: ok if you've flagged it as a future improvement, good enough for now.

dbcon/joblist/tupleannexstep.cpp
drrtuy: There are places where CS has to access fOrderByList[thread id]. I don't like this to be fOrderByList[thread id - 1] given that thread id are natural numbers starting from 1.

dbcon/joblist/joblistfactory.cpp
pleblanc1976: (slack convo) Short version: addOrderByAndLimit() is where the new code is, and it can handle it when there is no limit.  For the future, maybe change the name of the fcn.

pleblanc1976 (approved):

LinuxJedi (approved):

PR #918, opened 2019-10-31
MCOL-3563: fix compiler warnings / type errors.

benthompson15:
A previous commit made posix::read changed return value
from bool to int. However some callers still expected
bool and compared a bool<0 which is always false.

### DIFF
--- a/storage-manager/src/unit_tests.cpp
+++ b/storage-manager/src/unit_tests.cpp
@@ -33,6 +33,7 @@
 #include "S3Storage.h"
 #include "Utilities.h"
 #include "Synchronizer.h"
+#include "ProcessTask.h"
 
 #include <iostream>
 #include <stdlib.h>
@@ -189,7 +190,7 @@ void makeConnection()
     t.join();
 }
     
-bool opentask(bool connectionTest=false)
+bool opentask(bool connectionTest=false, bool errorHandle=false)
 {
     // going to rely on msgs being smaller than the buffer here
 	int err=0;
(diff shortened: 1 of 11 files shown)


pleblanc1976 (changes requested):
- opentask and pingtask need the bool->int fix
 - not sure the unit test is doing what it's supposed to do, take another look at that please.  At minimum add some english to describe why it's reading from closed sockets, etc.  If it works as intended, nobody would figure this out.  :D

LinuxJedi (approved):

storage-manager/src/unit_tests.cpp
```diff
@@ -209,20 +210,39 @@ bool opentask(bool connectionTest=false)
...
-    	err = ::recv(sessionSock, buf, 1024, MSG_DONTWAIT);
-    	assert(err == -1);
+        // make sure the thread is waiting for the rest of the data
+        // then kill the connection. This will trigger the task thread
+        // to exit on an error handling path
+        sleep(1);
+        close(sessionSock);
+        close(clientSock);
+        err = ::recv(sessionSock, buf, 1024, MSG_DONTWAIT);
+        assert(err == -1);
```
pleblanc1976, line 240: still confused about what we're doing with the close-recv bit.  Closing a socket and then reading from it will always return -1, doesn't say anything about the SM code.

IIRC, we're verifying correct behavior by 'manually' watching a log file.  If we don't want to automate that part (too much trouble is a totally valid reason right now), could you add a printout here telling the user what to look for?

storage-manager/src/unit_tests.cpp
```diff
@@ -509,19 +532,38 @@ void unlinktask(bool connectionTest=false)
...
```
pleblanc1976, line 561: same comment as above

pleblanc1976: Looks like the mods to the unit test are mainly this, which is only checking that reading from a closed socket returns -1 AFAICT.

pleblanc1976 (changes requested):
Looks OK except for the unit test changes.

pleblanc1976 (approved):

PR #919, opened 2019-10-31
MCOL-2276 "Unable to insert certain row." Fix

jmrojas2332:
When we initialize m_freeSpace we subtract m_totalHdrBytes from it so the comparison 

`(m_freeSpace >= (size + m_totalHdrBytes)`

is incorrect and was actually causing the bug. This bug occurs whenever we're looking for a block of size at most 14 bytes (m_totalHdrBytes) smaller than 8176 bytes (the block boundary size for TEXT) .


LinuxJedi (approved):
Great find!

PR #923, opened 2019-11-05
Mcol 3448 fix RowItem compares

davidhall586:
RowItem compares, such as (c1,c2) IN ((v11, v21), (v12, v22)) gave bad answers due to a typo getting the wrong datatype.


drrtuy (changes requested):
Good catch.
Could I ask you to reduce a number of commits from 3 to 1? Could you cherry-pick the relevant commit on top of the current develop-1.2?

PR #927, opened 2019-11-08
MCOL-3605 Blocks are not being completely filled in dictionaries

jmrojas2332:
m_totalHdrBytes is already subtracted from m_freeSpace upon initialization, so subtracting m_totalHdrBytes when comparing if signature size fits in a block is counting the header twice and allowing blocks to possibly have up to 12 free bytes when filling accross multiple blocks. We instead should only be checking to make sure the 2 byte offset fits.

### DIFF
--- a/writeengine/dictionary/we_dctnry.cpp
+++ b/writeengine/dictionary/we_dctnry.cpp
@@ -663,9 +663,9 @@ int Dctnry::insertDctnry2(Signature& sig)
 
     while (sig.size > 0)
     {
-        if (sig.size > (m_freeSpace - m_totalHdrBytes))
+        if (sig.size > (m_freeSpace - HDR_UNIT_SIZE))
         {
-            write_size = (m_freeSpace - m_totalHdrBytes);
+            write_size = (m_freeSpace - HDR_UNIT_SIZE);
         }
         else
         {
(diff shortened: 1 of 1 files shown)


writeengine/dictionary/we_dctnry.cpp
```diff
@@ -663,9 +663,9 @@ int Dctnry::insertDctnry2(Signature& sig)
 
     while (sig.size > 0)
     {
-        if (sig.size > (m_freeSpace - m_totalHdrBytes))
+        if (sig.size > (m_freeSpace - HDR_UNIT_SIZE))
```
jmrojas2332, line 666: HDR_UNIT_SIZE is 2 bytes. Only need to make sure offset fits since m_freeSpace has already taken the header into account

writeengine/dictionary/we_dctnry.cpp
```diff
@@ -887,8 +887,8 @@ int Dctnry::insertDctnry(const char* buf,
         //...String not found in cache, so proceed.
         //   If room is available in current block then insert into block.
         // @bug 3960: Add MAX_OP_COUNT check to handle case after bulk rollback
-        if ( ((totalUseSize <= m_freeSpace) ||
-                ((curSig.size > 8176) && (m_freeSpace > m_totalHdrBytes))) &&
+        if ( ((totalUseSize <= m_freeSpace - HDR_UNIT_SIZE) ||
```
jmrojas2332, line 890: we need to make sure offset fits here

writeengine/dictionary/we_dctnry.cpp
```diff
@@ -1111,14 +1111,14 @@ int Dctnry::insertDctnry(const int& sgnature_size,
     for (i = m_lastFbo; i < m_numBlocks; i++)
     {
         // @bug 3960: Add MAX_OP_COUNT check to handle case after bulk rollback
-        if ( ((m_freeSpace >= size) ||
-                ((size > 8176) && (m_freeSpace > m_totalHdrBytes))) &&
+        if ( ((m_freeSpace - HDR_UNIT_SIZE >= size) ||
```
jmrojas2332, line 1114: same as above: offset needs to fit

writeengine/dictionary/we_dctnry.cpp
```diff
@@ -1111,14 +1111,14 @@ int Dctnry::insertDctnry(const int& sgnature_size,
...
+                ((size > 8176) && (m_freeSpace > HDR_UNIT_SIZE))) &&
                 (m_curOp    <  (MAX_OP_COUNT - 1)) )
         {
             // found the perfect block; signature size fit in this block
-            if (size > (m_freeSpace - m_totalHdrBytes))
+            if (size > (m_freeSpace - HDR_UNIT_SIZE))
```
jmrojas2332, line 1119: only need to make sure offset fits since m_freeSpace has already taken the header into account

LinuxJedi (approved):

PR #929, opened 2019-11-09
MCOL-3606 Make ColumnStore use generic paths

LinuxJedi:
ColumnStore now uses standard bin/lib paths for pretty much everything.
Data path is now hard-coded to /var/lib/columnstore.

This patch also:

* Removes v1 decompression
* Removes a bunch of unneeded files
* Removes COLUMNSTORE_INSTALL_DIR / $INSTALLDIR
* Makes my.cnf.d work for all platforms (MCOL-3558)
* Changes configcpp to use recursive mutex (fixes possible config write deadlock)
* Fixes MCOL-3599 Fix regr functions, The library was installed in the wrong location
* Fixes a bunch of Ubuntu packaging issues
* Changes the binary names of several of the executables so as not to
clash with potential executables from other packages


LinuxJedi:
Still some testing to do before this should be reviewed

LinuxJedi:
Testing done, ready for review

oam/post/CMakeLists.txt
```diff
@@ -1,3 +1,7 @@
+configure_file("${CMAKE_CURRENT_SOURCE_DIR}/mcstest-001.sh.in" "${CMAKE_CURRENT_SOURCE_DIR}/mcstest-001.sh" @ONLY)
```
drrtuy, line 1: Do we really need these files?

LinuxJedi: Apparently, yes. They appear to be used by ProcMon to make sure the node is really up.

ProcMon calls startupTests.sh which calls mcstest-*.sh

utils/configcpp/configcpp.h
```diff
@@ -230,8 +230,7 @@ class Config
     xmlDocPtr fDoc;
     const std::string fConfigFile;
     time_t fMtime;
-    mutable boost::mutex fLock;
-    const std::string fInstallDir;
+    mutable boost::recursive_mutex fLock;
```
drrtuy, line 233: Don't we want to make it std::recursive_mutex?

LinuxJedi: No, because this is used for scoped locks which don't come into std until C++17

drrtuy (approved):

PR #930, opened 2019-11-12
MCOL-3607 Add long double support to math functions

davidhall586:


utils/funcexp/func_math.cpp
```diff
@@ -348,6 +378,32 @@ double Func_atan::getDoubleVal(Row& row,
         }
         break;
 
+        case execplan::CalpontSystemCatalog::LONGDOUBLE:
+        {
+            // null value is indicated by isNull
+            long double value = parm[0]->data()->getLongDoubleVal(row, isNull);
+
+            if (isNull || (value < -1.0 || value > 1.0))
```
jmrojas2332: does this conditional contain a typo, perhaps? should it not match line 358 above?

jmrojas2332: Val can be less than -1 and greater than 1 when taking the atan(val) if i'm not mistaken

utils/funcexp/func_math.cpp
```diff
@@ -1316,6 +1514,21 @@ double Func_tan::getDoubleVal(Row& row,
         }
         break;
 
+        case execplan::CalpontSystemCatalog::LONGDOUBLE:
+        {
+            // null value is indicated by NaN
+            long double value = parm[0]->data()->getLongDoubleVal(row, isNull);
+
+            if (isNull || value <= 0.0)
```
jmrojas2332: This may be a typo. I believe it should match line 1507. We could take the tangent of a negative

jmrojas2332 (commented):
Made a couple of comments on this regarding possibly a couple of typos.

utils/funcexp/func_math.cpp
```diff
@@ -491,6 +547,21 @@ double Func_cos::getDoubleVal(Row& row,
         }
         break;
 
+        case execplan::CalpontSystemCatalog::LONGDOUBLE:
+        {
+            // null value is indicated by isNull
+            long double value = parm[0]->data()->getLongDoubleVal(row, isNull);
+
+            if (isNull || (value < -1.0 || value > 1.0))
```
jmrojas2332: I believe this conditional should match line 540. Domain of cosine can be less than 1 and greater than 1

utils/funcexp/func_math.cpp
LinuxJedi: −π/2 and +π/2 radians are the limits.

davidhall586: You are correct. Typo

utils/funcexp/func_math.cpp
davidhall586: Correct.

utils/funcexp/func_math.cpp
davidhall586: Also correct

davidhall586 (commented):
Corrected typos

jmrojas2332 (approved):
Looks good.

PR #944, opened 2019-11-19
MCOL-2072 QueryStats creates a mysql_use_result() warning

jmrojas2332:
Added a flag to LibMysql::run to be used to suppress warning when no result is expected. The flag is a default parameter called resultExpected which is set to true by default but is set to false when called from querystats.

### DIFF
--- a/utils/libmysql_client/libmysql_client.cpp
+++ b/utils/libmysql_client/libmysql_client.cpp
@@ -96,7 +96,7 @@ int LibMySQL::init(const char* h, unsigned int p, const char* u, const char* w, 
 }
 
 
-int LibMySQL::run(const char* query)
+int LibMySQL::run(const char* query, bool resultExpected)
 {
     int ret = 0;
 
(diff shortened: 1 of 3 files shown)


drrtuy (approved):
The patch looks good for me but I don't like that BB failed to build it. It seem that BB fails to call mcsadmin shutdown.
Plz take a look that all BB builds fails only on 'Shutdown complete' stop and merge it.

benthompson15:
> The patch looks good for me but I don't like that BB failed to build it. It seem that BB fails to call mcsadmin shutdown.
> Plz take a look that all BB builds fails only on 'Shutdown complete' stop and merge it.

This is my fault. I broke after making changes for 1.4 because of path changes. please ignore this mcsadmin shutdown error. If this is not merged when fixed I will rerun these builds.

drrtuy:
Much appreciated.

ср, 20 нояб. 2019 г., 22:33 benthompson15 <notifications@github.com>:

PR #946, opened 2019-11-20
MCOL-3577

benthompson15:


pleblanc1976:
Gave Ben feedback in person.  As is it will only sync the node it's running on, and on the SM side, the original syncNow(prefix) can deadlock, and should just call his new syncNow() (everything) fcn for now.

pleblanc1976 (approved):
Hard to say whether those procmon/mgr are right.  Looks OK, but will have to rely on your testing.

PR #947, opened 2019-11-21
MCOL-3585 update some defaults in config file

jmrojas2332:
Updated PmMaxMemorySmall, FifoSizeLargeSize, and FifoSize settings from Columnstore.xml and Columnstore.xml.singleserver.

Decreasing FifoSize and FifoSizeLargeSide did not cause any performance hit. I ran regression suite test001 to confirm this w/ and w/o new settings.

RowAggrThreads and RowAggrBuckets are already being automated. I just updated what the values were being set to inside resourcemanager.cpp Now automating MaxOutstandingRequests as well.


pleblanc1976 (approved):
LGTM

PR #949, opened 2019-11-21
Implement bit_count() distributed function and handle hexadecimal string literals in buildReturnedColumn().

tntnatbry:
Fixes working_tpch1/qa_fe_cnxFunctions/bug4703.sql.

### DIFF
--- a/utils/funcexp/func_bitwise.cpp
+++ b/utils/funcexp/func_bitwise.cpp
@@ -359,5 +359,49 @@ int64_t Func_bitxor::getIntVal(Row& row,
 }
 
 
+//
+// BIT COUNT
+//
+
+
+CalpontSystemCatalog::ColType Func_bit_count::operationType( FunctionParm& fp, CalpontSystemCatalog::ColType& resultType )
+{
+    return resultType;
+}
+
+int64_t Func_bit_count::getIntVal(Row& row,
+                                  FunctionParm& parm,
+                                  bool& isNull,
+                                  CalpontSystemCatalog::ColType& operationColType)
+{
+    if ( parm.size() != 1 )
+    {
+        isNull = true;
+        return 0;
+    }
+
+    uint64_t val = 0;
+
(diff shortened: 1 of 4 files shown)


tntnatbry:
I will add a test case for bit_count() in the regression suite, please don't merge yet.

utils/funcexp/func_bitwise.cpp
```diff
@@ -359,5 +359,47 @@ int64_t Func_bitxor::getIntVal(Row& row,
...
+        return 0;
+    }
+
+    uint64_t val = 0;
+
+    if (!getUIntValFromParm(row, parm[0], val, isNull, fTimeZone))
+    {
+        std::ostringstream oss;
+        oss << "bit_count: datatype of " << execplan::colDataTypeToString(operationColType.colDataType);
+        throw logging::IDBExcept(oss.str(), ERR_DATATYPE_NOT_SUPPORT);
```
drrtuy, line 389: I wonder if this throw finishes the query?

utils/funcexp/func_bitwise.cpp
```diff
@@ -359,5 +359,47 @@ int64_t Func_bitxor::getIntVal(Row& row,
...
+    }
+
+    int64_t count = 0;
+
+    while ( val )
+    {
+        val &= val - 1;
```
drrtuy: I wonder if the Hacker's delight reciep on 'counting 1-bits in a word' is better?

utils/funcexp/func_bitwise.cpp
tntnatbry: The query is killed in case an exception is thrown here.

utils/funcexp/func_bitwise.cpp
tntnatbry: Hacker's Delight algo has much better performance compared to the original algo used, so the PR is updated with a new implementation.

LinuxJedi (approved):

PR #950, opened 2019-11-21
Threaded um join squashed

pleblanc1976:
Made UM hash table construction use (# cores) concurrent jobs for the main use case.  The path disk-join uses is still single threaded; did not want to get into that yet.  So, both versions of the code are still there.


oam/etc/Columnstore.xml.singleserver
```diff
@@ -241,6 +241,7 @@
 		<TempSaveSize>128M</TempSaveSize> <!-- default SWSDL max element save size -->
 		<WaitPeriod>10</WaitPeriod> <!-- in seconds -->
 		<TempFileDir>/tmp/columnstore_tmp_files</TempFileDir>
+		<CalpontHome>$INSTALLDIR</CalpontHome>
```
pleblanc1976: This was introduced in a merge, will take it out.

dbcon/joblist/tuplehashjoin.cpp
```diff
@@ -200,140 +202,237 @@ void TupleHashJoinStep::join()
...
-    JoinType jt;
-    RowGroupDL* smallDL;
-    uint32_t smallIt;
-    RowGroup smallRG;
-    boost::shared_ptr<TupleJoiner> joiner;
+    boost::shared_ptr<TupleJoiner> joiner = joiners[index];
+    ssize_t memBefore = 0, memAfter = 0;
+    bool gotMem;
+
+    boost::unique_lock<boost::mutex> scoped(memTrackMutex);
```
drrtuy, line 214: Do we really need boost for both shared and unique ptr types?

dbcon/joblist/tuplehashjoin.cpp
```diff
@@ -116,11 +116,13 @@ TupleHashJoinStep::TupleHashJoinStep(const JobInfo& jobInfo) :
     else
         allowDJS = false;
 
+    numCores = sysconf(_SC_NPROCESSORS_ONLN);
+    if (numCores <= 0)
+        numCores = 8;
```
LinuxJedi, line 121: You could have also got this information from resourceManager which does exactly the same thing but also appears to support CGroups

dbcon/joblist/tuplehashjoin.cpp
```diff
@@ -200,140 +202,237 @@ void TupleHashJoinStep::join()
...
+        boost::posix_time::microsec_clock::universal_time();
+
+    stopMemTracking = false;
+    uint64_t jobs[numCores];
+    uint64_t memMonitor = jobstepThreadPool.invoke([this, index] { this->trackMem(index); });
+    // starting 1 thread when in PM mode, since it's only inserting into a
+    // vector of rows.  The rest will be started when converted to UM mode.
+    if (joiner->inUM())
+        for (int i = 0; i < numCores; i++)
+            jobs[i] = jobstepThreadPool.invoke([this, i, index, &jobs] { this->smallRunnerFcn(index, i, jobs); });
```
drrtuy, line 299: Just to highlight. The first lambda in the code :)

dbcon/joblist/tuplehashjoin.cpp
```diff
@@ -200,140 +202,237 @@ void TupleHashJoinStep::join()
...
+    stopMemTracking = true;
+    memTrackDone.notify_one();
+    memTrackMutex.unlock();
+    jobstepThreadPool.join(memMonitor);
+
+    /* If there was an error or an abort, drain the input DL,
+        do endOfInput on the output */
+    if (cancelled())
+    {
+//		cout << "HJ stopping... status is " << status() << endl;
```
drrtuy, line 320: I would remove all needless debugging instructions.

dbcon/joblist/tuplehashjoin.cpp
```diff
@@ -200,140 +202,237 @@ void TupleHashJoinStep::join()
...
+        if (largeBPS)
+            largeBPS->abort();
+
+        bool more = true;
+        RGData oneRG;
+        while (more)
+            more = smallDLs[index]->next(smallIts[index], &oneRG);
+    }
+    //joiner->doneInserting();
```
drrtuy: Or unused code.

dbcon/joblist/tuplehashjoin.cpp
```diff
@@ -200,140 +202,237 @@ void TupleHashJoinStep::join()
...
+
+    end_time = boost::posix_time::microsec_clock::universal_time();
+    if (!(fSessionId & 0x80000000))
+        cout << "hash table construction time = " << end_time - start_time <<
```
drrtuy, line 333: It looks like this line should be removed.

dbcon/joblist/tuplehashjoin.cpp
```diff
@@ -200,140 +202,237 @@ void TupleHashJoinStep::join()
...
-
             if (smallRG.getRowCount() == 0)
                 goto next;
 
-            smallRG.getRow(0, &r);
-
-            memUseBefore = joiner->getMemUsage() + rgDataSize;
-
             // TupleHJ owns the row memory
+            utils::getSpinlock(rgdLock);
```
drrtuy, line 398: It looks like a presumtion that no thread could be slowdown while holding this lock. Could condition_variable be a replacement?

LinuxJedi (approved):

dbcon/joblist/tuplehashjoin.cpp
```diff
@@ -200,140 +202,237 @@ void TupleHashJoinStep::join()
...
+            {
+                joiner->setInUM(rgData[index]);
+                for (int i = 1; i < numCores; i++)
+                    jobs[i] = jobstepThreadPool.invoke([this, i, index, jobs]
+                        { this->smallRunnerFcn(index, i, jobs); });
+            }
 next:
-//  		cout << "inserted one rg into the joiner, rowcount = " <<
-//  			smallRG.getRowCount() << endl;
+            dlMutex.lock();
```
drrtuy, line 436: That is what you were referering to comparing consequences of internal and external locking for FIFO?

drrtuy: Why do you need to lock given that smallDL is a dedicated FIFO serving the thread?

dbcon/joblist/tuplehashjoin.h
```diff
@@ -614,10 +611,19 @@ class TupleHashJoinStep : public JobStep, public TupleDeliveryStep
     std::vector<boost::shared_ptr<joiner::TupleJoiner> > tbpsJoiners;
     std::vector<boost::shared_ptr<joiner::TupleJoiner> > djsJoiners;
     std::vector<int> djsJoinerMap;
-    boost::scoped_array<uint64_t> memUsedByEachJoin;
+    boost::scoped_array<ssize_t> memUsedByEachJoin;
```
drrtuy, line 614: Why not use here unique_ptr for the vector? It [seems](http://boost.2283326.n4.nabble.com/shared-array-Why-not-in-C-11-td4649418.html) that scoped_array never made into standart b/c there was no reason for this specific structure.

utils/common/fixedallocator.h
```diff
@@ -89,10 +97,46 @@ class FixedAllocator
...
+    bool useLock;
+    std::atomic<bool> lock;
 };
 
+inline void* FixedAllocator::allocate()
+{
+    void* ret;
+
+    if (useLock)
+        getSpinlock(lock);
```
drrtuy, line 109: This spinlock potentially can be very painfull.

utils/threadpool/threadpool.cpp
```diff
@@ -321,6 +322,7 @@ uint64_t ThreadPool::invoke(const Functor_T& threadfunc)
 
 void ThreadPool::beginThread() throw()
 {
+    utils::setThreadName("Idle");
```
drrtuy, line 325: Very good idea.

utils/joiner/tuplejoiner.h
```diff
@@ -344,13 +361,13 @@ class TupleJoiner
...
     TupleJoiner& operator=(const TupleJoiner&);
+    void getBucketCount();
 
     rowgroup::RGData smallNullMemory;
 
-
-    boost::scoped_ptr<hash_t> h;  // used for UM joins on ints
-    boost::scoped_ptr<sthash_t> sth;  // used for UM join on ints where the backing table uses a string table
-    boost::scoped_ptr<ldhash_t> ld;  // used for UM join on long double
+    boost::scoped_array<boost::scoped_ptr<hash_t> > h;  // used for UM joins on ints
```
drrtuy, line 368: This nest of boost entities attracts my attention a lot. I wonder if std::unique_ptr<std::vector<std::unique_ptr<hash_t> > > will do the trick?

utils/joiner/tuplejoiner.cpp
```diff
@@ -42,30 +43,49 @@ TupleJoiner::TupleJoiner(
...
+    getBucketCount();
+    m_bucketLocks.reset(new boost::mutex[bucketCount]);
+
     if (smallRG.getColTypes()[smallJoinColumn] == CalpontSystemCatalog::LONGDOUBLE)
     {
-        STLPoolAllocator<pair<const long double, Row::Pointer> > alloc;
-        _pool = alloc.getPoolAllocator();
-
-        ld.reset(new ldhash_t(10, hasher(), ldhash_t::key_equal(), alloc));
+        ld.reset(new boost::scoped_ptr<ldhash_t>[bucketCount]);
```
drrtuy, line 59: The same anti-boost story. Could we replace this with std::unique_ptr?

utils/joiner/tuplejoiner.cpp
```diff
@@ -199,6 +235,166 @@ bool TupleJoiner::operator<(const TupleJoiner& tj) const
...
+    bucketCount = (numCores == 1 ? 1 : (1 << (32 - __builtin_clz(numCores - 1))));
+    bucketMask = bucketCount - 1;
+}
+
+template<typename buckets_t, typename hash_table_t>
+void TupleJoiner::bucketsToTables(buckets_t *buckets, hash_table_t *tables)
+{
+    uint i;
+
+    bool done = false, wasProductive;
```
drrtuy, line 254: I like the name of the variable - wasProductive :)

utils/joiner/tuplejoiner.cpp
```diff
@@ -199,6 +235,166 @@ bool TupleJoiner::operator<(const TupleJoiner& tj) const
...
+                continue;
+            }
+            for (auto &element : buckets[i])
+                tables[i]->insert(element);
+            m_bucketLocks[i].unlock();
+            wasProductive = true;
+            buckets[i].clear();
+        }
+        if (!done && !wasProductive)
+            ::usleep(1000 * numCores);
```
drrtuy, line 276: This potentially leads to an alignment of separate threads with a skewed data set. A non-linear delay with a smaller constant could be of use or use ASM instruction sequence number as a random value. But the latter sounds like an overkill.

utils/joiner/tuplejoiner.cpp
```diff
@@ -212,57 +408,61 @@ void TupleJoiner::insert(Row& r, bool zeroTheRid)
...
         }
         else
         {
-            int64_t smallKey = r.getIntField(smallKeyColumns[0]);
+            int64_t smallKey;
 
+            if (!r.isUnsigned(smallKeyColumns[0]))
+                smallKey = r.getIntField(smallKeyColumns[0]);
+            else
+                smallKey = (int64_t) r.getUintField(smallKeyColumns[0]);
```
drrtuy: The key could be a negative value?

utils/joiner/tuplejoiner.cpp
```diff
@@ -516,38 +693,47 @@ void TupleJoiner::doneInserting()
...
-            ldit = ld->begin();
+            ldit = ld[bucket]->begin();
         else if (!smallRG.usesStringTable())
-            hit = h->begin();
+            hit = h[bucket]->begin();
         else
-            sthit = sth->begin();
+            sthit = sth[bucket]->begin();
 
         for (i = 0; i < rowCount; i++)
```
drrtuy, line 708: I wonder if this loop is called often and for large amounts of Rows?

dbcon/joblist/tuplehashjoin.cpp
pleblanc1976: :D  Technically no...

dbcon/joblist/tuplehashjoin.cpp
pleblanc1976: :man_dancing:

dbcon/joblist/tuplehashjoin.cpp
pleblanc1976: They've been pretty useful to me, and also serve as inline documentation.

dbcon/joblist/tuplehashjoin.cpp
pleblanc1976: yeah, taking that one out.  Missed it in the cleanup.

dbcon/joblist/tuplehashjoin.cpp
pleblanc1976: yes, debugging code that slipped through the cracks, will take it out.

dbcon/joblist/tuplehashjoin.cpp
pleblanc1976: mutex would be the replacement.  For typically very short waits like this, a spinlock is more efficient, for anything longer, mutex is more efficient b/c it frees up CPU cycles.  In this case I benchmarked it both ways and couldn't observe a difference.  Went with the one that doesn't involve an interrupt.

dbcon/joblist/tuplehashjoin.cpp
pleblanc1976: will make that change.

dbcon/joblist/tuplehashjoin.cpp
pleblanc1976: There are potentially numCores threads running this code at once. 

Yeah that's what I was talking about.  fifo::next() doesn't use its lock in such a way that multiple threads could use the same iterator at once -> need to lock externally to do that.

dbcon/joblist/tuplehashjoin.h
pleblanc1976: It could be a plain vector.  I left it how it was.  Not that important.

utils/common/fixedallocator.h
pleblanc1976: True.  When it does allocate add'l mem, threads could pile up there.  For the typical case when it's not allocating, much less risk.  I found it was better to copy *allocator instances than to have threads share one, but decided to leave the mods in there b/c they're harmless with the defaults.

utils/threadpool/threadpool.cpp
pleblanc1976: :+1:

utils/joiner/tuplejoiner.h
pleblanc1976: No need for the first unique_ptr even.  Yes it would do the trick, but it's not necessary right now.

utils/joiner/tuplejoiner.cpp
pleblanc1976: Heh, I don't share your zeal for eliminating boost ptrs. :)  That said, I bet we could use sed to do mass substitutions like boost::scoped_ptr -> std::unique_ptr, and it'll mostly work.

utils/joiner/tuplejoiner.cpp
pleblanc1976: :dancer:

utils/joiner/tuplejoiner.cpp
pleblanc1976: I think I see what you're getting at.  There's already a good deal of randomness in the system though.  But true, threads could consistently block on a few # of a buckets or even a single bucket, and bucket(s) may not be fully utilized b/c of the sleeps.  I'm going to test performance when inserting millions of duplicates to see if there's a regression there.  Will mod if there's a substantial performance loss.

utils/joiner/tuplejoiner.cpp
pleblanc1976: Yeah this was one of the uglier mods i made for this.  At the end of small-table construction, it will iterate over all of the rows in the small table, stopping when it sees a certain # of unique values.  I can't say how often it scans the whole thing in real use cases.

utils/joiner/tuplejoiner.cpp
drrtuy: I might use inapproriate term. There could be a potential syncronization of threads with a skewed data set so they will gather together waiting for the lock.

utils/joiner/tuplejoiner.cpp
drrtuy: What is the value for this #?

drrtuy (approved):

utils/joiner/tuplejoiner.cpp
pleblanc1976: The total # of rows on the small side as returned by TupleJoiner::size().

PR #953, opened 2019-11-25
Selected optimizer rewrites addition.

drrtuy:
…y_joins.

    Disabled 4th if block in buildOuterJoin to handle non-optimized MDB query
    structures.

    Broke getSelectPlan into pieces: processFrom, processWhere.

MCOL-3593 UNION processing depends on two flags isUnion that comes as
arg of getSelectPlan and unionSel that is a local variable in
getSelectPlan. Modularization of getSelectPlan broke the mechanizm.
This patch is supposed to partially fix it.

MCOL-3593 Removed unused if condition from buildOuterJoin that allows
unsupported construct subquery in ON expression.
Fixed an improper if condition that ignors tableMap entries w/o condition
in external_lock thus external_lock doesn't clean up when the query
finishes.
Fixed wrong logging for queries processed in tableMode. Now rnd_init
properly sends queryText down to ExeMgr to be logged.

MCOL-3593 Unused attribute FromSubQuery::fFromSub was removed.
 getSelectPlan has been modularized into: setExecutionParams,
 processFrom, processWhere. SELECT, HAVING, GROUP BY, ORDER BY
 still lives in getSelectPlan.
Copied optimization function simplify_joins_ into our pushdown
 code to provide the plugin code with some rewrites from MDB it
 expects.
The columnstore_processing_handlers_fallback session variable
 has been removed thus CS can't fallback from SH to partial
 execution paths, e.g. DH, GBH or plugin API.

MCOL-3602 Moved MDB optimizer rewrites into a separate file.

Add SELECT_LEX::optimize_unflattened_subqueries() call to fix IN
 into EXISTS rewrite for semi-JOINs with subqueries.

disable_indices_for_CEJ() add index related hints to disable
 index access methods in Cross Engine Joins.

create_SH() now flattens JOIN that has both physical tables and
 views. This fixes most of views related tests in the regression.

### DIFF
--- a/dbcon/mysql/ha_mcs_execplan.cpp
+++ b/dbcon/mysql/ha_mcs_execplan.cpp
@@ -16,11 +16,6 @@
    Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
    MA 02110-1301, USA. */
 
-/*
- * $Id: ha_mcs_execplan.cpp 9749 2013-08-15 04:00:39Z zzhu $
- */
- 
-/** @file */
 //#define DEBUG_WALK_COND
 #include <my_config.h>
 #include <string>
@@ -61,7 +56,6 @@ using namespace logging;
 #include "ha_mcs_impl_if.h"
 #include "ha_mcs_sysvars.h"
 #include "ha_subquery.h"
-//#include "ha_view.h"
 using namespace cal_impl_if;
 
 #include "calpontselectexecutionplan.h"
@@ -132,7 +126,6 @@ public:
     gp_walk_info* fgwip;
 };
 
-//#define OUTER_JOIN_DEBUG
 namespace
 {
 string lower(string str)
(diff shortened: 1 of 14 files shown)


dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -6299,89 +6256,80 @@ int getSelectPlan(gp_walk_info& gwi, SELECT_LEX& select_lex,
...
+ * RETURNS
+ *  error id as an int
+ ***********************************************************/
+int processWhere(SELECT_LEX &select_lex,
+    gp_walk_info &gwi,
+    SCSEP &csep,
+    List<Item> &on_expr_list)
+{
+    JOIN* join = select_lex.join;
+    Item_cond* icp = 0;
```
LinuxJedi, line 6298: Minor niggle, not for now, this should probably be NULL or nullptr. I suspect some compiler or static code analyser would flag it.

LinuxJedi (approved):

PR #954, opened 2019-11-26
Enable unhex function and make minor implementation changes.

tntnatbry:
Fixes working_tpch1/qa_fe_cnxFunctions/unhex.sql.

### DIFF
--- a/utils/funcexp/func_unhex.cpp
+++ b/utils/funcexp/func_unhex.cpp
@@ -36,17 +36,24 @@ using namespace execplan;
 
 namespace
 {
-int64_t hex_to_int(char c)
+inline int hex_to_int(char c, bool& isNull)
 {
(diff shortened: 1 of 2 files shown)


utils/funcexp/func_unhex.cpp
```diff
@@ -65,48 +66,42 @@ string Func_unhex::getStrVal(rowgroup::Row& row,
...
 {
     const string& from = parm[0]->data()->getStrVal(row, isNull);
-    char* to = new char[2 + from.length() / 2];
-
-    if (!to)
-    {
-        isNull = true;
-        return "";
-    }
+    char* to = new char[2 + from.size() / 2];
```
drrtuy, line 79: I would use std::unique_ptr here. If the code before delete throws this allocation will leak.

drrtuy (approved):

PR #955, opened 2019-11-26
MCOL-3585 Enabled MCOL-498 by default for all dbroots.

drrtuy:


pleblanc1976 (commented):
I could be misinterpreting this, but these changes seem to make OFF the default not ON.

Only if it's explicitly set to ON, the dbroot gets put in s_PreallocSpace.  In IDBPolicy::PreallocSpace(), if it finds the dbroot in s_PreallocSpace, it returns false.  So 'on' -> false.  'off' or nothing -> true.

I suspect but can't say for sure that this isn't the intended behavior.

drrtuy:
I renamed the method and made appropriate changes.

LinuxJedi (approved):

pleblanc1976 (approved):
looks good, will merge

PR #957, opened 2019-12-03
MCOL 3474 Fix TIMEDIFF w/ non-temporal data types

jmrojas2332:
The following changes were made to non-temporal data types after investigating the inconsistencies when running queries using MDB and columnstore engine. New regression tests added as well to verify changes allow columnstore to match mdb.

Regression tests PR: https://github.com/mariadb-corporation/mariadb-columnstore-regression-test/pull/168

### DIFF
--- a/utils/funcexp/func_timediff.cpp
+++ b/utils/funcexp/func_timediff.cpp
@@ -45,7 +45,7 @@ namespace funcexp
 namespace helpers
 {
 
-const string timediff( int64_t time1, int64_t time2)
+const string timediff( int64_t time1, int64_t time2, bool isDateTime)
 {
     long long seconds;
     long long microseconds;
@@ -56,9 +56,9 @@ const string timediff( int64_t time1, int64_t time2)
         l_sign = -l_sign;
 
     if ( time1 > time2 )
-        helpers::calc_time_diff(time1, time2, l_sign, &seconds, &microseconds);
+        helpers::calc_time_diff(time1, time2, l_sign, &seconds, &microseconds, isDateTime);
     else
-        helpers::calc_time_diff(time2, time1, l_sign, &seconds, &microseconds);
+        helpers::calc_time_diff(time2, time1, l_sign, &seconds, &microseconds, isDateTime);
 
     long t_seconds;
     int hour = seconds / 3600L;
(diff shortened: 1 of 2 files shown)


utils/funcexp/funchelpers.h
```diff
@@ -371,7 +371,7 @@ inline uint32_t calc_mysql_week( uint32_t year, uint32_t month, uint32_t day,
     return days / 7 + 1;
 }
 
-inline bool calc_time_diff(int64_t time1, int64_t time2, int l_sign, long long* seconds_out, long long* microseconds_out)
+inline bool calc_time_diff(int64_t time1, int64_t time2, int l_sign, long long* seconds_out, long long* microseconds_out, bool isDateTime = true)
```
jmrojas2332, line 374: isDateTime determines whether to treat time1 and time2 as full DateTimes or as Times

utils/funcexp/func_timediff.cpp
```diff
@@ -162,16 +162,35 @@ string Func_timediff::getStrVal(rowgroup::Row& row,
         case execplan::CalpontSystemCatalog::CHAR:
         case execplan::CalpontSystemCatalog::TEXT:
             text = parm[0]->data()->getStrVal(row, isNull);
-            isTime1 = std::count(text.begin(), text.end(), '-') <= 1; // Time can have at most 1 dash (signifies negative)
-            val1 = dataconvert::DataConvert::stringToDatetime(text, &isDate1);
+            
+            if (text.length() >= 14) // datetime has length at least 14
```
jmrojas2332, line 166: This is the major difference between results for MDB and columnstore. We should be converting all non-temporal data types to times if length is less than 14. There was a special case i found while testing results. see comment on line 170

drrtuy: I wonder if this magic value 14 differ for some of locales? Did you investigate this case?

drrtuy: BTW did you consider the case when the table uses UTF-8 encoding for varchar/text columns? This must produce widechar string but it worths to check b/c magic values tend to change in some corner cases.

utils/funcexp/func_timediff.cpp
```diff
@@ -162,16 +162,35 @@ string Func_timediff::getStrVal(rowgroup::Row& row,
...
+            {
+                val1 = dataconvert::DataConvert::stringToDatetime(text, &isDate1);
+            }
+            else if (text.length() == 8 && text.find("-") == string::npos && text.find(":") == string::npos) // special case, this is exactly a date
```
drrtuy, line 170: The previous comment is aplicable here also.

utils/funcexp/funchelpers.h
```diff
@@ -393,21 +393,21 @@ inline bool calc_time_diff(int64_t time1, int64_t time2, int l_sign, long long*
...
-    msec1 = (uint32_t)((time1 & 0xfffff));
-
-    year2 = (uint32_t)((time2 >> 48) & 0xffff);
-    month2 = (uint32_t)((time2 >> 44) & 0xf);
-    day2 = (uint32_t)((time2 >> 38) & 0x3f);
-    hour2 = (uint32_t)((time2 >> 32) & 0x3f);
-    min2 = (uint32_t)((time2 >> 26) & 0x3f);
-    sec2 = (uint32_t)((time2 >> 20) & 0x3f);
-    msec2 = (uint32_t)((time2 & 0xfffff));
+    year1 = isDateTime ? (uint32_t)((time1 >> 48) & 0xffff) : 0;
```
drrtuy, line 396: I wonder if multiple isDataTime conditions produce lots of jmp instructions in ASM. If so I would better have one if-else.

jmrojas2332:
New PR https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/962

PR #958, opened 2019-12-03
MCOL-3645 - Make ExeMgr destroy joblists in a different thread

pleblanc1976:
Prevents joblist destruction from stalling the main session thread by putting destruction in a different thread.  Will improve user exp when running queries that use a lot of mem, and/or when running queries in a script.

### DIFF
--- a/dbcon/joblist/resourcemanager.cpp
+++ b/dbcon/joblist/resourcemanager.cpp
@@ -80,6 +80,7 @@ ResourceManager::ResourceManager(bool runningInExeMgr) :
     fJlProcessorThreadsPerScan(defaultProcessorThreadsPerScan),
     fJlNumScanReceiveThreads(defaultScanReceiveThreads),
     fTwNumThreads(defaultNumThreads),
+    fJlMaxOutstandingRequests(defaultMaxOutstandingRequests),
     fHJUmMaxMemorySmallSideDistributor(fHashJoinStr,
                                        "UmMaxMemorySmallSide",
                                        getUintVal(fHashJoinStr, "TotalUmMaxMemorySmallSide", defaultTotalUmMemory),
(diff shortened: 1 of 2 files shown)


dbcon/joblist/resourcemanager.cpp
```diff
@@ -80,15 +80,15 @@ ResourceManager::ResourceManager(bool runningInExeMgr) :
     fJlProcessorThreadsPerScan(defaultProcessorThreadsPerScan),
     fJlNumScanReceiveThreads(defaultScanReceiveThreads),
     fTwNumThreads(defaultNumThreads),
+    fJlMaxOutstandingRequests(defaultMaxOutstandingRequests),
```
jmrojas2332, line 83: Thanks! I believe this warning was my fault.

exemgr/main.cpp
```diff
@@ -1061,13 +1080,28 @@ class SessionThread
...
-                                      li);
+                    // puts the real destruction in another thread to avoid
+                    // making the whole session wait.  It can take several seconds.
+                    int stmtID = csep.statementID();
+                    std::unique_lock<std::mutex> scoped(jlMutex);
+                    // C7's compiler complains about the msgLog capture here
+                    // msgLog is global scope, and passed by copy, so, unclear
+                    // what the warning is about.
+                    destructing++;
+                    std::thread bgdtor([jl, &jlMutex, &jlCleanupDone, stmtID, &li, msgLog, &destructing] {
```
jmrojas2332: My intellisense flags this msgLog parameter. 

> a variable with static storage duration cannot be captured in a lambda

Since it is static you can call from inside lambda without passing it into lambda in the first place.

jmrojas2332: This isn't really an error, but after compiling, I noticed it causes an extra warning.

jmrojas2332 (changes requested):
Looks good. Only issue I see is the static msgLog being passed into lambda which causes a warning. See my line comment.

exemgr/main.cpp
pleblanc1976: good call.  I tried many different combinations of captures and capture types.  This one doesn't give a warning.  Will add.

jmrojas2332 (approved):

PR #962, opened 2019-12-04
MCOL 3474 Fix TIMEDIFF w/ non-temporal data types

jmrojas2332:
The following changes were made to non-temporal data types after investigating the inconsistencies when running queries using MDB and columnstore engine. New regression tests added as well to verify changes allow columnstore to match mdb.

Regression tests PR: mariadb-corporation/mariadb-columnstore-regression-test#168

### DIFF
--- a/utils/dataconvert/dataconvert.cpp
+++ b/utils/dataconvert/dataconvert.cpp
@@ -3005,6 +3005,14 @@ int64_t DataConvert::intToDatetime(int64_t data, bool* date)
     else
     {
         y = atoi(year.c_str());
+
+        // special handling for 2-byte year
+        if (year.length() == 2)
+        {
+            y += 2000;
+            if (y > 2069)
+                y -= 100;
+        }
     }
 
     m = atoi(month.c_str());
@@ -3056,10 +3064,10 @@ int64_t DataConvert::intToTime(int64_t data, bool fromString)
 #if defined(__GNUC__) && __GNUC__ >= 7
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wformat-truncation="
-    snprintf( buf, 15, "%llu", (long long unsigned int)data);
+    snprintf( buf, 15, "%lld", (long long int)data);
 #pragma GCC diagnostic pop
 #else
-    snprintf( buf, 15, "%llu", (long long unsigned int)data);
+    snprintf( buf, 15, "%lld", (long long int)data);
 #endif
 
     //string date = buf;
@@ -3090,8 +3098,8 @@ int64_t DataConvert::intToTime(int64_t data, bool fromString)
 
         case 7:
             hour = string(bufread, 3);
-            min = string(bufread + 2, 2);
-            sec = string(bufread + 4, 2);
+            min = string(bufread + 3, 2);
+            sec = string(bufread + 5, 2);
             break;
 
         case 6:
@@ -3152,7 +3160,7 @@ int64_t DataConvert::intToTime(int64_t data, bool fromString)
     atime.msecond = ms;
     atime.is_neg = isNeg;
 
-    return *(reinterpret_cast<uint64_t*>(&atime));
+    return *(reinterpret_cast<int64_t*>(&atime));
 }
 
 int64_t DataConvert::stringToTime(const string& data)
--- a/utils/funcexp/func_timediff.cpp
+++ b/utils/funcexp/func_timediff.cpp
@@ -94,6 +94,14 @@ const string timediff( int64_t time1, int64_t time2, bool isDateTime)
 }
 }
 
+bool treatIntAsDatetime(const std::string &text)
+{
+    // min used when converting into to datetime is YYYYMMDD
+    // note: time diffing an int perceived to be in the format YYMMDD is actually treated as HHMMSS. Same functionality as MDB
+    bool isNeg = text.find("-") == 0;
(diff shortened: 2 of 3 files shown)


utils/funcexp/funchelpers.h
```diff
@@ -371,7 +371,7 @@ inline uint32_t calc_mysql_week( uint32_t year, uint32_t month, uint32_t day,
     return days / 7 + 1;
 }
 
-inline bool calc_time_diff(int64_t time1, int64_t time2, int l_sign, long long* seconds_out, long long* microseconds_out)
+inline bool calc_time_diff(int64_t time1, int64_t time2, int l_sign, long long* seconds_out, long long* microseconds_out, bool isDateTime = true)
```
jmrojas2332, line 374: isDateTime determines whether to treat time1 and time2 as full DateTimes or as Times

utils/funcexp/func_timediff.cpp
```diff
@@ -162,16 +162,35 @@ string Func_timediff::getStrVal(rowgroup::Row& row,
         case execplan::CalpontSystemCatalog::CHAR:
         case execplan::CalpontSystemCatalog::TEXT:
             text = parm[0]->data()->getStrVal(row, isNull);
-            isTime1 = std::count(text.begin(), text.end(), '-') <= 1; // Time can have at most 1 dash (signifies negative)
-            val1 = dataconvert::DataConvert::stringToDatetime(text, &isDate1);
+            
+            if (text.length() >= 14) // datetime has length at least 14
```
jmrojas2332: This is the major difference between results for MDB and columnstore. We should be converting all non-temporal data types to times if length is less than 14. There was a special case i found while testing results. see comment on line 170

jmrojas2332: https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/957#discussion_r353731147

jmrojas2332: https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/957#discussion_r353734092

utils/funcexp/funchelpers.h
```diff
@@ -393,21 +393,21 @@ inline bool calc_time_diff(int64_t time1, int64_t time2, int l_sign, long long*
...
-    msec1 = (uint32_t)((time1 & 0xfffff));
-
-    year2 = (uint32_t)((time2 >> 48) & 0xffff);
-    month2 = (uint32_t)((time2 >> 44) & 0xf);
-    day2 = (uint32_t)((time2 >> 38) & 0x3f);
-    hour2 = (uint32_t)((time2 >> 32) & 0x3f);
-    min2 = (uint32_t)((time2 >> 26) & 0x3f);
-    sec2 = (uint32_t)((time2 >> 20) & 0x3f);
-    msec2 = (uint32_t)((time2 & 0xfffff));
+    year1 = isDateTime ? (uint32_t)((time1 >> 48) & 0xffff) : 0;
```
jmrojas2332: https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/957#discussion_r353739320

jmrojas2332: @drrtuy  It's possible compiler will optimize this, however in case it doesn't I have changed to your recommendation. Please see new commit.

utils/funcexp/func_timediff.cpp
jmrojas2332: @drrtuy Have updated magic numbers to account for all locales considered during data type conversions between string types and datetimes/times, as well as int types and datetimes/times. Please see new commit

utils/dataconvert/dataconvert.cpp
```diff
@@ -3005,6 +3005,14 @@ int64_t DataConvert::intToDatetime(int64_t data, bool* date)
     else
     {
         y = atoi(year.c_str());
+
+        // special handling for 2-byte year
+        if (year.length() == 2)
```
jmrojas2332, line 3010: Found a bug while going through a test case. Year can be in format YY so this is needed so isDateValid does not flag this year as invalid. These magic numbers are taken from function mysql_str_to_datetime lines 711-714 of this same file

utils/dataconvert/dataconvert.cpp
```diff
@@ -3056,10 +3064,10 @@ int64_t DataConvert::intToTime(int64_t data, bool fromString)
 #if defined(__GNUC__) && __GNUC__ >= 7
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wformat-truncation="
-    snprintf( buf, 15, "%llu", (long long unsigned int)data);
+    snprintf( buf, 15, "%lld", (long long int)data);
```
jmrojas2332, line 3067: A bug I found while going through negative test cases. Since time can be negative this was causing a loss of sign.

utils/dataconvert/dataconvert.cpp
```diff
@@ -3090,8 +3098,8 @@ int64_t DataConvert::intToTime(int64_t data, bool fromString)
 
         case 7:
             hour = string(bufread, 3);
-            min = string(bufread + 2, 2);
-            sec = string(bufread + 4, 2);
+            min = string(bufread + 3, 2);
```
jmrojas2332, line 3101: Another bug I found. Trivial.

LinuxJedi (approved):

drrtuy (approved):

PR #963, opened 2019-12-04
MCOL 3585 Related. Updated Comment in config file

jmrojas2332:
Updated comments in config files.

Went ahead and updated the default PmMaxMemorySmallSide value from resourcemanger.h in case somebody commits a user error and deletes that value from config file.


oam/etc/Columnstore.xml
```diff
@@ -510,6 +510,8 @@
 			  ProcessorThreadsPerScan * MaxOutstandingRequests should be at least
 			  as many threads are available across all PMs.  -->
 		<!-- <ProcessorThreadsPerScan>16</ProcessorThreadsPerScan> --> 
+		<!-- MaxOutstandingRequests is going to default to the num of cores available 
+			 across all performance modules * 4 divided by the ProcessorThreadsPerScan -->
```
pleblanc1976: Thinking about this default again, I think we need to set a minimum of 2 * # of PMs as well.  ex, this is an odd configuration, but imagine a system with 2-cores and 2 pms.  4 * 4 / 16 = 1 so, only one extent's worth of jobs will be issued at once.  That will only 'liven' one PM at a time.  For the best performance, each PM would never run dry, and would always have some work waiting in its queue.

pleblanc1976 (changes requested):
Realized in some configurations we need to set a minimum value for MaxOutstandingRequests to prevent PMs from idling during a query.

oam/etc/Columnstore.xml
jmrojas2332: Updated the minimum to be max( 2 * # of PMs, 20). Twenty was the old default.

pleblanc1976 (approved):

PR #967, opened 2019-12-06
MCOL-1734 This patch fixes the error for NOT IN + subquery that return empty set

drrtuy:
This patch fixes the error for NOT IN + subquery  when outer query has NULLs in a key 
 column.
 The if statement that decides whether to add the Row into a result set
 or not has a logic error thus produces false positives for the case
 described in MCOL-1734.


LinuxJedi (approved):

pleblanc1976 (approved):
Looks good to me.

PR #979, opened 2019-12-12
MCOL 3596 conv() result string not matching reference

jmrojas2332:
This patch fixes a bug where only a max of 21 characters would be displayed when returned from conv function to MDB.

Test cases from working_tpch1/qa_fe_cnxFunctions/conv.sql from MCS Regression Test were used to verify working patch.

### DIFF
--- a/dbcon/mysql/ha_mcs_impl.cpp
+++ b/dbcon/mysql/ha_mcs_impl.cpp
@@ -649,6 +649,7 @@ int fetchNextRow(uchar* buf, cal_table_info& ti, cal_connection_info* ci, bool h
                             break;
 
                         default:
+                            (*f)->field_length = 255; // reserve enough space for varchar, was previously displaying only 21 characters
                             f2->store((const char*)row.getStringPointer(s), row.getStringLength(s), f2->charset());
                     }
 


dbcon/mysql/ha_mcs_impl.cpp
```diff
@@ -649,6 +649,7 @@ int fetchNextRow(uchar* buf, cal_table_info& ti, cal_connection_info* ci, bool h
                             break;
 
                         default:
+                            (*f)->field_length = 255; // reserve enough space for varchar, was previously displaying only 21 characters
```
jmrojas2332, line 652: After spending a great deal of time digging into CS and MDB trying to figure out why the result was getting truncated to length 21, I noticed we were doing something similar to this patch on lines 756 and 778.

drrtuy: It looks like we changing the parameter sent by MDB to us. This might be dangerous thus we need to change the length of the expected result in MDB code similar to what had done in MDEV-27320.

drrtuy:
Plz don't merge this yet.

jmrojas2332:
Confirmed this change needs to be in server. Server is setting length inside item_strfunc.h. Specifically, the function Item_func_conv::fix_length_and_dec() was setting max_length to 64 which was being used for field_length.

PR #981, opened 2019-12-13
MCOL-3672 Fix regression in deletes

LinuxJedi:
Deletes appear to only use the direct delete path. This allows that to
happen.


LinuxJedi:
Force merged as no one is available to review. Builds work and this is a blocker for other things.

PR #985, opened 2019-12-17
MCOL-3674: Fix row-based replication detection

pleblanc1976:
Changed the row-based repl detection mechanism so stmt-based repl works
again, however, the detection mechanism is still wrong somehow.  What it
should be is currently unknown.


pleblanc1976:
Debating whether we should just comment out the row-based check altogether.  Leaning toward that since we don't know what the current check really means.  Let me know.

LinuxJedi (approved):

PR #997, opened 2020-01-03
MCOL-1950: join & aggregation mem reduction

pleblanc1976:
Made our allocators capable of distinguishing between nodes and tables being allocatoed.  Nodes get allocated from the pool, tables get allocated (and more importantly, deallocated) by the system allocator.  In my test case, it reduced mem usage by 70% and boosted performance somewhat.  See the ticket for more info.


pleblanc1976:
Will improve the aesthetics of this PR and reissue.

PR #999, opened 2020-01-06
Fixed a couple minor things I ran into working on another task

pleblanc1976:
- Made TupleUnion not allocate 64MB per frame.
 - Saw a crash, where a THJS would be destroyed w/o being used.
   Fixed the bad reference.

### DIFF
--- a/dbcon/joblist/tuplehashjoin.cpp
+++ b/dbcon/joblist/tuplehashjoin.cpp
@@ -132,8 +132,9 @@ TupleHashJoinStep::~TupleHashJoinStep()
     if (ownsOutputDL)
         delete outputDL;
 
-    for (uint i = 0 ; i < smallDLs.size(); i++)
-        resourceManager->returnMemory(memUsedByEachJoin[i], sessionMemLimit);
+    if (memUsedByEachJoin)
+        for (uint i = 0 ; i < smallDLs.size(); i++)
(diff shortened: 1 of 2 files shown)


dbcon/joblist/tuplehashjoin.cpp
```diff
@@ -132,8 +132,9 @@ TupleHashJoinStep::~TupleHashJoinStep()
     if (ownsOutputDL)
         delete outputDL;
 
-    for (uint i = 0 ; i < smallDLs.size(); i++)
-        resourceManager->returnMemory(memUsedByEachJoin[i], sessionMemLimit);
+    if (memUsedByEachJoin)
```
jmrojas2332, line 135: Not sure what our coding standard says regarding this, but I would definitely add braces around this if statement considering it contains a for loop inside it and not a single statement. Other than that i have no issue with change.

jmrojas2332 (commented):
See comment in review

dbcon/joblist/tuplehashjoin.cpp
pleblanc1976: I think you're right, our coding standard says always add brackets.  I added them in the join abort fix PR, so will not add them to this one.

jmrojas2332 (approved):

PR #1000, opened 2020-01-07
MCOL-1734 CS now behaves similar to MDB in case of NOT IN + correlated subquery if the subquery returns empty set.

drrtuy:
CS now returns full outer record set.

gcc 8.2 complains about unused variable in ha_mcs_impl.cc

### DIFF
--- a/dbcon/mysql/ha_mcs_impl.cpp
+++ b/dbcon/mysql/ha_mcs_impl.cpp
@@ -2303,7 +2303,7 @@ int ha_mcs_impl_direct_update_delete_rows(bool execute, ha_rows *affected_rows)
         *affected_rows = ci->affectedRows;
     }
 
-    return 0;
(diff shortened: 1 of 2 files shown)


pleblanc1976 (changes requested):
Roman will add similar code for um joins

pleblanc1976:
Be sure to add a query or two to the regression test for this one.

LinuxJedi (approved):

pleblanc1976:
Update on this one?  Does disk-join produce correct results?

drrtuy:
Yes, UM-side memory and disk based joins also produce expected results.

вт, 14 янв. 2020 г. в 19:43, Patrick LeBlanc <notifications@github.com>:

pleblanc1976 (approved):

PR #1001, opened 2020-01-07
MCOL-3713:  Join abort fix

pleblanc1976:
Noticed a couple problems.  1) segregateJoiners() would double-lock djsMutex if any join exceeded the mem limit; 2) the logic to detect the mem overflow might have flagged joins as disk-joins even if that were disabled, requiring a longer than necessary abort process.  Also it was fairly unreadble; cleaned it up & added a little inline documentation.

To verify I set the session var columnstore_um_mem_limit first between the PM and UM join limits and executed a large join, then set it below the PM join limit.  Verified with gdb that no threads were orphaned; verified with valgrind & top that no memory was leaked.  Enabled disk join and verified the same.


dbcon/joblist/tuplehashjoin.cpp
```diff
@@ -1921,15 +1929,17 @@ void TupleHashJoinStep::segregateJoiners()
...
-    		joinIsTooBig = true;
+            joiner[i]->setConvertToDiskJoin();
     		djsJoiners.push_back(joiners[i]);
     		djsJoinerMap.push_back(i);
     	}
     	return;
     }
     #endif
 
+    boost::mutex::scoped_lock sl(djsLock);
```
drrtuy, line 1942: Why do we take lock here?

pleblanc1976: the lock protects djs and djsJoiners.  Abort() needs a consistent view of those, and it can come asynchronously.

jmrojas2332 (approved):

PR #1004, opened 2020-01-17
Runtime disable oam

pleblanc1976:
Made the SKIP_OAM_INIT build parameter a runtime parameter.  Now you toggle that behavior by setting or unsetting SKIP_OAM_INIT as an envvar.

I apologize for my editor changing various bits of whitespace.

### DIFF
--- a/cmake/configureEngine.cmake
+++ b/cmake/configureEngine.cmake
@@ -716,9 +716,9 @@ IF (NOT INLINE)
 SET (inline "")
 ENDIF()
 
-IF($ENV{SKIP_OAM_INIT})
-    set(SKIP_OAM_INIT 1 CACHE BOOL "Skip OAM initialization" FORCE)
-ENDIF()
+#IF($ENV{SKIP_OAM_INIT})
+#    set(SKIP_OAM_INIT 1 CACHE BOOL "Skip OAM initialization" FORCE)
+#ENDIF()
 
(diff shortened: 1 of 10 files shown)


LinuxJedi (approved):

oamapps/postConfigure/installer.cpp
```diff
@@ -711,6 +711,12 @@ int main(int argc, char* argv[])
             mysqlSetup();
             sleep(5);
```
pleblanc1976, line 713: Let's try moving the SKIP check to before that spurious sleep(5);

cmake/configureEngine.cmake
```diff
@@ -716,9 +716,9 @@ IF (NOT INLINE)
 SET (inline "")
 ENDIF()
 
-IF($ENV{SKIP_OAM_INIT})
-    set(SKIP_OAM_INIT 1 CACHE BOOL "Skip OAM initialization" FORCE)
-ENDIF()
+#IF($ENV{SKIP_OAM_INIT})
+#    set(SKIP_OAM_INIT 1 CACHE BOOL "Skip OAM initialization" FORCE)
+#ENDIF()
```
pleblanc1976: let's get rid of the SKIP stuff in the cmake files.

LinuxJedi (approved):

PR #1006, opened 2020-01-21
MCOL-128 Support ALTER TABLE...ENGINE=Columnstore

LinuxJedi:
Also implements:

* ALTER TABLE from Columnstore to another engine
* MCOL-3349 CREATE TABLE ... AS SELECT.

### DIFF
--- a/dbcon/mysql/ha_mcs_ddl.cpp
+++ b/dbcon/mysql/ha_mcs_ddl.cpp
@@ -2400,14 +2400,10 @@ int ha_mcs_impl_create_(const char* name, TABLE* table_arg, HA_CREATE_INFO* crea
         return 1;
     }
 
-    // @bug 3908. error out primary key for now.
+    // Send notice if primary key specified that it is not supported
     if (table_arg->key_info && table_arg->key_info->name.length && string(table_arg->key_info->name.str) == "PRIMARY")
     {
-        string emsg = logging::IDBErrorInfo::instance()->errorMsg(ERR_CONSTRAINTS);
-        setError(thd, ER_CHECK_NOT_IMPLEMENTED, emsg);
-        ci.alterTableState = cal_connection_info::NOT_ALTER;
-        ci.isAlter = false;
-        return 1;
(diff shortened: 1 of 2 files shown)


dbcon/mysql/ha_mcs_ddl.cpp
```diff
@@ -2600,7 +2607,7 @@ int ha_mcs_impl_delete_table_(const char* db, const char* name, cal_connection_i
 
     string emsg;
 
-    if (thd->lex->sql_command == SQLCOM_DROP_DB)
+    if ((thd->lex->sql_command == SQLCOM_DROP_DB) || (thd->lex->sql_command == SQLCOM_ALTER_TABLE))
```
pleblanc1976, line 2610: This is for the 'alter table .. engine = non-CS-engine' case I assume?

pleblanc1976 (changes requested):
Buildbot had a problem in test000.  I built locally to see if it was a BB problem or a code problem.  Looks like a code problem.

server commit = 8a931e4d16 (current 10.4 branch of server repo)
engine commit = d0a910cbdd6 (commit BB checked out)

Getting this running simple DDL:
`MariaDB [tpch1]> create table test (i int) engine=columnstore;
ERROR 1178 (42000): The storage engine for the table doesn't support IDB-4016: DML and DDL statements for Columnstore tables can only be run from the replication master.
MariaDB [tpch1]>`

dbcon/mysql/ha_mcs_ddl.cpp
LinuxJedi: Correct, the final phase is to drop the table in the old engine. If you don't do this it succeeds but generates an error and then you are in a weird state where the table is in both ColumnStore and the other engine.

LinuxJedi:
Weird, I didn't touch that code and test000 passed locally. I'll look into it.

pleblanc1976 (approved):
C7 BB is happy now, so will merge.

PR #1010, opened 2020-01-23
MCOL-3680 mysqld will abort if Columnstore.xml is missing.

tntnatbry:


davidhall586:
makeconfig is called all over the place without a try block. This is probably OK so long as the first call is in a try block. This patch only considers the connector library. What about all the other executables, especially on the PM? What happens when those processes start without an xml file available?

tntnatbry:
First call to makeConfig() is made throughout the code across all components of the system. Adding try-catch blocks on the first calls would be a much bigger change. Even within the connector code, there could be several sources of a call to makeConfig(), such as cal_connection_info() ctor, tid2sid() calls, etc. All components on the UM and PM would still abort, as they all require the xml file on startup. The idea here is to not impact mysqld if the xml file is missing so that users can still execute non-ColumnStore related queries. Thoughts?

LinuxJedi (approved):
It meets the requirements to fix the bug (to prevent mysqld from crashing when the XML file is missing)

PR #1013, opened 2020-01-23
Mcol 3632 Issue correct subselect error

davidhall586:
The error should be IDB-3016: Function or Operator with sub query on the SELECT clause is currently not supported.
At some point Server changed what they were sending and the error became too generic. This restores the old error message.
This fixes the problem with q0035 comparison in the WindowFunction regression test.

### DIFF
--- a/cmake/configureEngine.cmake
+++ b/cmake/configureEngine.cmake
@@ -716,9 +716,9 @@ IF (NOT INLINE)
 SET (inline "")
 ENDIF()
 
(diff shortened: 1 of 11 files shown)


pleblanc1976:
Is develop the intended target?  This PR includes more than an error msg fix.

tntnatbry:
@dhall-MariaDB The PR adds a lot more commits than intended for this issue as @pleblanc1976 mentioned. Can you re-submit this?

PR #1018, opened 2020-01-28
MCOL-2139 Fix subquery not being filtered when using view

jmrojas2332:
columnMap was empty when building predicate, which fails to build predicate on update or delete because it assumes this operation is on two constants (see https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/1018/files#diff-01ac5c8ba99c34e7b0cfd87552768908R2188) to view this code that fails to create predicate.

PR into develop-1.2 https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/1019

Code bases are a bit different (e.g., ha_mcs_execplan.cpp was ha_calpont_execplan.cpp)

### DIFF
--- a/dbcon/mysql/ha_mcs_execplan.cpp
+++ b/dbcon/mysql/ha_mcs_execplan.cpp
@@ -5664,7 +5664,9 @@ void gp_walk(const Item* item, void* arg)
                 
                 if (col->type() == Item::FIELD_ITEM)
                 {
-                    gwip->columnMap.insert(CalpontSelectExecutionPlan::ColumnMap::value_type(string(((Item_field*)item)->field_name.str), scsp));
+                    const auto &field_name = string(((Item_field*)item)->field_name.str);
+                    auto col = CalpontSelectExecutionPlan::ColumnMap::value_type(field_name, scsp);
+                    gwip->columnMap.insert(col);
                 }
             }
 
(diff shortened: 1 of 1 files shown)


dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -5661,6 +5661,11 @@ void gp_walk(const Item* item, void* arg)
             {
                 boost::shared_ptr<SimpleColumn> scsp(sc->clone());
                 gwip->scsp = scsp;
+                
+                if (col->type() == Item::FIELD_ITEM)
+                {
+                    gwip->columnMap.insert(CalpontSelectExecutionPlan::ColumnMap::value_type(string(((Item_field*)item)->field_name.str), scsp));
```
jmrojas2332: Fixed issue by inserting new column mapping when Item is Item::Ref_Item and coltype is Field_Item.

For reference, here is query:

`update dairy_collection_animal a, (select collection_id from dairy_collection_view where collection_id = 2)b set a.animal_id = 1 where a.collection_id = b.collection_id;`

The subquery's lhs of WHERE condition was not being mapped previously.

pleblanc1976: Looks OK to me, but I don't know this code very well, so lets wait for Roman's input before merging.

drrtuy: Looks OK for me too. However we need to run full regression with this oneliner.
I just have aestetic request for now. Could you make break this oneliner to make it less then 80 symbols to make it readable in vim with multiply panes? Here and in 1.2 PR also.

pleblanc1976 (approved):

PR #1023, opened 2020-02-01
MCOL-3718 Use systemd instead of mysql-Columnstore

LinuxJedi:
MariaDB server is now managed using systemd instead of our own
mysql-Columnstore script. This means that MariaDB server can be stopped
/ started independently of ColumnStore


LinuxJedi:
This appears to break Buildbot because our RPMs don't appear to have MariaDB's systemd unit. I'm not sure why.

benthompson15 (changes requested):
in void Oam::actionMysqlCalpont(MYSQLCALPONT_ACTION action)
check logic for MYSQL_STATUS 
before it was logging to mysql.status which is checked at end of function by
        if (checkLogStatus(mysqlStatus, "MySQL running"))
when that fails its set to MAN_OFFLINE and DEGRADED even though mysqld is running

benthompson15 (approved):

PR #1024, opened 2020-02-01
MCOL-3721 Allow collate and warn on ORDER BY

LinuxJedi:
COLLATE is now allowed but ignored on DDL and ORDER BY. For an ORDER BY
query a note is generated.


drrtuy (changes requested):
We decided to add column definition level support for COLLATE in DDL offline.

drrtuy (approved):

PR #1029, opened 2020-02-05
MCOL-3749 Assertion on error

davidhall586:
Correct error code return stops assertion


jmrojas2332 (approved):
Approved (Same as PR into develop branch)

PR #1031, opened 2020-02-06
MCOL-3520: Fix importing files from S3 for mode 1 imports.

pleblanc1976:


writeengine/splitter/we_cmdargs.cpp
```diff
@@ -217,10 +217,10 @@ std::string WECmdArgs::getCpImportCmdLine()
     if (fbTruncationAsError)
         aSS << " -S ";
 
-    if (!fS3Key.empty())
+    if (!fS3Key.empty() && !(fMode == 0 || fMode == 1))
     {
```
pleblanc1976, line 221: these cmdline params shouldn't be propagated when this process is acting as a distributor.

writeengine/splitter/we_cmdargs.h
```diff
@@ -37,7 +37,10 @@ class WECmdArgs
 public:
     WECmdArgs(int argc, char** argv);
     virtual ~WECmdArgs() {}
-public:
```
pleblanc1976, line 40: reorganized this file a little.

writeengine/splitter/we_filereadthread.cpp
```diff
@@ -200,37 +223,33 @@ bool WEFileReadThread::chkForListOfFiles(std::string& FileName)
...
+        if (fSdh.getDebugLvl() > 1)
+            cout << "Next Input File " << aFile << endl;
 
-            start = end + 1;
-            fInfileList.push_back(aFile);
-        }
-        else
+        if ((!doS3Import && access(aFile.c_str(), O_RDONLY) != 0) ||
+            (doS3Import && ms3_status(s3Connection, s3Bucket.c_str(), 
+                aFile.c_str(), &ms3status) != 0))
```
pleblanc1976, line 243: consolidated the if-stmt and moved the file access check to here from we_sdlhandler.

writeengine/splitter/we_filereadthread.cpp
```diff
@@ -451,9 +477,48 @@ void WEFileReadThread::openInFile()
...
+                else 
+                    os << "Download of '" << fInFileName << "' failed.  Got '" << ms3_error(err)
+                        << "'.";
+                throw runtime_error(os.str());
+            }
+            
+            arrSource.reset(new boost::iostreams::array_source((char *) buf, bufLen));
+            s3Stream.reset(new boost::iostreams::stream<boost::iostreams::array_source>(*arrSource));
+            fInFile.rdbuf(s3Stream->rdbuf());
+        }
```
pleblanc1976, line 519: The s3 lib currently requires downloading a whole object into mem.  This can change once add'l functionality is added to the s3 lib.

writeengine/splitter/we_filereadthread.cpp
```diff
@@ -463,12 +528,11 @@ void WEFileReadThread::openInFile()
                 cout << aGreenCol
                      << "trying to read from STDIN... "
                      << aDefCon << endl;
+             fInFile.rdbuf(cin.rdbuf());
```
pleblanc1976, line 531: random bug I noticed in testing.  if you specified an input list like "file1 stdin", it wouldn't read from stdin.  This line fixes it for mode-1 imports.

writeengine/splitter/we_sdhandler.cpp
```diff
@@ -501,14 +501,9 @@ void WESDHandler::setup()
...
     fLog.setLogFileName(aLogName.c_str(), aErrLogName.c_str(), getConsoleLog());
 
-    // In mode 0 and Mode 1, we need to check for local file availability
-    if ((0 == fRef.fCmdArgs.getMode()) || (1 == fRef.fCmdArgs.getMode()))
-    {
-        if (!check4InputFile(fRef.getLocFile()))
-        {
-            throw (runtime_error("Could not open Input file " + fRef.getLocFile()));
-        }
-    }
```
pleblanc1976, line 511: This used to do the existence/access check (wrongly).  Fixed it and moved it into readthread::chkForListOfFiles()

writeengine/splitter/we_cmdargs.cpp
```diff
@@ -1233,7 +1235,12 @@ void WECmdArgs::parseCmdLineArgs(int argc, char** argv)
             throw (runtime_error("No schema or local filename specified."));
         }
     }
-
+    
+    /* check for all-or-nothing cmdline args to enable S3 import */
+    int s3Tmp = (fS3Key.empty() ? 0 : 1) + (fS3Bucket.empty() ? 0 : 1) +
+        (fS3Secret.empty() ? 0 : 1) + (fS3Region.empty() ? 0 : 1);
+    if (s3Tmp != 0 && s3Tmp != 4)
+        throw runtime_error("The access key, secret, bucket, and region are all required to import from S3");
```
tntnatbry, line 1243: At other places, throw is used as throw ((runtime_error)), i.e. with an extra pair of paranthesis. Can we make this consistent?

writeengine/splitter/we_cmdargs.h
```diff
@@ -252,11 +241,44 @@ class WECmdArgs
...
+        return fS3Host;
+    }
+    std::string getS3Secret() const
+    {
+        return fS3Secret;
+    }
+    std::string getS3Region() const 
+    {
+        return fS3Region;
+    }
```
tntnatbry, line 268: I think the string returns should be const references here.

writeengine/splitter/we_filereadthread.cpp
```diff
@@ -122,6 +125,18 @@ WEFileReadThread::~WEFileReadThread()
...
+
+    if (doS3Import)
+    {
+        ms3_deinit(s3Connection);
+        ms3_library_deinit();
+        if (buf)
+        {
+            s3Stream.reset();
+            arrSource.reset();
+            free(buf);
```
tntnatbry, line 137: buf = NULL; here?

writeengine/splitter/we_cmdargs.cpp
tntnatbry: Just a side note not related to the PR, even though cpimport accepts mode 0, cpimport -h does not mention about it.

tntnatbry: Also, since Mode 2 and S3 is not allowed, the only case these options are propagated to cpimport.bin is Mode 3?

tntnatbry (changes requested):

writeengine/splitter/we_cmdargs.cpp
pleblanc1976: @roman yeah I saw references to mode 0 in the code, I'm not sure what that is or was.  I think it is an alias for mode 1 now, but might have originally had a different purpose.

@gagan Yes that's right.  Mode 1 means the splitter distributes data to a cpimport process on each PM.  Mode 2 means the splitter tells the cpimport processes to import local file F, which they all have by external means (ETL dropping a file with the same name on each node).  Obviously having them all fetch the same file from S3 doesn't make sense.  Mode 3 is a single-node import.  On a single-node cluster all modes are effectively mode 3, and so the args for importing from s3 need to be propagated,

writeengine/splitter/we_cmdargs.cpp
pleblanc1976: No need.  Many hands have been at work here I think, and there are a lot of stylistic conflicts.  In some cases I tried to make it look consistent, just not here.  :)

writeengine/splitter/we_filereadthread.cpp
pleblanc1976: not necessary, this is the dtor.  But good catch.

writeengine/splitter/we_cmdargs.cpp
tntnatbry: @pleblanc1976 This is very helpful, thanks for the clarification.

writeengine/splitter/we_cmdargs.cpp
tntnatbry: Sounds good to me :)

tntnatbry (approved):

PR #1043, opened 2020-02-11
MCOL-3748: Modify signal handling for different default actions.

benthompson15:


pleblanc1976 (changes requested):
list of signals handled is incomplete, and using the same var for both handlers is probably something we don't want.  Would allow a sequence like
1) start a graceful shutdown
2) it segfaults
3) segfault is ignored, where if run 5s earlier it wouldn't have been

pleblanc1976 (approved):

PR #1044, opened 2020-02-11
Merge pull request #1015 from pleblanc1976/mcol-3716

pleblanc1976:
Mcol-3716: Wrong min/max after cpimport which causes wrong select results


jmrojas2332 (approved):
LGTM

PR #1048, opened 2020-02-13
MCOL-3760 + Fix multiple rands in statement with or without ORDER BY

jmrojas2332:
After investigating MCOL-3760, and the rand function, it turns out MDB Server uses separate seeds for each rand() in a statement, while columnstore uses only one set of seeds used across all rands.

This was hidden from us in develop-1.2 regressions because 
a. we did not have tests using multiple rands in a statement
b. we did not handle ORDER BY rand (this was done on server) so our test that included a
`SELECT rand(2) FROM table ORDER BY rand(1)` passed in 1.2.

Solution: For each rand function in statement, store a set of seeds used when computing rand for that column. 
e.g., `SELECT rand(1), rand(2) FROM table` will contain separate seeds for rand(1) and rand(2). 

I added some regression tests inside working_tpch1/qa_fe_cnxFunctions/RAND.sql. PR is 
https://github.com/mariadb-corporation/mariadb-columnstore-regression-test/pull/184

This will also be needed in develop-1.2, in the case of multiple rands in a select statement, which i will PR into shortly after this is approved.

### DIFF
--- a/utils/funcexp/func_rand.cpp
+++ b/utils/funcexp/func_rand.cpp
@@ -53,6 +53,7 @@ double Func_rand::getRand()
         fSeed1 += 23;
 
     fSeed2 = (fSeed1 + fSeed2 + 33) % maxValue;
+    fSeeds[fSeedIndex] = std::make_pair(fSeed1, fSeed2);
     return (((double) fSeed1) / (double)maxValue);
 }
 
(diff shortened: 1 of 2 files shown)


davidhall586 (approved):

utils/funcexp/func_rand.cpp
```diff
@@ -70,6 +71,37 @@ double Func_rand::getDoubleVal(rowgroup::Row& row,
...
+    {
+        // No longer on first row. Disable additional seeding
+        if (!fMultipleSeedsSet)
+        {
+            fMultipleSeedsSet = true;
+            fSeedIndex = 0;
+        }
+        else
+        {
+            fSeedIndex += 1;
```
drrtuy, line 96: I wonder if we need to take a lock working with Func_rand intsances from multiple threads or they are separate?

drrtuy (approved):

PR #1057, opened 2020-02-20
MCOL-3788 Remove save/restore_group_list impl since the optimizer call does not happen in create_SH anymore

tntnatbry:


drrtuy (approved):

drrtuy (changes requested):
Something is wrong with BB CI. Could you check what's wrong?

drrtuy (approved):

PR #1060, opened 2020-02-24
1.4: MCOL-3829: CS not starting correctly if rebooted at specific point

pleblanc1976:
Found that the IDBDataFile path in BRM journal writing code needs to seek
to the end of the file before writing.

If save_brm is run (as it is during init), it truncates the journal file,
but the workernode retained its original offset, resulting in the front
of the file being 0-filled on the next journal write.  Load_brm can't
load that.


VERSION
```diff
@@ -1,4 +1,4 @@
 COLUMNSTORE_VERSION_MAJOR=1
 COLUMNSTORE_VERSION_MINOR=4
 COLUMNSTORE_VERSION_PATCH=3
-COLUMNSTORE_VERSION_RELEASE=1
+COLUMNSTORE_VERSION_RELEASE=2
```
pleblanc1976, line 4: we'll need to bump the version in dev-1.4 anyway.  ignore this intermediate update

jmrojas2332 (approved):

PR #1062, opened 2020-02-24
Make this update cmake cache for out of source builds to correctly work.

benthompson15:


dbcon/mysql/CMakeLists.txt
```diff
@@ -41,7 +41,7 @@ else ()
     add_library(ha_columnstore SHARED ${libcalmysql_SRCS})
     SET_TARGET_PROPERTIES(ha_columnstore PROPERTIES PREFIX "")
 
-    target_link_libraries(ha_columnstore ${ENGINE_LDFLAGS} ${ENGINE_WRITE_LIBS} ${NETSNMP_LIBRARIES} ${SERVER_BUILD_DIR}/libservices/libmysqlservices.a threadpool)
+    target_link_libraries(ha_columnstore ${ENGINE_LDFLAGS} ${ENGINE_WRITE_LIBS} ${NETSNMP_LIBRARIES} mysqlservices threadpool)
```
drrtuy, line 44: What if we search for this library path using cmake's FIND_LIBRARY?

drrtuy: Same question is valid for threadpoool.

PR #1072, opened 2020-02-27
1.4.2 columnstore-pre-uninstall missing this or storagemanager.cnf is lost

benthompson15:


benthompson15:
No one is around and this isn't protected branch and for sake of getting builds to QA for my now later this AM

PR #1074, opened 2020-02-27
MCOL-3485 

jmrojas2332:
Fix ORDER BY not working with multiple aggregates of same type in SELECT clause

Issue turned out to be the sum aggregates in the test case were sharing the same expression id (second aggregates expression id was overriding the first), which was causing the order by to sort by the second aggregate (see example). 

Altered code to correctly check for same aggregate on the select list

### DIFF
--- a/dbcon/mysql/ha_mcs_execplan.cpp
+++ b/dbcon/mysql/ha_mcs_execplan.cpp
@@ -4941,7 +4941,8 @@ ReturnedColumn* buildAggregateColumn(Item* item, gp_walk_info& gwi)
         {
             for (uint32_t i = 0; i < gwi.returnedCols.size(); i++)
             {
-                if (*ac == gwi.returnedCols[i].get())
+                if (*ac == gwi.returnedCols[i].get() 
+                        && ac->alias() == gwi.returnedCols[i].get()->alias())
                     ac->expressionId(gwi.returnedCols[i]->expressionId());
             }
         }


dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -4941,7 +4941,8 @@ ReturnedColumn* buildAggregateColumn(Item* item, gp_walk_info& gwi)
         {
             for (uint32_t i = 0; i < gwi.returnedCols.size(); i++)
             {
-                if (*ac == gwi.returnedCols[i].get())
+                if (*ac == gwi.returnedCols[i].get() 
+                        && ac->alias() == gwi.returnedCols[i].get()->alias())
```
drrtuy, line 4945: Could you double check the case where projection column alias is empty, e.g. select sum(...) as `` ...?

jmrojas2332: Sorts correctly when the sorting column is x or y and the other is ``. It does not sort by a column that has an empty column alias (existing behavior before chnage).

drrtuy:
Even if you reference column from projection by its number?

пт, 28 февр. 2020 г., 20:39 Jose Rojas <notifications@github.com>:

jmrojas2332:
@drrtuy Confirmed it sorts correctly when sorting by it's number

drrtuy (approved):
Nicely done.

PR #1083, opened 2020-03-04
MCOL-3870 PostConfigure exit with mysqld running

davidhall586:
"MariaDB Server is currently running on PID 1. Cannot run postConfigure whilst this is running"
Even though mysqld is not running

### DIFF
--- a/oamapps/postConfigure/postConfigure.cpp
+++ b/oamapps/postConfigure/postConfigure.cpp
@@ -566,14 +566,22 @@ int main(int argc, char* argv[])
     }
 
     char buf[512];
-    FILE *cmd_pipe = popen("pidof -s mysqld", "r");
-    fgets(buf, 512, cmd_pipe);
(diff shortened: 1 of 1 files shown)


jmrojas2332 (approved):

oamapps/postConfigure/postConfigure.cpp
```diff
@@ -566,14 +566,22 @@ int main(int argc, char* argv[])
     }
 
     char buf[512];
-    FILE *cmd_pipe = popen("pidof -s mysqld", "r");
-    fgets(buf, 512, cmd_pipe);
-    pid_t pid = strtoul(buf, NULL, 10);
-    pclose(cmd_pipe);
-
-    if (pid)
+    *(int64_t*)buf = 0;
```
drrtuy, line 569: I would suggest to avoid using C-style cast.

PR #1084, opened 2020-03-05
MCOL-3777 Cleanup left-over values from a previously failed insert.

tntnatbry:


dbcon/mysql/ha_mcs_impl.h
```diff
@@ -32,7 +32,7 @@ extern int ha_mcs_impl_close(void);
 extern int ha_mcs_impl_rnd_init(TABLE* table);
 extern int ha_mcs_impl_rnd_next(uchar* buf, TABLE* table);
 extern int ha_mcs_impl_rnd_end(TABLE* table, bool is_derived_hand = false);
-extern int ha_mcs_impl_write_row(const uchar* buf, TABLE* table);
+extern int ha_mcs_impl_write_row(const uchar* buf, TABLE* table, uint64_t rows_changed);
```
drrtuy, line 35: I wonder what is the meaning of rows_changed given that write_row writes one row? Is it possible to write more then or less then one row with write_row call?

tntnatbry: rows_changed is a variable that MariaDB server uses to keep track of the number of successful calls to write_row. If we have a multi-value insert such as `insert into t1 values (val1), (val2), (val3)`, then there would be 3 calls to write_row, and if all calls are successful, rows_changed would be set to 3. On the ColumnStore side however, all the 3 rows would be collected and inserted in a single batch insert.

drrtuy: Who sets the variable: plugin or MDB core? Do you mean we have rows_changed set to 3 on the 3d call to ha_mcs_impl_write_row() or this is a wrong thinking?

davidhall586 (approved):

dbcon/mysql/ha_mcs_impl.h
tntnatbry: MDB core increments rows_changed after a successful call to write_row(). rows_changed in ha_mcs_impl_write_row() will be 2 on the 3rd call, when write_row() returns, it will be incremented to 3 in the server.

drrtuy: So what happens if the plugin fails when MDB calls write_row() third time. Does MDB inserts the 4th row and hits the original bug or MDB errors out?

tntnatbry: It's all or nothing. If the plugin fails on the 3rd call, MDB will error out and no rows will be inserted.

PR #1091, opened 2020-03-06
MCOL-3760 Fix rand func in columnstore not matching mdb

jmrojas2332:
Construction of Func_rand only happens once so we need to initialize fFirstRow and fSeeds just like fSeedSet, inside seedSet(), otherwise fSeeds and fFirstRow never get reinitialized.

### DIFF
--- a/utils/funcexp/functor_export.h
+++ b/utils/funcexp/functor_export.h
@@ -46,6 +46,8 @@ public:
     {
         fSeedSet = seedSet;
         fMultipleSeedsSet = seedSet;
+        fFirstRow = nullptr;
+        fSeeds = {};
     }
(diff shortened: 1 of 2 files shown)


utils/funcexp/functor_export.h
```diff
@@ -46,6 +46,8 @@ class Func_rand : public Func
     {
         fSeedSet = seedSet;
         fMultipleSeedsSet = seedSet;
+        fFirstRow = nullptr;
+        fSeeds = {};
```
jmrojas2332: This line will have to be slightly different on develop-1.2

davidhall586 (approved):

jmrojas2332:
Fix previous change breaking bug3145. Tested running regression suite test001, specifically looking at working_tpch1_calpontonly/misc/bug3145.sql. Now passing

fSeeds re-initialization was not thread-safe. Also added check before accessing invalid index of fSeeds

PR #1094, opened 2020-03-07
MCOL-3769 This commit divides create_SH in two parts: before rewrites…

drrtuy:
This commit divides create_SH in two parts : before rewrites and 
after rewrites. We can safely fallback from SH to DH and table mode from
'before rewrites' part. We also returns meaningful warning.

### DIFF
--- a/dbcon/mysql/ha_mcs_pushdown.cpp
+++ b/dbcon/mysql/ha_mcs_pushdown.cpp
@@ -27,7 +27,7 @@ void disable_indices_for_CEJ(THD *thd_)
     TABLE_LIST* global_list;
     for (global_list = thd_->lex->query_tables; global_list; global_list = global_list->next_global)
     {
-    // MCOL-652 - doing this with derived tables can cause bad things to happen
+        // MCOL-652 - doing this with derived tables can cause bad things to happen
         if (!global_list->derived)
         {
             global_list->index_hints= new (thd_->mem_root) List<Index_hint>();
@@ -42,7 +42,7 @@ void disable_indices_for_CEJ(THD *thd_)
     }
 }
 
-bool optimize_unflattened_subqueries_(SELECT_LEX *select_lex)
+bool optimize_unflattened_subqueries_mcs(SELECT_LEX *select_lex)
 {
     bool result = false;
     TABLE_LIST *tbl;
@@ -52,7 +52,7 @@ bool optimize_unflattened_subqueries_(SELECT_LEX *select_lex)
         if (tbl->is_view_or_derived())
         {
             SELECT_LEX *dsl = tbl->derived->first_select();
-            result = optimize_unflattened_subqueries_(dsl);
+            result = optimize_unflattened_subqueries_mcs(dsl);
         }
     }
 
@@ -516,7 +516,7 @@ create_columnstore_derived_handler(THD* thd, TABLE_LIST *derived)
     //To search for CROSS JOIN-s we use tree invariant
     //G(V,E) where [V] = [E]+1
     List<Item> join_preds_list;
-    TABLE_LIST *tl;   
+    TABLE_LIST *tl;
     for (tl = sl->get_table_list(); !unsupported_feature && tl; tl = tl->next_local)
     {
         Item_cond* where_icp= 0;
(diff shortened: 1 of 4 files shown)


dbcon/mysql/ha_mcs_pushdown.cpp
```diff
@@ -759,65 +759,36 @@ create_columnstore_select_handler(THD* thd, SELECT_LEX* select_lex)
     // MCOL-2178 Disable SP support in the select_handler for now.
     // Check the session variable value to enable/disable use of
     // select_handler
+    if (!get_select_handler(thd))
+    {
+        return handler;
+    }
+
+    // Disable SP support in the select_handler for now.
```
tntnatbry, line 766: This comment is duplicated above.

dbcon/mysql/ha_mcs_pushdown.cpp
```diff
@@ -830,31 +801,55 @@ create_columnstore_select_handler(THD* thd, SELECT_LEX* select_lex)
...
         }
     }
 
-    if (!unsupported_feature)
+    // We apply dedicated rewrites from MDB here so MDB's data structures
+    // becomes dirty and CS has to raise an error in case of any problem
+    // or unsupported feature.
+    handler= new ha_columnstore_select_handler(thd, select_lex);
+    JOIN *join= select_lex->join;
     {
```
tntnatbry, line 813: Do we need these braces now that the if condition is gone?

dbcon/mysql/ha_mcs_pushdown.cpp
```diff
@@ -830,31 +801,55 @@ create_columnstore_select_handler(THD* thd, SELECT_LEX* select_lex)
...
-        int rc= 0;
-        // Skip execution for EXPLAIN queries
-        if (!thd->lex->describe)
+        disable_indices_for_CEJ(thd);
+
+        if (select_lex->handle_derived(thd->lex, DT_MERGE))
         {
-            rc= ha_cs_impl_pushdown_init(&mhi, handler->table);
+            unsupported_feature = true;
+            handler->err_msg.assign("create_columnstore_select_handler(): Internal error happens in SL::handle_derived()");
```
tntnatbry: Instead of happens, occurred would be a better word here.

dbcon/mysql/ha_mcs_pushdown.cpp
```diff
@@ -830,31 +801,55 @@ create_columnstore_select_handler(THD* thd, SELECT_LEX* select_lex)
...
+                join->conds, TRUE, FALSE);
+        }
+        // MCOL-3747 IN-TO-EXISTS rewrite inside MDB didn't add
+        // an equi-JOIN condition.
+        if (!unsupported_feature
+            && optimize_unflattened_subqueries_mcs(select_lex))
+        {
             unsupported_feature = true;
-        return handler;
+            handler->err_msg.assign("create_columnstore_select_handler: Internal error happens in optimize_unflattened_subqueries_mcs()");
```
tntnatbry: Instead of happens, occurred would be a better word here.

dbcon/mysql/ha_mcs_pushdown.cpp
```diff
@@ -868,7 +863,8 @@ ha_columnstore_select_handler::ha_columnstore_select_handler(THD *thd,
                                                              SELECT_LEX* select_lex)
   : select_handler(thd, mcs_hton)
 {
-  select = select_lex;
+  select= select_lex;
+  rewrite_error= 0;
```
tntnatbry: should be = false

dbcon/mysql/ha_mcs_pushdown.cpp
```diff
@@ -892,10 +888,26 @@ int ha_columnstore_select_handler::init_scan()
...
 
-    // Dummy init for SH. Actual init happens in create_SH
-    // to allow fallback to other handlers if SH fails.
     int rc = 0;
 
+    if (!rewrite_error)
+    {
+        // handler::table is the place for the result set
+        // Skip execution for EXPLAIN queries
+        if (!thd->lex->describe)
```
tntnatbry, line 898: Curious, why was this moved from create_SH to here?

tntnatbry (changes requested):

dbcon/mysql/ha_mcs_pushdown.cpp
drrtuy: We do. This is context where we guard variables we don't need outside.

dbcon/mysql/ha_mcs_pushdown.cpp
drrtuy: Good question.
B/c create_SH is inappropriate place to execute the query(we can't even return and error w/o breaking the execution and init_SH is the place to execute the query.
I placed execution in create_SH in the first place to have a fallback but later understood it could break execution when we partially rewrote SELECT_LEX and then fallback to table API

dbcon/mysql/ha_mcs_pushdown.cpp
drrtuy: Thanks to rebase that might be error prone.

dbcon/mysql/ha_mcs_pushdown.cpp
drrtuy: To be honest it at this point the error happens for real not occured :)

pleblanc1976 (changes requested):
It looks good to me, but I don't know this code very well.  If it doesn't break anything in our regression test I'm OK with it.  Gagan can take the lead on this one.

drrtuy:
All relevant comments had been fixed.

tntnatbry (approved):

tntnatbry:
There is compilation errors in BB. Can you please rebase on top of current develop?

drrtuy:
Rebased.

tntnatbry (approved):

pleblanc1976 (approved):

PR #1095, opened 2020-03-07
MCOL-3769 This commit divides create_SH() in two parts.

drrtuy:
This commit divides create_SH in two parts : before rewrites and 
after rewrites. We can safely fallback from SH to DH and table mode from
'before rewrites' part. We also returns meaningful warning.

### DIFF
--- a/dbcon/mysql/ha_mcs_pushdown.cpp
+++ b/dbcon/mysql/ha_mcs_pushdown.cpp
@@ -27,7 +27,7 @@ void disable_indices_for_CEJ(THD *thd_)
     TABLE_LIST* global_list;
     for (global_list = thd_->lex->query_tables; global_list; global_list = global_list->next_global)
     {
-    // MCOL-652 - doing this with derived tables can cause bad things to happen
+        // MCOL-652 - doing this with derived tables can cause bad things to happen
         if (!global_list->derived)
         {
             global_list->index_hints= new (thd_->mem_root) List<Index_hint>();
@@ -42,7 +42,7 @@ void disable_indices_for_CEJ(THD *thd_)
     }
 }
 
-bool optimize_unflattened_subqueries_(SELECT_LEX *select_lex)
+bool optimize_unflattened_subqueries_mcs(SELECT_LEX *select_lex)
 {
     bool result = false;
     TABLE_LIST *tbl;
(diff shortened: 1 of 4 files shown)


tntnatbry (changes requested):
Similar comments to https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/1094 also apply here.

dbcon/mysql/ha_mcs_pushdown.cpp
```diff
@@ -892,7 +863,8 @@ ha_columnstore_select_handler::ha_columnstore_select_handler(THD *thd,
                                                              SELECT_LEX* select_lex)
   : select_handler(thd, mcs_hton)
 {
-  select = select_lex;
+  select= select_lex;
+  rewrite_error= 0;
```
tntnatbry: = false;

dbcon/mysql/ha_mcs_pushdown.cpp
```diff
@@ -811,74 +801,55 @@ create_columnstore_select_handler(THD* thd, SELECT_LEX* select_lex)
...
-    if (!unsupported_feature)
     {
         disable_indices_for_CEJ(thd);
 
         if (select_lex->handle_derived(thd->lex, DT_MERGE))
         {
-            // early quit b/c of the error in handle_derived
-            return handler;
+            unsupported_feature = true;
+            handler->err_msg.assign("create_columnstore_select_handler(): Internal error happens in SL::handle_derived()");
```
tntnatbry: occurred here instead of happens :)

dbcon/mysql/ha_mcs_pushdown.cpp
```diff
@@ -811,74 +801,55 @@ create_columnstore_select_handler(THD* thd, SELECT_LEX* select_lex)
...
         // MCOL-3747 IN-TO-EXISTS rewrite inside MDB didn't add
         // an equi-JOIN condition.
-        optimize_unflattened_subqueries_(select_lex);
-    
-        if (conds)
+        if (!unsupported_feature
+            && optimize_unflattened_subqueries_mcs(select_lex))
+        {
+            unsupported_feature = true;
+            handler->err_msg.assign("create_columnstore_select_handler: Internal error happens in optimize_unflattened_subqueries_mcs()");
```
tntnatbry: occurred instead of happens

tntnatbry (changes requested):

drrtuy:
I pushed the changes into a wrong remote. Now should be fine.

tntnatbry (approved):

PR #1096, opened 2020-03-10
MCOL-3760 Related. Fix bug3145 regression broken on rand()

jmrojas2332:
Tested running regression suite test001, specifically looking at working_tpch1_calpontonly/misc/bug3145.sql. Now passing

fSeeds re-initialization was not thread-safe. Also added check before accessing invalid index of fSeeds.

### DIFF
--- a/utils/funcexp/functor_export.h
+++ b/utils/funcexp/functor_export.h
@@ -38,7 +38,7 @@ namespace funcexp
 class Func_rand : public Func
 {
 public:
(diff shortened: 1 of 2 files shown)


utils/funcexp/functor_export.h
```diff
@@ -38,16 +38,16 @@ namespace funcexp
...
     virtual ~Func_rand() {}
 
     double getRand();
     void seedSet(bool seedSet)
     {
         fSeedSet = seedSet;
         fMultipleSeedsSet = seedSet;
-        fFirstRow = nullptr;
-        fSeeds = {};
+        fFirstRow = NULL;
```
jmrojas2332, line 49: keeping consistent with 1.2 fix

davidhall586 (approved):

PR #1103, opened 2020-03-13
Mcol 3251 1.5

pleblanc1976:
I could not reproduce the original problem, so I went to the code to generate the filenames.  It looks like the only way a filename with no leading dbroot path can be generated is if the dbroot went away in a narrow window between the jobs being issued to PrimProc, and when the job starts processing.

I decided to replace the method used by PrimProc.  The existing method does a Lot of Stuff, which primproc doesn't need, but which other callers might need.  The new method is much simpler than the old method, and will return a more useful file-not-found error if a dbroot suddenly goes offline, as what might be happening in the ticket.

### DIFF
--- a/primitives/blockcache/iomanager.cpp
+++ b/primitives/blockcache/iomanager.cpp
@@ -1356,18 +1356,69 @@ ioManager::ioManager(FileBufferMgr& fbm,
     }
 
     fThreadCount = thrCount;
+    lastConfigMTime = 0;
+    loadDBRootCache();
     go();
 }
 
-void ioManager::buildOidFileName(const BRM::OID_t oid, const uint16_t dbRoot, const uint16_t partNum, const uint32_t segNum, char* file_name)
+void ioManager::loadDBRootCache()
 {
+    if (fConfig->getLastMTime() == lastConfigMTime)
+        return;
+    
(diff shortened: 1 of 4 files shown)


primitives/blockcache/iomanager.cpp
```diff
@@ -1359,15 +1359,13 @@ ioManager::ioManager(FileBufferMgr& fbm,
     go();
 }
 
-void ioManager::buildOidFileName(const BRM::OID_t oid, const uint16_t dbRoot, const uint16_t partNum, const uint32_t segNum, char* file_name)
+void ioManager::buildOidFileName(const BRM::OID_t oid, uint16_t dbRoot, const uint16_t partNum, const uint32_t segNum, char* file_name)
```
tntnatbry: partNum is uint32_t and segNum is uint16_t space in the dbrm code. It is reversed here, please change.

tntnatbry (changes requested):

primitives/blockcache/iomanager.cpp
pleblanc1976: Good catch!  This must have been here since the beginning.

tntnatbry (approved):

PR #1106, opened 2020-03-16
MCOL-2096 Fix IN clause combined with a subquery and window function

jmrojas2332:
filter was never being applied with this combination. after a large amount of investigation the culprit was the IN clause was never being built as a function.

regression tests; https://github.com/mariadb-corporation/mariadb-columnstore-regression-test/pull/193


drrtuy:
I wonder if this issue also affects IN + correlated/non-correlated subqueries?

jmrojas2332:
Issue not yet fixed. PR has stalled.

PR #1118, opened 2020-03-20
MCOL-3828 This commit replaces the method that calls JOIN::optimise()

drrtuy:
internally(thus can remove equi-JOIN predicates) with a recursive f()
to inject equi-JOIN conditions for IN + correlated subquery Item.
This change both solves MCOL-3747 issue and remove issues I first
faced in union.sql test.

This commit also fixes a clean up for conditions pushed for table API
execution path and replaces NULL with nullptr in ha_mcs_impl.cpp.

### DIFF
--- a/dbcon/mysql/ha_mcs_opt_rewrites.cpp
+++ b/dbcon/mysql/ha_mcs_opt_rewrites.cpp
@@ -1,4 +1,4 @@
-/* Copyright (C) 2019 MariaDB Corporation
+/* Copyright (C) 2019-20 MariaDB Corporation
 
    This program is free software; you can redistribute it and/or
    modify it under the terms of the GNU General Public License
@@ -14,6 +14,9 @@
    along with this program; if not, write to the Free Software
    Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
    MA 02110-1301, USA. */
+
+#include <typeinfo>
+
 #include "ha_mcs_opt_rewrites.h"
 
 // Search simplify_joins() function in the server's code for detail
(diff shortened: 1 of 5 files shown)


dbcon/mysql/ha_mcs_opt_rewrites.cpp
```diff
@@ -243,3 +246,86 @@ simplify_joins_mcs(JOIN *join, List<TABLE_LIST> *join_list, COND *conds, bool to
...
+{
+    bool* result= reinterpret_cast<bool*>(arg);
+    if (*result) return;
+
+    Item* item= const_cast<Item*>(item_arg);
+
+    JOIN* join= nullptr;
+    if (typeid(*item) == typeid(Item_in_subselect))
+    {
+        Item_in_subselect* sub= reinterpret_cast<Item_in_subselect*>(item);
```
tntnatbry, line 271: I don't think reinterpret_cast is appropriate here, since we are downcasting from a base class to one of its derived classes.

dbcon/mysql/ha_mcs_opt_rewrites.cpp
```diff
@@ -243,3 +246,86 @@ simplify_joins_mcs(JOIN *join, List<TABLE_LIST> *join_list, COND *conds, bool to
...
+
+/*@brief  in_subselect_rewrite_walk - Rewrites Item_in_subselect*/
+/************************************************************
+* DESCRIPTION:
+* It traverses filter predicates searching for
+* Item_in_subselect and rewrites it adding equi-join predicate
+* to finalise IN_2_EXISTS rewrite.
+* PARAMETERS:
+*    item_arg - Item to check.
+*    arg - unused dummy arg.
```
tntnatbry: We are using arg, so it shouldn't be dummy here.

dbcon/mysql/ha_mcs_opt_rewrites.cpp
```diff
@@ -243,3 +246,86 @@ simplify_joins_mcs(JOIN *join, List<TABLE_LIST> *join_list, COND *conds, bool to
...
+    }
+}
+
+/*@brief  in_subselect_rewrite - Rewrites Item_in_subselect*/
+/************************************************************
+* DESCRIPTION:
+* It traverses TABLE_LISTs running in_subselect_rewrite_walk
+* PARAMETERS:
+*   select_lex
+* RETURN:
```
tntnatbry, line 308: Mention about the return value here?

dbcon/mysql/ha_mcs_opt_rewrites.cpp
```diff
@@ -243,3 +246,86 @@ simplify_joins_mcs(JOIN *join, List<TABLE_LIST> *join_list, COND *conds, bool to
...
+
+    Item* item= const_cast<Item*>(item_arg);
+
+    JOIN* join= nullptr;
+    if (typeid(*item) == typeid(Item_in_subselect))
+    {
+        Item_in_subselect* sub= reinterpret_cast<Item_in_subselect*>(item);
+        // MCS 1.4.3 doesn't support IN + subquery with UNION so
+        // we safe to take this JOIN.
+        join= sub->unit->first_select()->join;
```
tntnatbry, line 274: Curious, how does this logic work for the case of an IN + derived subquery? e.g.:
select field1 from t1 where field2 in (select * from (select field3 from t2) a);

dbcon/mysql/ha_mcs_pushdown.cpp
```diff
@@ -843,9 +813,19 @@ create_columnstore_select_handler(THD* thd, SELECT_LEX* select_lex)
...
             join->conds = conds;
         }
+
+        // MCOL-3747 IN-TO-EXISTS rewrite inside MDB didn't add
+        // an equi-JOIN condition.
+        if (!unsupported_feature && in_subselect_rewrite(select_lex))
+        {
+            unsupported_feature = true;
+            handler->err_msg.assign("create_columnstore_select_handler(): \
+                Internal error occured in optimize_unflattened_subqueries_mcs()");
```
tntnatbry: in_subselect_rewrite instead of optimize_unflattened_subqueries_mcs

tntnatbry (changes requested):

dbcon/mysql/ha_mcs_opt_rewrites.cpp
drrtuy: +1

dbcon/mysql/ha_mcs_opt_rewrites.cpp
drrtuy: Why so? We avoid needless checks of dynamic_cast given we have an appropriate if predicate.

dbcon/mysql/ha_mcs_opt_rewrites.cpp
drrtuy: There will be no join inside Item_in_subselect that describes IN subquery so we just inject equi-join predicate into IN subquery.

dbcon/mysql/ha_mcs_pushdown.cpp
drrtuy: +1

dbcon/mysql/ha_mcs_opt_rewrites.cpp
tntnatbry: This is good to know.

tntnatbry (approved):

dbcon/mysql/ha_mcs_opt_rewrites.cpp
tntnatbry: Not related to the PR, but I am curious what the difference in the runtime cost is for using a typeid vs a dynamic_cast.

PR #1120, opened 2020-03-20
MCOL-3809 Missed change when porting from 1.4 to 1.2

jmrojas2332:
Missed this minor change. verified ticket's tests are now working.


pleblanc1976 (approved):
LGTM.  I'm curious to see if BB will work, will merge after it finishes.

PR #1121, opened 2020-03-20
MCOL-3756 Implement IS(NOT)TRUE() and IS(NOT)FALSE() functions.

tntnatbry:


utils/funcexp/functor_bool.h
```diff
@@ -241,6 +242,94 @@ class Func_isnull : public Func_Bool
...
+public:
+    Func_Truth(const std::string& funcName, bool a_value, bool a_affirmative) :
+        Func_Bool(funcName), value(a_value), affirmative(a_affirmative) {}
+
+    virtual ~Func_Truth() {}
+
+    execplan::CalpontSystemCatalog::ColType operationType(FunctionParm& fp, execplan::CalpontSystemCatalog::ColType& resultType)
+    {
+        assert (fp.size() == 1);
+        return fp[0]->data()->resultType();
```
drrtuy, line 258: Shouldn't this return the type that corresponds with a boolean?

drrtuy (changes requested):
The logic is perfectly fine but branch misprediction is pain so I recalled discrete math and it seems this is a binary func of 4 args so I simplified it and got in the end: f(v1,v2,i,a) = !a && (!i || v1 != v2 ) || (i != a). Where v1 value, v2 val, i isNull, a affirmative. Given the value domain is {0,1} so we can even resort to bit ops:
f(v1,v2,i,a) = !a & (!i | v1 ^ v2 ) | (i ^ a). There might be needless parenthesis though.

utils/funcexp/functor_bool.h
tntnatbry: I am honestly not sure how operationType() is even used in the first place. I used the same code here as in Func_isnull::operationType

tntnatbry:
Very good feedback on the use of bitwise operation for this function. I have incorporated it and updated the PR.

drrtuy (approved):
The patch is OK to be merged. Let's wait untill BB will be able to test build it.

davidhall586 (approved):

PR #1126, opened 2020-03-26
MCOL-3903 Enable Select Handler to run query part of INSERT..SELECT.

drrtuy:
Original SH implementation sends the result set back to the client
thus it can't be used in INSERT..SELECT, SELECT INTO OUTFILE,CREATE
TABLE AS SELECT etc.
CLX-77 feature has been backported into MDB to enable SH to run
query part of the mentioned queries.

### DIFF
--- a/dbcon/mysql/ha_mcs_execplan.cpp
+++ b/dbcon/mysql/ha_mcs_execplan.cpp
@@ -3117,7 +3117,10 @@ CalpontSystemCatalog::ColType colType_MysqlToIDB (const Item* item)
     return ct;
 }
 
-ReturnedColumn* buildReturnedColumn(Item* item, gp_walk_info& gwi, bool& nonSupport, bool pushdownHand, bool isRefItem)
+ReturnedColumn* buildReturnedColumn(
+    Item* item, gp_walk_info& gwi,
+    bool& nonSupport,
+    bool isRefItem)
 {
     ReturnedColumn* rc = NULL;
 
@@ -3275,9 +3278,9 @@ ReturnedColumn* buildReturnedColumn(Item* item, gp_walk_info& gwi, bool& nonSupp
             }
 
             if (func_name == "+" || func_name == "-" || func_name == "*" || func_name == "/" )
-                return buildArithmeticColumn(ifp, gwi, nonSupport, pushdownHand);
+                return buildArithmeticColumn(ifp, gwi, nonSupport);
             else
-                return buildFunctionColumn(ifp, gwi, nonSupport, pushdownHand);
+                return buildFunctionColumn(ifp, gwi, nonSupport);
         }
 
         case Item::SUM_FUNC_ITEM:
@@ -3410,8 +3413,7 @@ ReturnedColumn* buildReturnedColumn(Item* item, gp_walk_info& gwi, bool& nonSupp
 ArithmeticColumn* buildArithmeticColumn(
     Item_func* item,
     gp_walk_info& gwi,
-    bool& nonSupport,
-    bool pushdownHand)
+    bool& nonSupport)
 {
     if (get_fe_conn_info_ptr() == NULL)
         set_fe_conn_info_ptr((void*)new cal_connection_info());
@@ -3435,7 +3437,7 @@ ArithmeticColumn* buildArithmeticColumn(
     {
         if (gwi.clauseType == SELECT || /*gwi.clauseType == HAVING || */gwi.clauseType == GROUP_BY || gwi.clauseType == FROM) // select list
         {
-            lhs = new ParseTree(buildReturnedColumn(sfitempp[0], gwi, nonSupport, pushdownHand));
+            lhs = new ParseTree(buildReturnedColumn(sfitempp[0], gwi, nonSupport));
 
             if (!lhs->data() && (sfitempp[0]->type() == Item::FUNC_ITEM))
             {
@@ -3443,7 +3445,7 @@ ArithmeticColumn* buildArithmeticColumn(
                 Item_func* ifp = (Item_func*)sfitempp[0];
                 lhs = buildParseTree(ifp, gwi, nonSupport);
             }
-            else if(pushdownHand && !lhs->data() && (sfitempp[0]->type() == Item::REF_ITEM))
+            else if(!lhs->data() && (sfitempp[0]->type() == Item::REF_ITEM))
             {
                 // There must be an aggregation column in extended SELECT
                 // list so find the corresponding column.
@@ -3454,7 +3456,7 @@ ArithmeticColumn* buildArithmeticColumn(
                 if(rc)
                     lhs = new ParseTree(rc);
             }
-            rhs = new ParseTree(buildReturnedColumn(sfitempp[1], gwi, nonSupport, pushdownHand));
+            rhs = new ParseTree(buildReturnedColumn(sfitempp[1], gwi, nonSupport));
 
             if (!rhs->data() && (sfitempp[1]->type() == Item::FUNC_ITEM))
             {
@@ -3462,7 +3464,7 @@ ArithmeticColumn* buildArithmeticColumn(
                 Item_func* ifp = (Item_func*)sfitempp[1];
                 rhs = buildParseTree(ifp, gwi, nonSupport);
             }
-            else if(pushdownHand && !rhs->data() && (sfitempp[1]->type() == Item::REF_ITEM))
+            else if(!rhs->data() && (sfitempp[1]->type() == Item::REF_ITEM))
             {
                 // There must be an aggregation column in extended SELECT
                 // list so find the corresponding column.
@@ -3638,7 +3640,6 @@ ReturnedColumn* buildFunctionColumn(
     Item_func* ifp,
     gp_walk_info& gwi,
     bool& nonSupport,
-    bool pushdownHand,
     bool selectBetweenIn)
 {
     if (get_fe_conn_info_ptr() == NULL)
@@ -3680,7 +3681,7 @@ ReturnedColumn* buildFunctionColumn(
     // Arithmetic exp
     if (funcName == "+" || funcName == "-" || funcName == "*" || funcName == "/" )
     {
-        ArithmeticColumn* ac = buildArithmeticColumn(ifp, gwi, nonSupport, pushdownHand);
+        ArithmeticColumn* ac = buildArithmeticColumn(ifp, gwi, nonSupport);
         return ac;
     }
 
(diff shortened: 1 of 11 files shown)


dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -7879,129 +7866,13 @@ int getSelectPlan(gp_walk_info& gwi, SELECT_LEX& select_lex,
...
-        // We don't currently support limit with correlated subquery
-        if (gwi.subQuery && !gwi.correlatedTbNameVec.empty() && csep->hasOrderBy())
-        {
-            gwi.fatalParseError = true;
-            gwi.parseErrorText = IDBErrorInfo::instance()->errorMsg(ERR_NON_SUPPORT_LIMIT_SUB);
-            setError(gwi.thd, ER_INTERNAL_ERROR, gwi.parseErrorText, gwi);
-            return ER_CHECK_NOT_IMPLEMENTED;
-        }
-    } // LIMIT processing finishes here
+        processLimitAndOffset(select_lex, gwi, csep, unionSel, isUnion, isSelectHandlerTop);
```
tntnatbry: We need to check for return value here, and return if it's an error.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -6544,25 +6544,167 @@ int processWhere(SELECT_LEX &select_lex,
...
+
+#else
+            limitOffset = (join->unit)->offset_limit_cnt;
+            limitNum = (join->unit)->select_limit_cnt - (join->unit)->offset_limit_cnt;
+#endif
+        }
+        else
+        {
+            if (select_lex.master_unit()->global_parameters()->offset_limit)
+                {
```
tntnatbry: The indentation from here to the rest of the function needs to be moved back 4 spaces.

utils/loggingcpp/ErrorMessage.txt
```diff
@@ -17,7 +17,7 @@
 # Non support errors 1000 ~ 2000. 
 # The query will go through the optimizer again with some optimization turned off
 1000	ERR_MISS_JOIN	%1% not joined.
-1001	ERR_NON_SUPPORTED_FUNCTION	Function '%1%' can only be used in the outermost select or order by clause and cannot be used in conjunction with an aggregate function.
+1001	ERR_NON_SUPPORTED_FUNCTION	Function '%1%' isn't supported.
```
tntnatbry, line 20: "is not" would be proper wording here.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -6544,25 +6544,167 @@ int processWhere(SELECT_LEX &select_lex,
...
+            setError(gwi.thd, ER_INTERNAL_ERROR, gwi.parseErrorText, gwi);
+            return ER_CHECK_NOT_IMPLEMENTED;
+        }
+
+        // MDB applies OFFSET on its own for SH processing.
+        // See related MDEV-16327 for details.
+        if (isSelectHandlerTop && csep->limitStart())
+        {
+            csep->limitNum(csep->limitNum()+csep->limitStart());
+            csep->limitStart(0);
```
tntnatbry, line 6691: I don't think this gives us correct results back.

tntnatbry (changes requested):
I ran a couple random queries on the patch and it seems like it introduced some regressions:

```sql
MariaDB [test]> create table t1 (a int)engine=columnstore;
Query OK, 0 rows affected (0.400 sec)

MariaDB [test]> insert into t1 values (11), (1), (3), (4), (2);
Query OK, 5 rows affected (0.283 sec)
Records: 5  Duplicates: 0  Warnings: 0

MariaDB [test]> select * from t1 limit 2,1;
+------+
| a    |
+------+
|   11 |
|    1 |
|    3 |
+------+
3 rows in set (0.038 sec)

MariaDB [test]> create table i1 as select * from t1;
+------+
| a    |
+------+
|   11 |
|    1 |
|    3 |
|    4 |
|    2 |
+------+
5 rows in set (0.037 sec)

```

drrtuy:
I don't see the degradation when I run the same test:

MariaDB [test]> create table t1 (a int)engine=columnstore;
Query OK, 0 rows affected (0.533 sec)

MariaDB [test]> insert into t1 values (11), (1), (3), (4), (2);
Query OK, 5 rows affected (0.424 sec)
Records: 5  Duplicates: 0  Warnings: 0

MariaDB [test]> select * from t1 limit 2,1;
+------+
| a    |
+------+
|    3 |
+------+
1 row in set (0.083 sec)

MariaDB [test]> create table i1 as select * from t1;
Query OK, 5 rows affected (0.048 sec)
Records: 5  Duplicates: 0  Warnings: 0

MariaDB [test]>

dbcon/mysql/ha_mcs_execplan.cpp
drrtuy: I would say it might be unsafe for future but as of now it returns 0 only.

dbcon/mysql/ha_mcs_execplan.cpp
drrtuy: Could you elaborate on this?

utils/loggingcpp/ErrorMessage.txt
drrtuy: That is the question of opinions :)

drrtuy:
I think I know what's the problem. You need to have server side patch to see how this patch works.

dbcon/mysql/ha_mcs_execplan.cpp
drrtuy: +1

pleblanc1976:
> I think I know what's the problem. You need to have server side patch to see how this patch works.

Gagan, the patch is currently in the enterprise repo in the 10.4-enterprise branch if you still need it.

tntnatbry:
I wasn't aware earlier that this patch has a server side counterpart. After I applied the server patch, I no longer see the above regressions.

dbcon/joblist/resourcemanager.h
```diff
@@ -36,6 +36,7 @@
 #include "resourcedistributor.h"
 #include "installdir.h"
 
+#define DEBUGRM 1
```
tntnatbry: I believe this is checked in accidentally.

tntnatbry (changes requested):

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -6544,25 +6544,176 @@ int processWhere(SELECT_LEX &select_lex,
...
+        setError(gwi.thd, ER_INTERNAL_ERROR, gwi.parseErrorText, gwi);
+        return ER_CHECK_NOT_IMPLEMENTED;
+    }
+
+    // MDB applies OFFSET on its own for SH processing.
+    // See related MDEV-16327 for details.
+    if (isSelectHandlerTop && csep->limitStart())
+    {
+        if (std::numeric_limits<uint64_t>::max()-csep->limitStart()
+            < csep->limitStart())
```
tntnatbry: I think the intention here is to check the overflow for the sum of csep->limitStart() and csep->limitNum().

tntnatbry (changes requested):

tntnatbry:
The 3rd insert statement in working_tpch1/misc/insert_select.sql is erroring out for me now:
```sql
MariaDB [tpch1]> insert into idb_lineitem (l_shipdate, l_partkey) (select  now(), l_partkey from lineitem where l_orderkey <= 1000 order by l_orderkey limit 100);
ERROR 1178 (42000): The storage engine for the table doesn't support IDB-1001: Function 'current_timestamp' isn't supported.
```

tntnatbry:
Hmm. Running the whole working_tpch1/misc/insert_select.sql test, 2 queries are giving non-matching results with the ref. One is mentioned above, the other is the 2nd last query which involves a CEJ.

```sql
root@tntnatbry-1:/home/gagan_goel_mariadb_com/mariadb-columnstore-regression-test/mysql/queries# time mysql -u root -f tpch1 < working_tpch1/misc/insert_select.sql > working_tpch1/misc/insert_select.sql.log
ERROR 1178 (42000) at line 31: The storage engine for the table doesn't support IDB-1001: Function 'current_timestamp' isn't supported.

real	0m3.690s
user	0m0.005s
sys	0m0.005s
root@tntnatbry-1:/home/gagan_goel_mariadb_com/mariadb-columnstore-regression-test/mysql/queries# diff -w working_tpch1/misc/insert_select.sql.log working_tpch1/misc/insert_select.sql.ref.log 
6c6
< 2004
---
> 2104
14c14
< 103
---
> 233
```

drrtuy:
I'm not sure that the failure comes with my patch.
CS doesn't support now() or current_timestamp() so this is an expected error given that INSERT..SELECT processing comes to CS. I need to address this in regression instead.

pleblanc1976:
> I'm not sure that the failure comes with my patch.
> CS doesn't support now() or current_timestamp() so this is an expected error given that INSERT..SELECT processing comes to CS. I need to address this in regression instead.

Were they working before the patch?

tntnatbry:
Yeah before the patch, this test ran fine. As Roman mentioned, it is due to the insert select now getting executed by the select handler instead of the table API, and since we don't support current_timestamp() function yet, hence the error.

I think it would be better if we run test001 before and after the patch to see if there are any other deviations.

pleblanc1976:
> Yeah before the patch, this test ran fine. As Roman mentioned, it is due to the insert select now getting executed by the select handler instead of the table API, and since we don't support current_timestamp() function yet, hence the error.
> 
> I think it would be better if we run test000 before and after the patch to see if there are any other deviations.

This may be a problem.  Inserting rows with a current timestamp isn't uncommon.  Let's be sure Greg & Todd know of this limitation.

drrtuy:
Inserting rows with now() isn't uncommon however it isn't very true about INSERT..SELECT.
The issue with now() and current_timestamp() is [known](https://jira.mariadb.org/browse/MCOL-3902). We don't have them in funcexp and we don't need to be honest. Any now() or current_timestamp() should be replaced by a constant. But this goes out of the scope of the issue. There is a way to disable SH for the potentially problematic queries though.

drrtuy:
> Yeah before the patch, this test ran fine. As Roman mentioned, it is due to the insert select now getting executed by the select handler instead of the table API, and since we don't support current_timestamp() function yet, hence the error.
> 
> I think it would be better if we run test001 before and after the patch to see if there are any other deviations.

Sure. You should run test001 if you see it necessary. 
I didn't see any regressions running it and introduced by the patch.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -6807,7 +6958,7 @@ int getSelectPlan(gp_walk_info& gwi, SELECT_LEX& select_lex,
                     // @bug3881. set_user_var can not be treated as constant function
                     // @bug5716. Try to avoid post process function for union query.
                     if ((gwi.subQuery || select_lex.group_list.elements != 0 ||
-                            !csep->unionVec().empty() || isUnion || isPushdownHand ) &&
+                            !csep->unionVec().empty() || isUnion ) &&
```
tntnatbry: I missed it in the first pass. Since isPushdownHand is true, it should set the whole block in parenthesis to true.

tntnatbry (changes requested):

drrtuy:
This inconsistency in insert_select.sql points to a bigger change. We now obey isolation levels rules when autocommit is disabled. That is why CEJ doesn't see records in tpch1.myisam_lineitem. This goes out of the scope and we don't have tests to cover this behavior in our regression.

tntnatbry (approved):

PR #1130, opened 2020-03-28
MCOL-2221: Support UTF-8 characters for reverse() sql function

sq2huang:
Change char buffer to wchar_t buffer to support UTF8 characters.

### DIFF
--- a/utils/funcexp/func_reverse.cpp
+++ b/utils/funcexp/func_reverse.cpp
@@ -26,6 +26,7 @@ using namespace std;
 
 #include "functor_str.h"
 #include "functioncolumn.h"
+#include "utils_utf8.h"
 using namespace execplan;
 
 #include "rowgroup.h"
@@ -39,11 +40,11 @@ using namespace joblist;
 namespace
 {
 
-void reverse( char* start, char* end )
+void reverse( wchar_t* start, wchar_t* end )
 {
     while ( start < end )
     {
-        char c = *start;
+        wchar_t c = *start;
         *start++ = *end;
         *end-- = c;
     }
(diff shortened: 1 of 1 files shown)


drrtuy:
Greetings. 
Thank you very much for the contribution. Could you post here that you contribute your patch under BSL license?

utils/funcexp/func_reverse.cpp
```diff
@@ -73,13 +74,15 @@ std::string Func_reverse::getStrVal(rowgroup::Row& row,
...
     // which this one derived.
-    int len = str.length();
-    char* pbuf = new char[len + 1];
-    strncpy(pbuf, str.c_str(), len);
-    pbuf[len] = 0;
-    char* end = pbuf + len - 1;
+    wstring wstr = utf8::utf8_to_wstring(str);
+    int wlen = wstr.length();
+    wchar_t* pbuf = new wchar_t[wlen + 1];
+    wcsncpy(pbuf, wstr.c_str(), wlen);
```
drrtuy: I think we should avoid this copy operation sending down 4 pointers into reverse(). begin and end pointers for the source string and two similar for the destination string.

sq2huang:
> Greetings.
> Thank you very much for the contribution. Could you post here that you contribute your patch under BSL license?

I contribute this patch under BSL license

utils/funcexp/func_reverse.cpp
```diff
@@ -73,13 +74,14 @@ std::string Func_reverse::getStrVal(rowgroup::Row& row,
...
-    char* end = pbuf + len - 1;
-    reverse(pbuf, end);
-    string rstr = pbuf;
+    wstring wstr = utf8::utf8_to_wstring(str);
+    int wlen = wstr.length();
+    wchar_t* pbuf = new wchar_t[wlen + 1];
+    pbuf[wlen] = 0;
+    wchar_t* end = pbuf + wlen - 1;
+    reverse(wstr.c_str(), wstr.c_str() + wlen - 1, pbuf, end);
+    wstring wrstr = pbuf;
```
drrtuy, line 83: I wonder if we really need this assignment  or we could use wrstr to instantiate rstr?

sq2huang: Do you mean get rid of pbuf and write the reversed string directly into wrstr?

drrtuy:
Thanks for the contribution again.

drrtuy (approved):

PR #1134, opened 2020-04-02
MCOL-2096 Fix evaluation for derived table inside ConstantFilter

jmrojas2332:
fDerivedTable inside ConstantFilter was not being set correctly. ConstantFilter::setDerivedTable should call setDerivedTable on each SimpleFilter and retrieve derivedTable for one of them (they will all be the same derived table so retrieving the first will suffice).

See mariadb-corporation/mariadb-columnstore-regression-test#193 for regression test updates.

### DIFF
--- a/dbcon/execplan/constantfilter.cpp
+++ b/dbcon/execplan/constantfilter.cpp
@@ -272,8 +272,19 @@ void ConstantFilter::setDerivedTable()
         fDerivedTable = "";
         return;
     }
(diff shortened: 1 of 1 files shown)


drrtuy (commented):
The change itself affects derivedTableOptimization() and the CSEP after the change looks appropriate.

drrtuy (approved):

PR #1135, opened 2020-04-06
MCOL-3842 -- Failover fixes with glusterFS and general failover fixes.

benthompson15:


procmgr/processmanager.cpp
```diff
@@ -9198,6 +9193,14 @@ int ProcessManager::switchParentOAMModule(std::string newActiveModuleName)
 
     //clear run standby flag;
     runStandby = false;
+    int retryCount = 0;
+    //sleep, give time for message thread to startup
+    while (!MsgThreadActive && retryCount < 5)
```
pleblanc1976: should this be 'retryCount < 10' like the other loop?

pleblanc1976 (approved):

PR #1143, opened 2020-04-09
MCOL-3924 Use the first parse error found

davidhall586:
Often, after an error is found, it cascades into a bunch of errors, and each overwrites the previous. This sometimes obfuscates the true problem.


jmrojas2332 (approved):
LGTM

PR #1147, opened 2020-04-10
ha_mcs_impl_direct_update_delete_rows handle delete with replication enabled

benthompson15:


pleblanc1976:
Don't merge yet, need to make sure it's the right fix, and that it's fixed everywhere it should be.

dbcon/mysql/ha_mcs_impl.cpp
```diff
@@ -2291,6 +2291,17 @@ int ha_mcs_impl_direct_update_delete_rows(bool execute, ha_rows *affected_rows)
...
+    if (thd->slave_thread && !get_replication_slave(thd) && (
+                thd->lex->sql_command == SQLCOM_INSERT ||
+                thd->lex->sql_command == SQLCOM_INSERT_SELECT ||
+                thd->lex->sql_command == SQLCOM_UPDATE ||
+                thd->lex->sql_command == SQLCOM_UPDATE_MULTI ||
+                thd->lex->sql_command == SQLCOM_DELETE ||
+                thd->lex->sql_command == SQLCOM_DELETE_MULTI ||
+                thd->lex->sql_command == SQLCOM_TRUNCATE ||
+                thd->lex->sql_command == SQLCOM_LOAD))
+        return 0;
```
pleblanc1976, line 2306: Before returning, add
'if (affected_rows)
   *affected_rows = 0;'

pleblanc1976 (changes requested):
Need to add the same DML check to ha_cs_impl_pushdown_init(), just before the call to doUpdateDelete(), and add the 'affected_rows' assignment mentioned in ha_mcs_impl_direct_update_delete_rows().

pleblanc1976 (approved):

PR #1149, opened 2020-04-14
MCOL-3915 Move columnstore-post-install and postConfigure logic into package manager's scripts.

jmrojas2332:
MCOL-3914, Create systemd unit files for Columnstore service, included as well.

columnstore-post-install is called upon package installation, where we setup configuration for a single node using a simplified postConfigure that requires no user input, and then columnstore-post-install finished by kicking off columnstore.service which starts up the other systemd units for Columnstore.

Note: postConfigure.cpp and installer.cpp could still probably be cleaned up a bit


jmrojas2332:
PR was requested into wrong branch. Closed

PR #1157, opened 2020-04-17
Restart DBRMController on failover

benthompson15:


procmgr/processmanager.cpp
```diff
@@ -3836,6 +3836,7 @@ void ProcessManager::reinitProcesses()
 {
     log.writeLog(__LINE__, "reinitProcesses... ", LOG_TYPE_DEBUG);
 
+    restartProcessType("DBRMControllerNode");
```
pleblanc1976, line 3839: The worker is normally started before the controller.  However, since the system is sort of up at this point and there may be BRM activity, what we might have to do to avoid races is 
1) stop the controller
2) stop the worker
3) start the worker
4) start the controller

pleblanc1976 (changes requested):

pleblanc1976 (approved):

PR #1170, opened 2020-04-23
fixes for cpimport and locking issues with processes restarting / reinit

benthompson15:


procmgr/main.cpp
```diff
@@ -2085,16 +2145,21 @@ void pingDeviceThread()
...
+                                    log.writeLog(__LINE__, "'dbrmctl halt' done", LOG_TYPE_DEBUG);
+
                                     //set query system state not ready
                                     processManager.setQuerySystemState(false);
 
                                     processManager.setSystemState(oam::BUSY_INIT);
 
-                                    processManager.reinitProcessType("cpimport");
+                                    //call dbrm control
+                                    oam.dbrmctl("reload");
```
pleblanc1976, line 2161: I see a few lines later there is a disableModule call.  Has the config file been updated at this point?

procmgr/main.cpp
```diff
@@ -2381,10 +2476,6 @@ void pingDeviceThread()
                                 }
                                 else
                                 {
-                                    processManager.distributeConfigFile("system");
```
pleblanc1976, line 2384: Between the previous change and this change, I don't see an unconditional call to distributeConfigFile().  Maybe it is called deeper in the call stack?

procmgr/processmanager.cpp
```diff
@@ -10098,8 +10076,11 @@ int ProcessManager::OAMParentModuleChange()
...
     }
 
+    reinitProcessType("cpimport");
+
     // waiting until dml are ACTIVE
-    while (true)
+    int retry = 0;
+    while (retry < 30)
     {
         ProcessStatus DMLprocessstatus;
```
pleblanc1976, line 10085: In this and all of the new wait-for-dml blocks, what is ProcessOpState if/when getProcessStatus() throws an exception?

procmon/processmonitor.cpp
```diff
@@ -809,6 +810,24 @@ void ProcessMonitor::processMessage(messageqcpp::ByteStream msg, messageqcpp::IO
...
+                            fgets(buf, 512, cmd_pipe);
+                            pid_t pid = strtoul(buf, NULL, 10);
+
+                            pclose( cmd_pipe );
+
+                            if (pid)
+                                sleep(2);
+                            else
+                                break;
+                        }
```
pleblanc1976, line 828: Looks like the purpose of this loop is to wait up to 20s for cpimport to go away before killing cpimport.bin on the local machine, is that right?

procmgr/main.cpp
benthompson15: This found to be needed when a failover occurs during cpimport to keep dbrm from deadlocking. it shouldn't be an issue on failovers without cpimport, it may hang a minute while the downed node timesout trying to reconnect. and continue on with failover processes.

pleblanc1976 (approved):
good enough given the circumstances.

PR #1171, opened 2020-04-23
MCOL-3959 1.4 make from_unixtimestamp msec aware

davidhall586:
When using the function with decimal or float types, the fractional portion is the msec.

PR #1172, opened 2020-04-23
MCOL-3959 1.5 make from_unixtimestamp msec aware

davidhall586:
When using the function with decimal or float types, the fractional portion is the msec.


tntnatbry (changes requested):
See my comment in https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/1171.

tntnatbry (approved):

PR #1176, opened 2020-04-24
MCOL-3839 Count as window function doesn't work with NULLs

jmrojas2332:
We were incorrectly handling NULL as an argument for wf_count. fFieldIndex was returning -1 but we were storing as unsigned int causing unexpected results and a ExeMgr crash

### DIFF
--- a/utils/windowfunction/wf_count.cpp
+++ b/utils/windowfunction/wf_count.cpp
@@ -120,11 +120,12 @@ void WF_count<T>::operator()(int64_t b, int64_t e, int64_t c)
             if (cc)
             {
                 bool isNull = false;
-                int val = cc->getIntVal(fRow, isNull);
+                cc->getIntVal(fRow, isNull);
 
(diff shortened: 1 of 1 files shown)


davidhall586 (changes requested):
This solution has the incorrect assumption that the -1 is caused by a NULL parameter. What is actually happening is that because the parameter is a constant, the field index is -1. Any constant will do this.
This solution results in any constant parameter being treated as NULL, which is incorrect.
A more correct solution is to look at the value of fConstantColumns[0] for the constant value. Unfortunately, the value inserted here for NULL is 0 (incorrect). Perhaps adding a line in ConstantColumn.cpp after if (fType == ConstantColumn::NULLDATA) to push the actual NULL value in would work.

jmrojas2332 (commented):
I have updated change to work for other constants as well as NULLs.

davidhall586 (approved):

PR #1187, opened 2020-05-06
MCOL-3982 Avoid using autotools for libmarias3.

drrtuy:


pleblanc1976 (approved):
I question the reason to do this, but assume others have decided it's the right way to handle the problem. 
 The change itself is OK.

PR #1191, opened 2020-05-07
MCOL-3982 Avoid using autotools for libmarias3 and remove libxml2 fro…

drrtuy:
…m libmarias3 dependencies.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,7 +24,6 @@ Makefile
 cmake_install.cmake
 install_manifest.txt
 CTestTestfile.cmake
-config.h
 config.status
 stamp-h1
 export/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -311,11 +311,6 @@ SET (ENGINE_COMMON_INCLUDES  ${ENGINE_DEFAULT_INCLUDES} ${Boost_INCLUDE_DIR} ${L
 
 ADD_SUBDIRECTORY(utils)
 
-add_library(marias3 SHARED IMPORTED)
-set_target_properties(marias3 PROPERTIES IMPORTED_LOCATION ${CMAKE_BINARY_DIR}/lib/libmarias3.so.3.1.2)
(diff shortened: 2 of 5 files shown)


pleblanc1976 (approved):

pleblanc1976:
Waiting on the end of the freeze on dev-1.4

pleblanc1976 (changes requested):
Dave is telling me this broke the build on his C7 machine, with the error msg:
```
/home/calpont/mariadb/engine/utils/libmarias3/libmarias3/src/debug.c:20:20: fatal error: config.h: No such file or directory
 #include "config.h"
```
That file is generated by autoconf so it makes sense that it would be missing.  I'm not sure why the C7 PR builder succeeded though.  Getting the wrong PR code?  Is it finding another config.h?  Need to follow up.  Will need to follow up with 1.5 also.

drrtuy:
B/c BB follows the way ENT builders use to produce packages, namely it builds MCS having it under storage/columnstore. Server team builds libmarias3 for ARIA engine also. Building the engine this way compiler finds config.h in MDB_SRC/include.
JFYI David builds it out-of-source as we did in 1.2 era.

pleblanc1976 (approved):
Tentatively thinking this is OK

pleblanc1976:
This might be a workaround for a weird SLES 15 build problem we're having in Jenkins.  It's failing with the error 
`checking C++ Compiler version... /home/jenkins/workspace/10.4e-RPM-DEV/label/sles-15/padding_for_CPACK_RPM_BUILD_SOURCE_DIRS_PREFIX_ON_ES_BACKUP_DEBUGSOURCE/storage/columnstore/columnstore/utils/libmarias3/libmarias3/configure: line 13400: --version: command not found`

... as if it's trying to execute '--version' instead of 'g++ --version'.  IDK if this will break out-of-source builds, but we can fix that if necessary afterward.

PR #1192, opened 2020-05-08
MCOL-641 Simple aggregates support: min, max, sum, avg for wide-DECIM…

drrtuy:
…ALs.

### DIFF
--- a/dbcon/joblist/groupconcat.cpp
+++ b/dbcon/joblist/groupconcat.cpp
@@ -635,7 +635,7 @@ int64_t GroupConcator::lengthEstimate(const rowgroup::Row& row)
             case CalpontSystemCatalog::DECIMAL:
             case CalpontSystemCatalog::UDECIMAL:
             {
-                fieldLen += row.getPrecision(*i)/2;
+                fieldLen += 1;
 
                 break;
             }
--- a/dbcon/joblist/groupconcat.cpp
+++ b/dbcon/joblist/groupconcat.cpp
@@ -454,17 +454,32 @@ void GroupConcator::outputRow(std::ostringstream& oss, const rowgroup::Row& row)
             case CalpontSystemCatalog::UDECIMAL:
             {
                 int scale = (int) row.getScale(*i);
-                //{
-                //    long double dblVal = intVal / pow(10.0, (double)scale);
-                //    oss << fixed << setprecision(scale) << dblVal;
-                //}
-                char buf[utils::MAXLENGTH16BYTES];
- 
-                int128_t* dec = row.getBinaryField<int128_t>(*i);
-                dataconvert::DataConvert::decimalToString(dec,
-                    static_cast<uint32_t>(scale), buf,
-                    sizeof(buf), types[*i]);
-                oss << fixed << buf;
+
+                if (LIKELY(row.getColumnWidth(*i) == datatypes::MAXDECIMALWIDTH))
+                {
+                    char buf[utils::MAXLENGTH16BYTES];
+
+                    int128_t* dec = row.getBinaryField<int128_t>(*i);
+                    dataconvert::DataConvert::decimalToString(dec,
+                        static_cast<uint32_t>(scale), buf,
+                        sizeof(buf), types[*i]);
+                    oss << fixed << buf;
+                }
+                else
+                {
+                    int64_t intVal = row.getIntField(*i);
+                    if (scale == 0)
+                    {
+                        oss << intVal;
+                    }
+                    else
+                    {
+                        char buf[utils::MAXLENGTH8BYTES];
+                        dataconvert::DataConvert::decimalToString(intVal,
+                             scale, buf, sizeof(buf), types[*i]);
+                        oss << fixed << buf;
+                    }
+                }
 
                 break;
             }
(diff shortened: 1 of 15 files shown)


utils/rowgroup/rowaggregation.cpp
```diff
@@ -1174,18 +1204,18 @@ void RowAggregation::makeAggFieldsNull(Row& row)
...
-                    int128_t nullValue = 0;
-                    utils::setWideDecimalNullValue(nullValue);
                     uint32_t offset = row.getOffset(colOut);
-                    row.setBinaryField_offset(&nullValue, sizeof(nullValue),
+                    row.setBinaryField_offset(
+                        const_cast<int128_t*>(&datatypes::Decimal128Null),
+                        colWidth,
                         offset);
                 }
+                else if (colWidth == datatypes::MAXLEGACYWIDTH)
```
tntnatbry: Shouldn't this be <=, instead of ==?

utils/rowgroup/rowaggregation.cpp
```diff
@@ -2666,14 +2703,38 @@ void RowAggregationUM::calculateAvgColumns()
...
+                    int128_t* sumPnt = fRow.getBinaryField_offset<int128_t>(offset);
+                    uint32_t scaleDiff = scale - fRow.getScale(colOut);
+                    // multiplication overflow check
+                    datatypes::MultiplicationOverflowCheck multOp;
+                    int128_t sum = 0;
+                    if (scaleDiff > 0)
+                        multOp(*sumPnt, datatypes::mcs_pow_10[scaleDiff], sum);
+                    else
+                        sum = *sumPnt;
+                    int128_t avg = sum / cnt;
```
tntnatbry, line 2762: I think we can have a rounding error here, e.g. if sum=14, cnt=5, avg=2.8. This should be rounded up to 3, and not 2. We can refer to the divisionExecute() code in mcs_decimal.cpp for similar rounding handling.

dbcon/joblist/groupconcat.cpp
```diff
@@ -606,27 +635,7 @@ int64_t GroupConcator::lengthEstimate(const rowgroup::Row& row)
...
-                }
-                else
-                {
-                    if (v < 0) fieldLen++;
-
-                    while ((v /= 10) != 0) fieldLen++;
-
-                    fieldLen += 1;;
-                }
+                fieldLen += row.getPrecision(*i)/2;
```
tntnatbry: Why are we dividing by 2 here?

tntnatbry (changes requested):

utils/rowgroup/rowaggregation.cpp
drrtuy: Good catch. Yes, it must be <=

utils/rowgroup/rowaggregation.cpp
drrtuy: That implementation has its drawbacks, namely float128_t looses its precision for some numbers that utilizes most of 38 digits available. With the rounding applied avg(38)_test1 returns 0.01122334445566778844556666161821143552 for the avg(d3) column whilst MDB and algo w/o rounding both return 0.01122334445566778844556666161821143636. The former has a higher measure of inaccuracy if we multiply the calculated average to get the original sum.
This is somehow expected given that float must have a space for its exponent. I'll try to come up with a solution or resort to divisionExecute() approach for math.

dbcon/joblist/groupconcat.cpp
drrtuy: To simplify expected row length calculation. Why I took / 2. This is an assumption that every element of a concatenated string uses half of its precision. However now I see it should be at least scale + half of the prec - scale.
I've also created an issue to refactor group_concat code removing length estimation.

tntnatbry: Originally, we were calculating the fieldLen as the complete length of the DECIMAL here, now we are setting it to 1. But it looks like for DOUBLE/FLOAT fields as well, we are setting fieldLen=1. I am not sure why.  I obviously don't have the full context on this function.

utils/rowgroup/rowaggregation.cpp
```diff
@@ -1054,10 +1054,15 @@ void RowAggregation::initMapData(const Row& rowIn)
...
                 }
-                else
+                else if (fRow.getColumnWidth(colIn) <= datatypes::MAXLEGACYWIDTH)
                 {
                     fRow.setIntField(rowIn.getIntField(colIn), colOut);
                 }
+                else
+                {
+                    idbassert(0);
+                    throw std::logic_error("RowAggregation::initMapData(): DECIMAL bad length.");
```
tntnatbry, line 1064: Just curious, why do we need idbassert(0) along with throw logic_error() here?

drrtuy: If we have decimals wider then 16  bytes we assert here for them.

dbcon/joblist/groupconcat.cpp
drrtuy: I bet this is because of complexity of the actual length calculationfor floats and doubles. We will optimize out this length estimation in future.

utils/rowgroup/rowaggregation.cpp
pleblanc1976: I noticed this PR targets a feature branch not a release branch, is that intentional?

drrtuy: Yes. The feature isn't finished yet.

tntnatbry (approved):

PR #1193, opened 2020-05-08
MCOL-3985: non-systemd mysqld process restart fix

benthompson15:
add logic to check mysqld is stopped before starting again when dealing with non-systemd enabled systems.

### DIFF
--- a/oam/oamcpp/liboamcpp.cpp
+++ b/oam/oamcpp/liboamcpp.cpp
@@ -7869,8 +7869,8 @@ void Oam::actionMysqlCalpont(MYSQLCALPONT_ACTION action)
                     else
                     {
                         writeLog("***mysqld waiting for shutdown complete", LOG_TYPE_DEBUG);
-                        sleep(2);
                     }
+                    sleep(2);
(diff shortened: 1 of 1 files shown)


oam/oamcpp/liboamcpp.cpp
```diff
@@ -7865,7 +7871,28 @@ void Oam::actionMysqlCalpont(MYSQLCALPONT_ACTION action)
...
+                    if (kill(pidStatus,0) == -1)
+                    {
+                        if (errno == ESRCH)
+                        {
+                            writeLog("***mysqld shutdown complete", LOG_TYPE_DEBUG);
+                            break;
+                        }
+                        else
+                        {
+                            writeLog("***mysqld errno = " + string(strerror(errno)), LOG_TYPE_DEBUG);
```
pleblanc1976, line 7906: if errno != ESRCH, this will loop like mad and be done instantly

pleblanc1976 (approved):

pleblanc1976 (approved):

PR #1196, opened 2020-05-09
MCOL-3813 Count with view is incorrect

jmrojas2332:
This is not actually an issue with count, however, it is an issue that pertains to a view used in subquery.

Previously a fix for MCOL-1349 reinstated some code 0b32f95 where outer joins with views were being processed twice. There is a case where a view inside a subquery was not processing it's outer join at all. Added a fix here to allow view in subquery to process outerjoin.

This fixes regression test queries/working_tpch1_compareLogOnly/misc/bug5764.sql and i have added a separate regression test to test MCOL-1349 is in working state. mariadb-corporation/mariadb-columnstore-regression-test#219


drrtuy:
I wonder if you tested for potential regressions b/c the change scope might be unpredictable.

davidhall586 (approved):

PR #1197, opened 2020-05-10
Improve batch inserts.

tntnatbry:
1) Instead of making dbrm calls to writeVBEntry() per block,
     we make these calls per batch. This can have non-trivial
     reductions in the overhead of these calls if the batch size
     is large.

  2) In dmlproc, do not deserialize the whole insertpackage, which
     consists of the complete record set per column, which would be
     wasteful as we only need some metadata fields from insertpackage
     here. This is only done for batch inserts at the moment, this
     should also be applied to single inserts.

### DIFF
--- a/dbcon/dmlpackage/dmltable.cpp
+++ b/dbcon/dmlpackage/dmltable.cpp
@@ -72,6 +72,28 @@ int DMLTable::read(messageqcpp::ByteStream& bytestream)
     return retval;
 }
 
+void DMLTable::readMetaData(messageqcpp::ByteStream& bytestream)
+{
+    // read the table name
+    bytestream >> fName;
+
+    // read the schema name
+    bytestream >> fSchema;
+}
+
+void DMLTable::readRowData(messageqcpp::ByteStream& bytestream)
+{
+    messageqcpp::ByteStream::quadbyte rowNum;
+    bytestream >> rowNum;
+
+    for (unsigned int i = 0; i < rowNum; i++)
+    {
+        Row* aRow = new Row();
+        aRow->read(bytestream);
+        fRows.push_back(aRow);
+    }
+}
+
 int DMLTable::write(messageqcpp::ByteStream& bytestream)
 {
     int retval = 1;
--- a/dbcon/dmlpackage/insertdmlpackage.cpp
+++ b/dbcon/dmlpackage/insertdmlpackage.cpp
@@ -66,15 +66,15 @@ int InsertDMLPackage::write(messageqcpp::ByteStream& bytestream)
     bytestream << (uint8_t)fLogging;
     bytestream << (uint8_t)fLogending;
 
-    if (fTable != 0)
-    {
-        retval = fTable->write(bytestream);
-    }
-
     bytestream << fTableOid;
     bytestream << static_cast<const messageqcpp::ByteStream::byte>(fIsInsertSelect);
     bytestream << static_cast<const messageqcpp::ByteStream::byte>(fIsBatchInsert);
     bytestream << static_cast<const messageqcpp::ByteStream::byte>(fIsAutocommitOn);
+
+    if (fTable != 0)
+    {
+        retval = fTable->write(bytestream);
+    }
 
     return retval;
 }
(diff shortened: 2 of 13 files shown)


dbcon/dmlpackage/dmltable.cpp
```diff
@@ -72,6 +72,28 @@ int DMLTable::read(messageqcpp::ByteStream& bytestream)
...
+}
+
+void DMLTable::readRowData(messageqcpp::ByteStream& bytestream)
+{
+    messageqcpp::ByteStream::quadbyte rowNum;
+    bytestream >> rowNum;
+
+    for (unsigned int i = 0; i < rowNum; i++)
+    {
+        Row* aRow = new Row();
```
pleblanc1976, line 91: where do these get deleted?

pleblanc1976: I see new's but no deletes in this patch.  If the lifecycle of the mem is unclear, feel free to use smart ptrs.

dbcon/dmlpackage/insertdmlpackage.cpp
```diff
@@ -100,15 +100,50 @@ int InsertDMLPackage::read(messageqcpp::ByteStream& bytestream)
...
+
+    bytestream >> fTableOid;
+    bytestream >> reinterpret_cast<messageqcpp::ByteStream::byte&>(fIsInsertSelect);
+    bytestream >> reinterpret_cast<messageqcpp::ByteStream::byte&>(fIsBatchInsert);
+    bytestream >> reinterpret_cast<messageqcpp::ByteStream::byte&>(fIsAutocommitOn);
+
+    fTable = new DMLTable();
+    fTable->readMetaData(bytestream);
+}
+
```
pleblanc1976, line 140: Is readMetadata() different than read()?  I don't see a difference except for the return val, am I just missing it?

pleblanc1976 (changes requested):
Looks ok, pending info on lifecycle of objects alloc'd in the patch.

versioning/BRM/dbrm.cpp
```diff
@@ -2226,6 +2226,42 @@ int DBRM::writeVBEntry(VER_t transID, LBID_t lbid, OID_t vbOID,
     return err;
 }
 
+int DBRM::bulkWriteVBEntry(VER_t transID,
+                           const std::vector<BRM::LBID_t>& lbids,
+                           OID_t vbOID,
+                           const std::vector<uint32_t>& vbFBOs) DBRM_THROW
```
drrtuy, line 2232: I'm not sure if DBRM_THROW still makes sense in C++11.

dbcon/dmlpackage/dmltable.cpp
drrtuy: You won't see them b/c it happens inside DMLTable dtor so we are relatively safe here. There will be a leak in case of exception though. But this is an issue with the original code that Gagan reused.

dbcon/dmlpackage/insertdmlpackage.cpp
drrtuy: It doesn't deser an array of records values from the stream.

drrtuy (approved):
The global downside of this change is that write VB become less granular and holds locks for a longer time for INSERT with many values. I can't guess whether it is good or not. With the workloads I am aware of it should be harmless.

dbcon/dmlpackage/dmltable.cpp
tntnatbry: This is correct. As Roman mentioned, I am reusing the existing code here. The respective deletes are happening in the dtors of InsertDMLPackage, DMLTable and Row objects.

dbcon/dmlpackage/insertdmlpackage.cpp
tntnatbry: Exactly. This is the main reason to add this method; to avoid deserializing the row data containing the vectors of column values, which is wasteful in dmlproc as we only need a few metadata fields from the dml package there, and not the actual row data.

pleblanc1976 (approved):

PR #1199, opened 2020-05-11
MCOL-3815 Related. Fix columnstore-post-install not running on Debian/Ubuntu

jmrojas2332:
columnstore-post-install should be running during package installation on debian-based systems just like during rpm installation in centos/redhat. This commit fixes the issue found in cpackEngineDeb.cmake.


pleblanc1976 (approved):
LGTM

pleblanc1976:
I like all these green checkmarks.  Haven't seen that in a long time.  Of course, it's not testing this, but, still better than red checkmarks.  :D

PR #1203, opened 2020-05-12
MCOL-4000 Allow columnstore_use_import_for_batchinsert to use a new value, ALWAYS.

tntnatbry:


dbcon/mysql/ha_mcs_sysvars.cpp
```diff
@@ -270,13 +270,28 @@ static MYSQL_THDVAR_ULONG(
...
+    array_elements(mcs_use_import_for_batchinsert_values) - 1,
+    "mcs_use_import_for_batchinsert_values",
+    mcs_use_import_for_batchinsert_values,
+    NULL
+};
+
+static MYSQL_THDVAR_ENUM(
     use_import_for_batchinsert,
-    PLUGIN_VAR_NOCMDARG,
+    PLUGIN_VAR_RQCMDARG,
```
drrtuy, line 289: Are we going to set it in running MDB?

tntnatbry:
Please don't merge yet, I am adding tests cases to the regression suite for this.

dbcon/mysql/ha_mcs_sysvars.cpp
tntnatbry: No that's not required. PLUGIN_VAR_RQCMDARG means if we use this as a command line option, then we need to provide argument to the option. But using this variable as a command line option is not mandatory.

drrtuy: I was wrong using PLUGIN_VAR_RQCMDARG. We must use PLUGIN_VAR_NOCMDOPT for all of our settings so plz start with changing it to PLUGIN_VAR_NOCMDOPT.

tntnatbry:
Corresponding regression suite tests for this are in https://github.com/mariadb-corporation/mariadb-columnstore-regression-test/pull/223

dbcon/mysql/ha_mcs_sysvars.cpp
tntnatbry: PLUGIN_VAR_NOCMDOPT means the user cannot set the value of this system variable through the command line when starting up mysqld. Are we sure we don't want this behaviour?

drrtuy: We are sure.

drrtuy (approved):

PR #1205, opened 2020-05-14
MCOL-3999: Must restart WES on failover.

benthompson15:
On failover cannot just reinit WES must restart. This is because of mutex allocation on create table cannot handle added dbroots after an initial table is created. To test: 
1. create multinode cluster with gluster. 
2. create a table. (if no table is created before the failover there is no error)
3. failover non primary PM.
4. create another table.


pleblanc1976 (approved):
Will have to take your word for it that this makes it all work  :smile:

PR #1206, opened 2020-05-14
MCOL-4002 Reset cal_connection_info members at the end of ha_mcs_impl_end_bulk_insert() for transactions as well.

tntnatbry:


tntnatbry:
Corresponding regression test case is in https://github.com/mariadb-corporation/mariadb-columnstore-regression-test/pull/221

drrtuy (approved):

PR #1208, opened 2020-05-15
MCOL-3888: use value based on upgrade config file if possible.

benthompson15:


oamapps/postConfigure/postConfigure.cpp
```diff
@@ -1421,13 +1421,16 @@ int main(int argc, char* argv[])
             mysqlRep = true;
 
         string answer = "y";
+        // MCOL-3888: default 'y', if its something other than 'n' leave it 'y'
+        if ( MySQLRep == "n" )
+            answer = "n";
```
pleblanc1976, line 1426: Any need to check for 'N' as well?

benthompson15: This values has only accepted lowercase 'y' or 'n' until this point and changing this would require changes else where in code that only validates against lowercase values. Seems like a separate ticket would be needed

pleblanc1976 (approved):

pleblanc1976:
I believe this problem must exist in 1.2 as well, plz follow up with the PR for 1.2 if true.  You can cherry-pick the merge commit from this PR easily I imagine.

PR #1216, opened 2020-05-22
MCOL-4005 Fix handling of utf8 and \ for TEXT data type

tntnatbry:
when performing LDI using cpimport.


dbcon/mysql/ha_mcs_dml.cpp
```diff
@@ -904,7 +908,12 @@ int ha_mcs_impl_write_batch_row_(const uchar* buf, TABLE* table, cal_impl_if::ca
                             bitmap_set_bit(table->read_set, field->field_index);
                             String attribute;
                             field->val_str(&attribute);
-                            escape.assign((char*)buf, attribute.length());
+
+                            if (ci.utf8)
+                                escape.assign((char*)buf, attribute.length() * 3);
```
pleblanc1976: I'm not sure about this change.  This String type (which is not std::string); does it return characters or bytes, any idea?

pleblanc1976 (approved):

PR #1219, opened 2020-05-22
Partial MCOL-3459 fix for 1.4

pleblanc1976:
We found a problem with SM where we could leave partially written data with inaccurate metadata under high load conditions.

This fix has 2 angles.
1) write-side fixes
    - try harder to write the data whole
    - if partially written data is unavoidable, rename the file (its name stores some meta)
2) read-side fixes
    - The crashes were assertion failures on the metadata stored in the filename
    - A file can still have wrong meta in the filename if SM gets killed while writing
    - Turned the assertion into a warning.  It should still function correctly even if the filename is wrong as long as the other metadata for the object is intact.

### DIFF
--- a/storage-manager/src/IOCoordinator.cpp
+++ b/storage-manager/src/IOCoordinator.cpp
@@ -667,7 +667,7 @@ ssize_t IOCoordinator::append(const char *_filename, const uint8_t *data, size_t
                 newObject.key = oldKey;
             }
         
-            metadata.updateEntry(newObject.offset, newObject.key, err + newObject.offset);
+            metadata.updateEntry(newObject.offset, newObject.key, err);
         }
         cache->newObject(firstDir, newObject.key,err);
         newObjectKeys.push_back(newObject.key);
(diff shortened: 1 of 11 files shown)


storage-manager/src/IOCoordinator.cpp
```diff
@@ -505,7 +505,14 @@ ssize_t IOCoordinator::_write(const boost::filesystem::path &filename, const uin
             dataRemaining -= err;
             count += err;
             iocBytesWritten += err;
-            metadata.updateEntryLength(newObject.offset, (err + objectOffset));
+            // get a new name for the object
+            
+            bf::path oldPath = firstDir/newObject.key;
+            newObject.key = metadata.getNewKeyFromOldKey(newObject.key, err + objectOffset);
+            ::rename(oldPath.string().c_str(), (firstDir/newObject.key).string().c_str());
```
pleblanc1976: adding some error detection here

storage-manager/src/IOCoordinator.cpp
```diff
@@ -1135,7 +1168,10 @@ boost::shared_array<uint8_t> IOCoordinator::mergeJournal(const char *object, con
```
pleblanc1976, line 1216: The changes to mergeJournal are just little things I found while waiting for information.  Mainly, saving errno, and propagating kpis to the caller.

pleblanc1976:
going to postpose this partial fix until 1.4.5 for an even more partial fix for 1.4.4.

PR #1220, opened 2020-05-22
MCOL-4015 ExeMgr now re-establishes its connections with PrimProcs.

drrtuy:


dbcon/joblist/distributedenginecomm.cpp
```diff
@@ -386,9 +387,22 @@ void DistributedEngineComm::Listen(boost::shared_ptr<MessageQueueClient> client,
...
+        std::this_thread::sleep_for(std::chrono::seconds(3));
+        Setup();
+        if (originalPMCount != pmCount)
+        {
+            ostringstream os;
+            os << "DEC: lost connection to " << client->addr2String();
+            writeToLog(__FILE__, __LINE__, os.str(), LOG_TYPE_ERROR);
+        }
+/*
     	MessageQueueMap::iterator map_tok;
```
pleblanc1976: side note, this may be one reason why aborts have been a problem since I got here.  If DEC is no longer communicating PM connection failures to the jobsteps, well, IDK what else has been changed around this.  I hope there's something that tells the jobsteps to stop.

pleblanc1976 (commented):
Between what has been commented out previously, the new code, and the new code that is commented,  I'm not sure what's supposed to happen now.  A few questions come to mind.

Do joblists still abort when a PM goes down?
Would this keep them from aborting, and not know that they are chatting with a new PM?

I suspect from reading this, the answers are no and yes, meaning that no notification to the jobsteps is given, and they will continue sending jobs as if nothing happened.  If so, there will be trouble in multi-node configurations at least.  Let me know if testing has shown otherwise.

dbcon/joblist/distributedenginecomm.cpp
drrtuy: Related Zhi code has been commented out since the initial commit so I can't answer you. I will test MCS bahavior removing PM when ExeMgr avaits PM results doing SELECT.

drrtuy: I've restored this 0 message to notify joblist it is done in case of DEC errors in DEC::Listen() and ::writeToClient().

drrtuy:
Joblist didn't abort with the previous version now it aborts the execution if PrimProc goes down(there is a corner case when ExeMgr messaging gets into a black hole. This is outside the scope of my current endevour so I'll describe it in the jira issue.)
If PrimProc went down then:
- there must be no active queries b/c they errors out
- new queries hang waiting for queries to build a syscat cache untill PM restores
However the whole State Machine looks very fragile to network outages that are often in the public cloud.

pleblanc1976 (approved):
I talked with Roman, and our belief is that this patch does what it's supposed to do, and doesn't cause new problems.  We'll have to address the other unknowns & known-bugs at some other time.

PR #1222, opened 2020-05-23
Fixed a storagemanager assertion crashing things

pleblanc1976:
This is the minimal fix necessary to fix the problem we saw in Sky with a customer.  There are other bugs fixed in the sm-crash branch, but we're just committing the minimal fix for the 1.4.4 release.


pleblanc1976:
Preemptively merging.  Ben can let me know later if there is a problem.

PR #1223, opened 2020-05-26
MCOL-4018 Disable legacy OAM by default.

drrtuy:


dbcon/mysql/ha_mcs_client_udfs.cpp
```diff
@@ -294,14 +294,16 @@ extern "C"
 
         try
         {
-            if (getenv("SKIP_OAM_INIT"))
+            // Disable legacy-OAM.
+            //if (getenv("SKIP_OAM_INIT"))
             {
```
jmrojas2332, line 299: Consider commenting out this set of braces. I don't see a reason to keep them since there aren't any new declarations in this block.

jmrojas2332 (approved):
LGTM

oam/oamcpp/liboamcpp.cpp
```diff
@@ -2371,9 +2371,9 @@ void Oam::setProcessStatus(const std::string process, const std::string module,
 
 void Oam::processInitComplete(std::string processName, int state)
 {
-//This method takes too long on Windows and doesn't do anything there anyway...
-    if (getenv("SKIP_OAM_INIT") != NULL)
-        return;
+    //This method takes too long on Windows and doesn't do anything there anyway...
+    // Disable legacy OAM
+    return;
     // get current Module name
```
pleblanc1976, line 2377: Jose's comment applies to everything after this return stmt as well

pleblanc1976 (approved):
I think this is OK.  I'll leave it up to you if you want to keep the commented and short-cutted blocks of code, or if you need them as a reference for now.  Eventually they should be removed, but it doesn't have to be now.  You can merge when ready.

oam/oamcpp/liboamcpp.cpp
drrtuy: This is not a general cleanup yet so I won't remove blocks of code.

PR #1227, opened 2020-05-27
partial fix for MCOL-3459 in 1.4

pleblanc1976:
Fixed several bugs, mostly around the partial write scenario, which I found reading the code investigating a customer issue in Sky.  They were not the cause of the problem, yet still problems.

### DIFF
--- a/storage-manager/src/IOCoordinator.cpp
+++ b/storage-manager/src/IOCoordinator.cpp
@@ -671,7 +671,7 @@ ssize_t IOCoordinator::append(const char *_filename, const uint8_t *data, size_t
             {
                 ostringstream oss;
                 char buf[80];
-                oss << "IOCoordinator::write(): Failed to rename " << (cachePath/firstDir/oldKey).string() << " to " << 
+                oss << "IOCoordinator::append(): Failed to rename " << (cachePath/firstDir/oldKey).string() << " to " << 
                     (cachePath/firstDir/newObject.key).string() << "!  Got " << strerror_r(renameErrno, buf, 80); 
                 logger->log(LOG_ERR, oss.str().c_str());
                 newObject.key = oldKey;
--- a/storage-manager/src/IOCoordinator.cpp
+++ b/storage-manager/src/IOCoordinator.cpp
@@ -488,6 +488,8 @@ ssize_t IOCoordinator::_write(const boost::filesystem::path &filename, const uin
             //log error and abort
             l_errno = errno;
             logger->log(LOG_ERR,"IOCoordinator::write(): Failed newObject.");
+            metadata.removeEntry(newObject.offset);
+            replicator->remove(firstDir/newObject.key);
             errno = l_errno;
             if (count == 0)   // if no data has been written yet, it's safe to return -1 here.
                 return -1;
@@ -509,14 +511,17 @@ ssize_t IOCoordinator::_write(const boost::filesystem::path &filename, const uin
             // get a new name for the object
             string oldKey = newObject.key;
             newObject.key = metadata.getNewKeyFromOldKey(newObject.key, err + objectOffset);
-            int renameErr = ::rename((firstDir/oldKey).string().c_str(), (firstDir/newObject.key).string().c_str());
+            ostringstream os;
+            os << "IOCoordinator::write(): renaming " << oldKey << " to " << newObject.key;
+            logger->log(LOG_DEBUG, os.str().c_str());
+            int renameErr = ::rename((cachePath/firstDir/oldKey).string().c_str(), (cachePath/firstDir/newObject.key).string().c_str());
             int renameErrno = errno;
             if (renameErr < 0)
             {
                 ostringstream oss;
                 char buf[80];
-                oss << "IOCoordinator::write(): Failed to rename " << (firstDir/oldKey).string() << " to " << 
-                    (firstDir/newObject.key).string() << "!  Got " << strerror_r(renameErrno, buf, 80); 
+                oss << "IOCoordinator::write(): Failed to rename " << (cachePath/firstDir/oldKey).string() << " to " << 
+                    (cachePath/firstDir/newObject.key).string() << "!  Got " << strerror_r(renameErrno, buf, 80); 
                 logger->log(LOG_ERR, oss.str().c_str());
                 newObject.key = oldKey;
             }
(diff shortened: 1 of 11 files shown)


storage-manager/src/IOCoordinator.cpp
```diff
@@ -627,21 +651,41 @@ ssize_t IOCoordinator::append(const char *_filename, const uint8_t *data, size_t
...
+            ostringstream os;
+            os << "IOCoordinator::append(): renaming " << oldKey << " to " << newObject.key;
+            logger->log(LOG_DEBUG, os.str().c_str());
+            int renameErr = ::rename((cachePath/firstDir/oldKey).string().c_str(), (cachePath/firstDir/newObject.key).string().c_str());
+            int renameErrno = errno;
+            if (renameErr < 0)
+            {
+                ostringstream oss;
+                char buf[80];
+                oss << "IOCoordinator::write(): Failed to rename " << (cachePath/firstDir/oldKey).string() << " to " <<
```
pleblanc1976: Wrong function name, fixing it

storage-manager/src/IOCoordinator.cpp
```diff
@@ -1001,10 +1048,13 @@ int IOCoordinator::copyFile(const char *_filename1, const char *_filename2)
         for (const auto &object : objects)
         {
             bf::path journalFile = journalPath/firstDir1/(object.key + ".journal");
-            // XXXPAT: Need to follow up on this.  this is the wrong length to use here, but changing it
-            // to use the right length is causing an error somewhere else.  Not sure why yet.
-            metadataObject newObj = meta2.addMetadataObject(filename2, object.length);
-            assert(newObj.offset == object.offset);
+            
+            // originalLength = the length of the object before journal entries.
+            // the length in the metadata is the length after journal entries
```
pleblanc1976, line 1053: The error mentioned in the old comment is that the new metadata entry would get the original size, not the merged size.  Fixed that in the next line if the original size != merged size.

storage-manager/src/IOCoordinator.cpp
```diff
@@ -1137,7 +1187,10 @@ boost::shared_array<uint8_t> IOCoordinator::mergeJournal(const char *object, con
```
pleblanc1976, line 1187: The fixes to mergeJournal and related functions were all around little KPI accounting flaws & error propagation.  Should not affect functionality.

storage-manager/src/LocalStorage.cpp
```diff
@@ -101,6 +101,7 @@ int LocalStorage::copy(const bf::path &source, const bf::path &dest)
     if (err)
     {
         errno = err.value();
+        ::unlink(dest.string().c_str());
```
pleblanc1976, line 104: decided if the copy fails, it should delete the copy if it exists.

storage-manager/src/Replicator.cpp
```diff
@@ -128,7 +128,7 @@ int Replicator::newObject(const boost::filesystem::path &filename, const uint8_t
     OPEN(objectFilename.c_str(), O_WRONLY | O_CREAT);
     size_t count = 0;
     while (count < length) {
-        err = ::pwrite(fd, &data[count], length - count, offset);
+        err = ::pwrite(fd, &data[count], length - count, offset + count);
```
pleblanc1976, line 131: if a partial write happened, it would continue writing at the original offset, not at the end of the last write.

storage-manager/src/Replicator.cpp
```diff
@@ -154,6 +154,67 @@ int Replicator::newNullObject(const boost::filesystem::path &filename,size_t len
     return err;
 }
 
+ssize_t Replicator::_pwrite(int fd, const void *data, size_t length, off_t offset)
```
pleblanc1976, line 157: _pwrite and _write are drop-in replacements for ::pwrite & ::write, which try harder to write the data intact.

storage-manager/src/Replicator.cpp
```diff
@@ -238,26 +299,32 @@ int Replicator::addJournalEntry(const boost::filesystem::path &filename, const u
...
                 if ((uint)rollbackErr == (headerRollback.length() + 1))
                     mpLogger->log(LOG_CRIT, "Replicator::addJournalEntry: Rollback of journal header success.");
                 else
                     mpLogger->log(LOG_CRIT, "Replicator::addJournalEntry: Rollback of journal header failed!");
-                return err;
+                errno = l_errno;
+                if (err < 0)
+                    return err;
+                else
+                    return 0;
```
pleblanc1976, line 319: Since no user data has been written yet, it should either return 0 or -1 indicating an error.  Returning a val > 0 means to the caller that some user data was written successfully.

storage-manager/src/Synchronizer.cpp
```diff
@@ -434,8 +435,11 @@ void Synchronizer::process(list<string>::iterator name)
             success = true;
         }
         catch(exception &e) {
-            logger->log(LOG_CRIT, "Synchronizer::process(): error sync'ing %s opFlags=%d, got '%s'.  Retrying...", key.c_str(),
-                pending->opFlags, e.what());
+            // these are often self-resolving, so we will suppress logging it for 10 iterations, then escalate
+            // to error, then to crit
+            //if (++retryCount >= 10)
+                logger->log((retryCount < 20 ? LOG_ERR : LOG_CRIT), "Synchronizer::process(): error sync'ing %s opFlags=%d, got '%s'.  Retrying...", key.c_str(),
+                    pending->opFlags, e.what());
```
pleblanc1976, line 442: This escalates the error level for persistent errors, rather than log everything (self-correcting or not) as a critical error.

storage-manager/src/Synchronizer.cpp
```diff
@@ -463,18 +467,33 @@ void Synchronizer::synchronize(const string &sourceFile, list<string>::iterator
...
+            size_t size = bf::file_size(objectPath);
+            replicator->remove(objectPath);
+            cache->deletedObject(prefix, cloudKey, size);
+            cs->deleteObject(cloudKey);
+        }
+        catch (exception &e)
+        {
+            logger->log(LOG_DEBUG, "synchronize(): failed to remove orphaned object '%s' from the cache, got %s", 
+                objectPath.string().c_str(), e.what());
+        }
```
pleblanc1976, line 496: I ran into this during one of my tests.  Need to delete orphaned objects.

storage-manager/src/Synchronizer.cpp
```diff
@@ -683,7 +725,7 @@ void Synchronizer::synchronizeWithJournal(const string &sourceFile, list<string>
         
         while (count < size)
         {
-            err = ::write(newFD, data.get(), size - count);
+            err = ::write(newFD, &data[count], size - count);
```
pleblanc1976, line 728: After a partial write, this would repeat the data on successive writes.

benthompson15 (approved):

PR #1229, opened 2020-05-28
enable drone pipelines for develop branch (1.5)

mariadb-RomanNavrotskiy:


pleblanc1976 (approved):
The new system is happy, so I'm happy.  We'll have to continue work on the 1.5 system though; only tests the build process atm.

PR #1242, opened 2020-05-30
MCOL-4023 Pushdown WHERE conditions for UPDATE/DELETE.

tntnatbry:
For certain queries, such as:
  update cs1 set i = 41 where i = 42 or (i is null and 42 is null);
the SELECT_LEX.where does not contain the required where conditions.
Server sends the where conditions in the call to cond_push(), so
we are storing them in a handler data member, condStack, and later
push them down to getSelectPlan() for UPDATES/DELETEs.


drrtuy (approved):
I think it is even better that we swallow the whole condition for DML now.

PR #1243, opened 2020-05-31
cmake fixes

vuvova:
don't FATAL_ERROR if prerequisites are not found, skip building the plugin instead

and cmake output cleanup:

1. reduce the number of cmake warnings
2. remove unused OLD policies
3. only warn about missing prerequisites once
4. start cmake output from columnstore version


utils/libmarias3/CMakeLists.txt
```diff
@@ -1,21 +1,5 @@
...
-        message(STATUS "Submodule update")
-        execute_process(COMMAND ${GIT_EXECUTABLE} submodule update --init --recursive
-            WORKING_DIRECTORY ${ENGINE_SRC_DIR}
-                        RESULT_VARIABLE GIT_SUBMOD_RESULT)
-        if(NOT GIT_SUBMOD_RESULT EQUAL "0")
-            message(FATAL_ERROR "git submodule update --init failed with ${GIT_SUBMOD_RESULT}, please checkout submodules")
-        endif()
-    endif()
-endif()
-
```
pleblanc1976, line 18: I don't see a replacement for this block of code in the new version.  Is submodule pulling a manual step in the community server?

vuvova: No, it's done by `cmake/submodules.cmake` for all submodules, so your separate `git submodule update` is redundant.

_the server has, like, seven (?) submodules now, and after the second submodule we've figured out that if all submodules will need this auto-updating code, we can as well put it in the server and spare submodules some troubles and avoid some copy-pasting_

pleblanc1976 (approved):

PR #1248, opened 2020-06-01
ci: set proper path to storage-manager unit test

mariadb-RomanNavrotskiy:


pleblanc1976:
Does this also need to go into develop?

pleblanc1976 (approved):

PR #1254, opened 2020-06-04
Columnstore cache

tntnatbry:


tntnatbry:
This PR has a dependency on the changes in the server branch bb-10.5-monty. Please don't merge until the relevant changes from this branch (https://github.com/MariaDB/server/commit/56972cd0a0a0fa2a4b35e5e6b9dbd549595fa8d9) are merged into server 10.5

dbcon/mysql/ha_mcs.cpp
```diff
@@ -1137,6 +1156,514 @@ void ha_mcs::cond_pop()
...
+*/
+
+int ha_mcs_cache::create(const char *name, TABLE *table_arg,
+                         HA_CREATE_INFO *ha_create_info)
+{
+  int error;
+  char cache_name[FN_REFLEN+8];
+  DBUG_ENTER("ha_mcs_cache::create");
+
+  create_cache_name(cache_name, name);
```
pleblanc1976, line 1348: any need to check the length of name here?

dbcon/mysql/ha_mcs.cpp
```diff
@@ -1137,6 +1156,514 @@ void ha_mcs::cond_pop()
...
+    ha_choice save_transactional= ha_create_info->transactional;
+    row_type save_row_type=       ha_create_info->row_type;
+    ha_create_info->transactional= HA_CHOICE_NO;
+    ha_create_info->row_type=      ROW_TYPE_DYNAMIC;
+
+    if ((error= cache_handler->create(cache_name, table_arg, ha_create_info)))
+      DBUG_RETURN(error);
+    ha_create_info->transactional= save_transactional;
+    ha_create_info->row_type=      save_row_type;
+  }
```
pleblanc1976, line 1360: does this need to be in its own { } block?

dbcon/mysql/ha_mcs.cpp
```diff
@@ -1137,6 +1156,514 @@ void ha_mcs::cond_pop()
...
+
+/*
+  Create a name for the cache table
+*/
+
+static void create_cache_name(char *to, const char *name)
+{
+  uint dir_length= dirname_length(name);
+  to= strnmov(to, name, dir_length);
+  strxmov(to, CACHE_PREFIX, name+ dir_length, NullS);
```
pleblanc1976, line 1225: I don't see any man pages for these, have to assume they're maria fcns and that they're being used correctly

dbcon/mysql/ha_mcs.cpp
```diff
@@ -1137,6 +1156,514 @@ void ha_mcs::cond_pop()
...
+
+  {
+    LEX_CSTRING name= { STRING_WITH_LEN("Aria") };
+    plugin_maria= ha_resolve_by_name(0, &name, 0);
+    mcs_maria_hton= plugin_hton(plugin_maria);
+    error= mcs_maria_hton == NULL;                   // Engine must exists!
+    if (error)
+      my_error(HA_ERR_INITIALIZATION, MYF(0),
+               "Could not find storage engine %s", name.str);
+  }
```
pleblanc1976, line 1648: another block that seems like it doesn't need to be there

pleblanc1976:
After Monty's branch is merged into the server, we'll need to rebase the bb-10.5-cs branch of the server on 10.5.  Our ci/cd system uses bb-10.5-cs.

pleblanc1976 (dismissed):
Adding a technical blocker to prevent accidental merging until the reqs are in place.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -6094,7 +6094,9 @@ bool isMCSTable(TABLE* table_ptr)
     string engineName = table_ptr->s->db_plugin->name.str;
 #endif
 
-    if (engineName == "Columnstore" || engineName == "InfiniDB")
+    if (engineName == "Columnstore" ||
+        engineName == "InfiniDB" ||
```
drrtuy: There will be no Infinidb tables anymore.

dbcon/mysql/ha_mcs.cpp
```diff
@@ -1155,11 +1682,26 @@ maria_declare_plugin(columnstore)
...
+{
+  MYSQL_STORAGE_ENGINE_PLUGIN,
+  &ha_mcs_cache_storage_engine,
+  "Columnstore_cache",
+  "MariaDB Corporation AB",
+  "Insert cache for ColumnStore",
+  PLUGIN_LICENSE_GPL,
+  ha_mcs_cache_init,            /* Plugin Init */
+  ha_mcs_cache_deinit,          /* Plugin Deinit */
+  0x0100,                       /* 1.0 */
```
drrtuy: We should align the version with MCS version.

dbcon/mysql/ha_mcs.cpp
```diff
@@ -1137,6 +1156,514 @@ void ha_mcs::cond_pop()
...
+
+  return error;
+}
+
+static int ha_mcs_cache_deinit(void *p)
+{
+  if (plugin_maria)
+  {
+    plugin_unlock(0, plugin_maria);
+    plugin_maria= 0;
```
drrtuy: If this is a pointer we better use nullptr instead.

dbcon/mysql/ha_mcs.cpp
drrtuy: There is a LEX_CSTRING name used in this block only so I disagree it is totally useless. Moreover the block might be implemented by Monty so he might have his own reasoning behind the explicit block.

dbcon/mysql/ha_mcs.cpp
drrtuy: Yes. It is defined in strings/strxmov.c

dbcon/mysql/CMakeLists.txt
```diff
@@ -35,16 +36,17 @@ set_source_files_properties(ha_mcs.cpp PROPERTIES COMPILE_FLAGS "-fno-implicit-t
 
 if (COMMAND mysql_add_plugin)
     mysql_add_plugin(columnstore ${libcalmysql_SRCS} STORAGE_ENGINE MODULE_ONLY DEFAULT
-        LINK_LIBRARIES ${ENGINE_LDFLAGS} ${ENGINE_WRITE_LIBS} ${NETSNMP_LIBRARIES} threadpool
+        LINK_LIBRARIES ${ENGINE_LDFLAGS} ${ENGINE_WRITE_LIBS} ${NETSNMP_LIBRARIES} threadpool aria myisam perfschema
```
drrtuy: What exactly do we use from perfschema?

drrtuy (approved):
Technically it is OK. 
I would use the same version as MCS plugin has to avoid ambiguities and check wether we really need to link ha_columnstore against perfschema.

drrtuy:
JFYI bb-10.5-cs has been rebased but not pushed yet so stay tuned.

drrtuy:
The rebased bb-10.5-cs has just been pushed so feel free to check if the relevant commits are there.

dbcon/mysql/ha_mcs.cpp
tntnatbry: I think we are safe here not to; this call is made by the server handler mechanism which should be doing this error handling.

dbcon/mysql/ha_mcs.cpp
tntnatbry: I had the same question. I don't see a need to put this in a dedicated code block. I decided to keep it as I am not sure what the original author of the code thought while doing this.

dbcon/mysql/ha_mcs_execplan.cpp
tntnatbry: Good catch. This is updated.

dbcon/mysql/ha_mcs.cpp
tntnatbry: Ok. I have changed this to use MCS versions here.

dbcon/mysql/ha_mcs.cpp
tntnatbry: Yes this is pointer to a pointer. I have changed it to use a NULL here instead, just to keep the NULL usage consistent with the rest of the source file. Although there are several other places in the code still where we are still using 0.

dbcon/mysql/CMakeLists.txt
tntnatbry: I originally had linker errors when trying to link aria symbols, which indirectly depended on symbols from the myisam and perfschema libs. This was when we still had references of ha_maria hton in the code here. Now that we don't explicitly need ha_maria, I found that the linker does not complain anymore. So I have removed this linkage altogether.

tntnatbry:
I have made changes based on the comments on the PR in the commit https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/1254/commits/1a5c6910da9cc33d64044f01439749de65877ea2 and pushed. 
Please review.

PR #1258, opened 2020-06-04
Mcol 3536

davidhall586:


dbcon/execplan/predicateoperator.cpp
```diff
@@ -54,62 +57,21 @@ namespace execplan
 /**
  * Constructors/Destructors
  */
-PredicateOperator::PredicateOperator()
+PredicateOperator::PredicateOperator() :
+    cs(NULL)
```
drrtuy, line 60: Null for pointers sent out of fashion. Now it is cool to use nullptr according with c++-11.

dbcon/execplan/predicateoperator.cpp
```diff
@@ -344,28 +308,37 @@ void PredicateOperator::setOpType(Type& l, Type& r)
               r.colDataType == execplan::CalpontSystemCatalog::VARCHAR ||
               r.colDataType == execplan::CalpontSystemCatalog::TEXT))
     {
+#if 0
```
drrtuy, line 310: I wonder if this commented block is for a future use or we can remove it?

dbcon/execplan/treenode.h
```diff
@@ -336,7 +336,7 @@ class TreeNode
         fRefCount = refCount;
     }
```
drrtuy, line 338: What does this mean?

CMakeLists.txt
```diff
@@ -247,6 +247,12 @@ ELSE()
     SET (MARIADB_CLIENT_LIBS    -L${SERVER_BUILD_INCLUDE_DIR}/../libmariadb/libmariadb/ libmariadb.so)
 ENDIF()
 
+IF (INSTALL_LAYOUT)
```
drrtuy, line 250: I wonder what is the semantics of INSTALL_LAYOUT variable?

dbcon/joblist/batchprimitiveprocessor-jl.h
```diff
@@ -154,8 +151,7 @@ class BatchPrimitiveProcessorJL
     /* Turn a ByteStream into ElementTypes or StringElementTypes */
     void getElementTypes(messageqcpp::ByteStream& in, std::vector<ElementType>* out,
```
drrtuy, line 152: It looks like we even don't use BatchPrimitiveProcessorJL::getElementTypes.

dbcon/joblist/expressionstep.cpp
```diff
@@ -398,8 +398,10 @@ void ExpressionStep::populateColumnInfo(SimpleColumn* sc, JobInfo& jobInfo)
 //XXX use this before connector sets colType in sc correctly.
 //    type of pseudo column is set by connector
         if (dynamic_cast<PseudoColumn*>(sc) == NULL)
```
drrtuy, line 400: Do we really need this for a PseudoColumn that is a result of our internal UDAF which result types can't be a char.

utils/rowgroup/rowgroup.h
```diff
@@ -1413,15 +1361,19 @@ class RowGroup : public messageqcpp::Serializeable
     // Used to map the projected column and rowgroup index
     std::vector<uint32_t> keys;
     std::vector<execplan::CalpontSystemCatalog::ColDataType> types;
-
-    // DECIMAL support.  For non-decimal fields, the values are 0.
+    // For string collation
+    std::vector<uint32_t> charsetNumbers;
+    std::vector<CHARSET_INFO*> charsets;
```
drrtuy, line 1366: Is it safe to assume CHARSET_INFO life cycle is bigger then that of the RowGroup?

dbcon/execplan/predicateoperator.cpp
davidhall586: Yeah. Whatever. I had a funny feeling this may have caused trouble and decided to be uncool.

dbcon/execplan/predicateoperator.cpp
davidhall586: Yes. My intent was to perhaps re-enable that someday. It makes sense if the collation is a binary collation. However, I don't have faith that STRINT is properly handled throughout the code path, so it will wait.

dbcon/execplan/treenode.h
davidhall586: This isn't a change

CMakeLists.txt
davidhall586: I would need to waste time studying...

dbcon/joblist/expressionstep.cpp
davidhall586: The line:
if (dynamic_cast<PseudoColumn*>(sc) == NULL)
asked the question if this is NOT a PseudoColumn. So yes, we need these lines for everything that is NOT a PseudoColumn.

utils/rowgroup/rowgroup.h
davidhall586: Yes. CHARSET_INFO is singleton life of the program.
There is a global unordered map that contains all the CHARSET_INFO objects created at startup. We merely keep pointers into that hash map to prevent continuously hashing. Saves a few cycles.

dbcon/joblist/batchprimitiveprocessor-jl.h
davidhall586: True. As well as a number of other "get" functions in that class. This is beyond the scope of this work. MCOL-4045 has been opened to track this.

dbcon/execplan/treenode.h
drrtuy: I mean the comment under this line.

primitives/linux-port/dictionary.cpp
```diff
@@ -199,41 +203,27 @@ void PrimitiveProcessor::p_TokenByScan(const TokenByScanRequestHeader* h,
             goto no_store;
         }
 
-        // BUG 5110: If it is utf, we need to create utf strings to compare
-        if (utf8)
+        if (h->COP1 & COMPARE_LIKE)
```
drrtuy, line 203: I think we can future reduce boilerplate code and make this a function.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -6706,7 +6719,8 @@ int processLimitAndOffset(
     }
 
     // We don't currently support limit with correlated subquery
-    if (gwi.subQuery && !gwi.correlatedTbNameVec.empty() && csep->hasOrderBy())
+    if (csep->limitNum() != (uint64_t) - 1 &&
```
drrtuy, line 6720: What are the cases that you found that forces you to add this condition?

dbcon/joblist/tupleaggregatestep.cpp
```diff
@@ -1507,6 +1522,7 @@ void TupleAggregateStep::prep1PhaseAggregate(
             precisionAgg.push_back(0);
             precisionAgg.push_back(0);
             typeAgg.push_back(CalpontSystemCatalog::UBIGINT);
+            csNumAgg.push_back(8);
```
drrtuy, line 1525: This 8 is a a migic to tell aggregate code the output will be a numerical scalar that doesn't depend on charset?

dmlproc/dmlproc.cpp
```diff
@@ -20,13 +20,17 @@
...
+#include <my_sys.h>
+
 #include <unistd.h>
 #include <signal.h>
 #include <string>
 #include <set>
 #include <clocale>
-#include "boost/filesystem/operations.hpp"
-#include "boost/filesystem/path.hpp"
+//#include "boost/filesystem/operations.hpp"
```
drrtuy, line 28: We can remove this.

utils/funcexp/func_if.cpp
```diff
@@ -53,22 +53,22 @@ bool boolVal(SPTP& parm, Row& row, const string& timeZone)
             case CalpontSystemCatalog::TEXT:
             case CalpontSystemCatalog::VARCHAR:
                 ret = (atoi((char*)(parm->data()->getStrVal(timeZone).c_str())) != 0);
-
+                break;
```
drrtuy, line 56: Did I get it right that we have at least 3 years old bug here?

utils/funcexp/func_least.cpp
```diff
@@ -22,6 +22,10 @@
 *
 ****************************************************************************/
 
+#include <mariadb.h>
```
drrtuy: What do you think if we add this into a common header instead?

dbcon/joblist/batchprimitiveprocessor-jl.h
pleblanc1976: Yeah all of the "ElementType" stuff is likely OBE.  It's from the very first, and low-performing version of calpont.  '08 or so.

dbcon/joblist/tupleaggregatestep.cpp
davidhall586: 8 is the magic number for latin1_swedish_ci, so yes. Something needs to be there to take the space. It will not be referenced in the current implementation.

utils/funcexp/func_if.cpp
davidhall586: um... Yes

utils/funcexp/func_least.cpp
davidhall586: I think it probably needs to be done.

dbcon/joblist/batchprimitiveprocessor-jl.h
tntnatbry: This is good to know. I also came across getElementType(), getStringElementTypes(), etc. at some point and wasn't sure if these were being used anywhere.

pleblanc1976 (approved):
It's not fully functional yet but has to be merged now due to scheduling constraints.  We'll have to fix remaining problems during the freeze period.

PR #1261, opened 2020-06-05
MCOL-4042 cleanup after MODA

davidhall586:


pleblanc1976 (approved):
Dave tells me valgrind is happy, no more leaks from moda

PR #1264, opened 2020-06-09
MCOL-4019 This patch avoids fork() in wokernode/controllernode that h…

drrtuy:
…angs in ThreadPool::~

waiting infinitely for fPruneThread->join() to return.


pleblanc1976 (approved):
no problems, but I am curious about the reason for it, will discuss with Roman directly.

drrtuy:
The reason is explained in the MCOL-4019.

PR #1267, opened 2020-06-11
MCOL-4043 Fix memory leaks - 1

tntnatbry:
simpleScalarFilterToParseTree() performs a dynamic allocation
of a ParseTree object, but this memory is never freed later.
We now keep track of this allocation and perform the delete
in the JobList dtor after the query finishes.

### DIFF
--- a/dbcon/execplan/simplefilter.cpp
+++ b/dbcon/execplan/simplefilter.cpp
@@ -150,8 +150,14 @@ SimpleFilter::SimpleFilter(const SimpleFilter& rhs) :
 SimpleFilter::~SimpleFilter()
 {
     //delete fOp;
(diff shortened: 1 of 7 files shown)


tntnatbry:
Test case to trigger the leak fixed by the PR:
```sql
create table cs1 (a int)engine=columnstore;
insert into cs1 values (123);
select * from cs1 where a=(select a from cs1);
select * from cs1 having a=(select a from cs1);
select * from cs1 where a=(select a from cs1) and a!=(select a from cs1);
select * from cs1 where a=(select a from cs1) or a!=(select a from cs1);
```

pleblanc1976 (approved):

PR #1269, opened 2020-06-11
MCOL-4043 Fix memory leaks - 2

tntnatbry:
Perform deletes of Compare objects in the OrderByData dtor,
which were allocated by CompareRule::compileRules().


tntnatbry:
Test case to trigger the leak fixed by the PR:
```sql
create table cs1 (a int, b bigint)engine=columnstore;
insert into cs1 values (123, 1234000);
select rank() over (partition by a,b order by a,b desc) as ranking from cs1;
```

pleblanc1976 (approved):

PR #1271, opened 2020-06-12
Installer now pre-setups S3 settings to bootstrap S3-backed MCS.

drrtuy:


pleblanc1976 (approved):
It makes a couple vars mandatory that should be optional (endpoint, key, secret).  Having key and secret in the config file is also a security risk.  SM pulls them from AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY already (secure way to do it) so we don't need to store keys at all technically.  We'll have to follow up on securing those.  Time is short, so unfortunately I have to merge this now as is.

PR #1273, opened 2020-06-12
MCOL-4056: reset ExtentMutexes if dbroot count changes.

benthompson15:


writeengine/shared/we_fileop.cpp
```diff
@@ -1975,11 +1975,22 @@ void FileOp::initDbRootExtentMutexes( )
...
-        Config::getRootIdList( rootIds );
+        //Something in configuration changed the size
+        //reset and rebuild this map
+        std::map<int, boost::mutex*>::iterator k = m_DbRootAddExtentMutexes.begin();
+        while (k != m_DbRootAddExtentMutexes.end() )
+        {
+            delete k->second;
+            ++k;
+        }
+    }
```
pleblanc1976, line 1990: I think this will fail in an odd scenario like this.
starting with dbroots 1, 2.
drop dbroot 2
add dbroot 3

the size check will pass, but the map will be wrong.

pleblanc1976 (changes requested):

writeengine/shared/we_fileop.cpp
pleblanc1976: also need to clear the map after the loop

benthompson15: This is valid point. I'll come up with better solution. Started with this so issue wasn't forgotten about. Somewhat low priority and low threat of happening.

PR #1281, opened 2020-06-15
MCOL-4033: removes softwareReport.sh and adds copy of storagemanager.cnf

benthompson15:


pleblanc1976 (approved):
The debian packing lists on the server side will have to be updated after this.

PR #1282, opened 2020-06-16
Disable THR_LOCK virtual functions for the ColumnStore Cache.

tntnatbry:


tntnatbry:
I am not sure about the relevance of these functions as they are not implemented in ColumnStore. Disabling these for now since cache_copy_status caused a server crash when executing working_dml/misc/columnstore_cache.sql (specifically, the query "INSERT INTO cache1 SELECT * FROM cache1;")

PR #1296, opened 2020-06-19
MCOL-4043 Fix memory leaks - 1 (second attempt)

tntnatbry:
simpleScalarFilterToParseTree() performs a dynamic allocation
of a ParseTree object, but this memory is never freed later.
We now keep track of this allocation and perform the delete
in ~CSEP/CSEP::unserialize() after the query finishes.

### DIFF
--- a/dbcon/execplan/calpontselectexecutionplan.cpp
+++ b/dbcon/execplan/calpontselectexecutionplan.cpp
@@ -173,6 +173,26 @@ CalpontSelectExecutionPlan::~CalpontSelectExecutionPlan()
 
     fFilters = NULL;
     fHaving = NULL;
+
+    if (!fDynamicParseTreeVec.empty())
+    {
+        for (auto& parseTree : fDynamicParseTreeVec)
+        {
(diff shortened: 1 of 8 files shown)


tntnatbry:
Below are the queries triggering various code paths to simpleScalarFilterToParseTree() that lead to a leak fixed by this patch:
```sql
drop table if exists cs1;
create table cs1 (a int)engine=columnstore;
insert into cs1 values (123);
select * from cs1 where a=(select a from cs1);
select * from cs1 having a=(select a from cs1);
select * from cs1 where a=(select a from cs1) and a!=(select a from cs1);
select * from cs1 where a=(select a from cs1) or a!=(select a from cs1);
select * from (select * from cs1 where a=(select a from cs1)) aa;
select * from cs1 aa where exists (select * from cs1 bb where bb.a=(select a from cs1));
select * from cs1 aa where exists (select * from cs1 bb where aa.a=bb.a and bb.a=(select a from cs1));
```

pleblanc1976 (approved):

PR #1306, opened 2020-06-24
MCOL-4030

benthompson15:


ottok:
Fixing this would help me work on https://jira.mariadb.org/browse/MCOL-4057 and currently test builds at https://salsa.debian.org/mariadb-team/mariadb-server/pipelines due to too long log output (and also too slow/long build in general, but fixing this would be a step forward).

pleblanc1976 (approved):
The main risk with changes like this is calling a different function than was intended.  Ben says this passes the regression test, so this should be safe.

ottok:
I was planning to do a rebuild now, but can't find any commits with "MCOL" in the title on the server 10.5 branch. Is there a docs somewhere on what is the ColumnStore -> Server workflow so I could understand how this works and what branch and repo I should track and base my work on?

Thanks.

ottok:
Allright, it is a submodule on the 10.5 repo so I just track what commit the submodule has:
```
± git log --oneline storage/columnstore/columnstore/
d742f282fc9 ColumnStore RPM packaging fixes
04df0bdae73 A few important fixes for columnstore.
ec2de1d58a6 update MCS
```

These commits on the `develop` branch are already there, so I'll try a rebuild now.

PR #1311, opened 2020-06-26
MCOL-4109 Error out instead of sending a warning

tntnatbry:
when a non-existing table is dropped.

This patch accomodates the changes made to server 10.5 as part
of MDEV-11412, where the server now tries to drop the table from
all storage engines when a .frm table does not exist.

We were earlier retuning a warning to the client and setting the
return code to 0. We now instead return ER_NO_SUCH_TABLE_IN_ENGINE
error code to the server if the table does not exist in
ColumnStore.

### DIFF
--- a/dbcon/mysql/ha_mcs_ddl.cpp
+++ b/dbcon/mysql/ha_mcs_ddl.cpp
@@ -2128,6 +2128,11 @@ int ProcessDDLStatement(string& ddlStatement, string& schema, const string& tabl
             ci->isAlter = false;
         }
 
+        if (b == ddlpackageprocessor::DDLPackageProcessor::DROP_TABLE_NOT_IN_CATALOG_ERROR)
+        {
+            return ER_NO_SUCH_TABLE_IN_ENGINE;
+        }
+
         if ((b != 0) && (b != ddlpackageprocessor::DDLPackageProcessor::WARNING))
         {
             thd->get_stmt_da()->set_overwrite_status(true);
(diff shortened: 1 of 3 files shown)


dbcon/mysql/ha_mcs_ddl.cpp
```diff
@@ -2583,7 +2588,7 @@ int ha_mcs_impl_delete_table_(const char* db, const char* name, cal_connection_i
     stmt += ";";
     int rc = ProcessDDLStatement(stmt, schema, tbl, tid2sid(thd->thread_id), emsg);
 
-    if (rc != 0)
+    if (rc != 0 && rc != ER_NO_SUCH_TABLE_IN_ENGINE)
```
pleblanc1976, line 2591: Do we not need to push_warning for other errors as well?

tntnatbry: We are pushing warnings for all errors, except when rc=ER_NO_SUCH_TABLE_IN_ENGINE. If I include rc=ER_NO_SUCH_TABLE_IN_ENGINE in the push_warning, I get the following:
```sql
MariaDB [test]> drop table if exists tttttt;
Query OK, 0 rows affected, 2 warnings (0.085 sec)

MariaDB [test]> show warnings;
+---------+------+--------------------------------------------------+
| Level   | Code | Message                                          |
+---------+------+--------------------------------------------------+
| Warning | 9999 | CAL0001: Table does not exist in ColumnStore.    |
| Note    | 1051 | Unknown table 'test.tttttt'                      |
+---------+------+--------------------------------------------------+
2 rows in set (0.000 sec)
```

The expected warning here should be:
```sql
MariaDB [test]> show warnings;
+-------+------+---------------------------+
| Level | Code | Message                   |
+-------+------+---------------------------+
| Note  | 1051 | Unknown table 'test.tttt' |
+-------+------+---------------------------+
```

This is why rc=ER_NO_SUCH_TABLE_IN_ENGINE is excluded.

pleblanc1976: ah I see.  I misread this the first time I looked at it.  I agree, this should warn on everything but no-such-table, which is handled a different way.

pleblanc1976 (approved):

PR #1316, opened 2020-06-29
MCOL-4124 return error when COMMIT|ROLLBACK error

davidhall586:


pleblanc1976 (approved):

tntnatbry:
@dhall-MariaDB 

On line 520 in this file, I see the following:

```sh
        rc = ProcessCommandStatement ( thd, command, ci, schema );
        rc = ci.rc;
```
If ProcessCommandStatement returns a non-zero value and ci.rc=0, then we incorrectly set rc to 0.

tntnatbry:
Also, I see one issue with the usage of `rc = max(rc, rc1)`; if either of the two values is less than 0 (error) and the other is 0, then we incorrectly set rc to 0.

PR #1319, opened 2020-06-30
fix auto-disabling of the columnstore for bintars

vuvova:
RPM|DEB is not a valid cmake condition

Also: DRY


vuvova:
currently it causes columnstore to be skipped in RPM builds, which fails upgrade tests in buildbot

pleblanc1976 (approved):

PR #1328, opened 2020-07-06
MCOL-4116 add long LONGDOUBLE support to update

davidhall586:


writeengine/server/we_dmlcommandproc.cpp
```diff
@@ -3121,6 +3121,20 @@ uint8_t WE_DMLCommandProc::processUpdate(messageqcpp::ByteStream& bs,
                             break;
                         }
 
+                        case CalpontSystemCatalog::LONGDOUBLE:
+                        {
+                            double dll = row.getDoubleField(fetchColPos);
```
tntnatbry: Shouldn't this be long double in the rest of this case statement?

writeengine/server/we_dmlcommandproc.cpp
```diff
@@ -3121,6 +3121,20 @@ uint8_t WE_DMLCommandProc::processUpdate(messageqcpp::ByteStream& bs,
...
+
+                            if (dll == std::numeric_limits<double>::infinity())
+                                continue;
+
+                            ostringstream os;
+                            //@Bug 3350 fix the precision.
+                            os << setprecision(16) << dll;
```
tntnatbry: I see setprecision(16) for the double case. Is this also accurate for long double?

tntnatbry (changes requested):

writeengine/server/we_dmlcommandproc.cpp
davidhall586: blech. Of course.

writeengine/server/we_dmlcommandproc.cpp
davidhall586: I'm not sure. I set it to 19. Hopefully that will be OK. 7 for FLOAT above also seems strange.

tntnatbry: https://en.cppreference.com/w/cpp/types/numeric_limits/digits10

Precision for float, double, and long double is 6, 15, and 18 respectively. We are setting them to one higher than the standard. I believe this is happening throughout the code base.

tntnatbry (approved):

PR #1329, opened 2020-07-06
Mcol 4126

davidhall586:
Since ci->tableOid isn't used until much farther in the logic, I moved it's creation down in the code. This resulted in fewer paths where cleanup is needed.


pleblanc1976:
Does this work with multi-table transactions?

tntnatbry:
Patch looks good to me. Can you address the above comment from @pleblanc1976 ?

davidhall586:
The test005 regression test has a number of multi-table transactions. It appears to work for them. I haven't studied the mechanism to figure out how multi-table transactions function.

tntnatbry (approved):

PR #1335, opened 2020-07-09
MCOL-4152 Clear condStack(s) at end of query

davidhall586:


drrtuy (changes requested):
We need to research the server's behavior w/o using workarounds b/c we can't predict when it fires back.

PR #1337, opened 2020-07-12
MCOL-4164 Multi-table DROP.

drrtuy:


dbcon/mysql/ha_mcs_ddl.cpp
```diff
@@ -107,6 +107,31 @@ inline uint32_t tid2sid(const uint32_t tid)
...
+    dbname_end++;
+
+  int cnt = dbname_end - dbname_start;
+  char *dbname = (char *)my_alloca(cnt + 1);
+  memcpy(dbname, dbname_start, cnt);
+  dbname[cnt] = '\0';
+  decode_objectname(decoded_dbname, dbname, FN_REFLEN);
+  my_afree(dbname);
+
+  char *tbname_start = dbname_end + 1;
```
tntnatbry, line 132: dbname_start/dbname_end/tbname_start should be const char, instead of char.

dbcon/mysql/ha_mcs_ddl.cpp
```diff
@@ -107,6 +107,31 @@ inline uint32_t tid2sid(const uint32_t tid)
...
+  decode_objectname(decoded_tbname, tbname_start, FN_REFLEN);
+}
```
tntnatbry, line 134: Indentation issue. `decode_file_path` function block is indented with 2 spaces, while `decode_objectname` is indented with 4 spaces.

tntnatbry (approved):

PR #1341, opened 2020-07-13
MCOL-641 Add support for functions (Part 2).

tntnatbry:


drrtuy (approved):
Something to look into:
We use getIntValue getter in all time related functions ignoring the width of the argument.

tntnatbry:
I think you are referring to code blocks of the form (in Func_hour::getIntVal, although other time related functions have similar code):

```sh
        {           
            if (parm[0]->data()->resultType().scale == 0)                                                                                    
            {   
                val = dataconvert::DataConvert::intToDatetime(parm[0]->data()->getIntVal(row, isNull));                                      
                
                if (val == -1)
                    isNull = true;                                                                                                           
            }                                                                                                                                
            else                                                                                                                             
            {                                                                                                                                
                isNull = true;                                                                                                               
            }                                                                                                                                
                                                                                                                                             
            break;                                                                                                                           
        }
```
Here, `parm[0]->data()->getIntVal(row, isNull)` calls TreeNode::getIntVal() which uses the column width of the DECIMAL type to determine whether to use fResult.decimalVal.s128Value or fResult.decimalVal.value for the result.

PR #1346, opened 2020-07-24
MCOL-4175 PP clean up

drrtuy:
Add int128_t support into ByteStream

Fixed UTs broken after collation patch

The main idea is to make generic changes appropriate for DTs wider than 16 bytes later. The patch is ready for review but don't merge it yet b/c I want to revive the primproc/linux-port/tdriver.cpp UT.


dbcon/execplan/calpontsystemcatalog.h
```diff
@@ -1021,7 +1021,6 @@ inline bool isUnsigned(const execplan::CalpontSystemCatalog::ColDataType type)
         case execplan::CalpontSystemCatalog::UMEDINT:
         case execplan::CalpontSystemCatalog::UINT:
         case execplan::CalpontSystemCatalog::UBIGINT:
-        // TODO MCOL-641 add decimal here
```
tntnatbry, line 1024: Shouldn't UDECIMAL/DECIMAL be part of isUnsigned/isSignedInteger check respectively?

dbcon/joblist/batchprimitiveprocessor-jl.cpp
```diff
@@ -117,9 +117,6 @@ void BatchPrimitiveProcessorJL::addFilterStep(const PseudoColStep& pcs)
     cc->setStepUuid(uuid);
     filterSteps.push_back(cc);
     filterCount++;
-    // TODO MCOL-641 How do we get to this execution path?
-    //if (utils::isWide(cc->getWidth()))
-    //    hasWideDecimalType = true;
```
tntnatbry, line 122: Why are we removing this here?

dbcon/joblist/batchprimitiveprocessor-jl.cpp
```diff
@@ -190,9 +187,6 @@ void BatchPrimitiveProcessorJL::addProjectStep(const PseudoColStep& step)
     colWidths.push_back(cc->getWidth());
     tupleLength += cc->getWidth();
     projectCount++;
-    // TODO MCOL-641 How do we get to this execution path?
-    //if (utils::isWide(cc->getWidth()))
-    //    hasWideDecimalType = true;
```
tntnatbry, line 195: Why are we removing this here?

dbcon/joblist/batchprimitiveprocessor-jl.cpp
```diff
@@ -775,21 +769,30 @@ void BatchPrimitiveProcessorJL::getRowGroupData(ByteStream& in, vector<RGData>*
...
+                {
+                    in >> tmp128;
+                    *min = tmp128;
+                    in >> tmp128;
+                    *max = tmp128;
+                }
+                else
+                {
+                    std::cout << __func__ << " WARNING!!! Not implemented for the data type "
+<< colType.colDataType << std::endl;
```
tntnatbry: I haven't seen this styling anywhere else in the code base. When the stream is spanning multiple lines, we have code of the form (this is from dbcon/joblist/jlf_tuplejoblist.cpp::joinToLargeTable):
```
                oss << smallKey.str() << " join on " << largeKey.str()
                    << " joinType: " << info->fJoinData.fTypes.front()
                    << "(" << joinTypeToString(info->fJoinData.fTypes.front()) << ")"
                    << (info->fJoinData.fTypeless ? " " : " !") << "typeless" << endl;
                oss << "smallSideIndex-largeSideIndex :" << smallIndex.str() << " --"
                    << largeIndex.str() << endl;
                oss << "small side RG" << endl << info->fRowGroup.toString() << endl;
```
Let's keep a consistent styling.

primitives/linux-port/column.cpp
```diff
@@ -278,18 +278,18 @@ inline bool isEmptyVal(uint8_t type, const uint8_t* val8);
 template<>
 inline bool isEmptyVal<32>(uint8_t type, const uint8_t* ival) // For BINARY
 {
-    const uint64_t* val = reinterpret_cast<const uint64_t*>(ival);
-    return ((val[0] == joblist::BINARYEMPTYVALUELOW)
-            && (val[1] == joblist::BINARYNULLVALUELOW)
-            && (val[2] == joblist::BINARYNULLVALUELOW)
-            && (val[3] == joblist::BINARYEMPTYVALUEHIGH));
+    std::cout << __func__ << " WARNING!!! Not implemented for 32 byte data types."
+<< std::endl;
```
tntnatbry: Styling issue for the stream, see the comment above.
Shouldn't we assert here, similar to ExeMgr?

primitives/linux-port/column.cpp
```diff
@@ -413,22 +413,22 @@ inline bool isEmptyVal<1>(uint8_t type, const uint8_t* ival)
...
 {
-    const uint64_t* val = reinterpret_cast<const uint64_t*>(ival); 
-    return ((val[0] == joblist::BINARYNULLVALUELOW)
-            && (val[1] == joblist::BINARYNULLVALUELOW)
-            && (val[2] == joblist::BINARYNULLVALUELOW)
-            && (val[3] == joblist::BINARYNULLVALUEHIGH));
+
+    std::cout << __func__ << " WARNING!!! Not implemented for 32 byte data types."
+<< std::endl;
+    return false;
```
tntnatbry, line 430: Styling issue for the stream, see the comment above.
Shouldn't we assert here, similar to ExeMgr?

primitives/linux-port/column.cpp
```diff
@@ -558,12 +558,9 @@ inline bool isNullVal(uint32_t length, uint8_t type, const uint8_t* val8)
 {
     switch (length)
     {
-        case 32:
-            return isNullVal<32>(type, val8);
```
tntnatbry, line 562: If isNullVal<32> is removed from here, do we need this function implemented above?

primitives/linux-port/column.cpp
```diff
@@ -576,7 +573,8 @@ inline bool isNullVal(uint32_t length, uint8_t type, const uint8_t* val8)
         case 1:
             return isNullVal<1>(type, val8);
     };
-
+    std::cout << __func__ << " WARNING!!! Not implemented for " << length << " bytes data types." << std::endl;
```
tntnatbry, line 575: Should this also assert, similar to ExeMgr?

primitives/linux-port/column.cpp
```diff
@@ -758,8 +756,7 @@ inline void store(const NewColRequestHeader* in,
         switch (in->DataSize)
         {
             case 32:
-                ptr2 += (rid << 5);
-                memcpy(ptr1, ptr2, 32);
+                std::cout << __func__ << " WARNING!!! Not implemented for 32 byte data types." << std::endl;
```
tntnatbry, line 758: Should we assert here?

primitives/linux-port/column.cpp
```diff
@@ -920,20 +916,23 @@ inline uint8_t* nextBinColValue(int type,
...
     }
 
-    *isNull = isNullVal<W>(type, &val8[*rid * W]);
-    *isEmpty = isEmptyVal<W>(type, &val8[*rid * W]);
-    //cout << "nextUnsignedColValue index " << *index <<  " rowid " << *rid << endl;
+    uint32_t curValueOffset = *rid * W;
+
+    *isNull = isNullVal<W>(type, &val8[curValueOffset]);
+    *isEmpty = isEmptyVal<W>(type, &val8[curValueOffset]);
+    //cout << "nextColBinValue " << *index <<  " rowid " << *rid << endl;
```
tntnatbry, line 922: It should be BinCol here instead of ColBin, and also the missing "index" :)

primitives/linux-port/column.cpp
```diff
@@ -920,20 +916,23 @@ inline uint8_t* nextBinColValue(int type,
...
+    return &val8[curValueOffset];
 
 #ifdef PRIM_DEBUG
             throw logic_error("PrimitiveProcessor::nextColBinValue() bad width");
 #endif
             return NULL;
 }
-}
+
+} // anon namespace end
```
tntnatbry: This is not the right place to end the anonymous namespace. It should be ending at the end of p_Col_bin_ridArray().

primitives/linux-port/column.cpp
```diff
@@ -1590,15 +1587,16 @@ inline void p_Col_bin_ridArray(NewColRequestHeader* in,
 
     if (out->ValidMinMax)
     {
+        // Assume that isUnsigned returns true for 8-bytes DTs only
```
tntnatbry, line 1588: I think we should include UDECIMAL/DECIMAL to isUnsigned/isSigned checks.

primitives/linux-port/column.cpp
```diff
@@ -1731,20 +1727,13 @@ inline void p_Col_bin_ridArray(NewColRequestHeader* in,
...
                     out->Max = val;
             }
-            else if (isUnsigned((CalpontSystemCatalog::ColDataType)in->DataType))
-            {
-                if (static_cast<uint128_t>(out->Min) > static_cast<uint128_t>(val))
-                    out->Min = val;
-
-                if (static_cast<uint128_t>(out->Max) < static_cast<uint128_t>(val))
-                    out->Max = val;
-            }
```
tntnatbry, line 1747: I think we should keep this branch here.

primitives/linux-port/column.cpp
```diff
@@ -1932,8 +1920,6 @@ boost::shared_ptr<ParsedColumnFilter> parseColumnFilter
                 case 8:
                     ret->prestored_argVals[argIndex] = *reinterpret_cast<const uint64_t*>(args->val);
                     break;  
-                case 16:
-                    cout << __FILE__<< ":" <<__LINE__ << " Fix for 16 Bytes ?" << endl;
```
tntnatbry, line 1936: As mentioned earlier in the comments, I think we should return true for UDECIMAL for this func, and hence implement case 16.

primitives/linux-port/primitiveprocessor.h
```diff
@@ -329,6 +329,7 @@ class PrimitiveProcessor
     int dict_OffsetIndex, currentOffsetIndex;		// used by p_dictionary
     int fDebugLevel;
     dbbc::Stats* fStatsPtr; // pointer for pmstats
+    // To be removed b/c always true
     bool logicalBlockMode;
```
tntnatbry, line 333: Are we removing this var in this branch?

primitives/primproc/batchprimitiveprocessor.cpp
```diff
@@ -1991,15 +1988,16 @@ void BatchPrimitiveProcessor::writeProjectionPreamble()
...
-            if (hasBinaryColumn)
+            if (UNLIKELY(hasWideColumnOut))
             {
-                *serialized << (uint8_t) 16; // width of min/max value
-                *serialized << (unsigned __int128) bigMinVal;
-                *serialized << (unsigned __int128) bigMaxVal;
+                // PSA width
+                *serialized << (uint8_t) wideColumnWidthOut;
+                *serialized << min128Val;
+                *serialized << max128Val;
```
tntnatbry, line 1997: These should be casted to unsigned before serializing, as the receiving end is deserializing uint128_t.

primitives/primproc/batchprimitiveprocessor.cpp
```diff
@@ -1991,15 +1988,16 @@ void BatchPrimitiveProcessor::writeProjectionPreamble()
         {
             *serialized << (uint8_t) 1;
             *serialized << lbidForCP;
-            if (hasBinaryColumn)
+            if (UNLIKELY(hasWideColumnOut))
             {
-                *serialized << (uint8_t) 16; // width of min/max value
-                *serialized << (unsigned __int128) bigMinVal;
-                *serialized << (unsigned __int128) bigMaxVal;
+                // PSA width
```
tntnatbry, line 1994: What does PSA mean?

primitives/primproc/batchprimitiveprocessor.cpp
```diff
@@ -2091,15 +2089,19 @@ void BatchPrimitiveProcessor::makeResponse()
...
+                // Remove the assert for >16 bytes DTs.
+                assert(wideColumnWidthOut == datatypes::MAXDECIMALWIDTH);
+                *serialized << (uint8_t) wideColumnWidthOut;
+                *serialized << min128Val;
+                *serialized << max128Val;
```
tntnatbry, line 2101: Again, should be casted to unsigned.

primitives/primproc/columncommand.cpp
```diff
@@ -282,11 +282,24 @@ void ColumnCommand::issuePrimitive()
         //if (wasVersioned && outMsg->ValidMinMax)
         //	cout << "CC: versioning overriding min max data\n";
         bpp->lbidForCP = lbid;
-        if (primMsg->DataSize > 8)
+        if (UNLIKELY(utils::isWide(primMsg->DataSize)))
         {
-            bpp->hasBinaryColumn = true;
-            bpp->bigMaxVal = outMsg->Max;
-            bpp->bigMinVal = outMsg->Min;
+            if (datatypes::Decimal::isWideDecimalType(colType))
```
tntnatbry, line 287: In the outer if, we are checking for primMsg->DataSize. While in the inner if, we are using colType.colWidth. Can they be unequal?

utils/messageqcpp/bytestream.cpp
```diff
@@ -429,6 +444,15 @@ void ByteStream::peek(uint128_t& o) const
     o = *((uint128_t*) fCurOutPtr);
 }
 
+void ByteStream::peek(int128_t& o) const
+{
+
+    if (length() < 16)
+        throw underflow_error("ByteStream>int128_t: not enough data in stream to fill datatype");
+
+    o = *((uint128_t*) fCurOutPtr);
```
tntnatbry: Should be int128_t*

utils/messageqcpp/bytestream.h
```diff
@@ -76,7 +77,7 @@ class ByteStream : public Serializeable
     typedef uint16_t doublebyte;
     typedef uint32_t quadbyte;
     typedef uint64_t octbyte;
-    typedef uint128_t hexbyte;
+    typedef int128_t hexbyte;
```
tntnatbry, line 80: Why are we changing it to signed here?

tntnatbry (changes requested):

dbcon/execplan/calpontsystemcatalog.h
drrtuy: It wasn't so far. Do you suggest to change the default and see where it fires back?

dbcon/joblist/batchprimitiveprocessor-jl.cpp
drrtuy: Did you put the comment here getting to this place actually in the first place?

dbcon/joblist/batchprimitiveprocessor-jl.cpp
drrtuy: Did you put the comment here getting to this place actually in the first place?

primitives/linux-port/column.cpp
drrtuy: I see no reason for that.

primitives/linux-port/column.cpp
drrtuy: I left it as an example.

primitives/linux-port/column.cpp
drrtuy: Change of the default. I won't do this now.

primitives/linux-port/column.cpp
drrtuy: Plz look the comment above.

primitives/linux-port/column.cpp
drrtuy: I had no intentions to make it actual. It seems nextUnsignedColValue changed to nextColBinValue with vim command.

primitives/linux-port/column.cpp
drrtuy: I see no reason to change the existing namespace.

primitives/linux-port/column.cpp
drrtuy: Big change of the default w/o visible benefit IMHO.

primitives/linux-port/column.cpp
drrtuy: Given that we aren't going to change isUnsigned I disagree.

primitives/linux-port/column.cpp
drrtuy: Plz look at the comment above.

primitives/linux-port/primitiveprocessor.h
drrtuy: No, not yet b/c it is also change of the default.

drrtuy: I plan to remove it merging column scanning code contribution later.

primitives/primproc/batchprimitiveprocessor.cpp
drrtuy: Precalculated Small Aggregate this is the term for what we call min/max.

primitives/primproc/batchprimitiveprocessor.cpp
drrtuy: I see no reason if ByteStream supports signed.

primitives/primproc/columncommand.cpp
drrtuy: As of now no, b/c colType is ctored inside ::prep and colType.colWidth is taken from primMsg->DataSize.

drrtuy: But I will make it consistent.

utils/messageqcpp/bytestream.h
drrtuy: B/c we use uint128_t nowhere and ByteStream::operator<< now supports int128_t arguments.

dbcon/joblist/batchprimitiveprocessor-jl.cpp
drrtuy: You should search for example in primitives/blockcache/iomanager.cpp, primitives/primproc/pseudocc.cpp, primitives/primproc/bppseeder.cpp, primitives/primproc/primproc.h
I think the question of style is of a personal preference. However I agree that we should follow it.

primitives/linux-port/column.cpp
drrtuy: No, we shouldn't assert here b/c this template was given as an example for future development and the code doesn't use it.

dbcon/execplan/calpontsystemcatalog.h
drrtuy: We won't be changing it doing wide-DECIMAL. This change is for datatype framework project IMHO.

utils/messageqcpp/bytestream.h
drrtuy: For wide-DECIMAL it shouldn't matter b/c we just send 16 bytes over the wire and sign shouldn't interfere with this operation.
Moreover if we consider int128_t as a container datatype for all 16 byte DTs there is no difference if we use signed or unsigned type b/c we'll treat it as char[16].
What was your thinking when you asked?

utils/messageqcpp/bytestream.cpp
drrtuy: Good catch.

dbcon/execplan/calpontsystemcatalog.h
tntnatbry: Ok. I have added a comment in MCOL-3609 to make this change.

dbcon/joblist/batchprimitiveprocessor-jl.cpp
tntnatbry: Yes, I remember putting this comment here earlier. I am not sure what leads a call to BatchPrimitiveProcessorJL::addFilterStep(const PseudoColStep& pcs).

dbcon/joblist/batchprimitiveprocessor-jl.cpp
tntnatbry: Yes, I remember putting this comment here earlier. I am not sure what leads a call to BatchPrimitiveProcessorJL::addProjectStep(const PseudoColStep& step).

dbcon/joblist/batchprimitiveprocessor-jl.cpp
drrtuy: I might didn't make the question clear. Have you seen this place in the call stack running a query?

dbcon/joblist/batchprimitiveprocessor-jl.cpp
tntnatbry: I have looked at all stream usages in the examples you mentioned, I don't see this styling anywhere. As an example, this is what I see in primitives/blockcache/iomanager.cpp (the starting stream operators '<<' are all aligned at the same column):
```
class FdEntry
{
public:
    FdEntry() : oid(0), dbroot(0), partNum(0), segNum(0),
        fp(0), c(0), inUse(0), compType(0)
    {
        cmpMTime = 0;
    }

    FdEntry(const BRM::OID_t o, const uint16_t d, const uint32_t p, const uint16_t s, const int ct, IDBDataFile* f) :
        oid(o), dbroot(d), partNum(p), segNum(s), fp(f), c(0), inUse(0), compType(0)
    {
        cmpMTime = 0;

        if (oid >= 1000)
            compType = ct;
    }

    ~FdEntry()
    {
        delete fp;
        fp = 0;
    }

    BRM::OID_t oid;
    uint16_t dbroot;
    uint32_t partNum;
    uint16_t segNum;
    IDBDataFile* fp;
    uint32_t c;
    int inUse;

    CompChunkPtrList ptrList;

    int compType;
    bool isCompressed() const
    {
        return (oid >= 1000 && compType != 0);
    }
    time_t cmpMTime;
    friend ostream& operator<<(ostream& out, const FdEntry& o)
    {
        out << " o: " << o.oid
            << " f: " << o.fp
            << " d: " << o.dbroot
            << " p: " << o.partNum
            << " s: " << o.segNum
            << " c: " << o.c
            << " t: " << o.compType
            << " m: " << o.cmpMTime;
        return out;
    }
};
```

primitives/linux-port/column.cpp
tntnatbry: I should clarify. We have changed the namespace by mistake at some point earlier while working on MCOL-641. Right place in develop where the anonymous namespace is ending is p_Col_ridArray(). So, in MCOL-641, we should end it after p_Col_bin_ridArray().

primitives/linux-port/column.cpp
tntnatbry: If/when we add UDECIMAL in isUnsigned, we will miss adding the support here in future, if we remove this branch. So, we should keep the isUnsigned check here.

utils/messageqcpp/bytestream.h
tntnatbry: The code in the vicinity, e.g. octbyte, is defined as unsigned. So I thought it would make sense to defined it that way. But I agree, the signedness doesn't matter in this case.

dbcon/joblist/batchprimitiveprocessor-jl.cpp
drrtuy: I wasn't able to neither get to this place nor crash the engine using pseudocolumn ops available.

primitives/linux-port/column.cpp
drrtuy: GTK

dbcon/joblist/batchprimitiveprocessor-jl.cpp
tntnatbry: I think the fundamental question here is whether BatchPrimitiveProcessorJL::addFilterStep(const PseudoColStep& pcs) or BatchPrimitiveProcessorJL::addProjectStep(const PseudoColStep& step) ever gets called from anywhere or not? In other words, are these two functions dead?

If not, then we should keep these commented out blocks for wide decimal, and later identify when `utils::isWide(cc->getWidth())` can return true.

drrtuy: The code is used by two of [these functions](https://mariadb.com/kb/en/columnstore-information-functions/) , namely idbExtentMax and idbExtentMin used in either projection or filtering. However I don't think the comments helps with that. The change must be done in pseudocolumn processing code in PP, joblist. I'll address this after I get back from vacation.

tntnatbry (approved):

PR #1349, opened 2020-07-27
Extent elimination refactoring and extent map upgrade from v4 to v5

tntnatbry:


tntnatbry:
@dhall-MariaDB Can you review the extent map upgrade code from v4 to v5 in versioning/BRM/extentmap.h(.cpp) along with @drrtuy ?

dbcon/joblist/lbidlist.cpp
```diff
@@ -296,13 +297,29 @@ bool LBIDList::GetMinMax(int64_t* min, int64_t* max, int64_t* seq,
 
         if (isUnsigned(colDataType))
```
drrtuy, line 298: It is safe but sUnsigned from CSC namespace returns false for DECIMAL.

davidhall586 (approved):

dbcon/joblist/lbidlist.cpp
drrtuy: *isUnsigned

drrtuy (changes requested):
Don't merge it yet.

dbcon/joblist/lbidlist.cpp
```diff
@@ -653,32 +680,46 @@ inline bool LBIDList::compareVal(const T& Min, const T& Max, const T& value, cha
...
+        }
+        else
+        {
+            return (static_cast<uint64_t>(value) >= static_cast<uint64_t>(min) &&
+                    static_cast<uint64_t>(value) <= static_cast<uint64_t>(max));
+        }
     }
     else
     {
         return (value >= min && value <= max);
```
drrtuy, line 711: Will we compare wide-DECIMALs getting into else branch?

drrtuy: I think the first question is: if it is possible to get here with wide-DECIMALs in the first place.

dbcon/joblist/lbidlist.cpp
```diff
@@ -759,12 +807,13 @@ bool LBIDList::CasualPartitionPredicate(const BRM::EMCasualPartition_t& cpRange,
...
                     value = static_cast<int64_t>(val);
                 }
+
                 case 16:
                 {
                     unsigned __int128 val;
-                    bigValuePod = reinterpret_cast<dataconvert::Int128Pod_t*>(&val);
-                    bigValuePod->lo = *reinterpret_cast<const uint64_t*>(MsgDataPtr);
-                    bigValuePod->hi = *(reinterpret_cast<const uint64_t*>(MsgDataPtr) + 1);
+                    int128Ptr = reinterpret_cast<uint64_t*>(&val);
```
drrtuy: This approach surely works. 
But as times goes by the vision had changed and I see this as a 16-bytes read from BS::buf() and subsequent int128 cast though. It's up to you to decide should we proceed with this version or not.

dbcon/joblist/lbidlist.cpp
```diff
@@ -918,6 +968,16 @@ bool LBIDList::GetMinMax<int64_t>(int64_t& min, int64_t& max, int64_t& seq, int6
                                   const std::vector<struct BRM::EMEntry>* pEMEntries,
                                   execplan::CalpontSystemCatalog::ColDataType colDataType);
 
+template
+bool LBIDList::GetMinMax<__int128>(__int128* min, __int128* max, int64_t* seq,
```
drrtuy, line 972: I believe the code doesn't compile but just out of curiosity did you try to compile it w/o this explicit template specialization?

drrtuy: The reason why I'm asking is that there is only one version of GetMinMax and that is templated. linker might ask for an explicit template to link against it though.

dbcon/joblist/rowestimator.cpp
```diff
@@ -140,12 +140,13 @@ uint64_t RowEstimator::adjustValue(const execplan::CalpontSystemCatalog::ColType
...
+template<typename T>
 uint32_t RowEstimator::estimateDistinctValues(const execplan::CalpontSystemCatalog::ColType& ct,
-        const uint64_t& min,
-        const uint64_t& max,
+        const T& min,
+        const T& max,
         const char cpStatus)
 {
-    uint64_t ret = 10;
+    T ret = 10;
```
drrtuy, line 149: It seems like an automated change has been applied here. 
We return uint32_t from this f() into uint32_t variable so ret should be uint32_t not even uint64_t. Given the calculations this should be safe.

dbcon/joblist/rowestimator.cpp
```diff
@@ -199,7 +200,7 @@ uint32_t RowEstimator::estimateDistinctValues(const execplan::CalpontSystemCatal
         ret = max - min + 1;
     }
 
-    if (ret > fRowsPerExtent)
+    if (ret > (T) fRowsPerExtent)
```
drrtuy: Could you ellaborate on this cast?

dbcon/joblist/rowestimator.cpp
```diff
@@ -238,7 +248,10 @@ float RowEstimator::estimateOpFactor(const T& min, const T& max, const T& value,
         case COMPARE_NLE:
             if (cpStatus == BRM::CP_VALID)
             {
-                factor = (1.0 * max - value) / (1.0 * max - min + 1);
+                if (!datatypes::Decimal::isWideDecimalType(ct))
+                    factor = (1.0 * max - value) / (1.0 * max - min + 1);
+                else
+                    factor = (1.0 * max - value) / (1.0 * max - min + 1);
```
drrtuy: I presume here math works w/o an explicit cast to float128?

dbcon/joblist/tuple-bps.cpp
```diff
@@ -1582,11 +1585,19 @@ bool TupleBPS::processSingleFilterString(int8_t BOP, int8_t colWidth, int64_t va
...
+                bigVal2 = *((int128_t*) filterString);
+                filterString += 16;
+                break;
+
             default:
                 throw logic_error("invalid column width");
         }
 
-        thisPredicate = compareSingleValue(COP, val, val2);
+        if (colWidth < 16)
```
drrtuy: I think we should replace this with a constant from datatypes::

drrtuy: We should reduce a number of hardcoded values around the codebase I believe.

dbcon/joblist/tuple-bps.cpp
```diff
@@ -1691,9 +1703,13 @@ bool TupleBPS::processPseudoColFilters(uint32_t extentIndex, boost::shared_ptr<m
                && (!hasSegmentDirFilter || processOneFilterType(8, emEntry.partitionNum, PSEUDO_SEGMENTDIR))
                && (!hasExtentIDFilter || processOneFilterType(8, emEntry.range.start, PSEUDO_EXTENTID))
                && (!hasMaxFilter || (emEntry.partition.cprange.isValid == BRM::CP_VALID ?
-                                     processOneFilterType(emEntry.range.size, emEntry.partition.cprange.hi_val, PSEUDO_EXTENTMAX) : true))
+                                     (!datatypes::Decimal::isWideDecimalType(fColType) ?
```
drrtuy, line 1706: Who is the author of this. I wish him to work with such condition expressions all his life.

dbcon/mysql/ha_mcs_partition.cpp
```diff
@@ -1184,16 +1318,16 @@ extern "C"
 
             if (partIt->second.status & CPINVALID)
             {
-                if (ct.colWidth <= 8)
+                if (!datatypes::Decimal::isWideDecimalType(ct))
                     output << setw(30) << "N/A" << setw(30) << "N/A";
                 else
-                    output << setw(40) << "N/A" << setw(40) << "N/A";
+                    output << setw(42) << "N/A" << setw(42) << "N/A";
```
drrtuy: Could you replace this hardcoded value with a corresponding const from datatypes:: namespace from decimal.h or I can do this after the merge.

dbcon/joblist/lbidlist.cpp
tntnatbry: Keeping this here for now, in case we add UDECIMAL to isUnsigned in the future.

dbcon/joblist/lbidlist.cpp
tntnatbry: Yes. The calling function, TupleBPS::addCPPredicates(), filters out extents of the large-side table, based on the CP info of the small-side tables. So if either the small-side or the large-side table contains wide-decimal columns as join keys (which we currently do not support), we will get here for wide-decimals.

dbcon/joblist/lbidlist.cpp
tntnatbry: Linker complains if I remove this template instantiation. Reason for this is the client of GetMinMax() is tuple-bps.cpp, which is a separate translation unit.

dbcon/joblist/rowestimator.cpp
tntnatbry: This is intentional. The code at the end of this function is this:

```
    else
    {
        ret = max - min + 1;
    }

    if (ret > (T) fRowsPerExtent)
    {
        ret = fRowsPerExtent;
    }

    return ret;
```

So the type of ret has to be same as max/min, since we perform arithmetic on these values. Subsequently, `if (ret > (T) fRowsPerExtent)` ensures the maximum value returned fits in an uint32_t.

dbcon/joblist/lbidlist.cpp
drrtuy: And I c that we don't treat value based on the dt used. Should't we have the same cimparison as in preceding blocks?

dbcon/joblist/rowestimator.cpp
tntnatbry: I agree, this cast is not required. I am updating this.

dbcon/joblist/lbidlist.cpp
tntnatbry: I agree. A direct cast to int128_t* should work just fine. I am changing this.

dbcon/joblist/rowestimator.cpp
tntnatbry: Good catch. I missed the cast to __float128 here and in the next case, fixing this.

dbcon/joblist/tuple-bps.cpp
tntnatbry: Agreed. Changing this.

dbcon/joblist/tuple-bps.cpp
tntnatbry: That's a million dollar question you have asked, I wish I had the answer. But I take part of the blame :)

dbcon/mysql/ha_mcs_partition.cpp
tntnatbry: Done. I've replaced the hard coded value with utils::MAXLENGTH16BYTES

tntnatbry:
Changes have been pushed based on the feedback, please review.

dbcon/joblist/lbidlist.cpp
tntnatbry: Not sure I understand the last comment. The preceding else block is for unsigned types, for which the comparison involves first casting to the unsigned equivalent of either in128_t or int64_t. For the last else condition (signed datatypes), we compare the values directly.

drrtuy (approved):

PR #1354, opened 2020-07-31
MCOL-4171

davidhall586:
Window Function changes for decimal(38)


dbcon/execplan/windowfunctioncolumn.cpp
```diff
@@ -387,9 +389,16 @@ void WindowFunctionColumn::adjustResultType()
         boost::iequals(fFunctionName, "AVG") ||
         boost::iequals(fFunctionName, "AVG_DISTINCT"))
     {
-        fResultType.colDataType = CalpontSystemCatalog::LONGDOUBLE;
-        fResultType.colWidth = sizeof(long double);
-        fResultType.precision = -1;
+        if (fFunctionParms[0]->resultType().colDataType == CalpontSystemCatalog::DECIMAL)
```
tntnatbry: Don't we also need to check for UDECIMAL here? As well as in the earlier if condition in this function?

dbcon/execplan/windowfunctioncolumn.cpp
```diff
@@ -661,18 +670,35 @@ void WindowFunctionColumn::evaluate(Row& row, bool& isNull)
...
-                default:
+                case 8:
                 {
                     if (row.equals<8>(BIGINTNULL, fInputIndex))
                         isNull = true;
                     else
                     {
-                        fResult.decimalVal.value = (int64_t)row.getUintField<8>(fInputIndex);
+                        fResult.decimalVal.value = row.getIntField<8>(fInputIndex);
+                        fResult.decimalVal.scale = (unsigned)fResultType.scale;
```
tntnatbry, line 683: Don't we also need to set fResult.decimalVal.precision here?

utils/messageqcpp/bytestream.cpp
```diff
@@ -236,6 +236,17 @@ ByteStream& ByteStream::operator<<(const uint64_t o)
 }
 
 // WIP MCOL-641
+ByteStream& ByteStream::operator<<(const int128_t& o)
```
tntnatbry: @drrtuy has already added support for int128_t in bytestream.h/.cpp in his PR: https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/1346/files

Can you undo this change?

utils/regr/moda.h
```diff
@@ -54,6 +55,38 @@
...
+struct hasher
+{
+    inline size_t operator()(T val) const
+    {
+        return fHasher((char*) &val, sizeof(T));
+    }
+
+private:
+    utils::Hasher fHasher;
+};
```
tntnatbry, line 69: I wasn't aware of this usage of the Hasher class from hasher.h. In the casual partitioning PR, I added Hash128 and Equal128 classes in hasher.h: https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/1349/files#diff-2f5b44d849a51ebb552b1793a8ed93cdR369

As you will see, I am merely returning the lower 64 bytes as the hash value, which does not look like the right approach, as it will lead to all 2^64 values in the upper 64 bytes getting the same hash values for the same bit pattern in the lower 64 bytes -> significant collisions.

We should consolidate this and just have this implementation at a single place.

utils/rowgroup/rowgroup.h
```diff
@@ -343,14 +343,17 @@ class Row
     template<int len> inline bool equals(uint64_t val, uint32_t colIndex) const;
     inline bool equals(long double val, uint32_t colIndex) const;
     bool equals(const std::string& val, uint32_t colIndex) const;
-
+    inline bool equals(int128_t val, uint32_t colIndex) const;
```
tntnatbry: Shouldn't val be a const ref?

utils/rowgroup/rowgroup.h
```diff
@@ -380,6 +383,7 @@ class Row
     inline void setFloatField(float val, uint32_t colIndex);
     inline void setDecimalField(double val, uint32_t colIndex) { };  // TODO: Do something here
     inline void setLongDoubleField(long double val, uint32_t colIndex);
+    inline void setInt128Field(int128_t val, uint32_t colIndex);
```
tntnatbry: Shouldn't val be a const ref?

utils/windowfunction/wf_percentile.cpp
```diff
@@ -90,6 +88,32 @@ boost::shared_ptr<WindowFunctionType> WF_percentile<T>::makeFunction(int id, con
                 break;
             }
 
+            case CalpontSystemCatalog::DECIMAL:
+            {
+                if (wc->functionParms()[0]->resultType().colWidth < 16)
+                {
+                    func.reset(new WF_percentile<int128_t>(id, name));
```
tntnatbry: Should be int64_t here.

utils/windowfunction/wf_percentile.cpp
```diff
@@ -90,6 +88,32 @@ boost::shared_ptr<WindowFunctionType> WF_percentile<T>::makeFunction(int id, con
...
+                }
+                break;
+            }
+            
+            case CalpontSystemCatalog::UDECIMAL:
+            {
+                if (wc->functionParms()[0]->resultType().colWidth < 16)
+                {
+                    func.reset(new WF_percentile<uint128_t>(id, name));
```
tntnatbry: Should be uint64_t here.

utils/windowfunction/wf_sum_avg.cpp
```diff
@@ -1,4 +1,4 @@
-/* Copyright (C) 2014 InfiniDB, Inc.
+/* CopyrighT (C) 2014 InfiniDB, Inc.
```
tntnatbry, line 1: The lower case 't' looked perfectly ok earlier :)

utils/windowfunction/wf_sum_avg.cpp
```diff
@@ -50,112 +50,116 @@ using namespace joblist;
...
-
-long double avgWithLimit(long double sum, uint64_t count, int scale, long double u, long double l)
-{
-    long double factor = pow(10.0, scale);
-    long double avg = sum / count;
-    avg *= factor;
-    avg += (avg < 0) ? (-0.5) : (0.5);
-
-    if (avg > u || avg < l)
+    int128_t factor = pow(10.0, scale);
```
tntnatbry: I don't think the pow() library function supports int128_t. You might want to look at how we are calculating this factor in some of the utils/funcexp support that we added. I am posting a code snippet from Func_ceil::getUintVal for easy reference (see the usage of datatypes::getScaleDivisor(scaleDivisor, d.scale);):

```
        {
            IDB_Decimal d = parm[0]->data()->getDecimalVal(row, isNull);

            if (isNull)
                break;

            // negative scale is not supported by CNX yet
            if (d.scale > 0)
            {
                if (d.scale > datatypes::INT128MAXPRECISION)
                {
                    std::ostringstream oss;
                    oss << "ceil: datatype of " << execplan::colDataTypeToString(op_ct.colDataType)
                        << " with scale " << (int) d.scale << " is beyond supported scale";
                    throw logging::IDBExcept(oss.str(), ERR_DATATYPE_NOT_SUPPORT);
                }

                if (op_ct.colWidth == datatypes::MAXDECIMALWIDTH)
                {
                    int128_t tmp = d.s128Value;
                    int128_t scaleDivisor;
                    datatypes::getScaleDivisor(scaleDivisor, d.scale);
                    d.s128Value /= scaleDivisor;

                    // Add 1 if this is a positive number and there were values to the right of the
                    // decimal point so that we return the largest integer value not less than X.
                    if ((tmp - (d.s128Value * scaleDivisor)) > 0)
                        d.s128Value += 1;

                    ret = datatypes::Decimal::getUInt64FromWideDecimal(d.s128Value);
                }
                else
                {
                    int64_t tmp = d.value;
                    d.value /= helpers::powerOf10_c[d.scale];

                    // Add 1 if this is a positive number and there were values to the right of the
                    // decimal point so that we return the largest integer value not less than X.
                    if ((tmp - (d.value * helpers::powerOf10_c[d.scale])) > 0)
                        d.value += 1;

                    ret = (uint64_t) d.value;
                }
            }
        }
        break;
```

utils/windowfunction/wf_sum_avg.cpp
```diff
@@ -50,112 +50,116 @@ using namespace joblist;
...
-
-template<>
-int64_t calculateAvg<int64_t>(int64_t sum, uint64_t count, int scale)
+template<typename T_IN, typename T_OUT>
+uint128_t WF_sum_avg<T_IN, T_OUT>::calculateAvg(uint128_t sum, uint64_t count, int scale)
 {
-    int64_t t = (int64_t) avgWithLimit(sum, count, scale,
-                                       numeric_limits<int64_t>::max(), numeric_limits<int64_t>::min());
-    return t;
+    uint128_t factor = pow(10.0, scale);
```
tntnatbry: Same here. See my comment above.

utils/windowfunction/wf_udaf.cpp
```diff
@@ -542,11 +542,13 @@ void WF_udaf::SetUDAFValue(static_any::any& valOut, int64_t colOut,
     static const static_any::any& intTypeId = (int)1;
     static const static_any::any& longTypeId = (long)1;
     static const static_any::any& llTypeId = (long long)1;
+    static const static_any::any& int128TypeId = (int128_t)1;
     static const static_any::any& ucharTypeId = (unsigned char)1;
     static const static_any::any& ushortTypeId = (unsigned short)1;
     static const static_any::any& uintTypeId = (unsigned int)1;
     static const static_any::any& ulongTypeId = (unsigned long)1;
     static const static_any::any& ullTypeId = (unsigned long long)1;
+    static const static_any::any& uint128TypeId = (int128_t)1;
```
tntnatbry: This should be uint128_t?

utils/windowfunction/windowfunctiontype.cpp
```diff
@@ -337,6 +357,11 @@ template<> void WindowFunctionType::setValue<long double>(uint64_t i, long doubl
     fRow.setLongDoubleField(t, i);
 }
 
+template<> void WindowFunctionType::setValue<int128_t>(uint64_t i, int128_t& t)
+{
+    fRow.setInt128Field(t, i);
+}
```
tntnatbry, line 363: Shouldn't t be a const ref here? And how about specialization of setValue for uint128_t?

utils/windowfunction/windowfunctiontype.cpp
```diff
@@ -483,6 +518,40 @@ void WindowFunctionType::implicit2T(uint64_t i, T& t, int s)
...
         }
 
+        case CalpontSystemCatalog::DECIMAL:
+        {
+            uint32_t w = fRow.getColumnWidth(i);
+            if (w < 16)
+                t = (T) fRow.getIntField(i);
+            else 
+                t = (T) fRow.getInt128Field(i);
+            pw = s - fRow.getScale(i); // pw is difference of scales, will be in [-18, 18]
```
tntnatbry: [-38, 38] ?

utils/windowfunction/windowfunctiontype.cpp
```diff
@@ -483,6 +518,40 @@ void WindowFunctionType::implicit2T(uint64_t i, T& t, int s)
...
+
+            if (pw > 0)
+                t *= IDB_pow[pw];
```
tntnatbry: IDB_pow does not handle int128_t if pw > 18. Refer to one of my earlier comments regarding this.

utils/windowfunction/windowfunctiontype.cpp
```diff
@@ -483,6 +518,40 @@ void WindowFunctionType::implicit2T(uint64_t i, T& t, int s)
...
+        }
+
+        case CalpontSystemCatalog::UDECIMAL:
+        {
+            uint32_t w = fRow.getColumnWidth(i);
+            if (w < 16)
+                t = (T) fRow.getUintField(i);
+            else 
+                t = (T) fRow.getUint128Field(i);
+            pw = s - fRow.getScale(i); // pw is difference of scales, will be in [-18, 18]
```
tntnatbry: [-38, 38] ?

utils/windowfunction/windowfunctiontype.cpp
```diff
@@ -483,6 +518,40 @@ void WindowFunctionType::implicit2T(uint64_t i, T& t, int s)
...
+
+            if (pw > 0)
+                t *= IDB_pow[pw];
```
tntnatbry: IDB_pow does not handle int128_t if pw > 18. Refer to one of my earlier comments regarding this.

utils/windowfunction/windowfunctiontype.h
```diff
@@ -238,6 +242,10 @@ class WindowFunctionType
     {
         fRow.setIntField(v, i);
     }
+    void    setInt128Value(int64_t i, int128_t v)
```
tntnatbry: Shouldn't v be a const ref here?

writeengine/wrapper/writeengine.cpp
```diff
@@ -837,7 +837,18 @@ int WriteEngineWrapper::deleteRow(const TxnID& txnid, const vector<CSCTypesList>
...
+        case WriteEngine::WR_VARBINARY : // treat same as char for now
+        case WriteEngine::WR_CHAR:
+        case WriteEngine::WR_BLOB:
+        case WriteEngine::WR_TEXT:
+            valArray = (char*) calloc(sizeof(char), totalRow * MAX_COLUMN_BOUNDARY);
+            break;
+        default:
+            valArray = calloc(totalRow, colWidth);
+            break;
+    }
```
tntnatbry, line 854: For WriteEngine::WR_TOKEN, is colWidth == sizeof(Token) ?

tntnatbry (changes requested):

dbcon/execplan/windowfunctioncolumn.cpp
davidhall586: Yes

utils/messageqcpp/bytestream.cpp
davidhall586: Yes

utils/rowgroup/rowgroup.h
davidhall586: It could be. I don't think it matters very much. If you like that better, I'll change it. It might save one mv assembly code, depending on the micro code implementation.

utils/windowfunction/windowfunctiontype.cpp
davidhall586: I don't see an earlier comment. Should we resize IDB_pow, or use std::pow()?

utils/windowfunction/windowfunctiontype.cpp
davidhall586: Lol. Of course I didn't think to update the comment.

writeengine/wrapper/writeengine.cpp
davidhall586: No, colwidth is the size specified in the SQL. CHAR(1) will have a colwidth 1. CHAR(28) will have a colwidth 28. Hence the hard coded Constant "8"

dbcon/execplan/windowfunctioncolumn.cpp
davidhall586: No. output precision is not specified. We don't want to limit the output precision of aggregated values. It could lead to all sorts of unintended truncation. The default precision is 0. This is the way it has worked for many years.

utils/regr/moda.h
davidhall586: Here's another implementation of it: TupleJoiner::hasher. This one doesn't handle int128 at all.

utils/messageqcpp/bytestream.cpp
davidhall586: Done

utils/rowgroup/rowgroup.h
davidhall586: I wet ahead and made these const ref, since it will save a couple of pico seconds

utils/windowfunction/windowfunctiontype.cpp
davidhall586: So, I created another constant table for int128 > 18 digits and everybody calls datatypes::getScaleDivisor.

utils/windowfunction/windowfunctiontype.cpp
davidhall586: This code is now obsolete, having been replaced by a call to datatypes::getScaleDivisor()

utils/windowfunction/windowfunctiontype.cpp
davidhall586: Replaced by datatypes::getScaleDivisor()

utils/windowfunction/windowfunctiontype.h
davidhall586: Changed

utils/regr/moda.h
davidhall586: I've created a MCOL to consolidate the hashers into one place. I believe most can be done with the template class I created for moda.

dbcon/execplan/windowfunctioncolumn.cpp
davidhall586: Done

datatypes/mcs_decimal.h
```diff
@@ -73,21 +136,21 @@ constexpr int128_t Decimal128Null = int128_t(0x8000000000000000LL) << 64;
...
-            divisor *= mcs_pow_10[maxPowOf10];
-            divisor *= mcs_pow_10[maxPowOf10];
-            break;
-        case 1:
-            divisor *= mcs_pow_10[maxPowOf10];
-        case 0:
-            divisor *= mcs_pow_10[scale%maxPowOf10];
-        default:
-            break;
+        divisor = mcs_pow_10_128[scale-18];
```
tntnatbry: Shouldn't this be `scale-19`?

utils/rowgroup/rowgroup.h
tntnatbry: @dhall-MariaDB const ref here as well, just as you did in the equals() above?

utils/windowfunction/wf_percentile.cpp
tntnatbry: @dhall-MariaDB ^^

utils/windowfunction/wf_percentile.cpp
tntnatbry: @dhall-MariaDB ^^

utils/windowfunction/wf_sum_avg.cpp
tntnatbry: @dhall-MariaDB ^^ . Now that you have added mcs_pow_10_128. That can be used here.

utils/windowfunction/wf_sum_avg.cpp
tntnatbry: @dhall-MariaDB ^^

utils/windowfunction/wf_udaf.cpp
tntnatbry: @dhall-MariaDB ^^

utils/windowfunction/windowfunctiontype.cpp
tntnatbry: @dhall-MariaDB ^^

utils/windowfunction/windowfunctiontype.h
tntnatbry: I don't see this change.

writeengine/wrapper/writeengine.cpp
tntnatbry: I am a bit confused. I am referring to the following:

```
        case WriteEngine::WR_TOKEN:
            valArray = (Token*) calloc(sizeof(Token), totalRow);
            break;
```

This is now replaced by the default case: `valArray = calloc(totalRow, colWidth);`

Is this correct?

utils/windowfunction/windowfunctiontype.cpp
```diff
@@ -480,7 +480,7 @@ template<typename T>
 void WindowFunctionType::implicit2T(uint64_t i, T& t, int s)
 {
     int ct = fRow.getColType(i);
-    int pw = 0;
+    int64_t divisor = 1;
```
tntnatbry: This should be int128_t?

utils/windowfunction/windowfunctiontype.cpp
```diff
@@ -596,6 +556,15 @@ void WindowFunctionType::implicit2T(uint64_t i, T& t, int s)
             break;
         }
     }
+
+    s -= fRow.getScale(i); // we scale only the difference of scales
```
tntnatbry, line 581: For DOUBLE/UDOUBLE/FLOAT/UFLOAT/LONGDOUBLE, we didn't have `s -= fRow.getScale(i);` in the individual case statements above. Is this still OK for these datatypes?

tntnatbry (changes requested):

utils/rowgroup/rowgroup.h
davidhall586: Didn't get pushed.

utils/windowfunction/wf_sum_avg.cpp
davidhall586: LOL

davidhall586: Reminds me to update the copyright notices

utils/windowfunction/wf_sum_avg.cpp
davidhall586: Yeah. I missed that one

utils/windowfunction/wf_sum_avg.cpp
davidhall586: Same

utils/windowfunction/wf_udaf.cpp
davidhall586: Fixed

utils/windowfunction/windowfunctiontype.cpp
davidhall586: It probably should be const, but since the parent template isn't const ref, this can't be unless we go back and change a bunch. It's probably not worth the effort. const only exists to protect programmers from programmers. The machine don't care.

davidhall586: Added uint128_t stuff. Required to add to Row as well.

utils/windowfunction/windowfunctiontype.h
davidhall586: Huh. I have it. Maybe it made it in this time.

writeengine/wrapper/writeengine.cpp
davidhall586: Sorry. I didn't parse your question. Now that I understand the question, the answer is: I can't definitely say. So I'll add token back into the case.

utils/windowfunction/windowfunctiontype.cpp
davidhall586: It should b T

datatypes/mcs_decimal.h
davidhall586: Yes. And after some thought, I changed msc_pow-10 and mcs_pow_10_128 to signed, as it was in IDB_pow. Because getScaleDivisor is template based, it is most often a signed T. With the way it works, we still get a maximum scale of 38.

utils/windowfunction/windowfunctiontype.cpp
davidhall586: The old code for double:
            if (s == 0)
                t = (T) fRow.getDoubleField(i);
            else
                t = (T) (fRow.getDoubleField(i) * IDB_pow[s]); // s is scale, [0, 18]
The only difference here is that scale is assumed to be positive because, for some reason, we're not taking the diff.
That being said, the only time s is non zero is when getting the default for LEAD|LAG (if not a constant). Currently, the MDB parser doesn't parse the default parameter for LEAD|LAG, so there's no way to test it. This is sleeping code -- unreachable -- until MDB  implements the full standard for Window Functions. Nonetheless, I changed a line in LEAD|LAG that I think looks wrong.

tntnatbry (approved):

PR #1360, opened 2020-08-12
MCOL-3814 add back lower to viewName assignment

davidhall586:


tntnatbry (changes requested):
Do we need to include `<vector>` in ha_mcs.h, just as you did for develop-1.4 PR?

davidhall586:
No. I added <vector> to 1.4 because it was needed for reasons I don't fully understand. It wouldn't compile without it. In 1.5, <vector> is being included somewhere, so it's not needed.

tntnatbry (approved):

PR #1361, opened 2020-08-12
Mcol 3814 1.4

davidhall586:


dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -2448,11 +2448,11 @@ SimpleColumn* buildSimpleColFromDerivedTable(gp_walk_info& gwi, Item_field* ifp)
                     sc->timeZone(gwi.thd->variables.time_zone->get_name()->ptr());
                     if (!viewName.empty())
                     {
-                            sc->viewName(viewName);
+                            sc->viewName(lower(viewName));
```
tntnatbry, line 2451: Do we need the TODO comment here as well, just as you added for the develop PR?

tntnatbry (changes requested):

dbcon/mysql/ha_mcs_execplan.cpp
davidhall586: Probably not. I sincerely doubt we'll backport MCOL-4144 to 1.4

tntnatbry (approved):

PR #1371, opened 2020-08-18
ColumnStore Cache fixes

tntnatbry:


dbcon/mysql/ha_mcs.cpp
```diff
@@ -1223,8 +1223,6 @@ bool ha_mcs::is_crashed() const
     lock.
 */
 
-static my_bool (*original_get_status)(void*, my_bool);
-
 my_bool get_status_and_flush_cache(void *param,
                                    my_bool concurrent_insert);
```
davidhall586, line 1158: my_bool get_status_and_flush_cache(void *param,
                                   my_bool concurrent_insert);
Why is this declaration here? Declarations of this sort usually exist in a header if used externally. They may be included like this if the function is used before the definition (C++ is a single pass compiler).  Neither appear  to be the case here.

davidhall586 (approved):

dbcon/mysql/ha_mcs.cpp
tntnatbry: Good catch. I don't see a need for this function declaration here. I will have it removed in a follow up commit.

PR #1387, opened 2020-08-27
Dataconvert::decimalToString badly fails w/o 20th member of mcs_pow_1…

drrtuy:
…0 so I returned it

WF::percentile runtime threw an exception b/c of wrong DT deduced from its argument

Replaced literals with constants

Tought WF_sum_avg::checkSumLimit to use refs instead of values

### DIFF
--- a/utils/regr/moda.cpp
+++ b/utils/regr/moda.cpp
@@ -21,6 +21,7 @@
 #include "moda.h"
 #include "bytestream.h"
 #include "objectreader.h"
+#include "columnwidth.h"
 
 using namespace mcsv1sdk;
 
@@ -186,7 +187,7 @@ mcsv1_UDAF::ReturnCode moda::init(mcsv1Context* context,
         {
             context->setColWidth(8);
         }
-        else
+        else if (utils::widthByPrecision(colTypes[0].precision))
         {
             context->setColWidth(16);
         }
--- a/utils/windowfunction/wf_nth_value.cpp
+++ b/utils/windowfunction/wf_nth_value.cpp
@@ -84,31 +84,19 @@ boost::shared_ptr<WindowFunctionType> WF_nth_value<T>::makeFunction(int id, cons
         }
 
         case CalpontSystemCatalog::DECIMAL:
+        case CalpontSystemCatalog::UDECIMAL:
         {
-            if (wc->functionParms()[0]->resultType().colWidth < 16)
+            if (wc->functionParms()[0]->resultType().colWidth < datatypes::MAXDECIMALWIDTH)
             {
                 func.reset(new WF_nth_value<int64_t>(id, name));
             }
-            else
+            else if (wc->functionParms()[0]->resultType().colWidth == datatypes::MAXDECIMALWIDTH)
             {
                 func.reset(new WF_nth_value<int128_t>(id, name));
             }
             break;
         }
         
-        case CalpontSystemCatalog::UDECIMAL:
-        {
-            if (wc->functionParms()[0]->resultType().colWidth < 16)
-            {
-                func.reset(new WF_nth_value<uint64_t>(id, name));
-            }
-            else
-            {
-                func.reset(new WF_nth_value<uint128_t>(id, name));
-            }
-            break;
-        }
-
         case CalpontSystemCatalog::DOUBLE:
         case CalpontSystemCatalog::UDOUBLE:
         {
--- a/utils/windowfunction/wf_percentile.cpp
+++ b/utils/windowfunction/wf_percentile.cpp
@@ -46,6 +46,8 @@ using namespace ordering;
 #include "constantcolumn.h"
 using namespace execplan;
 
+#include "mcs_decimal.h"
+
 #include "windowfunctionstep.h"
 using namespace joblist;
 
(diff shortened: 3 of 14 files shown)


davidhall586 (approved):

utils/regr/moda.cpp
```diff
@@ -186,7 +187,7 @@ mcsv1_UDAF::ReturnCode moda::init(mcsv1Context* context,
         {
             context->setColWidth(8);
         }
-        else
+        else if (utils::widthByPrecision(colTypes[0].precision))
```
tntnatbry, line 190: Shouldn't the returned value here be compared against datatypes::MAXDECIMALWIDTH?

utils/windowfunction/wf_nth_value.cpp
```diff
@@ -84,31 +84,19 @@ boost::shared_ptr<WindowFunctionType> WF_nth_value<T>::makeFunction(int id, cons
...
-            {
-                func.reset(new WF_nth_value<uint64_t>(id, name));
-            }
-            else
-            {
-                func.reset(new WF_nth_value<uint128_t>(id, name));
-            }
-            break;
-        }
-
```
tntnatbry, line 111: I am not sure if removing the separate case for UDECIMAL is correct now. As earlier we had `func.reset(new WF_nth_value<uint64_t>(id, name));` for UDECIMAL, now we have `func.reset(new WF_nth_value<int64_t>(id, name));`. Similarly for the wide decimal case.

@dhall-MariaDB can you confirm this change is correct?

utils/windowfunction/wf_percentile.cpp
```diff
@@ -89,31 +91,19 @@ boost::shared_ptr<WindowFunctionType> WF_percentile<T>::makeFunction(int id, con
...
-                {
-                    func.reset(new WF_percentile<uint64_t>(id, name));
-                }
-                else
-                {
-                    func.reset(new WF_percentile<uint128_t>(id, name));
-                }
-                break;
-            }
-
```
tntnatbry, line 116: Please see my previous comment on verifying the correctness of removing this case block for UDECIMAL.

@dhall-MariaDB FYI.

utils/windowfunction/wf_percentile.cpp
```diff
@@ -89,31 +91,19 @@ boost::shared_ptr<WindowFunctionType> WF_percentile<T>::makeFunction(int id, con
             }
 
             case CalpontSystemCatalog::DECIMAL:
+            case CalpontSystemCatalog::UDECIMAL:
             {
-                if (wc->functionParms()[0]->resultType().colWidth < 16)
+                if (wc->resultType().colWidth < datatypes::MAXDECIMALWIDTH)
```
tntnatbry, line 96: For a lot of other window functions, we are still using `wc->functionParms()[0]->resultType().colWidth`, and not `wc->resultType().colWidth`.

Is this deviation intentional for wf_percentile.cpp?

utils/windowfunction/wf_stats.cpp
```diff
@@ -82,30 +82,18 @@ boost::shared_ptr<WindowFunctionType> WF_stats<T>::makeFunction(int id, const st
...
-            if (wc->functionParms()[0]->resultType().colWidth < 16)
-            {
-                func.reset(new WF_stats<uint64_t>(id, name));
-            }
-            else
-            {
-                func.reset(new WF_stats<uint128_t>(id, name));
-            }
-            break;
-        }
```
tntnatbry, line 108: Please see my previous comment on verifying the correctness of removing this case block for UDECIMAL.

@dhall-MariaDB FYI.

utils/windowfunction/wf_sum_avg.cpp
```diff
@@ -52,36 +52,50 @@ using namespace joblist;
...
-    if (((sum >= 0) && ((fMax128 - sum) < val)) ||
-        ((sum <  0) && ((fMin128 - sum) > val)))
-    {
-        string errStr = "SUM(int128_t)";
-        errStr = IDBErrorInfo::instance()->errorMsg(ERR_WF_OVERFLOW, errStr);
-        cerr << errStr << endl;
-        throw IDBExcept(errStr, ERR_WF_OVERFLOW);
-    }
-}
+inline void WF_sum_avg<T_IN,T_OUT>::checkSumLimit(T_IN& val, T_OUT& sum)
```
tntnatbry, line 57: Shouldn't the function parameters here be const ref?

utils/windowfunction/wf_sum_avg.cpp
```diff
@@ -188,31 +167,19 @@ boost::shared_ptr<WindowFunctionType> WF_sum_avg<T_IN, T_OUT>::makeFunction(int
...
-            {
-                func.reset(new WF_sum_avg<uint64_t, uint128_t>(id, name));
-            }
-            else
-            {
-                func.reset(new WF_sum_avg<uint128_t, uint128_t>(id, name));
-            }
-            break;
-        }
-
```
tntnatbry, line 215: Please see my previous comment on verifying the correctness of removing this case block for UDECIMAL.

@dhall-MariaDB FYI.

utils/windowfunction/wf_sum_avg.h
```diff
@@ -58,19 +55,17 @@ class WF_sum_avg : public WindowFunctionType
...
+
+    void checkSumLimit(T_IN& val, T_OUT& sum);
+
     int128_t calculateAvg(int128_t sum, uint64_t count, int scale);
-    uint128_t calculateAvg(uint128_t sum, uint64_t count, int scale);
     long double calculateAvg(long double sum, uint64_t count, int scale);
 };
 
+template<>
+void WF_sum_avg<long double,long double>::checkSumLimit(long double& val, long double& sum);
```
tntnatbry, line 66: Did you mean an explicit template instantiation here, instead of template specialization? Also since this is a header file, I am not sure if this is needed in the first place.

tntnatbry (changes requested):

tntnatbry:
@drrtuy @dhall-MariaDB 

I have added some review comments, even though I know the PR has already been merged.

utils/regr/moda.cpp
drrtuy: It should be a different method.

utils/windowfunction/wf_percentile.cpp
drrtuy: Yes. Percentile has a decimal literal as an argument.

utils/windowfunction/wf_sum_avg.cpp
drrtuy: It should if it doesn't call for an explicit casts.

utils/windowfunction/wf_nth_value.cpp
drrtuy: Why exactly are you not sure?
My points are: 
- IDB_Decimal uses int64 to store the value and it is enough to store decimal/udecimal(18)
- All WFs changed(except avg) can't produce the value bigger than column can have. avg doesn't use the signedness
- If MCS runs WF on expression/aggregate there are overflow checks in the relevant code.

tntnatbry: I didn't have the full context on the changes here, thanks for this clarification.

PR #1390, opened 2020-08-28
::writeRow now treats WR_BINARY as int128 for 16 bytes DT only

drrtuy:
WF avg uses const & as arguments types

This should be merged only by the last reviewer.

### DIFF
--- a/writeengine/wrapper/writeengine.cpp
+++ b/writeengine/wrapper/writeengine.cpp
@@ -837,84 +837,22 @@ int WriteEngineWrapper::deleteRow(const TxnID& txnid, const vector<CSCTypesList>
 inline void allocateValArray(void*& valArray, ColTupleList::size_type totalRow,
                              ColType colType, int colWidth)
 {
+    // MCS allocates 8 bytes even for CHARs smaller then 8 bytes.
     switch (colType)
     {
         case WriteEngine::WR_VARBINARY : // treat same as char for now
         case WriteEngine::WR_CHAR:
         case WriteEngine::WR_BLOB:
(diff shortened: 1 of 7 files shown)


writeengine/wrapper/writeengine.cpp
```diff
@@ -837,84 +837,23 @@ int WriteEngineWrapper::deleteRow(const TxnID& txnid, const vector<CSCTypesList>
 inline void allocateValArray(void*& valArray, ColTupleList::size_type totalRow,
                              ColType colType, int colWidth)
 {
+    // All three cases effectively uses colWidth however left the switch
+    // for readability and potential Token size changes.
```
tntnatbry: This statement is not necessarily true. For a CHAR field, `colWidth` is not always equal to `MAX_COLUMN_BOUNDARY`. If we replace the entire switch block with a single line `valArray = calloc(totalRow, colWidth);` , we get a crash in test000.

tntnatbry (changes requested):

writeengine/wrapper/writeengine.cpp
drrtuy: Really. Could you tell me where exactly b/c I had no crashes running test000 with deb builds prior the David's WF patch.

tntnatbry: The crash was in mysql/scripts/create_tpch_4byte.sql while running test000 when we just used `calloc(totalRow, colWidth)`. Specifically `create table orders` statement.

tntnatbry (approved):

PR #1391, opened 2020-08-28
MCOL-4099 allow insert on first or last char

davidhall586:
Seems the previous fix for handling first character insert broke last character insert. I can't seem to do simple arithmetic.

### DIFF
--- a/utils/funcexp/func_insert.cpp
+++ b/utils/funcexp/func_insert.cpp
@@ -68,6 +68,7 @@ std::string Func_insert::getStrVal(rowgroup::Row& row,
     start = fp[1]->data()->getIntVal(row, isNull);
     if (isNull)
         return "";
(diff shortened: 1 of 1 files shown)


utils/funcexp/func_insert.cpp
```diff
@@ -83,14 +84,12 @@ std::string Func_insert::getStrVal(rowgroup::Row& row,
     int64_t strLen = cs->numchars(pos, end);
     
     // Return the original string if start isn't within the string.
-    if ((start < 1) || start >= strLen)
+    if ((start < 0) || start > strLen)
```
tntnatbry: Since start is now 0 based here, the 2nd part of the || should be `start >= strLen`

tntnatbry (changes requested):

tntnatbry (approved):

PR #1404, opened 2020-08-31
MCOL-4227 Related. Shared storage

jmrojas2332:
Prevent clusters using shared storage from initializing columnstore more than once


pleblanc1976 (approved):

pleblanc1976 (changes requested):
it looks OK, but drone failed to build.  Can merge once drone is happy.

pleblanc1976 (approved):
The error was caused by the previous PR

PR #1405, opened 2020-08-31
MCOL-4227 Related. Shared storage

jmrojas2332:
Prevent clusters using shared storage from initializing columnstore more than once


pleblanc1976:
For some reason the drone job for this has been blocked in the u16 build stage for 4.5 hours.  It passed on every other platform.

pleblanc1976 (approved):

PR #1408, opened 2020-08-31
MCOL-4012 Revert mcs-loadbrm related changes

jmrojas2332:
Note: Reverting for now. Cannot change files on server side now. Will reintroduce change in this reversion once develop opens again


jmrojas2332:
Working on PR to include adding back reversion of a previous PR

PR #1418, opened 2020-09-02
Several fixes to migrate this branch from community to enterprise.

pleblanc1976:


pleblanc1976:
Pushing this in now to make it into the nightly build.  Will revert if there are unexpected failures in the PR build.

PR #1451, opened 2020-09-14
MCOL-4172 Add support for wide-DECIMAL into statistical aggregate and…

drrtuy:
… regr_* UDAF functions

The patch fixes wrong results returned when multiple UDAF exist in projection


utils/rowgroup/rowaggregation.cpp
```diff
@@ -1032,15 +1039,15 @@ void RowAggregation::initMapData(const Row& rowIn)
             case execplan::CalpontSystemCatalog::DECIMAL:
             case execplan::CalpontSystemCatalog::UDECIMAL:
             {
-                if (LIKELY(fRow.getColumnWidth(colIn) == datatypes::MAXDECIMALWIDTH))
+                if (LIKELY(rowIn.getColumnWidth(colIn) == datatypes::MAXDECIMALWIDTH))
```
drrtuy, line 1044: This is to fix a bug when type detection was based on output DT and not on input DT.

drrtuy:
The main idea of the patch is to replace RowAggregation::fRGContext with a collection RowAggregation::fRGContextColl to set up UDAF contexts once per RowGroup.
The patch also contains a number of fixes to enable multi-arg UDAFs in different combinations in projection. JFYI there were issues having regr_avgx(c1,c2), regr_slope(c1,c2), regr_avgy(c1,c2) in projection previously.
It also fixes aggregations on constant columns.

datatypes/mcs_decimal.cpp
```diff
@@ -208,6 +208,21 @@ namespace datatypes
         return toString(const_cast<execplan::IDB_Decimal&>(value));
     }
 
+    // Compare perf with f(string&, ColTypeAlias&, int128_t&)
+    int128_t Decimal::int128FromString(const std::string& value,
```
drrtuy, line 212: This is a questionable interface approach since I return 16 byte from the function but I believe this should be fine from the performance perspective.

dbcon/joblist/tupleaggregatestep.cpp
```diff
@@ -1537,7 +1534,7 @@ void TupleAggregateStep::prep1PhaseAggregate(
     }
 
     // there is avg(k), but no count(k) in the select list
-    uint64_t lastCol = returnedColVec.size();
+    uint64_t lastCol = outIdx;
```
drrtuy, line 1537: This fixes the bug exists since MCOL-1201 and fires back for ExeMgr-side UDAF processing only.
I didn't put it into a separate issue b/c nobody complains since that time.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -5044,6 +5055,17 @@ ReturnedColumn* buildAggregateColumn(Item* item, gp_walk_info& gwi)
                 ac->resultType(parm->resultType());
             }
         }
+        else if (bIsConst && hasDecimalConst && isAvg)
```
drrtuy, line 5058: aggregates with literal arguments will now have a proper scale and precision to process it later in RowAggregationUM::doNotNullConstantAggregate

utils/rowgroup/rowaggregation.cpp
```diff
@@ -2346,18 +2386,20 @@ void RowAggregation::doUDAF(const Row& rowIn, int64_t colIn, int64_t colOut,
...
-    rc = fRGContext.getFunction()->nextValue(&fRGContext, valsIn);
-    fRGContext.setUserData(NULL);
+    rc = fRGContextColl[origFuncColsIdx].getFunction()->nextValue(&fRGContextColl[origFuncColsIdx],
+                                                              valsIn);
+    fRGContextColl[origFuncColsIdx].setUserData(NULL);
 
     if (rc == mcsv1sdk::mcsv1_UDAF::ERROR)
     {
-        RowUDAFFunctionCol* rowUDAF = dynamic_cast<RowUDAFFunctionCol*>(fFunctionCols[funcColsIdx].get());
+        RowUDAFFunctionCol* rowUDAF = dynamic_cast<RowUDAFFunctionCol*>(fFunctionCols[origFuncColsIdx].get());
```
drrtuy, line 2397: @dhall-MariaDB Does it sound for you?

dbcon/mysql/ha_mcs_partition.cpp
```diff
@@ -921,18 +921,8 @@ void partitionByValue_common(UDF_ARGS* args,								// input
...
-                                (static_cast<uint128_t>(partInfo.bigMin) < static_cast<uint128_t>(mapit->second.bigMin) ? partInfo.bigMin : mapit->second.bigMin);
-                            mapit->second.bigMax =
-                                (static_cast<uint128_t>(partInfo.bigMax) > static_cast<uint128_t>(mapit->second.bigMax) ? partInfo.bigMax : mapit->second.bigMax);
-                        }
-                        else
-                        {
-                            mapit->second.bigMin = (partInfo.bigMin < mapit->second.bigMin ? partInfo.bigMin : mapit->second.bigMin);
-                            mapit->second.bigMax = (partInfo.bigMax > mapit->second.bigMax ? partInfo.bigMax : mapit->second.bigMax);
-                        }
+                        mapit->second.int128Min = (partInfo.int128Min < mapit->second.int128Min ? partInfo.int128Min : mapit->second.int128Min);
```
drrtuy, line 924: int128_t range is enough to store both SIGNED and UNSIGNED wide decimals.

davidhall586 (approved):

utils/rowgroup/rowaggregation.cpp
```diff
@@ -720,31 +721,32 @@ void RowAggregation::setJoinRowGroups(vector<RowGroup>* pSmallSideRG, RowGroup*
...
 // thread safe.
 //------------------------------------------------------------------------------
-void RowAggregation::resetUDAF(RowUDAFFunctionCol* rowUDAF)
+void RowAggregation::resetUDAF(RowUDAFFunctionCol* rowUDAF, uint64_t funcColsIdx)
 {
     // RowAggregation and it's functions need to be re-entrant which means
     // each instance (thread) needs its own copy of the context object.
     // Note: operator=() doesn't copy userData.
-    fRGContext = rowUDAF->fUDAFContext;
+    fRGContextColl[funcColsIdx] = rowUDAF->fUDAFContext;
```
tntnatbry, line 729: @drrtuy This line is causing a PrimProc crash when running the following query:

```SQL
MariaDB [tpch1]> select o_orderstatus, moda(o_custkey) from orders group by o_orderstatus order by o_orderstatus;
```

I don't have a full context on the problem yet, but in RowAggregation::initialize(), the code path does not reach the else block containing `fRGContextColl.resize(fFunctionCols.size());`.

PR #1459, opened 2020-09-21
MCOL-4144-dev Enable lower_case_table_names

davidhall586:
Create tables and schemas with lower case name only if the flag is set.
During operations, convert to lowercase in plugin. Byt the time a query gets to ExeMgr, DDLProc etc., everything must be lower case if the flag is set, and undisturbed if not.


dbcon/execplan/calpontselectexecutionplan.h
```diff
@@ -539,16 +546,23 @@ class CalpontSelectExecutionPlan : public CalpontExecutionPlan
         return fSubType;
     }
 
-    void derivedTbAlias(const std::string derivedTbAlias)
+    void derivedTbAlias(const std::string derivedTbAlias, int lower_case_table_names=0)
```
tntnatbry, line 549: Is the default assignment here intentional?

dbcon/execplan/calpontsystemcatalog.cpp
```diff
@@ -1583,16 +1601,13 @@ const CalpontSystemCatalog::TableColName CalpontSystemCatalog::dictColName(const
 
     return tableColName;
 }
-uint64_t CalpontSystemCatalog::nextAutoIncrValue ( TableName aTableName)
+uint64_t CalpontSystemCatalog::nextAutoIncrValue (TableName aTableName, int lower_case_table_names)
 {
-    transform( aTableName.schema.begin(), aTableName.schema.end(), aTableName.schema.begin(), to_lower() );
-    transform( aTableName.table.begin(), aTableName.table.end(), aTableName.table.begin(), to_lower() );
-
```
tntnatbry, line 1590: I think you missed the `if (lower_case_table_names)` block here.

dbcon/execplan/calpontsystemcatalog.cpp
```diff
@@ -1691,16 +1706,13 @@ uint64_t CalpontSystemCatalog::nextAutoIncrValue ( TableName aTableName)
 
 }
 
-int32_t CalpontSystemCatalog::autoColumOid ( TableName aTableName)
+int32_t CalpontSystemCatalog::autoColumOid (TableName aTableName, int lower_case_table_names)
 {
-    transform( aTableName.schema.begin(), aTableName.schema.end(), aTableName.schema.begin(), to_lower() );
-    transform( aTableName.table.begin(), aTableName.table.end(), aTableName.table.begin(), to_lower() );
-
```
tntnatbry, line 1698: I think you missed the `if (lower_case_table_names)` block here.

dbcon/execplan/calpontsystemcatalog.cpp
```diff
@@ -2992,6 +2990,12 @@ const CalpontSystemCatalog::RIDList CalpontSystemCatalog::columnRIDs(const Table
         checkSysCatVer();
     }
 
+    if (lower_case_table_names)
+    {
+        boost::algorithm::to_lower(aTableName.schema);
+        boost::algorithm::to_lower(aTableName.table);
+    }
```
tntnatbry: This block should be moved earlier in the function, right after `TableName aTableName(tableName);`

dbcon/mysql/ha_mcs.cpp
```diff
@@ -63,14 +63,15 @@ static HASH mcs_open_tables;
...
+#if 0
 #ifdef DEBUG_ENTER
 #undef DEBUG_ENTER
 #endif
 #ifdef DEBUG_RETURN
 #undef DEBUG_ENTER
 #endif
 #define DEBUG_RETURN return
-
+#endif
```
tntnatbry: Can you explain why these undefs are disabled?

dbcon/mysql/ha_mcs_client_udfs.cpp
```diff
@@ -487,6 +489,11 @@ extern "C"
                 return result;
             }
         }
+        if (lower_case_table_names)
```
tntnatbry, line 492: Curious, where is `lower_case_table_names` defined?

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -2387,8 +2377,8 @@ SimpleColumn* buildSimpleColFromDerivedTable(gp_walk_info& gwi, Item_field* ifp)
                     // @bug 3003. Keep column alias if it has.
                     sc->alias(ifp->is_autogenerated_name() ? tcn.column : ifp->name.str);
 
-                    sc->tableAlias(lower(gwi.tbList[i].alias));
-                    sc->viewName(lower(viewName));
+                    sc->tableAlias(gwi.tbList[i].alias);
```
tntnatbry, line 2380: This should also have the `lower_case_table_names` as the second argument.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -2446,17 +2436,15 @@ SimpleColumn* buildSimpleColFromDerivedTable(gp_walk_info& gwi, Item_field* ifp)
                     sc->alias(ifp->is_autogenerated_name() ? cols[j]->alias() : ifp->name.str);
                     sc->tableName(csep->derivedTbAlias());
                     sc->colPosition(j);
-                    string tableAlias(csep->derivedTbAlias());
-                    sc->tableAlias(lower(tableAlias));
+                    sc->tableAlias(csep->derivedTbAlias());
```
tntnatbry, line 2439: This should also have the `lower_case_table_names` as the second argument.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -5231,7 +5239,7 @@ void gp_walk(const Item* item, void* arg)
                     break;
 
                 string aliasTableName(scp->tableAlias());
-                scp->tableAlias(lower(aliasTableName));
+                scp->tableAlias(aliasTableName);
```
tntnatbry, line 5242: This should have `lower_case_table_names` as the second argument.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -7968,7 +7985,7 @@ int getSelectPlan(gp_walk_info& gwi, SELECT_LEX& select_lex,
             sc1->columnName(sc->columnName());
             sc1->tableName(sc->tableName());
             sc1->tableAlias(sc->tableAlias());
-            sc1->viewName(lower(sc->viewName()));
+            sc1->viewName(sc->viewName());
```
tntnatbry, line 7988: This should have `lower_case_table_names` as the second argument.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -9123,7 +9153,7 @@ int getGroupPlan(gp_walk_info& gwi, SELECT_LEX& select_lex, SCSEP& csep, cal_gro
         funcFieldVec[i]->print(&str, QT_ORDINARY);
         sc->alias(string(str.c_ptr()));
         //sc->tableAlias(funcFieldVec[i]->table_name);
-        sc->tableAlias(sc->tableAlias());
+        sc->tableAlias(sc->alias());
```
tntnatbry, line 9156: This should have `lower_case_table_names` as the second argument.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -9920,7 +9950,7 @@ int getGroupPlan(gp_walk_info& gwi, SELECT_LEX& select_lex, SCSEP& csep, cal_gro
             sc1->columnName(sc->columnName());
             sc1->tableName(sc->tableName());
             sc1->tableAlias(sc->tableAlias());
-            sc1->viewName(lower(sc->viewName()));
+            sc1->viewName(sc->viewName());
```
tntnatbry, line 9953: This should have `lower_case_table_names` as the second argument.

dbcon/mysql/ha_mcs_impl.cpp
```diff
@@ -965,8 +965,6 @@ vector<string> getOnUpdateTimestampColumns(string& schema, string& tableName, in
     boost::shared_ptr<CalpontSystemCatalog> csc = CalpontSystemCatalog::makeCalpontSystemCatalog(sessionID);
     csc->identity(execplan::CalpontSystemCatalog::FE);
     CalpontSystemCatalog::TableName aTableName;
-    boost::algorithm::to_lower(schema);
-    boost::algorithm::to_lower(tableName);
```
tntnatbry, line 969: I think we need the `if (lower_case_table_names)` block here.

dbcon/mysql/ha_mcs_impl.cpp
```diff
@@ -1916,7 +1949,7 @@ uint32_t doUpdateDelete(THD* thd, gp_walk_info& gwi, const std::vector<COND*>& c
     CalpontSystemCatalog::ROPair roPair;
     try
     {
-        roPair = csc->tableRID( aTableName );
+        roPair = csc->tableRID(aTableName);
```
tntnatbry, line 1952: This should have `lower_case_table_names` as the second argument.

dbcon/mysql/ha_mcs_impl_if.h
```diff
@@ -335,6 +341,7 @@ struct cal_connection_info
     // MCOL-1101 remove compilation unit variable rmParms
     std::vector <execplan::RMParam> rmParms;
     long long affectedRows;
+    int lock_type;
```
tntnatbry: Changes in this file don't belong to MCOL-4144, probably you need to rebase your branch on top of latest develop.

dbcon/mysql/ha_view.cpp
```diff
@@ -108,8 +115,8 @@ void View::transform()
             else if (table_ptr->view)
             {
                 // for nested view, the view name is vout.vin... format
-                CalpontSystemCatalog::TableAliasName tn = make_aliasview(table_ptr->db.str, table_ptr->table_name.str, table_ptr->alias.str, viewName);
-                gwi.viewName = make_aliastable(table_ptr->db.str, table_ptr->table_name.str, viewName);
+                CalpontSystemCatalog::TableAliasName tn = make_aliasview(table_ptr->db.str, table_ptr->table_name.str, table_ptr->alias.str, viewName, lower_case_table_names);
```
tntnatbry: true before `lower_case_table_names` argument.

dbcon/mysql/idb_mysql.h
```diff
@@ -28,6 +28,9 @@ template <class T> bool isnan(T);
 #endif
 #endif
 
+#define INFINIDB_DEBUG
+#define DEBUG_WALK_COND
+
```
tntnatbry, line 33: I think these debugging flags are accidentally checked in.

dbcon/mysql/is_columnstore_tables.cpp
```diff
@@ -142,7 +142,7 @@ static int is_columnstore_tables_fill(THD* thd, TABLE_LIST* tables, COND* cond)
         if (tb_info.tablewithautoincr)
         {
             table->field[5]->set_notnull();
-            table->field[5]->store(systemCatalogPtr->nextAutoIncrValue((*it).second));
+            table->field[5]->store(systemCatalogPtr->nextAutoIncrValue((*it).second), lower_case_table_names);
```
tntnatbry: `lower_case_table_names` should be passed as an argument to nextAutoIncrValue(). I am not sure how this is compiling.

writeengine/server/we_ddlcommandproc.cpp
```diff
@@ -916,13 +912,11 @@ uint8_t WE_DDLCommandProc::writeSyscolumn(ByteStream& bs, std::string& err)
 
             if (SCHEMA_COL == column.tableColName.column)
             {
-                boost::to_lower(schema);
                 colTuple.data = schema;
                 tmpStr = schema;
             }
             else if (TABLENAME_COL == column.tableColName.column)
             {
-                boost::to_lower(tablename);
```
tntnatbry, line 925: Do we not need the `if (lower_case_table_names)` blocks in this file?

tntnatbry (changes requested):

dbcon/mysql/columnstore.cnf
```diff
@@ -1,7 +1,6 @@
 [mysqld]
 plugin-load-add=ha_columnstore.so
 plugin-maturity=beta
-lower_case_table_names=1
```
tntnatbry, line 4: I believe we need to leave this setting here.

drrtuy: We can't have this option enabled according to the server team.

dbcon/execplan/calpontselectexecutionplan.h
davidhall586: Yes. There are place where it's called with one parameter -- Where we know the first parameter is already set. In general, lower_case_table_names is defaulted where it's a new parameter. In some cases, we don't need it, in others, such as when the function is called in ExeMgr, DDLProc, etc, the global lower_case_table_names isn't available and the to_lower (if needed) should have been done before the query was sent to the process.

dbcon/execplan/calpontsystemcatalog.cpp
davidhall586: No, I didn't miss it. This function calls tableInfo and passes the flag on. I thought this would be enough, but maybe it isn't. I all the if (lower_case_table_names) here since aTableName is used later in the function.

davidhall586: Interestingly, this function is only called once with the flag set, and that is while processing system catalog, which is always lower case anyway. While the fix is made, it will have no functional change to the current behavior.

dbcon/execplan/calpontsystemcatalog.cpp
davidhall586: See above

dbcon/mysql/ha_mcs.cpp
davidhall586: Failure to revert a debug change, though I'm not sure why they're there.

dbcon/execplan/calpontsystemcatalog.cpp
tntnatbry: I am seeing the call to this func with the 2nd argument from is_columnstore_tables_fill in . Is this what you are referring to?

dbcon/execplan/calpontsystemcatalog.cpp
tntnatbry: I understand we are not calling autoColumnOid with the 2nd argument. But for completeness, we should add the if block here since aTableName is used later on in this func.

dbcon/execplan/calpontsystemcatalog.cpp
```diff
@@ -2709,9 +2705,7 @@ const CalpontSystemCatalog::RID CalpontSystemCatalog::constraintColRID(const Tab
     aTableColName.schema = tableColName.schema;
     aTableColName.table = tableColName.table;
     aTableColName.column = tableColName.column;
-    transform( aTableColName.schema.begin(), aTableColName.schema.end(), aTableColName.schema.begin(), to_lower() );
-    transform( aTableColName.table.begin(), aTableColName.table.end(), aTableColName.table.begin(), to_lower() );
-    transform( aTableColName.column.begin(), aTableColName.column.end(), aTableColName.column.begin(), to_lower() );
+    boost::algorithm::to_lower(aTableColName.column);
```
tntnatbry, line 2720: I know this code above is commented out, so the deletions above will have no effect. But in case this is activated in future, we will lose the changes where we might need to apply the to_lower conversions based on the value of `lower_case_table_names`

dbcon/mysql/ha_view.cpp
davidhall586: Most excellent spot. I can't believe I missed one of those. I did triple check for just this.

dbcon/mysql/idb_mysql.h
davidhall586: Yep

dbcon/mysql/is_columnstore_tables.cpp
davidhall586: It's defaulted. However, I added it in because, while I expect tablename to always be lower case here because it's a system catalog name, and we always create those lower case, there's no guarantee of that in the future.

writeengine/server/we_ddlcommandproc.cpp
davidhall586: No -- because the global lower_case_table_names, which we pass to all those types of functions, is not available outside the plugin. WriteEngineServer doesn't have it. Changes in the plugin ensure that the proper case is set before being sent to the other processes

dbcon/execplan/calpontsystemcatalog.cpp
davidhall586: Absolutely true. There are tons of commented out code having to do with indexes and constraints. I started down the path of changing them (removing the transforms), but gave up because of time and inability to test commented code, so I never added the if (lower_case_table_names) stuff in. If we ever enable these blocks, they will most likely be re-written anyway, so we just need to be aware.

dbcon/execplan/calpontsystemcatalog.cpp
davidhall586: It's fine where it is because all that happens in between are checks not relying on the case of aTableName. However, in the tiny chance that someday we change the case of calpontsystemcatalog, I will move it.

dbcon/execplan/calpontsystemcatalog.cpp
davidhall586: Yeah.

dbcon/execplan/calpontsystemcatalog.cpp
davidhall586: I did

dbcon/mysql/ha_mcs_client_udfs.cpp
davidhall586: lower_case_table_names is a global in Server space defined in mysqld.h:
extern MYSQL_PLUGIN_IMPORT uint lower_case_table_names;

dbcon/execplan/calpontsystemcatalog.cpp
```diff
@@ -1583,16 +1601,20 @@ const CalpontSystemCatalog::TableColName CalpontSystemCatalog::dictColName(const
...
-    transform( aTableName.table.begin(), aTableName.table.end(), aTableName.table.begin(), to_lower() );
-
     TableInfo tbInfo;
 
+    if (lower_case_table_names)
+    {
+        boost::algorithm::to_lower(aTableColName.schema);
+        boost::algorithm::to_lower(aTableColName.table);
+    }
+    boost::algorithm::to_lower(aTableColName.column);
```
tntnatbry: `aTableColName`?? Shouldn't this be `aTableName`?

dbcon/execplan/calpontsystemcatalog.cpp
```diff
@@ -1691,13 +1713,17 @@ uint64_t CalpontSystemCatalog::nextAutoIncrValue ( TableName aTableName)
...
+
```
tntnatbry, line 1724: `aTableColName`?? Shouldn't this be `aTableName`?

dbcon/execplan/calpontsystemcatalog.cpp
tntnatbry: I agree that we will need to be aware if and when this code is rewritten. But for us to be reminded of the changes needed in future, shouldn't we leave this code as is?

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -2446,17 +2436,15 @@ SimpleColumn* buildSimpleColFromDerivedTable(gp_walk_info& gwi, Item_field* ifp)
...
                     if (!viewName.empty())
                     {
-                        // TODO: Remove lower when MCOL-4144 fixed    
-                        sc->viewName(lower(viewName));
+                        sc->viewName(viewName, lower_case_table_names);
                     }
                     else
                     {
-                            sc->viewName(lower(csep->derivedTbView()));
+                        sc->viewName(csep->derivedTbView());
```
tntnatbry, line 2447: This should also have the `lower_case_table_names` as the second argument.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -2570,15 +2558,14 @@ void collectAllCols(gp_walk_info& gwi, Item_field* ifp)
                 SimpleColumn* sc = new SimpleColumn();
                 sc->columnName(cols[j]->alias());
                 sc->colPosition(j);
-                string tableAlias(csep->derivedTbAlias());
-                sc->tableAlias(lower(tableAlias));
-                sc->viewName(lower(gwi.tbList[i].view));
+                sc->tableAlias(csep->derivedTbAlias());
+                sc->viewName(gwi.tbList[i].view);
```
tntnatbry, line 2562: These 2 calls should have the `lower_case_table_names` as the second argument.

dbcon/mysql/ha_mcs_impl.cpp
```diff
@@ -1304,11 +1302,18 @@ uint32_t doUpdateDelete(THD* thd, gp_walk_info& gwi, const std::vector<COND*>& c
...
+                boost::algorithm::to_lower(tableName);
+                boost::algorithm::to_lower(tmpTableName);
+            }
+
             if (strcasecmp(tableName.c_str(), "") == 0)
             {
                 tableName = tmpTableName;
             }
-            else if (strcasecmp(tableName.c_str(), tmpTableName.c_str()) != 0)
+            else if (strcmp(tableName.c_str(), tmpTableName.c_str()) != 0)
```
tntnatbry, line 1316: Shouldn't the sensitivity of the string comparison be based on `lower_case_table_names`? That is:
```
if (lower_case_table_names) strcasecmp()
else strcmp()
```

dbcon/mysql/ha_mcs_impl.cpp
```diff
@@ -1821,7 +1854,7 @@ uint32_t doUpdateDelete(THD* thd, gp_walk_info& gwi, const std::vector<COND*>& c
 
             try
             {
-                colrids = csc->columnRIDs(deleteTableName);
+                colrids = csc->columnRIDs(deleteTableName, lower_case_table_names);
```
tntnatbry: We are missing the 2nd argument here, which should be false.

dbcon/mysql/ha_mcs_impl.cpp
```diff
@@ -3185,7 +3218,7 @@ void ha_mcs_impl_start_bulk_insert(ha_rows rows, TABLE* table, bool is_cache_ins
 
             try
             {
-                colrids = csc->columnRIDs(tableName);
+                colrids = csc->columnRIDs(tableName, lower_case_table_names);
```
tntnatbry: This should have `lower_case_table_names` as the second argument.

dbcon/mysql/idb_mysql.h
```diff
@@ -28,6 +28,9 @@ template <class T> bool isnan(T);
 #endif
 #endif
 
+//#define INFINIDB_DEBUG
+//d#define DEBUG_WALK_COND
```
tntnatbry: The 'd' above doesn't look nice.

dbcon/mysql/is_columnstore_tables.cpp
```diff
@@ -142,7 +142,7 @@ static int is_columnstore_tables_fill(THD* thd, TABLE_LIST* tables, COND* cond)
         if (tb_info.tablewithautoincr)
         {
             table->field[5]->set_notnull();
-            table->field[5]->store(systemCatalogPtr->nextAutoIncrValue((*it).second));
+            table->field[5]->store(systemCatalogPtr->nextAutoIncrValue((*it).second, lower_case_table_names), lower_case_table_names);
```
tntnatbry: You are passing `lower_case_table_names` to table->field[5]->store() as well here.

tntnatbry (changes requested):

tntnatbry:
I have added new review comments. In addition, the previous review also has comments that need to be resolved.

dbcon/mysql/ha_mcs_execplan.cpp
davidhall586: Anything placed into csep has already been converted.  No need to spend cpu doing it again.

dbcon/mysql/ha_mcs_execplan.cpp
davidhall586: Anything put into SimpleColumn has been converted if needed. No need to spend cpu cycles doing it again.

dbcon/mysql/ha_mcs_execplan.cpp
davidhall586: Anything put into SimpleColumn has been converted if needed. No need to spend cpu cycles doing it again.

dbcon/mysql/ha_mcs_execplan.cpp
davidhall586: Anything put into SimpleColumn has been converted if needed. No need to spend cpu cycles doing it again.

dbcon/mysql/ha_mcs_execplan.cpp
davidhall586: Anything put into SimpleColumn has been converted if needed. No need to spend cpu cycles doing it again.

dbcon/mysql/ha_mcs_impl.cpp
davidhall586: schema and tableName are guaranteed to have been converted before any call to getOnUpdateTimestampColumns

dbcon/mysql/ha_mcs_impl.cpp
davidhall586: aTableName was converted at line 1580

dbcon/mysql/ha_mcs_impl_if.h
davidhall586: Fixed

dbcon/execplan/calpontsystemcatalog.cpp
davidhall586: Maybe, but to retrofit would take a bit. Time constraints...

dbcon/execplan/calpontsystemcatalog.cpp
davidhall586: Yep

dbcon/execplan/calpontsystemcatalog.cpp
davidhall586: Yep

dbcon/mysql/ha_mcs_execplan.cpp
davidhall586: All values put into csep have already been converted

dbcon/mysql/ha_mcs_execplan.cpp
davidhall586: All values put into csep or gwi have already been converted

dbcon/mysql/ha_mcs_impl.cpp
davidhall586: tableName and tmpTableName are converted just above, so the strcasecmp is no longer needed.

dbcon/mysql/ha_mcs_impl.cpp
davidhall586: Fixed

dbcon/mysql/ha_mcs_impl.cpp
davidhall586: false added as second arg

dbcon/mysql/idb_mysql.h
davidhall586: Are you saying my fingers bounce?

dbcon/mysql/is_columnstore_tables.cpp
davidhall586: After some study, we should ignore lower_case_table_names for information schema.

dbcon/mysql/is_columnstore_tables.cpp
davidhall586: Changed again. It seems information schema should ignore the lower_case_table_names

dbcon/mysql/ha_mcs_execplan.cpp
davidhall586: all values added to gwi have already been converted

tntnatbry (approved):

PR #1460, opened 2020-09-21
MCOL-4144 1.5 enable lower_case_table_names

davidhall586:
Create tables and schemas with lower case name only if the flag is set.
During operations, convert to lowercase in plugin. Byt the time a query gets to ExeMgr, DDLProc etc., everything must be lower case if the flag is set, and undisturbed if not.


tntnatbry:
Review comments are added to https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/1459.

tntnatbry (changes requested):
Review comments are added to https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/1459.

PR #1462, opened 2020-09-21
MCOL-4282 Follow up on the previous commit.

tntnatbry:
Activate statement arena for:
  1. disable_indices_for_CEJ()
  2. in_subselect_rewrite() for prepared statements

### DIFF
--- a/dbcon/mysql/ha_mcs_pushdown.cpp
+++ b/dbcon/mysql/ha_mcs_pushdown.cpp
@@ -796,7 +796,13 @@ create_columnstore_select_handler(THD* thd, SELECT_LEX* select_lex)
     handler= new ha_columnstore_select_handler(thd, select_lex);
     JOIN *join= select_lex->join;
     {
+        Query_arena *arena, backup;
+        arena= thd->activate_stmt_arena_if_needed(&backup);
+
         disable_indices_for_CEJ(thd);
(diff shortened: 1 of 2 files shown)


dbcon/mysql/ha_mcs_pushdown.cpp
```diff
@@ -826,7 +831,10 @@ create_columnstore_select_handler(THD* thd, SELECT_LEX* select_lex)
                 sel->where= conds;
 
                 if (isPS)
-                    sel->prep_where= conds;
+                {
+                    sel->prep_where= conds ? conds->copy_andor_structure(thd) : 0;
+                    in_subselect_rewrite(sel);
```
drrtuy: We should check the return value of this in_subselect_rewrite() call and set unsupported_feature. See the block at line 859.

drrtuy (changes requested):
Plz look at the comment.

dbcon/mysql/ha_mcs_pushdown.cpp
tntnatbry: Good catch. This is now fixed.

tntnatbry:
The comment is addressed, please review the PR again.

drrtuy (approved):

PR #1474, opened 2020-09-25
MCOL-4282 Follow up version 2.

tntnatbry:
Mutate the optimizer flags for prepared statements in:
  1. ha_mcs::open
  2. ha_mcs::discover_check_version

This is done to ensure the optimizer flags are disabled before
JOIN::prepare() is called during "PREPARE stmt FROM ...".

### DIFF
--- a/dbcon/mysql/ha_mcs.cpp
+++ b/dbcon/mysql/ha_mcs.cpp
@@ -220,6 +220,20 @@ int ha_mcs::open(const char* name, int mode, uint32_t test_if_locked)
 {
     DBUG_ENTER("ha_mcs::open");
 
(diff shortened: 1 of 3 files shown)


drrtuy (commented):
I might have missed it but I see that we don't restore optimizer flags after prepare so what flags will be used in the scenario at the 3d step?
create table cs1(i bigint)engine=columnstore;
prepare stmt 'select * from cs1';
select * from innodb_t;

tntnatbry:
PR is updated to restore the optimizer flags in ha_mcs::reset(). Please review again.

drrtuy (approved):

PR #1484, opened 2020-09-30
MCOL-4314 Fix newly initialized nodes corrupting shared storagemanager data a previous node created.

jmrojas2332:
The problem this addresses is only an issue in scenarios with shared storagemanager.
Issue: when columnstore is installed on new nodes after columnstore is installed on a previous node, the new nodes corrupt data in storagemanager folder when they initialize as single nodes.


drrtuy:
Don't merge before CI starts to work.

drrtuy (approved):

PR #1504, opened 2020-10-03
MCOL-4328 There is a new option in both cpimport and cpimport.bin to …

drrtuy:
…asign

an owner for all data files created by cpimport

The patch consists of two parts: cpimport.bin changes, cpimport splitter
changes

cpimport.bin computes uid_t and gid_t early and propagates it down the stack
where MCS creates data files


abarkov (approved):
The patch is OK to push

PR #1518, opened 2020-10-09
Don't merge MCOL-4337 Controllernode now tries to establish connections to

drrtuy:
all workernodes when initialising
There is a new corresponding XML config options
DBRM_Controller.WorkerConnectionTimeout that controls how long
controllernode waits for wns before it proceeds


versioning/BRM/masterdbrmnode.cpp
```diff
@@ -171,11 +158,80 @@ void MasterDBRMNode::initMsgQueues(config::Config* config)
...
-        slaves.push_back(MessageQueueClientPool::getInstance(module));
+        ltmp = static_cast<int>(config::Config::fromText(stmp));
+        if (ltmp > 1)
+            connectTimeoutSecs = ltmp;
+    }
+}
+
+void MasterDBRMNode::connectToWorkers(const size_t connectTimeoutSecs)
+{
+    size_t timeoutMsecs = connectTimeoutSecs * 1000000;
```
jmrojas2332, line 195: minor: msecs can be interpreted as milliseconds, but I believe you mean microseconds. a rename of variable can clarify intention

jmrojas2332 (approved):
Everything looks good to me. This is much needed PR!

PR #1523, opened 2020-10-12
MCOL-4334 Enable Select Handler for queries run inside Stored Procedures

drrtuy:
There is another session variable to enable/disable SH in SP

### DIFF
--- a/dbcon/mysql/ha_mcs_sysvars.cpp
+++ b/dbcon/mysql/ha_mcs_sysvars.cpp
@@ -98,6 +98,15 @@ static MYSQL_THDVAR_BOOL(
     1
 );
 
+static MYSQL_THDVAR_BOOL(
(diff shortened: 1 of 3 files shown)


davidhall586 (approved):

dbcon/mysql/ha_mcs_sysvars.cpp
```diff
@@ -419,6 +429,16 @@ void set_group_by_handler(THD* thd, bool value)
     THDVAR(thd, group_by_handler) = value;
 }
 
+bool get_select_handler_in_stored_procedures(THD* thd)
+{
+    return ( thd == NULL ) ? false :
```
tntnatbry, line 434: Since the default value of the variable is `true`, we should return back a `true` here for the case when `thd == NULL`

tntnatbry (approved):

PR #1526, opened 2020-10-14
MCOL-641 This commit fixes a group_concat() regression

tntnatbry:
on a narrow decimal field, introduced by an earlier commit.

### DIFF
--- a/dbcon/joblist/groupconcat.cpp
+++ b/dbcon/joblist/groupconcat.cpp
@@ -471,16 +471,15 @@ void GroupConcator::outputRow(std::ostringstream& oss, const rowgroup::Row& row)
                 else
                 {
                     int64_t intVal = row.getIntField(*i);
+
                     if (scale == 0)
                     {
                         oss << intVal;
                     }
                     else
                     {
-                        char buf[utils::MAXLENGTH8BYTES];
-                        dataconvert::DataConvert::decimalToString(intVal,
-                             scale, buf, sizeof(buf), types[*i]);
-                        oss << fixed << buf;
+                        long double dblVal = intVal / pow(10.0, (double)scale);
+                        oss << fixed << setprecision(scale) << dblVal;
                     }
                 }
 


tntnatbry:
```sql
MariaDB [test]> drop table cs1;
Query OK, 0 rows affected (0.366 sec)

MariaDB [test]> create table cs1 (a decimal(4,2))engine=columnstore;
Query OK, 0 rows affected (0.400 sec)

MariaDB [test]> insert into cs1 values (0.08);
Query OK, 1 row affected (0.211 sec)
```

Incorrect output:
```sql
MariaDB [test]> select group_concat(a) from cs1;
+-----------------+
| group_concat(a) |
+-----------------+
| .08             |
+-----------------+
1 row in set (0.077 sec)
```

Expected output:
```sql
MariaDB [test]> select group_concat(a) from cs1;
+-----------------+
| group_concat(a) |
+-----------------+
| 0.08            |
+-----------------+
1 row in set (0.236 sec)
```

dbcon/joblist/groupconcat.cpp
```diff
@@ -471,16 +471,15 @@ void GroupConcator::outputRow(std::ostringstream& oss, const rowgroup::Row& row)
...
                     {
                         oss << intVal;
                     }
                     else
                     {
-                        char buf[utils::MAXLENGTH8BYTES];
-                        dataconvert::DataConvert::decimalToString(intVal,
-                             scale, buf, sizeof(buf), types[*i]);
-                        oss << fixed << buf;
+                        long double dblVal = intVal / pow(10.0, (double)scale);
```
drrtuy, line 481: This doesn't look right. I believe we should teach decimalToString to do the thing if it can't.

drrtuy: I believe you can restore the the old code replacing pow() with a ref to a dict of 10^n values.

tntnatbry: Fixing pre-existing bugs in narrow decimals should be outside the scope of MCOL-641.

For narrow decimals, performing a floating point division for further processing appears to be a common theme. As an example, see RowAggregation::doSum: here for narrow decimals, we are again using `valIn /= pow(10.0, scale);`. I am sure there will be a lot of other places as well that will need changes. This commit simply reverts back changes performed to narrow decimals to the original state.

drrtuy: I would disagree given that we are going to rewrite decimal processing to some degree. It is a very good point that the change must be introduced throughout the whole code base so I'll create an issue to replace all pow(10.0, (double)scale) expressions with a lookup in a values dictionary where the scale < 19 and assign it to you.

drrtuy:
I believe we also have a bug in dataconvert::DataConvert::decimalToString() this should be addressed also.

PR #1530, opened 2020-10-16
MCOL-4313 Very fragile but high speed approach with inline ASM

drrtuy:
GCC compiler uses aligned versions of SIMD instructions expecting
aligned memory blocks that is hard to implement now


abarkov:
I suggest to add macros  int128_store() and int128_get(). OK to push otherwise.

PR #1542, opened 2020-10-22
MCOL-641 Fixes for arithmetic operations.

tntnatbry:
1. Perform type promotion to wide decimal if the result
   of an arithmetic operation has a precision > 18.
2. Only set the decimal width of an arithmetic operation to wide
   if both the LHS and RHS of the operation are decimal types.

### DIFF
--- a/dbcon/mysql/ha_mcs_execplan.cpp
+++ b/dbcon/mysql/ha_mcs_execplan.cpp
@@ -3601,68 +3601,51 @@ ArithmeticColumn* buildArithmeticColumn(
 
     // @bug5715. Use InfiniDB adjusted coltype for result type.
     // decimal arithmetic operation gives double result when the session variable is set.
-    CalpontSystemCatalog::ColType mysql_type = colType_MysqlToIDB(item);
+    CalpontSystemCatalog::ColType mysqlType = colType_MysqlToIDB(item);
 
-    if (mysql_type.colDataType == CalpontSystemCatalog::DECIMAL ||
-        mysql_type.colDataType == CalpontSystemCatalog::UDECIMAL)
+    const CalpontSystemCatalog::ColType& leftColType = pt->left()->data()->resultType();
+    const CalpontSystemCatalog::ColType& rightColType = pt->right()->data()->resultType();
+
+    // Only tinker with the type if all columns involved are decimal
+    if (datatypes::Decimal::isDecimalOperands(mysqlType.colDataType,
+            leftColType.colDataType, rightColType.colDataType))
     {
-        int32_t leftColWidth = pt->left()->data()->resultType().colWidth;
-        int32_t rightColWidth = pt->right()->data()->resultType().colWidth;
+        int32_t leftColWidth = leftColType.colWidth;
+        int32_t rightColWidth = rightColType.colWidth;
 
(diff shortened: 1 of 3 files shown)


dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -3601,68 +3601,55 @@ ArithmeticColumn* buildArithmeticColumn(
 
     // @bug5715. Use InfiniDB adjusted coltype for result type.
     // decimal arithmetic operation gives double result when the session variable is set.
-    CalpontSystemCatalog::ColType mysql_type = colType_MysqlToIDB(item);
+    CalpontSystemCatalog::ColType mysqlType = colType_MysqlToIDB(item);
 
-    if (mysql_type.colDataType == CalpontSystemCatalog::DECIMAL ||
-        mysql_type.colDataType == CalpontSystemCatalog::UDECIMAL)
+    CalpontSystemCatalog::ColType leftColType = pt->left()->data()->resultType();
```
drrtuy: resultType() returns a const IIRC so leftColType can be a const symbol.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -3601,68 +3601,55 @@ ArithmeticColumn* buildArithmeticColumn(
...
+    CalpontSystemCatalog::ColType rightColType = pt->right()->data()->resultType();
+
+    // Only tinker with the type if all columns involved are decimal
+    if ((mysqlType.colDataType == CalpontSystemCatalog::DECIMAL ||
```
drrtuy, line 3610: Plz make an ORed comparison an inline function to reduce a boilerplate code.

drrtuy (commented):
Could you ellaborate on the goals behind the change in tupleaggregatestep.cpp ?

dbcon/mysql/ha_mcs_execplan.cpp

tntnatbry: Also replaced the OR code with an inline isDecimalOperands().

tntnatbry:
There are 2 changes to tupleaggregastep.cpp:
  1. Addition of a for loop condition to avoid a potential segfault in the case `scale.size() < jobInfo.nonConstCols.size()` or `scale.size()  < jobInfo.nonConstDelCols.size()`
  2. Replaced `oidsAgg.push_back()`, `keysAgg.push_back()` calls in `wideDecimalOrLongDouble()` which were duplicated in both the if and else blocks, with a single set of calls in the calling function.

drrtuy:
> 2. Replaced `oidsAgg.push_back()`, `keysAgg.push_back()` calls in `wideDecimalOrLongDouble()` which were duplicated in both the if and else blocks, with a single set of calls in the calling function.

The second change divides assignment into parts so one can easily overlook this. I would suggest to put keysAgg.push_back() and oidsAgg.push_back() into the function after if-else.

dbcon/mysql/ha_mcs_execplan.cpp
drrtuy: Thx a lot.

drrtuy (approved):

PR #1543, opened 2020-10-22
Columnstore syslog file creation should be with mysql as owner

jmrojas2332:
Columnstore syslog file creation should be with mysql as owner

Along with this I have created cmake variables for default user/group to increase maintainability of user/group. Allows us to stop hardcoding "mysql" as user and will allow us in the future to simplistically change default.


drrtuy:
With all that static mysql put here and there we'll become static. I believe someday we'll have to replace mysql with mariadb and that will be a great day full of useless efforts.

drrtuy (approved):

jmrojas2332:
@drrtuy @benthompson15 observe new changes, making default user/group cmake variables

PR #1544, opened 2020-10-22
Don't merge yet. Columnstore syslog file creation should be with mysql as owner

jmrojas2332:
Columnstore syslog file creation should be with mysql as owner

Along with this I have created cmake variables for default user/group to increase maintainability of user/group. Allows us to stop hardcoding "mysql" as user and will allow us in the future to simplistically change default.


drrtuy (changes requested):
Very good.
However there are some files left behind: mcs-storagemanager.service.in, mariadb-columnstore-start.sh.in, mcs-loadbrm.py.in(has both constant and literals in the code).

drrtuy (approved):

PR #1549, opened 2020-10-27
Updated the documentation in rowgroup.h

pleblanc1976:
Conflicts:
	utils/rowgroup/rowgroup.h


pleblanc1976:
I noticed a couple lines that shouldn't be in the version for 1.4, I'll remove them.

PR #1552, opened 2020-10-28
Stalled. Mcol 3771 benchmark re2 to replace boost regex

mariadb-SergeyZefirov:
Relevant JIRA ticket: https://jira.mariadb.org/browse/MCOL-3771?page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel&focusedCommentId=169986

What I did here is to remove a bunch of "#ifdef __linux__" conditionals from different parts of code and moved them into single header file (utils/common/regexp-port.h). That file defines a regex containing structure and everyone creates that structures and uses unified interface.

To utilize re2 I had to link it somehow and I added it as an external CMake project. This is second relatively big change.

I also did a dozen or so fixes of warnings, most probably I will have to revert some of them if not all.


drrtuy:
With the latest patch on collation-aware LIKE MCS started to use MDB like facilities. JFYI MCS already uses PCRE provided by MDB for RLIKE processing.

PR #1556, opened 2020-10-30
Mcol 4313 macros

drrtuy:


datatypes/mcs_decimal.cpp
```diff
@@ -209,6 +217,23 @@ namespace datatypes
...
+        if (value == Decimal128Null)
+        {
+            return std::string("NULL");
+        }
+        else if (value == Decimal128Empty)
+        {
+            return std::string("EMPTY");
+        }
+        int128_t& constLessValue = const_cast<int128_t&>(value);
+        dataconvert::DataConvert::decimalToString(&constLessValue,
```
tntnatbry, line 232: I noticed the first parameter to `dataconvert::DataConvert::decimalToString` is declared as int128_t*. I think it should be const int128_t*. Also `decimalToString` is changing the source value from negative to positive. I don't think that's intended.

utils/common/simd_asm.h
```diff
@@ -0,0 +1,77 @@
...
+
+// Inline asm has three argument lists: output, input and clobber list
+
+#if defined(__GNUC__) && (__GNUC___ > 7)
+    #define MACRO_VALUE_PTR_128(dst, \
+                                dst_restrictions, \
+                                src, \
+                                src_restrictions, \
+                                clobb) \
+        ::memcpy((dst), (src), sizeof(__int128));
```
tntnatbry: This should be `&(src)`.

utils/rowgroup/rowgroup.h
```diff
@@ -1165,12 +1152,7 @@ inline void Row::setLongDoubleField(const long double& val, uint32_t colIndex)
 
 inline void Row::setInt128Field(const int128_t& val, uint32_t colIndex)
 {
-    *((int128_t*)&data[offsets[colIndex]]) = val;
-}
-
-inline void Row::setUint128Field(const uint128_t& val, uint32_t colIndex)
-{
-    *((uint128_t*)&data[offsets[colIndex]]) = val;
+    setBinaryField<int128_t>(&(const_cast<int128_t&>(val)), colIndex);
```
tntnatbry: I think the first parameter of `setBinaryField` should be a `const T*`.

utils/windowfunction/frameboundrange.cpp
```diff
@@ -299,21 +299,26 @@ void FrameBoundExpressionRange<T>::validate()
...
 
             case execplan::CalpontSystemCatalog::DECIMAL:
+            case execplan::CalpontSystemCatalog::UDECIMAL:
             {
-                if (this->fRow.getColumnWidth(this->fIndex[1]) < 16)
+                if (UNLIKELY(this->fRow.getColumnWidth(this->fIndex[1])
+                                < datatypes::MAXDECIMALWIDTH))
                 {
-                    int64_t tmp = this->fRow.getIntField(this->fIndex[1]);
+                    uint64_t tmp = this->fRow.getIntField(this->fIndex[1]);
```
tntnatbry: This is not correct for narrow decimals; the rhs returns an int64_t for an UDECIMAL which is not correct. I think we should keep the original two separate cases for DECIMAL and UDECIMAL.

tntnatbry (changes requested):

utils/windowfunction/frameboundrange.cpp
drrtuy: I didn't get the point to be honest.

utils/rowgroup/rowgroup.h
drrtuy: Agree.

datatypes/mcs_decimal.cpp
drrtuy: It should be but outside the scope of the change. And yes, decimalToString has this flaw.

utils/common/simd_asm.h
drrtuy: JFYI The macro will be replaced with the ASM variant from #else preprocessor branch.

PR #1557, opened 2020-10-30
Properly initialize hasCollation data member of Row and RowGroup classes.

tntnatbry:


tntnatbry:
@dhall-MariaDB Please note that I am assuming the default value for `hasCollation` as `false` for initialisation. It is only getting set to `true` at 2 places: `RowGroup::RowGroup` non-default ctor, and `RowGroup::truncate`, based on whether the Row has a CHAR/VARCHAR/TEXT field. Since you added `hasCollation` support in the first place, so I just wanted to mention how I am initialising it.

davidhall586 (approved):

PR #1567, opened 2020-11-03
Mcol 4255 1.5

davidhall586:


utils/funcexp/func_round.cpp
```diff
@@ -136,7 +136,18 @@ uint64_t Func_round::getUintVal(Row& row,
                                 bool& isNull,
                                 CalpontSystemCatalog::ColType& op_ct)
 {
-    return parm[0]->data()->getUintVal(row, isNull);
+    uint64_t x;
+    if (UNLIKELY(op_ct.colDataType == execplan::CalpontSystemCatalog::DATE))
+    {
+        IDB_Decimal d = getDecimalVal(row, parm, isNull, op_ct);
+        x = d.value;
```
drrtuy: Sorry that I get in but IDB_Decimal::value is an int64. I believe the assumption is that DATE can't have negative int values but it is better to explicitly add static_cast just to avoid complications and make it easier to read.

davidhall586: Fixed as suggested

tntnatbry (approved):

PR #1570, opened 2020-11-05
MCOL-4188 Fix regression in a union subquery involving a numeric field.

tntnatbry:
Since we now perform type promotion to wide decimals for aggregations
involving numeric fields, we need to check for wide decimal in
in and out ROWs and call the appropriate setter and getter functions.

### DIFF
--- a/utils/dataconvert/dataconvert.cpp
+++ b/utils/dataconvert/dataconvert.cpp
@@ -3160,6 +3160,9 @@ DataConvert::joinColTypeForUnion(datatypes::SystemCatalog::TypeHolderStd &unione
                         unionedType.colDataType = datatypes::SystemCatalog::DECIMAL;
                     }
 
+                    if (type.precision > unionedType.precision)
+                        unionedType.precision = type.precision;
+
                     unionedType.scale = (type.scale > unionedType.scale) ? type.scale : unionedType.scale;
                     break;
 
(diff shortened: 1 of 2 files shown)


drrtuy:
I might had missed that but the patch that promotes to wide decimal should work for expressions that have decimal arguments, isn't it? Why do we check for width processing non-decimal types?

tntnatbry:
I didn't fully understand the comment. This patch checks for whether we are processing a wide decimal field for either the `in` or the `out` Row, and calls either setInt128Field or getInt128Field if we are working with wide decimal. Earlier, this code would only call Row::getIntField or Row::setIntField which raises an exception on width==16.

drrtuy:
But in some parts we are using if conditions in cases that covers
non-decimal types. Why are we doing this?

пт, 6 нояб. 2020 г., 21:20 Gagan Goel <notifications@github.com>:

tntnatbry:
We are only checking `if column width == 16` for DECIMAL/UDECIMAL types. Which parts are you referring to?

drrtuy:
I was reffering to if condition you added in DataConvert::joinColTypeForUnion.

utils/dataconvert/dataconvert.cpp
```diff
@@ -3427,9 +3430,12 @@ DataConvert::joinColTypeForUnion(datatypes::SystemCatalog::TypeHolderStd &unione
...
                 case datatypes::SystemCatalog::UDOUBLE:
-                    unionedType.colDataType = datatypes::SystemCatalog::DOUBLE;
-                    unionedType.scale = 0;
-                    unionedType.colWidth = sizeof(double);
+                    if (!unionedType.isWideDecimalType())
+                    {
+                        unionedType.colDataType = datatypes::SystemCatalog::DOUBLE;
+                        unionedType.scale = 0;
+                        unionedType.colWidth = sizeof(double);
+                    }
```
tntnatbry, line 3417: @drrtuy I believe you are referring to this block. Here I am checking if unionedType is a wide decimal, if so, then we don't reduce the type of unionedType to double. Same goes in a similar case block for long double.

tntnatbry:
DECIMAL/UDECIMAL are separated out in their own case blocks. Please review.

drrtuy (approved):

PR #1586, opened 2020-11-10
MCOL-4188 Fix regressions in CEIL()/CHAR() for narrow decimals.

tntnatbry:
In addition, a regression in a WHERE clause with a WF field
as the LHS and an addition operation on two WF fields on the RHS
is also fixed. The issue was SimpleColumn::getDecimalVal() was
setting precision = 19, with the value of one of the operands of the
addition operation being set in VDecimal::value instead of
VDecimal::s128Value. addSubtractExecute() in mcs_decimal.cpp makes the
assumption that if precision > 18 and precision <= 38, we need to
fetch the wide s128Value, not the narrow value field. So we are
fixing the precision set in SimpleColumn::getDecimalVal().

### DIFF
--- a/dbcon/execplan/simplecolumn.h
+++ b/dbcon/execplan/simplecolumn.h
@@ -335,20 +335,18 @@ public:
         //      the original decimal scale is stored in scale field, which is no use for double.
         if (fResultType.precision == -1)
         {
-            IDB_Decimal rv;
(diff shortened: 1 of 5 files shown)


tntnatbry:
Refactored simplecolumn.h/windowfunctioncolumn.cpp as suggested. Please review.

drrtuy (approved):

PR #1588, opened 2020-11-11
Fix warnings

mariadb-AlexeyAntipovsky:


storage-manager/src/unit_tests.cpp
```diff
@@ -348,8 +348,9 @@ void metadataJournalTest(std::size_t size, off_t offset)
     w.run();
 
     // verify response
-    err = ::recv(sessionSock, buf, 1024, MSG_DONTWAIT);
-    sm_response *resp = (sm_response *) buf;
+    uint8_t bufRead[1024];
+    err = ::recv(sessionSock, bufRead, 1024, MSG_DONTWAIT);
```
drrtuy: Could you replace literal with a constexpr sizeof(bufRead) here and in the other similar places?

utils/common/checks.h
```diff
@@ -0,0 +1,39 @@
+#ifndef UTILS_COMMON_CHECKS_H
+#define UTILS_COMMON_CHECKS_H
+
+namespace utils {
+
+template <typename T>
+inline bool is_nonnegative (T val);
+
+template <> inline bool is_nonnegative(uint8_t)       { return true; }
```
drrtuy: Clever approach indeed. 
What do you think if we employ type traits here to future reduce boilerplat. Something like
template<typename T, typename = std::enable_if<std::is_unsigned<T>::value> >
nline bool is_nonnegative(T)       { return true; }
template<typename T, typename = std::enable_if<std::is_signed<T>::value> >
nline bool is_nonnegative(T)       { return v >=0; }

utils/common/vlarray.h
```diff
@@ -0,0 +1,81 @@
+#ifndef UTILS_COMMON_VLARRAY_H
+#define UTILS_COMMON_VLARRAY_H
+
+#include <boost/noncopyable.hpp>
+
+namespace utils {
+
+template <typename T, size_t SIZE=256>
+class VLArray : boost::noncopyable
```
drrtuy: Very nice approach. 
What do you think if we reduce boost importance here and employ an approach like [this](https://www.codeproject.com/Tips/1220451/Cplusplus-Tip-Make-Your-Class-Non-copyable-Without)? Will this work or not?

utils/common/vlarray.h
```diff
@@ -0,0 +1,81 @@
+#ifndef UTILS_COMMON_VLARRAY_H
+#define UTILS_COMMON_VLARRAY_H
+
+#include <boost/noncopyable.hpp>
+
+namespace utils {
+
+template <typename T, size_t SIZE=256>
```
drrtuy: Could you reduce this down to 64 that should be enough for the most cases?

utils/common/checks.h
mariadb-AlexeyAntipovsky: I just forgot that std::enable_if appeared in c++11, not c++14

mariadb-RomanNavrotskiy:
just noticed Debian packaging missing files

```
dh_install: Cannot find (any matches for) "usr/bin/ProcMgr" (tried in "." and "debian/tmp")
dh_install: mariadb-plugin-columnstore missing files: usr/bin/ProcMgr
dh_install: Cannot find (any matches for) "usr/bin/ProcMon" (tried in "." and "debian/tmp")
dh_install: mariadb-plugin-columnstore missing files: usr/bin/ProcMon
```
https://ci.columnstore.mariadb.net/mariadb-corporation/mariadb-columnstore-engine/1113/8/4

drrtuy:
These are expected errors. This depends on your work Roman that you are
doing with deb packaging.

чт, 12 нояб. 2020 г., 19:22 mariadb-RomanNavrotskiy <
notifications@github.com>:

drrtuy (approved):

PR #1606, opened 2020-11-18
MCOL-4397 dev call refreshShm() for calshowpartitions

davidhall586:


tntnatbry (changes requested):
I think the following UDFs also need these refresh calls: caldisablepartitions(), calenablepartitions(), caldroppartitions().

davidhall586:
caldisablepartitions(), calenablepartitions(), caldroppartitions() do not access the extent map in the connector. The issue only arises if the extent map is accessed in the connector. These functions make tcp calls to DDLProc, which was refreshed during columnstore start.

tntnatbry (approved):

PR #1607, opened 2020-11-19
MCOL-4394 __float128 related code had been moved into a separate file

drrtuy:
Trim to double and to long double conversions for Decimal

### DIFF
--- a/datatypes/mcs_decimal.h
+++ b/datatypes/mcs_decimal.h
@@ -25,6 +25,7 @@
 #include "exceptclasses.h"
 #include "widedecimalutils.h"
 #include "mcs_int128.h"
+#include "mcs_float128.h"
 
 
 namespace datatypes
@@ -334,36 +335,6 @@ class Decimal
             return static_cast<uint64_t>(value);
         }
 
-         /**
-            @brief The method converts a wide decimal value to a double.
-        */
-        static inline double getDoubleFromWideDecimal(const int128_t& value, int8_t scale)
-        {
-            int128_t scaleDivisor;
-
-            getScaleDivisor(scaleDivisor, scale);
-
-            __float128 tmpval = (__float128) value / scaleDivisor;
-
-            return getDoubleFromFloat128(tmpval);
-        }
-
-        /**
-            @brief The method converts a wide decimal value to a double.
-        */
-        static inline double getDoubleFromWideDecimal(const int128_t& value)
-        {
-            return getDoubleFromFloat128(static_cast<__float128>(value));
-        }
-
-        /**
-            @brief The method converts a wide decimal value to a long double.
-        */
-        static inline long double getLongDoubleFromWideDecimal(const int128_t& value)
-        {
-            return getLongDoubleFromFloat128(static_cast<__float128>(value));
-        }
-
         /**
             @brief The method converts a wide decimal value to an int64_t,
             saturating the value if necessary.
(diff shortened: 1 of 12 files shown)


abarkov (approved):

datatypes/mcs_decimal.h
```diff
@@ -598,6 +569,35 @@ class VDecimal: public TSInt128
...
+    inline operator double() const
+    {
+        return toDouble();
+    }
+
+    inline long double toLongDouble() const
+    {
+        datatypes::TFloat128 y(s128Value);
+        return static_cast<long double>(y);
+    }
```
tntnatbry, line 599: The semantics of `toLongDouble()` should be the same as `toDouble()`, i.e., we should use the `scale` here to construct `TFloat128`.

Also, we should also add `operator long double()` as a member function here.

datatypes/mcs_int128.h
```diff
@@ -203,6 +180,36 @@ class TSInt128
       return s128Value == static_cast<int128_t>(x);
     }
 
+    inline operator double() const
+    {
+      return toDouble();
+    }
+
+    inline long double toDouble() const
```
tntnatbry, line 188: Return type here should be `double` and not `long double`.

datatypes/mcs_int128.h
```diff
@@ -203,6 +180,36 @@ class TSInt128
...
+    {
+      return static_cast<double>(s128Value);
```
tntnatbry, line 190: I think this conversion can overflow, this should be `return static_cast<double>(TFloat128(s128Value))`.

datatypes/mcs_int128.h
```diff
@@ -203,6 +180,36 @@ class TSInt128
...
+    }
+
+    inline operator long double() const
+    {
+      return toLongDouble();
+    }
+
+    inline long double toLongDouble() const
+    {
+      return static_cast<long double>(s128Value);
```
tntnatbry, line 200: I think this conversion can overflow, this should be `return static_cast<long double>(TFloat128(s128Value))`.

utils/rowgroup/rowaggregation.cpp
```diff
@@ -2045,8 +2045,9 @@ void RowAggregation::doStatistics(const Row& rowIn, int64_t colIn, int64_t colOu
         case execplan::CalpontSystemCatalog::UDECIMAL:  // handle scale later
             if (LIKELY(fRowGroupIn.getColumnWidth(colIn) == datatypes::MAXDECIMALWIDTH))
             {
-                int128_t* val128InPtr = rowIn.getBinaryField<int128_t>(colIn);
-                valIn = Dec::getLongDoubleFromWideDecimal(*val128InPtr);
+                // To save from unaligned memory
+                datatypes::TSInt128 val128In(rowIn.getBinaryField<int128_t>(colIn));
+                valIn =  static_cast<long double>(val128In.toTFloat128());
```
tntnatbry, line 2050: The `.toTFloat128()` call here is unnecessary.

tntnatbry:
@drrtuy Please look at the review comments I added to this PR.

datatypes/mcs_decimal.h
drrtuy: To be honest we don't need this function at this point so I'll remove it.

datatypes/mcs_int128.h
drrtuy: Good catch. I wonder how did this pass through the compiler.

datatypes/mcs_int128.h
drrtuy: This is a very good question. As you can see I didn't change the behavior so I trusted whatever logic existed before I wrote this. And I believe the logic had been introduced by you in the first place.

utils/rowgroup/rowaggregation.cpp
drrtuy: Not exactly. TSInt128::operator long double() shouldn't be used here b/c I wish  to remove all casts from int128 -> real types b/c they doesn't look right. TFloat128 should be used instead. We'll have extra 16 byte asignment step here but we won't be casting from int128 into a real type.

PR #1608, opened 2020-11-19
DO NOT MERGE: MCOL-3917 DDLProc/DMLProc re-establish connections on Primproc restart.

jmrojas2332:
Whenever Primproc is restarted, or crashes and recovers, both DMLProc and DDLProc are unusable. This patch allows DMLProc/DDLProc to Primproc reconnection to happen when the error inside DistributedEngineComm is hit.

This patch also removes all systemd service dependencies from systemd units except the following:

mcs-exemgr restarts when mcs-controllernode restarts so "system is ready to accept ready" flag is set to true on controllernode restarts.
mcs-dmlproc restarts when mcs-controllernode restarts so "systemready" flag  is set to true during controllernode restart

mcs-loadbrm is started when mcs-workernode@1 or mcs-workernode@2 is restarted/started.

The latter will likely stay as-is, but the other two will be addressed in a future PR.

### DIFF
--- a/dbcon/joblist/distributedenginecomm.cpp
+++ b/dbcon/joblist/distributedenginecomm.cpp
@@ -1054,7 +1054,10 @@ int DistributedEngineComm::writeToClient(size_t index, const ByteStream& bs, uin
         alarmItem.append(" PrimProc");
         alarmMgr.sendAlarmReport(alarmItem.c_str(), oam::CONN_FAILURE, SET);
         */
-        throw runtime_error("DistributedEngineComm::write: Broken Pipe error");
+
+        // Re-establish connection before throwing error 
+        Setup();
+        throw runtime_error("DistributedEngineComm::write: Broken Pipe error. Attempting to re-establish connection.");
     }
 }
 
--- a/dmlproc/dmlprocessor.cpp
+++ b/dmlproc/dmlprocessor.cpp
@@ -1097,8 +1097,6 @@ void PackageHandler::run()
     }
     catch (std::exception& e)
     {
-
-
         cout << "dmlprocessor.cpp PackageHandler::run() package type("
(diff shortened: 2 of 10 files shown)


dmlproc/dmlprocessor.cpp
```diff
@@ -1112,6 +1110,9 @@ void PackageHandler::run()
         ml.logErrorMessage(message);
         result.result = DMLPackageProcessor::COMMAND_ERROR;
         result.message = message;
+
+        DistributedEngineComm* dec = DistributedEngineComm::instance(frm);
+        dec->Setup();
```
jmrojas2332, line 1115: The call to setup inside DEC works the majority of the time, but I hit a case where this point is reached without queries being runnable. Call setup again to avoid the edge case (only happens for dmlproc).

dbcon/joblist/distributedenginecomm.cpp
```diff
@@ -1054,7 +1054,10 @@ int DistributedEngineComm::writeToClient(size_t index, const ByteStream& bs, uin
         alarmItem.append(" PrimProc");
         alarmMgr.sendAlarmReport(alarmItem.c_str(), oam::CONN_FAILURE, SET);
         */
-        throw runtime_error("DistributedEngineComm::write: Broken Pipe error");
+
+        // Re-establish connection before throwing error 
+        Setup();
```
drrtuy, line 1059: I wonder if Setup() can throw on its own? Control flow will go into another exceptions handler so we'll lost the error message and luckily if the new exception also belongs to runtime_error class.

dmlproc/dmlprocessor.cpp
drrtuy: This is the most important part of the change in dmlprocessor. I don't know how to test it properly to be honest.

PR #1610, opened 2020-11-20

tntnatbry:
1. Add wide decimal support to AggregateColumn::evaluate
and TreeNode::getDecimalVal().
2. Use the pm aggregate attributes to determine um aggregate
attributes in TupleAggregateStep::prep2PhasesAggregate.

### DIFF
--- a/datatypes/mcs_datatype.h
+++ b/datatypes/mcs_datatype.h
@@ -230,6 +230,16 @@ public:
           }
       }
 
+      /**
+          @brief The method returns true if the column precision
+          belongs to a wide decimal range.
+      */
+      inline bool isWideDecimalPrecision() const
+      {
+          return ((precision > INT64MAXPRECISION) &&
+              (precision <= INT128MAXPRECISION));
+      }
+
   };
 
   class TypeHolderStd: public TypeAttributesStd
--- a/dbcon/execplan/treenode.h
+++ b/dbcon/execplan/treenode.h
@@ -1003,41 +1003,82 @@ inline IDB_Decimal TreeNode::getDecimalVal()
             throw logging::InvalidConversionExcept("TreeNode::getDecimalVal: non-support conversion from binary string");
 
         case CalpontSystemCatalog::BIGINT:
+        {
+            idbassert(fResultType.scale == 0);
+
+            if (fResultType.isWideDecimalPrecision())
+                fResult.decimalVal = IDB_Decimal(0, 0, fResultType.precision,
+                                         (int128_t)fResult.intVal);
+            else
+                fResult.decimalVal = IDB_Decimal(fResult.intVal, 0,
+                                         fResultType.precision);
+
+            break;
+        }
+
         case CalpontSystemCatalog::MEDINT:
         case CalpontSystemCatalog::INT:
         case CalpontSystemCatalog::SMALLINT:
         case CalpontSystemCatalog::TINYINT:
-            fResult.decimalVal.value = (int64_t)(fResult.intVal * pow((double)10.0, fResultType.scale));
-            fResult.decimalVal.scale = fResultType.scale;
-            fResult.decimalVal.precision = fResultType.precision;
+        {
+            idbassert(fResultType.scale == 0);
+
+            fResult.decimalVal = IDB_Decimal(fResult.intVal, 0,
+                                     fResultType.precision);
+
             break;
+        }
 
         case CalpontSystemCatalog::UBIGINT:
+        {
+            idbassert(fResultType.scale == 0);
+
+            if (fResultType.isWideDecimalPrecision())
+                fResult.decimalVal = IDB_Decimal(0, 0, fResultType.precision,
+                                         (int128_t)fResult.uintVal);
+            else
+                fResult.decimalVal = IDB_Decimal((int64_t)fResult.uintVal, 0,
+                                         fResultType.precision);
+
+            break;
+        }
+
         case CalpontSystemCatalog::UMEDINT:
         case CalpontSystemCatalog::UINT:
         case CalpontSystemCatalog::USMALLINT:
         case CalpontSystemCatalog::UTINYINT:
-            fResult.decimalVal.value = (int64_t)(fResult.uintVal * pow((double)10.0, fResultType.scale));
-            fResult.decimalVal.scale = fResultType.scale;
-            fResult.decimalVal.precision = fResultType.precision;
+        {
+            idbassert(fResultType.scale == 0);
+
+            fResult.decimalVal = IDB_Decimal((int64_t)fResult.uintVal, 0,
+                                     fResultType.precision);
+
             break;
+        }
 
         case CalpontSystemCatalog::LONGDOUBLE:
+        {
+            long double dlScaled = fResult.longDoubleVal;
+
+            if (fResultType.scale > 0)
             {
-                long double dlScaled = fResult.longDoubleVal;
-                if (fResultType.scale > 0)
-                {
-                    dlScaled= fResult.longDoubleVal * pow((double)10.0, fResultType.scale);
-                }
-                if (dlScaled > (double)MAX_BIGINT)
-                {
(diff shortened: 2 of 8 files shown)


dbcon/execplan/treenode.h
```diff
@@ -1031,41 +1034,75 @@ inline IDB_Decimal TreeNode::getDecimalVal()
...
         case CalpontSystemCatalog::MEDINT:
         case CalpontSystemCatalog::INT:
         case CalpontSystemCatalog::SMALLINT:
         case CalpontSystemCatalog::TINYINT:
-            fResult.decimalVal.value = (int64_t)(fResult.intVal * pow((double)10.0, fResultType.scale));
-            fResult.decimalVal.scale = fResultType.scale;
-            fResult.decimalVal.precision = fResultType.precision;
+            fResult.decimalVal = IDB_Decimal(
+                                     (int64_t)(fResult.intVal * pow((double)10.0, fResultType.scale)),
+                                     fResultType.scale, fResultType.precision);
```
abarkov: It seems fResultType.scale should always be 0 here. I think we can remove the pow() part and add this:

`idbassert(fResultType.scale == 0);`

dbcon/execplan/treenode.h
```diff
@@ -1031,41 +1034,75 @@ inline IDB_Decimal TreeNode::getDecimalVal()
...
         case CalpontSystemCatalog::UMEDINT:
         case CalpontSystemCatalog::UINT:
         case CalpontSystemCatalog::USMALLINT:
         case CalpontSystemCatalog::UTINYINT:
-            fResult.decimalVal.value = (int64_t)(fResult.uintVal * pow((double)10.0, fResultType.scale));
-            fResult.decimalVal.scale = fResultType.scale;
-            fResult.decimalVal.precision = fResultType.precision;
+            fResult.decimalVal = IDB_Decimal(
+                                     (int64_t)(fResult.uintVal * pow((double)10.0, fResultType.scale)),
+                                     fResultType.scale, fResultType.precision);
```
abarkov: Same here: I'd remove the pow() part and add `idbassert(fResultType.scale==0);`

dbcon/execplan/treenode.h
```diff
@@ -1031,41 +1034,75 @@ inline IDB_Decimal TreeNode::getDecimalVal()
...
-                fResult.decimalVal.value = (int64_t)roundl((fResult.longDoubleVal * pow((double)10.0, fResultType.scale)));
-                fResult.decimalVal.scale = fResultType.scale;
-                fResult.decimalVal.precision = fResultType.precision;
+                dlScaled= fResult.longDoubleVal * pow((double)10.0, fResultType.scale);
             }
+            if (dlScaled > (double)MAX_BIGINT)
+            {
+                throw logging::InvalidConversionExcept("TreeNode::getDecimalVal: decimal overflow.");
+            }
+
```
abarkov, line 1067: Is the above `if` block still correct with wide decimal support added? I think it should now check for max int128_t value.

The below code could probably do:
- if dlScaled is greater than max int128_t, then error.
- if dlScaled is greated than MAX_BIGINT, the go with the wide-decimal constructor.
- otherwise, go with the narrow decimal constructor.

abarkov (changes requested):

datatypes/mcs_datatype.h
```diff
@@ -231,6 +231,16 @@ class SystemCatalog
           }
       }
 
+      /**
+          @brief The method returns true if the column precision
+          belongs to a wide decimal range.
+      */
+      inline bool isWideDecimalPrecision() const
```
drrtuy, line 237: We can't trust precision detecting DECIMAL type b/c there is a division case( decimal(15) / decimal(38) ) that produces wide decimal with a precision < 18. I would suggest to use width as a marker in the context of this patch where it is possible.

dbcon/execplan/treenode.h
tntnatbry: I agree, `fResultType.scale` here should be 0. Updated the patch.

dbcon/execplan/treenode.h
tntnatbry: Done.

dbcon/execplan/treenode.h
tntnatbry: Good catch, the original `if` block was incorrect. I am now checking if `dlScaled` is in the range of an `int64_t`. If it is, then I call the narrow decimal ctor. If it is outside the range of `int64_t`, I am silently saturating to `int128_t` and calling the wide decimal ctor. I think raising the exception is also an option here, instead of saturating.

datatypes/mcs_datatype.h
tntnatbry: We are only using this function when the source type is a BIGINT/UBIGINT whose precision is sometimes set as 19. This causes issues when the caller detects that the value is wide decimal based on precision (as an example, in our implementations of the arithmetic operations in mcs_decimal.h). Width cannot be correctly used here since BIGINT/UBIGINT are 8 bytes.

drrtuy:
I got the point.
My point was we should discourage using precision as a marker not only here
but everywhere where it is possible and use width instead.

ср, 25 нояб. 2020 г., 3:47 Gagan Goel <notifications@github.com>:

dbcon/execplan/treenode.h
```diff
@@ -32,6 +32,8 @@
 #include <stdlib.h>
 #include <unistd.h>
 
+#include <quadmath.h>
```
drrtuy: We surely will remove quadmath from the code b/c it brings additional dependency so we shouldn't use it.

dbcon/execplan/treenode.h
```diff
@@ -1031,41 +1036,83 @@ inline IDB_Decimal TreeNode::getDecimalVal()
...
-                fResult.decimalVal.value = (int64_t)roundl((fResult.longDoubleVal * pow((double)10.0, fResultType.scale)));
-                fResult.decimalVal.scale = fResultType.scale;
-                fResult.decimalVal.precision = fResultType.precision;
+                dlScaled = fResult.longDoubleVal * pow((double)10.0, fResultType.scale);
             }
+
+            if ((dlScaled > (long double)INT64_MAX) || (dlScaled < (long double)(INT64_MIN)))
+            {
+                int128_t temp = datatypes::Decimal::getInt128FromFloat128(
+                                    roundq(__float128(dlScaled)));
```
drrtuy: We surely will remove quadmath from the code b/c it brings additional dependency so we shouldn't use it.

tntnatbry: I have removed the `roundq` call from this block. The narrow decimal version uses a `lroundl` call however.

dbcon/execplan/treenode.h
tntnatbry: I have removed the quadmath dependency from this file.

drrtuy (approved):

abarkov (approved):
Looks ok. Thanks for addressing previous suggestions.

tntnatbry:
@abarkov I have pushed one more change to this PR since your last review, in dbcon/mysql/ha_mcs_execplan.cpp: In `buildAggregateColumn()`: If the aggregate function is a COUNT, we are setting the column type as BIGINT and `ct.scale = parm->resultType().scale;`. Since we added `idbassert(fResultType.scale == 0);` check in TreeNode::getDecimalVal(), we had an assertion failure if `parm->resultType().scale` is non-zero. Fix is to set `ct.scale` to 0 in `buildAggregateColumn()`.

abarkov:
@tntnatbry, I agree. COUNT is an integer result aggregate function, so setting scale to 0 rather than parm->resultType().scale sounds good.

drrtuy:
@tntnatbry Could you plz rebase the patch so I can proceed with merging?

tntnatbry:
@drrtuy Rebase is done. Please review.

drrtuy:
@tntnatbry Thanks a lot.

PR #1612, opened 2020-11-20

tntnatbry:
1. Make PredicateOperator::setOpType() function wide decimal aware.
  2. Added support for wide decimal in jlf_subquery.cpp::getColumnValue()
     used in scalar subqueries.
  3. Fixed the column index used for fetching wide decimal values from a Row
     when a wide decimal field is used in the order by clause.

### DIFF
--- a/dbcon/execplan/predicateoperator.cpp
+++ b/dbcon/execplan/predicateoperator.cpp
@@ -215,7 +215,9 @@ void PredicateOperator::setOpType(Type& l, Type& r)
             {
                 // should following the result type that MySQL gives
                 fOperationType = l;
-                fOperationType.scale = (l.scale > r.scale ? l.scale : r.scale);
+                fOperationType.scale = std::max(l.scale, r.scale);
+                fOperationType.precision = std::max(l.precision, r.precision);
+                fOperationType.colWidth = std::max(l.colWidth, r.colWidth);
                 break;
             }
 
@@ -229,7 +231,9 @@ void PredicateOperator::setOpType(Type& l, Type& r)
             case execplan::CalpontSystemCatalog::UBIGINT:
                 fOperationType.colDataType = execplan::CalpontSystemCatalog::DECIMAL;
                 fOperationType.scale = l.scale;
-                fOperationType.colWidth = 8;
+                fOperationType.precision = l.precision;
+                fOperationType.colWidth = (l.colWidth == datatypes::MAXDECIMALWIDTH) ?
+                    l.colWidth : 8;
                 break;
 
             default:
@@ -261,7 +265,9 @@ void PredicateOperator::setOpType(Type& l, Type& r)
             case execplan::CalpontSystemCatalog::UBIGINT:
                 fOperationType.colDataType = execplan::CalpontSystemCatalog::DECIMAL;
                 fOperationType.scale = r.scale;
-                fOperationType.colWidth = 8;
+                fOperationType.precision = r.precision;
+                fOperationType.colWidth = (r.colWidth == datatypes::MAXDECIMALWIDTH) ?
+                    r.colWidth : 8;
                 break;
 
             case execplan::CalpontSystemCatalog::LONGDOUBLE:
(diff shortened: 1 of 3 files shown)


dbcon/joblist/jlf_subquery.cpp
```diff
@@ -90,19 +90,32 @@ void getColumnValue(ConstantColumn** cc, uint64_t i, const Row& row, const strin
...
-            if (row.getScale(i) > 0)
+            if (row.getColumnWidth(i) == datatypes::MAXDECIMALWIDTH)
             {
-                if (data > 0)
-                    oss << "." << (data % IDB_pow[row.getScale(i)]);
-                else if (data < 0)
-                    oss << "." << (-data % IDB_pow[row.getScale(i)]);
+                int128_t val;
+                row.getInt128Field(i, val);
+                IDB_Decimal dec(0, row.getScale(i), row.getPrecision(i), val);
```
tntnatbry: I will remove the `row.getInt128Field()` in favour of `row.getTSInt128Field()` once the new IDB_Decimal ctor taking a `TSInt128` as an argument is available as part of https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/1610

tntnatbry:
This PR needs an update which is dependent on the merge of https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/1610.

dbcon/execplan/predicateoperator.cpp
```diff
@@ -216,6 +216,10 @@ void PredicateOperator::setOpType(Type& l, Type& r)
                 // should following the result type that MySQL gives
                 fOperationType = l;
                 fOperationType.scale = (l.scale > r.scale ? l.scale : r.scale);
+                fOperationType.precision = (l.precision > r.precision ?
```
drrtuy: Why not use std::max() here?

dbcon/joblist/jlf_subquery.cpp
```diff
@@ -90,19 +90,32 @@ void getColumnValue(ConstantColumn** cc, uint64_t i, const Row& row, const strin
...
+
+                oss << dec.toString(true);
```
drrtuy: I believe we don't need oss here.

dbcon/joblist/jlf_subquery.cpp
drrtuy: You can even use a ctor with 3 arguments that doesn't set the int64 value in the Decimal

dbcon/joblist/jlf_subquery.cpp
```diff
@@ -90,19 +90,32 @@ void getColumnValue(ConstantColumn** cc, uint64_t i, const Row& row, const strin
...
+                if (row.getScale(i) > 0)
+                {
+                    if (data > 0)
+                        oss << "." << (data % IDB_pow[row.getScale(i)]);
+                    else if (data < 0)
+                        oss << "." << (-data % IDB_pow[row.getScale(i)]);
+                }
+
+                *cc = new ConstantColumn(oss.str(),
+                                         IDB_Decimal(data, row.getScale(i), row.getPrecision(i)));
```
drrtuy: Ohh, I didn't realize what you are doing here.
I've moved dataconvert::decimalToString into Decimal::toString() so it can print out both narrow and wide decimals and you need to only call Decimal ctor with appropriate arguments and call Decimal::toString() of the resulting instance calling ConstantColumn ctor.

drrtuy:
Good job. 
I don't see other issues with the PR so we can proceed after the comments will be addressed.

dbcon/execplan/predicateoperator.cpp
tntnatbry: Good point. Done.

dbcon/joblist/jlf_subquery.cpp
tntnatbry: Done.

dbcon/joblist/jlf_subquery.cpp
tntnatbry: I agree. I am now constructing the IDB_Decimal object and calling the narrow decimal version of dec.toString().

tntnatbry:
Comments are addressed. There is one change related to the use of `row.getInt128Field(i, val);` in this patch. This will be addressed once either PR https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/1618 or PR https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/1610 is merged. Until then, please don't merge this PR.

drrtuy (approved):

PR #1617, opened 2020-11-25

tntnatbry:
1. In TupleAggregateStep::configDeliveredRowGroup(), use
jobInfo.projectionCols instead of jobInfo.nonConstCols
for setting scale and precision if the source column is
wide decimal.

2. Tighten rules for wide decimal processing. Specifically:
  a. Replace (precision > INT64MAXPRECISION) checks with
     (precision > INT64MAXPRECISION && precision <= INT128MAXPRECISION)
  b. At places where (colWidth == MAXDECIMALWIDTH) is not enough to
     determine if a column is wide decimal or not, also add a check on
     type being DECIMAL/UDECIMAL.

### DIFF
--- a/datatypes/mcs_decimal.h
+++ b/datatypes/mcs_decimal.h
@@ -598,16 +598,14 @@ class VDecimal: public TSInt128
 
     bool operator==(const VDecimal& rhs) const
     {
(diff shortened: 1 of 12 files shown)


abarkov (approved):
The patch looks OK for me. Thanks.

PR #1618, opened 2020-11-25
Mcol 4409 2

drrtuy:


utils/funcexp/func_from_unixtime.cpp
```diff
@@ -67,10 +67,9 @@ DateTime getDateTime(rowgroup::Row& row,
 
             if (parm[0]->data()->resultType().colWidth == datatypes::MAXDECIMALWIDTH)
             {
-                int128_t scaleDivisor;
-                datatypes::getScaleDivisor(scaleDivisor, dec.scale);
-                val = datatypes::Decimal::getInt64FromWideDecimal(dec.s128Value / scaleDivisor);
-                msec = datatypes::Decimal::getUInt32FromWideDecimal(dec.s128Value % scaleDivisor);
+                auto integralAndFractional = dec.getIntegralAndFractional<datatypes::TSInt128>();
+                val = static_cast<int64_t>(integralAndFractional.first);
+                msec = static_cast<uint32_t>(integralAndFractional.second);
```
abarkov, line 72: This part looks suspicious. getIntegralAndFractional() divides the fractional part to scale:

```
+    inline std::pair<TSInt128, D> getIntegralAndFractional() const
+    {
+        int128_t scaleDivisor;
+        getScaleDivisor(scaleDivisor, scale);
+        return std::make_pair(TSInt128(s128Value / scaleDivisor),
+                              D((typename get_integral_type<D>::type)(s128Value % scaleDivisor) / scaleDivisor));
+    }
```
Won't msec  always be zero?

It seems we need two separate methods:
getIntegralAndFractional() - without division
getIntegralAndDividedFractional() - with division

datatypes/mcs_decimal.h
```diff
@@ -596,6 +526,71 @@ class VDecimal: public TSInt128
...
+    }
+
+    inline TSInt128 getIntegralPart() const
+    {
+        int128_t scaleDivisor = 0;
+        int128_t result = 0;
+        if(LIKELY(utils::is_nonnegative(scale)))
+        {
+            getIntegralPart(scaleDivisor, result);
+        }
```
abarkov: I prefer `getScaleDivisor()` and `getIntegralPart(scale)`  to be non-void functions returning the result (instead of having the result as an output parameter), so we can write the code as short as this:

```
if(LIKELY(utils::is_nonnegative(scale)))
    return TSInt128(getIntegralPart(getScaleDivisor()));
return TSInt128(getIntegralPartNegativeScale(getScaleDivisor()));
```
Also, perhaps rename the low level `getIntergralPart()` to `getIntegralPartNonNegativeScale()`

datatypes/mcs_float128.h
```diff
@@ -66,6 +85,40 @@ class TFloat128
...
+    inline int64_t toTSInt64() const
+    {
+      if (value > static_cast<__float128>(INT64_MAX))
+          return INT64_MAX;
+      else if (value < static_cast<__float128>(INT64_MIN))
+          return INT64_MIN;
+
+      return static_cast<int64_t>(value);
+    }
+
```
abarkov, line 97: toTSInt64() returns int64_t. I think this method should be renamed somehow. We'll eventually have classes TSInt32 / TUInt32 / TSInt64 / TUInt64. So methods like toTSInt64() will return instances of these classes rather than trivial types like int64_t.

I suggest either of these:
- Introduce these classes right now and have toTSInt64() return TSInt64()
- Or rename toTSInt64() to something else
- Or simply move this code inside operator int64_t.

datatypes/mcs_float128.h
```diff
@@ -66,6 +85,40 @@ class TFloat128
...
+          return 0;
+
+      return static_cast<uint64_t>(value);
+    }
+
+    inline operator uint64_t() const
+    {
+      return toTUInt64();
+    }
+
```
abarkov, line 117: The previous comments also apply to toTUint64().

datatypes/mcs_int128.cpp
```diff
@@ -117,20 +117,17 @@ namespace datatypes
...
-          return INT64_MAX;
-      else if (s128Value < static_cast<int128_t>(INT64_MIN))
-          return INT64_MIN;
+    if (s128Value > static_cast<int128_t>(INT32_MAX))
+        return INT32_MAX;
+    else if (s128Value < static_cast<int128_t>(INT32_MIN))
+        return INT32_MIN;
 
-      return static_cast<int64_t>(s128Value);
+    return static_cast<int32_t>(s128Value);
```
abarkov: Same here: this method returns int32_t rather than TSInt32.

Also, why this code is inside *.cpp file rather than *.h?  It should be in *h, to be inlined.

datatypes/mcs_int128.h
```diff
@@ -200,11 +200,40 @@ class TSInt128
       return static_cast<long double>(s128Value);
     }
 
+    int32_t toTSInt32() const;
+    inline operator int32_t() const
```
abarkov, line 203: Same here:
- This method returns int32_t rather than TSInt32
- Why the actual code is inside *.cpp rather than *.h ?

datatypes/mcs_int128.h
```diff
@@ -200,11 +200,40 @@ class TSInt128
...
+    {
+      return toTSInt64();
+    }
+
+    uint64_t toTUInt64() const;
+    inline operator uint64_t() const
+    {
+      return toTUInt64();
+    }
+
```
abarkov, line 242: Same comment for all these methods ^^^^

dbcon/execplan/treenode.h
```diff
@@ -61,8 +61,17 @@ class IDB_Decimal: public datatypes::VDecimal
...
+              int8_t s,
+              uint8_t p,
+              const int128_t &val128 = 0)
+    : VDecimal(val, s, p, val128)
+  { }
+   IDB_Decimal(const TSInt128& val128,
+               int8_t s,
+               uint8_t p)
+        : VDecimal(val128, s, p)
+  { }
```
abarkov: I find it confusing that in MCS we pass attributes in this order: scale, precision.
- Server passes the other way around: precision, scale.
- SQL defines the data type also the other way around:  DECIMAL(20,2), where 20 precision and 2 is scale.

If we're doing massive code changes anyway, I suggest to use the natural attribute order in new classes and methods, so we'll gradually get rid of the old order.

utils/funcexp/func_ceil.cpp
```diff
@@ -592,6 +575,10 @@ IDB_Decimal Func_ceil::getDecimalVal(Row& row,
                     // decimal point so that we return the largest integer value not less than X.
                     if ((tmp - (ret.s128Value * scaleDivisor)) > 0)
                         ret.s128Value += 1;
+*/
+                    ret = IDB_Decimal(ret.getRoundedIntegralPart(),
+                                      ret.scale,
+                                      ret.precision);
```
abarkov, line 570: Why did you put the old code into a comment rather than just delete it?

utils/funcexp/func_char.cpp
```diff
@@ -154,15 +156,18 @@ string Func_char::getStrVal(Row& row,
                         tmpval++;
 
                     value = datatypes::Decimal::getInt32FromWideDecimal(tmpval);
+*/
+                    // rounding by the left over
+                    value = static_cast<int32_t>(d.getRoundedIntegralPart(roundingFactor));
```
abarkov, line 145: Same here, why not just delete the old code? Will it be needed later?

utils/funcexp/func_floor.cpp
```diff
@@ -160,7 +160,7 @@ int64_t Func_floor::getIntVal(Row& row,
 
             if (op_ct.colWidth == datatypes::MAXDECIMALWIDTH)
             {
-                ret = datatypes::Decimal::getInt64FromWideDecimal(tmp.s128Value);
+                ret = static_cast<int64_t>(tmp.toTSInt128());
```
abarkov, line 163: I prefer this:

`return static_cast<int64_t>(tmp);`

Reasoning:
- We have the conversion operator defined already. Why not use it?
- There are no reasons to break the switch and do "return" immediately after the switch. Just return right here.

utils/funcexp/func_maketime.cpp
```diff
@@ -212,22 +182,7 @@ string Func_maketime::getStrVal(rowgroup::Row& row,
...
-                int128_t lefto = (d.s128Value - tmpval * scaleDivisor) / scaleDivisor2;
-
-                if (tmpval >= 0 && lefto > 4)
-                    tmpval++;
-
-                if (tmpval < 0 && lefto < -4)
-                    tmpval--;
-
-                sec = datatypes::Decimal::getInt64FromWideDecimal(tmpval);
+                sec = static_cast<int64_t>(d.getPosNegRoundedIntegralPart(4));
```
abarkov, line 185: Great job! So much duplicate code removed. I like this very much.

utils/funcexp/func_mod.cpp
```diff
@@ -237,12 +237,12 @@ double Func_mod::getDoubleVal(Row& row,
...
-                    int128_t value = d.s128Value / scaleDivisor;
-                    int128_t lefto = d.s128Value % scaleDivisor;
-                    datatypes::TFloat128 tmp((__float128) (value % div) + (__float128) lefto / scaleDivisor);
-                    mod = static_cast<double>(tmp);
+                    auto intAndFract = d.getIntegralAndFractional<datatypes::TFloat128>();
+
+                    datatypes::TSInt128 integralRemainder = intAndFract.first % div;
+                    mod = static_cast<double>(integralRemainder.toTFloat128() +
+                                                   intAndFract.second);
+
```
abarkov: I suggest to move this code to a new private/protected method in Func_mod. Say:

`IDB_Decimal doDecimal();`

So here you can just do:

`return static_cast<double>(doDecimal())`

utils/funcexp/func_mod.cpp
```diff
@@ -360,12 +360,11 @@ long double Func_mod::getLongDoubleVal(Row& row,
...
-                    datatypes::getScaleDivisor(scaleDivisor, d.scale);
-                    int128_t value = d.s128Value / scaleDivisor;
-                    int128_t lefto = d.s128Value % scaleDivisor;
-                    datatypes::TFloat128 tmp((__float128) (value % div) + (__float128) lefto / scaleDivisor);
-                    mod = static_cast<long double>(tmp);
+                    auto intAndFract = d.getIntegralAndFractional<datatypes::TFloat128>();
+
+                    datatypes::TSInt128 integralRemainder = intAndFract.first % div;
+                    mod = static_cast<long double>(integralRemainder.toTFloat128() +
+                                                   intAndFract.second);
```
abarkov: `return static_cast<long double>(doDecimal());`

utils/funcexp/func_mod.cpp
```diff
@@ -486,12 +485,10 @@ int64_t Func_mod::getIntVal(Row& row,
...
-                    int128_t scaleDivisor;
-                    datatypes::getScaleDivisor(scaleDivisor, d.scale);
-                    int128_t value = d.s128Value / scaleDivisor;
-                    int128_t lefto = d.s128Value % scaleDivisor;
-                    __float128 tmp = (__float128) (value % div) + (__float128) lefto / scaleDivisor;
-                    mod = datatypes::Decimal::getInt64FromFloat128(tmp);
+                    auto intAndFract = d.getIntegralAndFractional<datatypes::TFloat128>();
+                    datatypes::TSInt128 integralRemainder = intAndFract.first % div;
+                    mod = static_cast<int64_t>(integralRemainder.toTFloat128() +
+                                               intAndFract.second);
```
abarkov: `return static_cast<int64_t>(doDecimal());`

utils/funcexp/func_mod.cpp
```diff
@@ -603,12 +600,11 @@ uint64_t Func_mod::getUIntVal(Row& row,
...
-                    datatypes::getScaleDivisor(scaleDivisor, d.scale);
-                    int128_t value = d.s128Value / scaleDivisor;
-                    int128_t lefto = d.s128Value % scaleDivisor;
-                    __float128 tmp = (__float128) (value % div) + (__float128) lefto / scaleDivisor;
-                    mod = datatypes::Decimal::getUInt64FromFloat128(tmp);
+                    auto intAndFract = d.getIntegralAndFractional<datatypes::TFloat128>();
+                    datatypes::TSInt128 integralRemainder = intAndFract.first % div;
+                    mod = static_cast<uint64_t>(integralRemainder.toTFloat128() +
+                                               intAndFract.second);
+
```
abarkov: `return static_cast<uint64_t>(doDecimal());`

abarkov: by the way, why aren't you fixing `Func_mod::getDecimalVal()`. It has exactly the same piece of code in here:
```
        // special case integer division
        else if (div.scale == 0)
        {
            int128_t scaleDivisor;
            datatypes::getScaleDivisor(scaleDivisor, d.scale);
            int128_t value = dividendInt / scaleDivisor;
            int128_t lefto = dividendInt % scaleDivisor;
            int128_t mod = (value % divInt) * scaleDivisor + lefto;
            retValue.s128Value = mod;
        }
```
it can be replaced to `return doDecimal();`

utils/funcexp/func_round.cpp
```diff
@@ -133,18 +133,7 @@ int64_t Func_round::getIntVal(Row& row,
...
-                x.s128Value /= 10;
-        }
-        else
-        {
-            while (x.scale++ < 0)
-                x.s128Value *= 10;
-        }
-
-        return datatypes::Decimal::getInt64FromWideDecimal(x.s128Value);
+        return static_cast<int64_t>(x.getIntegralPart());
```
abarkov, line 136: I find the old code strange. It actually truncates, not rounds. So does your new code. Why is that?

abarkov (changes requested):

dbcon/execplan/treenode.h
drrtuy: This change will be too massive and just for nothing. I won't be doing it now.

drrtuy: But I agree this looks more natural.

utils/funcexp/func_ceil.cpp
drrtuy: Missed that part.

utils/funcexp/func_floor.cpp
drrtuy: This calls for an unrelated change in the else branch that I don't want to do.

utils/funcexp/func_mod.cpp
drrtuy: Good point. I will deliver this outside the scope of this patch.

utils/funcexp/func_mod.cpp
drrtuy: Will do the latter.

utils/funcexp/func_round.cpp
drrtuy: I stick to whatever code we have. I believe the investigation should go as a separate issue that I'll create.

utils/funcexp/func_floor.cpp
drrtuy: I would say that this behavior looks very strange we returning actual int128 value not looking at its scale.  That is why we casting tmp.toTSInt128() to int64 in the first place. I believe that this calls for additional investigation outside the scope of this issue.

utils/funcexp/func_mod.cpp
drrtuy: Implemented both suggestion.

utils/funcexp/func_mod.cpp
drrtuy: Done.

utils/funcexp/func_mod.cpp
drrtuy: Done.

utils/funcexp/func_mod.cpp
drrtuy: Done.

utils/funcexp/func_from_unixtime.cpp
drrtuy: Done.

utils/funcexp/func_char.cpp
drrtuy: Removed.

utils/funcexp/func_ceil.cpp
drrtuy: Removed.

datatypes/mcs_int128.h
drrtuy: Removed.

datatypes/mcs_float128.h
drrtuy: Renamed methods into a set of operatorX methods.

datatypes/mcs_decimal.h
drrtuy: Done.

datatypes/mcs_int128.h
drrtuy: Done.

datatypes/mcs_decimal.h
```diff
@@ -375,31 +375,43 @@ class Decimal: public TSInt128
...
             getScaleDivisor(scaleDivisor, scale);
             return std::make_pair(TSInt128(s128Value / scaleDivisor),
-                                  D((typename get_integral_type<D>::type)(s128Value % scaleDivisor) / scaleDivisor));
+                                  TFloat128((__float128)(s128Value % scaleDivisor) / scaleDivisor));
+        }
+
+        // This method returns integral part as a TSInt128 and
+        // fractional part as a TFloat128
+        inline std::pair<TSInt128, TSInt128> getIntegralAndFractional() const
+        {
```
abarkov, line 391: The comment does not match. Should be "fractional part as a TSInt128".

datatypes/mcs_decimal.h
```diff
@@ -439,6 +449,20 @@ class Decimal: public TSInt128
...
+            if (!isScaled())
+            {
+                return s128Value % div.getValue();
+            }
+            // Scale the value and calculate
+            // (LHS.value % RHS.value) * LHS.scaleMultiplier + LHS.scale_div_remainder
+            auto integralFractionalDivisor = getIntegralFractionalAndDivisor();
+            return (std::get<0>(integralFractionalDivisor) % div.getValue()) * std::get<2>(integralFractionalDivisor) + std::get<1>(integralFractionalDivisor);
+        }
+
```
abarkov, line 465: It works differently in MariaDB vs ColumnStore. Not sure what to do with that. I'd fix it at once
```
DROP TABLE IF EXISTS t1;
CREATE TABLE t1 (a DECIMAL(10,2)) ENGINE=MyISAM;
INSERT INTO t1 VALUES (1.3);
SELECT a % 1.3 FROM t1;
```
```
+---------+
| a % 1.3 |
+---------+
|    0.00 |
+---------+
```
```
DROP TABLE IF EXISTS t1;
CREATE TABLE t1 (a DECIMAL(10,2)) ENGINE=ColumnStore;
INSERT INTO t1 VALUES (1.3);
SELECT a % 1.3 FROM t1;
```
```
+---------+
| a % 1.3 |
+---------+
|    0.30 |
+---------+
```

datatypes/mcs_int128.h
abarkov: thanks for addressing this.

datatypes/mcs_int128.h
```diff
@@ -200,45 +200,81 @@ class TSInt128
...
     }
 
-    int32_t toTSInt32() const;
     inline operator int32_t() const
     {
-      return toTSInt32();
+      if (s128Value > static_cast<int128_t>(INT32_MAX))
+          return INT32_MAX;
+      else if (s128Value < static_cast<int128_t>(INT32_MIN))
+          return INT32_MIN;
```
abarkov, line 208: The "else" keyword is not really needed here. There are more similar redundant "else"s in the patch.

utils/funcexp/func_from_unixtime.cpp
```diff
@@ -67,7 +67,7 @@ DateTime getDateTime(rowgroup::Row& row,
 
             if (parm[0]->data()->resultType().colWidth == datatypes::MAXDECIMALWIDTH)
             {
-                auto integralAndFractional = dec.getIntegralAndFractional<datatypes::TSInt128>();
+                auto integralAndFractional = dec.getIntegralAndFractional();
```
abarkov, line 70: I think this code is wrong. Both the narrow decimal and the wide decimal branch:
```
DROP TABLE IF EXISTS t1;
CREATE TABLE t1 (a DECIMAL(10,7)) ENGINE=ColumnStore;
INSERT INTO t1 VALUES (1.3);
SELECT FROM_UNIXTIME(a) FROM t1;
+----------------------------+
| FROM_UNIXTIME(a)           |
+----------------------------+
| 1970-01-01 04:00:01.902848 |
+----------------------------+
```

The correct result is:
```
DROP TABLE IF EXISTS t1;
CREATE TABLE t1 (a DECIMAL(10,7)) ENGINE=MyISAM;
INSERT INTO t1 VALUES (1.3);
SELECT FROM_UNIXTIME(a) FROM t1;
+----------------------------+
| FROM_UNIXTIME(a)           |
+----------------------------+
| 1970-01-01 04:00:01.300000 |
+----------------------------+
```

The problem is that the code does not take into account how the scale relates to 6 (which is the divisor  for microsecords).
The code is correct when scale is exactly 6.
But it's wrong for other scale values.

utils/funcexp/func_from_unixtime.cpp
abarkov: Thanks. Unfortunately, there are more problems in this code. See the new comment.

utils/funcexp/functor_real.h
```diff
@@ -427,6 +427,35 @@ class Func_mod : public Func_Real
...
+                auto intAndFract = d.getIntegralAndDividedFractional();
+
+                datatypes::TSInt128 integralRemainder = intAndFract.first % div;
+                return static_cast<ModType>(integralRemainder.toTFloat128() +
+                                               intAndFract.second);
+
+            }
+        }
+        int64_t value = d.value / pow(10.0, d.scale);
+        return value % div;
```
abarkov, line 456: `dec1 % dec2`  should work as follows:
- Adjust both dec1 and dec2 to the same scale - the maximum one.
- Perform % in integer format
- Adjust the result according to the function result scale.

abarkov: The same logic should apply to `int MOD dec` and `dec MOD int`.

abarkov (changes requested):

utils/funcexp/functor_real.h
drrtuy: Agree but this goes out of the scope.

utils/funcexp/func_from_unixtime.cpp
drrtuy: Agree but this goes out of the scope.

drrtuy: I'll create an issue for this.

PR #1619, opened 2020-11-27

tntnatbry:
1. In BatchPrimitiveProcessorJL::countThisMsg(), exit early if the
return code from primproc is non-zero.
2. For statistical window functions (stddev/var), properly handle
the case when there is only one value for the calculation.

### DIFF
--- a/dbcon/joblist/batchprimitiveprocessor-jl.cpp
+++ b/dbcon/joblist/batchprimitiveprocessor-jl.cpp
@@ -703,6 +703,12 @@ bool BatchPrimitiveProcessorJL::countThisMsg(messageqcpp::ByteStream& in) const
     const uint8_t* data = in.buf();
     uint32_t offset = sizeof(ISMPacketHeader) + sizeof(PrimitiveHeader); // skip the headers
 
+    ISMPacketHeader* hdr = (ISMPacketHeader*)(data);
+
+    // Exit early if PrimProc has thrown an exception
+    if (hdr->Status > 0)
+        return true;
+
     if (_hasScan)
     {
         if (data[offset] != 0)
--- a/utils/windowfunction/wf_stats.cpp
+++ b/utils/windowfunction/wf_stats.cpp
@@ -189,8 +189,7 @@ void WF_stats<T>::operator()(int64_t b, int64_t e, int64_t c)
             fCount++;
         }
 
-        if ((fCount > 0) &&
-                !(fCount == 1 && (fFunctionId == WF__STDDEV_SAMP || fFunctionId == WF__VAR_SAMP)))
+        if (fCount > 1)
         {
             int scale = fRow.getScale(colIn);
             long double factor = pow(10.0, scale);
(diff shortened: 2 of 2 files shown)


dbcon/joblist/batchprimitiveprocessor-jl.cpp
```diff
@@ -703,6 +703,12 @@ bool BatchPrimitiveProcessorJL::countThisMsg(messageqcpp::ByteStream& in) const
     const uint8_t* data = in.buf();
     uint32_t offset = sizeof(ISMPacketHeader) + sizeof(PrimitiveHeader); // skip the headers
 
+    ISMPacketHeader* hdr = (ISMPacketHeader*)(data);
+
+    // Exit early if PrimProc has thrown an exception
+    if (hdr->Status > 0)
+        return true;
```
drrtuy, line 710: This is a very good change but you should also log a message explaining that PP throws an exception running a primitive.

utils/windowfunction/wf_stats.cpp
```diff
@@ -225,9 +224,17 @@ void WF_stats<T>::operator()(int64_t b, int64_t e, int64_t c)
     {
         setValue(CalpontSystemCatalog::DOUBLE, b, e, c, (double*) NULL);
     }
-    else if (fCount == 1 && (fFunctionId == WF__STDDEV_SAMP || fFunctionId == WF__VAR_SAMP))
+    else if (fCount == 1)
     {
-        setValue(CalpontSystemCatalog::DOUBLE, b, e, c, (double*) NULL);
+        if (fFunctionId == WF__STDDEV_SAMP || fFunctionId == WF__VAR_SAMP)
```
drrtuy, line 229: I believe we are fixing something that is outside of the scope of 4188 here or am I wrong?

dbcon/joblist/batchprimitiveprocessor-jl.cpp
tntnatbry: This check is added because the below assertion in this function `idbassert(in.length() > offset);` was failing when the buffer contained an error message. We need to skip the rest of this function execution in case of an error.

The error logging is handled by the calling function TupleBPS::receiveMultiPrimitiveMessages. Specifically this code block:

```
                if (bs->length() == 0 || hdr->Status > 0)
                {
                    /* PM errors mean this should abort right away instead of draining the PM backlog */
                    tplLock.lock();

                    if (bs->length() == 0)
                    {
                        errorMessage(IDBErrorInfo::instance()->errorMsg(ERR_PRIMPROC_DOWN));
                        status(ERR_PRIMPROC_DOWN);
                    }
                    else
                    {
                        string errMsg;

                        bs->advance(sizeof(ISMPacketHeader) + sizeof(PrimitiveHeader));
                        *bs >> errMsg;
                        status(hdr->Status);
                        errorMessage(errMsg);
                    }

                    abort_nolock();
                    goto out;
                }
```

utils/windowfunction/wf_stats.cpp
tntnatbry: This change is fixing a regression in working_tpch1_compareLogOnly/windowFunctions/MCOL-3423.sql. I did not see this test failing in develop prior to the MCOL-641 merge. So it's possible some change done to this file as part of MCOL-641 introduced this regression. Now, we are handling `fCount == 0` and `fCount == 1` here as it is handled in the non-window function versions of these functions in `RowAggregationUM::calculateStatisticsFunctions`

dbcon/joblist/batchprimitiveprocessor-jl.cpp
drrtuy: If PP has thrown an exception there will be no error message from PP and it looks like an assertion in ExeMgr w/o any note about PP. That is why I asked about an explicit message.

tntnatbry: This is not true. PP sends the error info (code and message) to ExeMgr in case an exception is raised by PP. Refer to the catch blocks in `BatchPrimitiveProcessor::execute()`. These blocks call `writeErrorMsg` function which log the error info to bytestream, which are then read by the ExeMgr on the receiving side.

drrtuy: I suggest you to try to throw an exception from BatchPrimitiveProcessor::execute() and see how the logging output helps you to understand that PP throws.

drrtuy: I believe that was a good moment to finally start logging something usefull in case of PP throw. I believe I should do it on my own.

drrtuy (approved):

PR #1638, opened 2020-12-02
MCOL-2055 Fix. Flush PrimProc Cache during batchinserts

jmrojas2332:
From Ticket: insert..select inserts shorter strings when columnstore_use_import_for_batchinsert is disabled, for data types BLOCK LONGBLOB TEXT LONGTEXT

After investigating, I discovered the first block of the dictionary writeengineserver writes to during the insert/select is not found when trying to retrieve the newly inserted value(s). Restarting DMLProc after the insert/select would fix this issue due to a primproc cache flush.

This fix flushes the dict oids from the primproc cache after the batch insert, which allows retrieval of the entire text/blob without having to restart DMLProc.

### DIFF
--- a/writeengine/wrapper/writeengine.cpp
+++ b/writeengine/wrapper/writeengine.cpp
@@ -1681,6 +1681,13 @@ int WriteEngineWrapper::insertColumnRecs(const TxnID& txnid,
         // Write row(s) to database file(s)
         //----------------------------------------------------------------------
         rc = writeColumnRec(txnid, cscColTypeList, colStructList, colOldValueList, rowIdArray, newColStructList, colNewValueList, tableOid, useTmpSuffix); // @bug 5572 HDFS tmp file
+
+        if (rc == NO_ERROR)
+        {
+            rc = cacheutils::flushPrimProcCache();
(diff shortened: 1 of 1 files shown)


writeengine/wrapper/writeengine.cpp
```diff
@@ -1681,6 +1681,13 @@ int WriteEngineWrapper::insertColumnRecs(const TxnID& txnid,
         // Write row(s) to database file(s)
         //----------------------------------------------------------------------
         rc = writeColumnRec(txnid, cscColTypeList, colStructList, colOldValueList, rowIdArray, newColStructList, colNewValueList, tableOid, useTmpSuffix); // @bug 5572 HDFS tmp file
+
+        if (rc == NO_ERROR)
+        {
+            rc = cacheutils::flushPrimProcCache();
```
drrtuy: I wonder if you understand how heavy is the operation for a big caches?

jmrojas2332: I understand your concern. I have updated to only flush the dictionary oids from the cache

davidhall586 (approved):

PR #1642, opened 2020-12-03
MCOL-2000 fix up bulk import

davidhall586:
With the changes to field length imposed by MCOL-2000, the length bytes for bulk insert require a slightly different calculation.


mariadb-AlexeyAntipovsky (approved):

drrtuy (changes requested):
We shouldn't ask our internal syscat about how many bytes the length of the buffer sent by MDB is.

drrtuy (approved):

PR #1644, opened 2020-12-03
Trim up Decimal comparison

drrtuy:


datatypes/mcs_decimal.h
```diff
@@ -441,226 +442,93 @@ class Decimal: public TSInt128
...
+                }
                 else
-                    return (datatypes::Decimal::compare(*this, rhs) == 0);
+                {
+                    //const_cast<Decimal&>(rhs).s128Value = rhs.value;
+                    return op64(datatypes::Decimal::compare(*this, 
+                                                            Decimal(scale,
+                                                                    precision,
+                                                                    (int128_t) rhs.value)),
+                                0);
```
abarkov: Should not it be `Decimal(rhs.scale, rhs.precision, (int128_t) rhs.value))` here?

abarkov (changes requested):

abarkov (approved):

PR #1651, opened 2020-12-04
Fix for a behavior change + additional logging

drrtuy:


abarkov (approved):

tntnatbry:
This change is causing failure of working_tpch1_compareLogOnly/sub/bug3780.sql. Here is an example query from the test:

```sql
MariaDB [tpch1]> select * from customer where c_nationkey = (select n_nationkey from nation where n_regionkey = c_nationkey) and c_custkey < 30;
ERROR 1815 (HY000): Internal error: st: 1 TupleBPS::receiveMultiPrimitiveMessages() caught  an exception originally thrown by PrimProc:
```

Actual returned output should be:

```sql
ERROR 1815 (HY000) at line 1: Internal error: IDB-3002: Subquery returns more than 1 row.
```

drrtuy:
The change of the behavior when countThisMsg suddenly starts to return
instead of throwing an exception looks more dangerous from my perspective.

ср, 9 дек. 2020 г. в 19:42, Gagan Goel <notifications@github.com>:

drrtuy:
But the case should be researched more if it produces a behavior difference.

ср, 9 дек. 2020 г. в 20:30, Roland Noland <drrtuy@gmail.com>:

PR #1653, opened 2020-12-04
Leverage MDB Field methods to output char or binary-based types when …

drrtuy:
…processing batch inserts

### DIFF
--- a/dbcon/mysql/ha_mcs_dml.cpp
+++ b/dbcon/mysql/ha_mcs_dml.cpp
@@ -884,93 +884,6 @@ int ha_mcs_impl_write_batch_row_(const uchar* buf, TABLE* table, cal_impl_if::ca
 
                     break;
                 }
-
-                case CalpontSystemCatalog::CHAR:
-                {
-                    uint32_t colWidthInBytes = ci.columnTypes[colpos].colWidth;
-
-                    if (nullVal && (ci.columnTypes[colpos].constraintType != CalpontSystemCatalog::NOTNULL_CONSTRAINT))
-                    {
-                        fprintf(ci.filePtr, "%c", ci.delimiter);
-                    }
-                    else
-                    {
-                        if (current_thd->variables.sql_mode & MODE_PAD_CHAR_TO_FULL_LENGTH)
-                        {
-                            // Pad to the full length of the field
-                            escape.assign((char*)buf, colWidthInBytes);
-
-                            boost::replace_all(escape, "\\", "\\\\");
-
-                            fprintf(ci.filePtr, "%c%.*s%c%c", ci.enclosed_by, (int)escape.length(),
-                                    escape.c_str(), ci.enclosed_by, ci.delimiter);
-                        }
-                        else
-                        {
-                            // Get the actual data length
-                            Field* field = table->field[colpos];
-                            bitmap_set_bit(table->read_set, field->field_index);
-                            String attribute;
-                            field->val_str(&attribute);
-
-                            escape.assign((char*)buf, attribute.length());
-                            boost::replace_all(escape, "\\", "\\\\");
-
-                            fprintf(ci.filePtr, "%c%.*s%c%c", ci.enclosed_by, (int)escape.length(),
-                                    escape.c_str(), ci.enclosed_by, ci.delimiter);
-                        }
-                    }
-
-                    buf += colWidthInBytes;
-
-                    break;
-                }
-
-                case CalpontSystemCatalog::VARCHAR:
-                {
-                    uint32_t colWidthInBytes = ci.columnTypes[colpos].colWidth;
-
-                    if (nullVal && (ci.columnTypes[colpos].constraintType != CalpontSystemCatalog::NOTNULL_CONSTRAINT))
-                    {
-                        fprintf(ci.filePtr, "%c", ci.delimiter);
-
-                        if (colWidthInBytes < 256)
-                        {
-                            buf++;
-                        }
-                        else
-                        {
-                            buf = buf + 2 ;
-                        }
-                    }
-                    else
-                    {
-                        // Maximum number of bytes allowed for a VARCHAR
-                        // field is 65532, so the max length fits in 2 bytes.
-                        // dataLength is length in bytes, not length in chars
-                        uint16_t dataLength = 0;
-
-                        if (colWidthInBytes < 256)
-                        {
-                            dataLength = *(uint8_t*) buf;
-                            buf++;
-                        }
-                        else
-                        {
-                            dataLength = *(uint16_t*) buf;
-                            buf = buf + 2 ;
-                        }
-
-                        escape.assign((char*)buf, dataLength);
-                        boost::replace_all(escape, "\\", "\\\\");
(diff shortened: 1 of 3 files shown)


dbcon/mysql/ha_mcs_dml.cpp
```diff
@@ -1659,134 +1572,44 @@ int ha_mcs_impl_write_batch_row_(const uchar* buf, TABLE* table, cal_impl_if::ca
...
                         {
-                            dataLength = *(uint8_t*) buf;
-                            buf++;
+                            char* ucharptr = value.c_ptr();
+                            for (uint32_t i = 0; i < value.length(); i++)
+                            {
+                                fprintf(ci.filePtr, "%02x", *(uint8_t*)ucharptr);
+                                    ucharptr++;
+                            }
+                            fprintf(ci.filePtr, "%c", ci.delimiter);
```
abarkov, line 1599: - The variable name "ucharptr" implies it's "a pointer to unsigned char". While in fact it's a "a pointer to signed char". For me it sounds confusing.

- Calling value.ptr() instead of value.c_ptr() should be OK here, to avoid realloc.

- There's no a need to increment two variables: i and ucharptr. Rewriting the code to increment only one variable should be better.

- I'd put this loop into a function. It looks too generic and useful in other cases. It will be needed for BINARY as well, when we fix it.

Collecting all together, I suggest this method in "ci":
```
void fprintBinary(const char *str, size_t length)
{
  for (const char *end= str + length ; ptr < end; ptr++)
    fprintf(filePtr, "%02x", *(uint8_t*) ptr);
}
```
and the call will be:
```
  ci.fprintBinary(value.ptr(), value.length());
```

dbcon/mysql/ha_mcs_dml.cpp
```diff
@@ -1659,134 +1572,44 @@ int ha_mcs_impl_write_batch_row_(const uchar* buf, TABLE* table, cal_impl_if::ca
...
-                    if (isBlob)
-                    {
-                        for (uint32_t i = 0; i < dataLength; i++)
-                        {
-                            fprintf(ci.filePtr, "%02x", *(uint8_t*)ucharptr);
-                            ucharptr++;
+                            escape.assign(value.ptr(), value.length());
+                            boost::replace_all(escape, "\\", "\\\\");
+                            fprintf(ci.filePtr, "%c%.*s%c%c", ci.enclosed_by, (int)escape.length(),
+                                escape.c_str(), ci.enclosed_by, ci.delimiter);
```
abarkov, line 1606: Double copy happens here:
- From value to escape on "assign()".
- To a new buffer inside boost::replace_all()

As discussed earlier in slack, I think we need a method in "ci" like this:
```
void fprintEscaped(const char *str, size_t length)
{
  fprintf(filePtr, "%c", enclosed_by);
  for (const char *end= str + length; str < end; str++)
  {
    if (*str == '\\')
      fprintf(filePtr, "%c", '\\');
    fprintf(filePtr, "%c", *str);
  }
  fprintf(filePtr, "%c%c", enclosed_by, delimiter);
}
```

For now, it can still escape only backslash as before.
But later we'll need to escape more characters, as cpimport currently fails when the string has characters 0x07, 0x0A and 0x11.

dbcon/mysql/ha_mcs_impl.cpp
```diff
@@ -3669,8 +3677,8 @@ int ha_mcs_impl_end_bulk_insert(bool abort, TABLE* table)
              ((thd->lex)->sql_command == SQLCOM_INSERT) || ((thd->lex)->sql_command == SQLCOM_LOAD) ||
              ((thd->lex)->sql_command == SQLCOM_INSERT_SELECT) || ci->isCacheInsert) )
         {
+            bitmap_clear_all(table->read_set);
```
abarkov: I find it too annoying to add bitmap_clear_all() in so many places.

Please consider adding a class to clear the bitmap automatically when the execution leaves the block.
```
class Scoped_bitmap_set_and_clear_all
{
  MY_BITMAP *mBitmap;
public:
  Scoped_bitmap_clear_all(MY_BITMAP *bitmap)
     :mBitmap(bitmap)
  {
    bitmap_set_all(mBitmap);
  }
  ~Scoped_bitmap_clear_all()
  {
    bitmap_clear_all(mBitmap);
  }
};
```

abarkov (changes requested):

dbcon/mysql/ha_mcs_dml.cpp
mariadb-AlexeyAntipovsky: Double-copy is much cheaper than calling fprintf(3) on each char. I suspect that the most optimal would be function something like that (I haven't tested it, just as an example):
```cpp
void fwriteEscaped(FILE* fp, const char* str, size_t len)
{
    fprintf(fp, "%c", enclosed_by);
    const char* first_not_written = str;
    const char* s = str;
    for (const char *end = str + len; s < end; s++)
    {
        if (*s == '\\')
        {
            if (first_not_written != str)
            {
                fwrite(first_not_written, s - first_not_written, 1, fp);
            }
            fputc('\\', fp);
            fputc(*str, fp);
            first_not_written = s + 1;
        }
    }                                                                                                                                                                                    
                                                                                                                                                                                         
    if (first_not_written != str + len)
    {
        fwrite(first_not_written, str + len - first_not_written, 1, fp);                                                                                                                 
    }                                                                                                                                                                                    
    fprintf(fp, "%c%c", enclosed_by, delimiter);
}
```

mariadb-AlexeyAntipovsky: Quick and dumb benchmark (each function called 100000 times):
```
fwriteEscaped(bs): 177271us
fprintEscaped(bs): 1099690us
stringEscaped(bs): 205665us
fwriteEscaped(nobs): 13391us
fprintEscaped(nobs): 567740us
stringEscaped(nobs): 22922us
fwriteEscaped(somebs): 31323us
fprintEscaped(somebs): 612498us
stringEscaped(somebs): 83735us
```
* `bs` — 255 `\`
* `nobs` — 255 `a`
* `somebs` — 255 `a`, every 17th char replaced with `\`

abarkov: Thanks. Good catch. ```fputc(*str, f)``` is much better than ```fprintf("%c", *str)```

dbcon/mysql/ha_mcs_impl.cpp
drrtuy: I don't see how it would fit in here w/o adding another member into ci variable that must be properly cleared in many places.

dbcon/mysql/ha_mcs_dml.cpp
mariadb-AlexeyAntipovsky: Anyway, in fact the less external function calls the faster the code. A naive `fputc`-version is the fastest only when the string contains only `\` (obviously). In more realistic case `nobs` it is slower than both `stringEscape` and `fwriteEscape`, and in `somebs` case it is about 15% faster than `stringEscape`, but it's still twice as slow as `fwriteEscape`

drrtuy: I'm grateful for the important input but I'll let you step in and introduce this optimization on your own b/c it goes future then I wish to optimize with this patch.

mariadb-AlexeyAntipovsky (approved):

PR #1666, opened 2020-12-10
Refactoring bit functions to follow InnoDB results for the same queries

drrtuy:


abarkov (commented):
The patch does not fully reproduce MariaDB behavior:

```
DROP TABLE IF EXISTS t1;
CREATE TABLE t1 (a TIME(1)) ENGINE=ColumnStore;
INSERT INTO t1 VALUES ('00:00:01.9');
SELECT 1<<a, 1.0<<a, 1 & a, 1.0 & a FROM t1;
+------+--------+-------+---------+
| 1<<a | 1.0<<a | 1 & a | 1.0 & a |
+------+--------+-------+---------+
|    2 |      2 |     1 |       1 |
+------+--------+-------+---------+
```

```
DROP TABLE IF EXISTS t1;
CREATE TABLE t1 (a TIME(1)) ENGINE=MyISAM;
INSERT INTO t1 VALUES ('00:00:01.9');
SELECT 1<<a, 1.0<<a, 1 & a, 1.0 & a FROM t1;
+------+--------+-------+---------+
| 1<<a | 1.0<<a | 1 & a | 1.0 & a |
+------+--------+-------+---------+
|    2 |      2 |     0 |       0 |
+------+--------+-------+---------+
```

######################################

```
DROP TABLE IF EXISTS t1;
CREATE TABLE t1 (a VARCHAR(10)) ENGINE=ColumnStore;
INSERT INTO t1 VALUES ('1.9');
SELECT 1<<a, 1.0<<a, 1 & a, 1.0 & a FROM t1;
+------+--------+-------+---------+
| 1<<a | 1.0<<a | 1 & a | 1.0 & a |
+------+--------+-------+---------+
|    2 |      2 |     1 |       1 |
+------+--------+-------+---------+
```
```
DROP TABLE IF EXISTS t1;
CREATE TABLE t1 (a VARCHAR(10)) ENGINE=MyISAM;
INSERT INTO t1 VALUES ('1.9');
SELECT 1<<a, 1.0<<a, 1 & a, 1.0 & a FROM t1;
+------+--------+-------+---------+
| 1<<a | 1.0<<a | 1 & a | 1.0 & a |
+------+--------+-------+---------+
|    2 |      2 |     0 |       0 |
+------+--------+-------+---------+
```
#####################################

```
DROP TABLE IF EXISTS t1;
CREATE TABLE t1 (a DECIMAL(30,0) NOT NULL) ENGINE=ColumnStore;
INSERT INTO t1 VALUES (1000000000000000000000000000);
SELECT a<<0, a <<1, a << 2 FROM t1;
+----------------------+-------+----------------------+
| a<<0                 | a <<1 | a << 2               |
+----------------------+-------+----------------------+
| 18446744073709551615 |     0 | 18446744073709551612 |
+----------------------+-------+----------------------+
```
```
DROP TABLE IF EXISTS t1;
CREATE TABLE t1 (a DECIMAL(30,0) NOT NULL) ENGINE=MyISAM;
INSERT INTO t1 VALUES (1000000000000000000000000000);
SELECT a<<0, a <<1, a << 2 FROM t1;
+----------------------+----------------------+----------------------+
| a<<0                 | a <<1                | a << 2               |
+----------------------+----------------------+----------------------+
| 18446744073709551615 | 18446744073709551614 | 18446744073709551612 |
+----------------------+----------------------+----------------------+
```
########################################

```
DROP TABLE IF EXISTS t1;
CREATE TABLE t1 (a INT) ENGINE=ColumnStore;
INSERT INTO t1 VALUES (65);
SELECT 1<<a FROM t1;
+------+
| 1<<a |
+------+
|    2 |
+------+
```
```
DROP TABLE IF EXISTS t1;
CREATE TABLE t1 (a INT) ENGINE=MyISAM;
INSERT INTO t1 VALUES (65);
SELECT 1<<a FROM t1;
+------+
| 1<<a |
+------+
|    0 |
+------+
```

########################################

ColumnStore performs eager operand evaluation in bit-AND and bit-OR:
```
DROP TABLE IF EXISTS t1;
CREATE TABLE t1 (a INT) ENGINE=ColumnStore;
INSERT INTO t1 VALUES (NULL);
SELECT a & SLEEP(3) FROM t1;
+--------------+
| a & SLEEP(3) |
+--------------+
|         NULL |
+--------------+
1 row in set (3.106 sec)  <-- notice timing
```

MariaDB performs lazy operand evaluation in bit-AND and bit-OR
(if the first operand is NULL, the second operand is not evaluated)
```
DROP TABLE IF EXISTS t1;
CREATE TABLE t1 (a INT) ENGINE=MyISAM;
INSERT INTO t1 VALUES (NULL);
SELECT a & SLEEP(3) FROM t1;
+--------------+
| a & SLEEP(3) |
+--------------+
|         NULL |
+--------------+
1 row in set (0.001 sec)  <-- notice timing
```

abarkov:
We should think how to share the code between the server and MCS.

The server could have some class templates which perform operations, so the server side and MCS side function implementations could reuse these templates passing their own classes as template parameters.

abarkov:
Please create an MCOL for this task.

abarkov:
I propose to implement this change in a slightly different way:

- Add a class TUInt64 - a complement for TSInt64, but for unsigned values.
- Add a class TBitOperand, deriving from TUInt64 and adding "bool mIsNull" flag.
- Add a static method TBitOperand::make_from_int(...) to create it from the  
  expression's INT representation.
- Add a static method TBitOperand::make_from_decimal(...) to create it from the  
  expression's DECIMAL representation.

The two methods TBitOperand::make_from_xxx() would take into account all tweaks performed by MariadDB inside Item_func_xxx that implement bit operations.


The body of say the bit-AND operation would look like:

```
// MariaDB has two different code branches:
// for INT and for DECIMAL (and other) argument types
if (use_int_format)
{
  TBitOperand a= TBitOperand::make_from_int(parm[0]);
  if (isNull= a.isNull())
    return 0;
  TBitOperand b= TBitOperand::make_from_int(parm[1]));
  return isNull= b.isNull() ? 0 : a & b;
}
else
{
  TBitOperand a= TBitOperand::make_from_decimal(parm[0]);
  if (isNull= a.isNull())
    return 0;
  TBitOperand b= TBitOperand::make_from_decimal(parm[1]));
  return isNull= b.isNull() ? 0 : a & b;
}
```


use_int_format should be set once, during the query preparation stage.


The body of other operations would look similar, with minor differences: for example:
- bit shift operations always extract INT value for parm[1] - even if parm[0] is DECIMAL
- bit shift operations perform eager evaluation rather than lazy evaluation.


Rationale:
- It will look very close to MariaDB implementation. Now we can make sure easier that
MCS and MDB work in the same way. Later we can go even farther: reuse the code between the server and MCS using templates.
- Later, the two code branches (contolled by use_int_format) would go to separate FuncHandlers (when we add them). The code style that I propose make it very easy to split itself into two different FuncHandlers, like it is done in MariaDB.
- I believe the overall object code size will be much smaller than the current implementation proposed in the patch (with a pretty large BitwiseFunctional template
which is compiled in five copies).

PR #1672, opened 2020-12-10
MCOL-4177 Add support for bulk insertion for wide decimals.

tntnatbry:
1. This patch adds support for wide decimals with/without scale
     to cpimport. In addition, INSERT ... SELECT and LDI are also
     now supported.
  2. Logic to compute the number of bytes to convert a binary
     representation in the buffer to a narrow decimal is also
     simplified.
  3. Support for selective block loading for 16 bytes columns
     is also added to PrimProc.

### DIFF
--- a/datatypes/mcs_datatype.h
+++ b/datatypes/mcs_datatype.h
@@ -167,6 +167,11 @@ public:
           scale(0),
           precision(-1)
       {}
+      TypeAttributesStd(int32_t w, int32_t s, int32_t p)
+         :colWidth(w),
+          scale(s),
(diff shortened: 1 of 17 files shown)


tntnatbry:
Closing this PR as it is now split into two PRs: For handling bulk insertion for wide decimals (https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/1677), and for fixing the PrimProc issue for selective block loading for 16 byte columns (https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/1676).

PR #1677, opened 2020-12-11
MCOL-4177 Add support for bulk insertion for wide decimals.

tntnatbry:
1. This patch adds support for wide decimals with/without scale
     to cpimport. In addition, INSERT ... SELECT and LDI are also
     now supported.
  2. Logic to compute the number of bytes to convert a binary
     representation in the buffer to a narrow decimal is also
     simplified.

### DIFF
--- a/dbcon/mysql/ha_mcs_datatype.h
+++ b/dbcon/mysql/ha_mcs_datatype.h
@@ -188,6 +188,20 @@ public:
 
 class WriteBatchFieldMariaDB: public WriteBatchField
 {
+  // Maximum number of decimal digits that can be represented in 4 bytes
+  static const int DIG_PER_DEC = 9;
+  // See strings/decimal.c
+  const int dig2bytes[DIG_PER_DEC+1]={0, 1, 1, 2, 2, 3, 3, 4, 4, 4};
+
+
+  // Returns the number of bytes required to store a given number
+  // of decimal digits
+  int numDecimalBytes(int digits)
+  {
+    return (((digits/DIG_PER_DEC) * 4) + dig2bytes[digits % DIG_PER_DEC]);
+  }
+
+
 public:
   Field *m_field;
   const CalpontSystemCatalog::ColType &m_type;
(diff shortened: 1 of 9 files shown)


dbcon/mysql/ha_mcs_datatype.h
```diff
@@ -188,6 +188,20 @@ class StoreFieldMariaDB: public StoreField
...
+  static const int DIG_PER_DEC = 9;
+  // See strings/decimal.c
+  const int dig2bytes[DIG_PER_DEC+1]={0, 1, 1, 2, 2, 3, 3, 4, 4, 4};
+
+
+  // Returns the number of bytes required to store a given number
+  // of decimal digits
+  int numDecimalBytes(int digits)
+  {
+    return (((digits/DIG_PER_DEC) * 4) + dig2bytes[digits % DIG_PER_DEC]);
```
drrtuy, line 201: I might had loost the point but my math tells me this f() returns 17 bytes for DECIMAL(38). Did I get it right that this f() works for precision <= 18?

dbcon/mysql/ha_mcs_datatype.h
```diff
@@ -539,264 +553,28 @@ class WriteBatchFieldMariaDB: public WriteBatchField
...
+    else if (m_type.precision > datatypes::INT64MAXPRECISION)
+    {
+      // TODO MCOL-641 The below else block for narrow decimal
+      // i.e. (m_type.precision <= datatypes::INT64MAXPRECISION)
+      // converts the decimal binary representation in buf directly
+      // to a string, while here, the my_decimal ctor first calls
+      // bin2decimal() on buf, and then we construct the string from
+      // the my_decimal. This approach might be a bit slower than the
+      // narrow decimal approach.
+      my_decimal dec(buf, m_type.precision, m_type.scale);
```
drrtuy, line 573: The approach looks logical IMHO but I'll ask bar if he can suggest/contribute to this part.

utils/dataconvert/dataconvert.cpp
```diff
@@ -111,7 +111,7 @@ void number_int_value(const string& data,
                       const datatypes::SystemCatalog::TypeAttributesStd& ct,
                       bool& pushwarning,
                       bool noRoundup,
-                      T& intVal)
+                      T& intVal, bool* saturate)
```
drrtuy, line 114: I wonder why do you choose ptr and not ref here?

utils/dataconvert/dataconvert.cpp
```diff
@@ -304,11 +304,17 @@ void number_int_value(const string& data,
             {
                 intVal = MIN_TINYINT;
                 pushwarning = true;
+
+                if (saturate)
```
drrtuy, line 308: I am curios if there are reasons why we can't move the boilerplate code outside the switch/case block?

writeengine/xml/we_xmljob.cpp
```diff
@@ -1087,13 +1087,24 @@ void XMLJob::fillInXMLDataNotNullDefault(
...
             case execplan::CalpontSystemCatalog::UDECIMAL:
             {
-                col.fDefaultInt = Convertor::convertDecimalString(
-                                      col_defaultValue.c_str(),
-                                      col_defaultValue.length(),
-                                      colType.scale);
+                if (LIKELY(colType.colWidth == datatypes::MAXDECIMALWIDTH))
+                {
+                    bool dummy = false;
+                    dataconvert::number_int_value(col_defaultValue, colType.colDataType,
```
drrtuy: I would like to discourage you from using dataconvert functions and use TypeHandler classes' methods instead. Could you leverage SystemCatalog::TypeAttributesStd::decimal128FromString() here?

dbcon/mysql/ha_mcs_datatype.h
tntnatbry: That's correct, the server sends 17 bytes in the binary buffer if `digits == 38`. Yes, this function also works correctly for precision <= 18. I have verified that for every possible combination of precision and scale, the bytes computed with the old logic and the new simplified logic are the same.

utils/dataconvert/dataconvert.cpp
tntnatbry: Default assignment to a non-const ref parameter is not possible (default assignment so as to avoid changing all occurrences of this function called), that's why I am using a ptr instead.

utils/dataconvert/dataconvert.cpp
tntnatbry: We need to set the flag to true if the value falls outside the min/max range for a given data type, and we only do that check inside the switch case block.

writeengine/xml/we_xmljob.cpp
tntnatbry: Thanks for pointing this out, I wasn't aware of this function. Patch is updated to use `decimal128FromString` here.

drrtuy (approved):

PR #1678, opened 2020-12-12
MCOL-4180 Add some missing support for wide decimals to dbcon/execplan classes.

tntnatbry:


drrtuy (approved):

dbcon/execplan/treenode.h
```diff
@@ -829,8 +836,16 @@ inline float TreeNode::getFloatVal()
             return (float)fResult.doubleVal;
 
         case CalpontSystemCatalog::DECIMAL:
+        case CalpontSystemCatalog::UDECIMAL:
```
drrtuy, line 839: Did I get it right that previously we returned nothing for UDECIMAL?

dbcon/execplan/treenode.h
```diff
@@ -974,8 +989,15 @@ inline long double TreeNode::getLongDoubleVal()
...
-            // this may not be accurate. if this is problematic, change to pre-calculated power array.
-            return (long double)(fResult.decimalVal.value / pow((long double)10, fResult.decimalVal.scale));
+            if (fResultType.colWidth == datatypes::MAXDECIMALWIDTH)
+            {
+                return static_cast<long double>(fResult.decimalVal);
+            }
+            else
+            {
+                // this may not be accurate. if this is problematic, change to pre-calculated power array.
+                return (long double)(fResult.decimalVal.value / pow((long double)10, fResult.decimalVal.scale));
```
drrtuy, line 999: I would like to make this a method of the Decimal class but I think you would like to change it globally in a separate patch.

dbcon/execplan/treenode.h
tntnatbry: Yes, previously we were throwing an exception for UDECIMAL.

dbcon/execplan/treenode.h
tntnatbry: I agree, this change should be outside the scope of this patch.

PR #1681, opened 2020-12-14
MCOL-4263 return int for func_floor on datetime

davidhall586:
For TIMESTAMP, it should do similar. However, it didn't work. For some reason, MDB has the function set as DATETIME, which for cs, isn't the same thing. Added a kludge to ha_mcs_execplan.cpp to handle it.


davidhall586:
Seems cherry-pick doesn't work anymore from develop-5 to develop

PR #1684, opened 2020-12-16
MCOL-4031 More accurate memory usage counting while sorting

mariadb-AlexeyAntipovsky:


utils/windowfunction/idborderby.cpp
```diff
@@ -744,7 +744,7 @@ void IdbOrderBy::initialize(const RowGroup& rg)
     // initialize rows
     IdbCompare::initialize(rg);
 
-    uint64_t newSize = fRowsPerRG * rg.getRowSize();
```
drrtuy, line 747: I'm curios why do you account the header size also? We should reuse an existing RowGroup just applying RGData to it.

dbcon/joblist/limitedorderby.cpp
```diff
@@ -181,7 +208,7 @@ void LimitedOrderBy::finalize()
     {
         // *DRRTUY Very memory intensive. CS needs to account active
         // memory only and release memory if needed.
-        uint64_t memSizeInc = fRowsPerRG * fRowGroup.getRowSize();
+        uint64_t memSizeInc = fRowGroup.getSizeWithStrings();
```
drrtuy: I'm curios why do you account the header size also? I believe we reuse an existing RowGroup just applying RGData to it so that we don't need to count the header size every time.

mariadb-AlexeyAntipovsky: Yes, I missed that point.

drrtuy (approved):

PR #1699, opened 2020-12-26
upd ci: autobake-deb patch, mtr branches, mtr files owner

mariadb-RomanNavrotskiy:


ottok (changes requested):
Remove the TRAVIS-CI env, you should not piggyback on it since Drone nor ColumnStore is not Travis.

ottok:
Ideally, before you do any changes to the CI, it should be in a fully green state. Currently non-deb builds are failing on MTR..

mariadb-RomanNavrotskiy:
> Remove the TRAVIS-CI env, you should not piggyback on it since Drone nor ColumnStore is not Travis.

we use that env just to skip some things and speed up builds. we really know what Drone and Travis and columnstore is =)

we can use any env or method to get the same. you can propose something if using that env confusing you.

ottok (changes requested):
The commit message is subpar. See here tips on how to write good commit messages: https://github.com/Seravo/gnitpick#recommended-reads

Looking at the contents, it includes two unrelated changes. This should thus be two different commits. The 'local Pipeline' changes are obsoleted by #1700, so this PR would suffice with just the mtr changes.

The mtr changes are not described in the commit message nor in inline comments. Are they perhaps related to fixing columnstore/basic.mcs38_select_limit?

mariadb-RomanNavrotskiy:
> The commit message is subpar. See here tips on how to write good commit messages: https://github.com/Seravo/gnitpick#recommended-reads
> 
> Looking at the contents, it includes two unrelated changes. This should thus be two different commits. The 'local Pipeline' changes are obsoleted by #1700, so this PR would suffice with just the mtr changes.
> 
> The mtr changes are not described in the commit message nor in inline comments. Are they perhaps related to fixing columnstore/basic.mcs38_select_limit?

this PR has 'autobake patch' changes and 'mtr' changes.
I squashed this to reduce noise in commits history. it is only about CI updates.

I updated commit message.

drrtuy (approved):

PR #1705, opened 2020-12-30
MCOL-4478 MCS now rounds the last digit of an avg() result

drrtuy:


utils/rowgroup/rowaggregation.cpp
```diff
@@ -2800,8 +2800,22 @@ void RowAggregationUM::calculateAvgColumns()
                         multOp(*sumPnt, datatypes::mcs_pow_10[scaleDiff], sum);
                     else
                         sum = *sumPnt;
-                    int128_t avg = sum / cnt;
-                    fRow.setBinaryField_offset(&avg, sizeof(avg), offset);
+                    datatypes::lldiv_t_128 avgAndRem = datatypes::lldiv128(sum, cnt);
+                    // Round the last digit
+                    if (avgAndRem.rem * 2 > (int128_t)cnt)
```
mariadb-AlexeyAntipovsky: I think that `llabs(avgAndRem.rem) * 2 >= (int128_t)cnt` looks like a more expected behavior, i.e.
* 1.2345 => 1.235
* -1.2345 => -1.235

mariadb-AlexeyAntipovsky (changes requested):

utils/rowgroup/rowaggregation.cpp
drrtuy: Good catch.

mariadb-AlexeyAntipovsky (approved):

PR #1709, opened 2021-01-07
MCOL-4472 Fix func_floor to match func_ceil

davidhall586:
Currently, for TIME type, func_floor is incorrect as of MariaDB 10.5


davidhall586:
Need to remove commented code

PR #1717, opened 2021-01-13
A cleanup for MCOL-4464 Bitwise operations not like in MariaDB

abarkov:
CI with RelWithDebInfo builds revealed a problem in the main
patch for MCOL-4464, which did not show up with Debug builds.

Methods like:
- getDoubleVal()
- getDateIntVal()
- getDatetimeIntVal()
- getTimestampIntVal()
- getTimeIntVal()
- getUintVal()
- getIntVal()
- getStrVal()
require the caller to initialize the isNull argument to false.
This fact was not taken into account in MCOL-4464.

Adding proper initializations.

### DIFF
--- a/dbcon/execplan/treenode.h
+++ b/dbcon/execplan/treenode.h
@@ -267,13 +267,13 @@ public:
     }
     datatypes::TUInt64Null toTUInt64Null(rowgroup::Row& row)
     {
-        bool isNull;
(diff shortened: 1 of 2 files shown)


drrtuy (approved):

drrtuy:
The failed test in the CI's MTR run isn't connected with the patch.

PR #1719, opened 2021-01-15
MCOL-4065 DISTINCT is case sensitive

abarkov:
This patch makes DISTINCT and GROUP BY collation aware.


drrtuy:
The failed test contains non-deterministic reference set.

drrtuy (approved):

PR #1720, opened 2021-01-15
MCOL-4412

drrtuy:
The main idea is to replace literal empty magic values used in PrimProc and WriteEngineServer with pointers to constant values that are TypeHandler-specific so processing avoids type-based switch-case blocks.


datatypes/mcs_datatype.h
```diff
@@ -2294,160 +2329,14 @@ class TypeHandlerSLongDouble: public TypeHandlerReal
     throw logging::QueryDataExcept("convertColumnData: unknown column data type.", logging::dataTypeErr);
     return boost::any();
   }
-};
```
drrtuy, line 2297: The classes were moved to come after getEmptyTypeHandlerStr.

writeengine/wrapper/writeengine.cpp
```diff
@@ -775,9 +781,13 @@ int WriteEngineWrapper::deleteRow(const TxnID& txnid, const vector<CSCTypesList>
             cscColType = cscColTypeList[i];
             Convertor::convertColType(&curColStruct);
 
-            m_colOp[op(curColStruct.fCompressionType)]->
-                getEmptyRowValue(curColStruct.colDataType, curColStruct.colWidth, (uint8_t*)&emptyVal);
-            curTuple.data = emptyVal;
+            const uint8_t* emptyVal = m_colOp[op(curColStruct.fCompressionType)]->
+                getEmptyRowValue(curColStruct.colDataType, curColStruct.colWidth);
+
+            if (curColStruct.colWidth == datatypes::MAXDECIMALWIDTH)
```
mariadb-AlexeyAntipovsky, line 787: Shouldn't we also check the column data type here?

drrtuy: We shouldn't b/c we check coltype getting empty value ptr from BlockOp::getEmptyRowValue()

mariadb-AlexeyAntipovsky (approved):

PR #1722, opened 2021-01-19
MCOL-4496 Upmerged the commit 7fa5ca3a6 to generalise CHAR|VARCHAR|TE…

drrtuy:
…XT|BLOB

bulk insertion processing and fix the original issue introduced by MCOL-2000


mariadb-AlexeyAntipovsky (approved):

drrtuy:
The failure in MTR caused by MCOL-4456 and will be fixed with the PR that is coming today.

PR #1723, opened 2021-01-20
MCOL-4456 Fix for bulk insertion of BLOBs

drrtuy:


mariadb-AlexeyAntipovsky (approved):

drrtuy:
MTR fails b/c of forgotten results not included in the MTR .result file.

PR #1725, opened 2021-01-20
MCOL-4505 Fix DISTINCT with sub query gets wrong results

mariadb-AlexeyAntipovsky:


dbcon/joblist/tupleannexstep.cpp
```diff
@@ -512,31 +523,48 @@ void TupleAnnexStep::executeNoOrderByWithDistinct()
...
 
+                    // skip first limit-start rows
+                    if (distinctMap->size() <= fLimitStart)
+                    {
+                        rowGroupSkip.incRowCount();
+                        rowSkip.nextRow();
+                        if (UNLIKELY(rowGroupSkip.getRowCount() >= rowgroup::rgCommonSize))
+                        {
+                            dataVecSkip.push_back(rgDataSkip);
+                            rgDataSkip.reinit(rowGroupSkip);
```
drrtuy, line 562: We are allocating a new memory block for the RGData. Despite the fact the number of RGDatas with data to skip should be small we at least must left the commentary here.

drrtuy (approved):

PR #1726, opened 2021-01-20
MCOL-4505 Fix DISTINCT with sub query gets wrong results

mariadb-AlexeyAntipovsky:


dhall-MariaDB:
MCOL-4505 requested a fix to the query returning too many rows in some case. The fix submitted looks to be limit with an offset.

davidhall586 (approved):

PR #1739, opened 2021-01-27
MCOL-4493 Add ON expressions for WHERE processing when the JOIN

tntnatbry:
type is not LEFT/RIGHT.


drrtuy (approved):
We still didn't find out why 1.4, 1.5 and later 5.x needs this and 1.2 doesn't. The plausable explanation is that the OUTER JOIN rewrites I borrowed from MDB works in a slightly different way. 
The patch doesn't break regression and does what is expected.

PR #1742, opened 2021-01-29
MCOL-4031 More accurate memory usage counting while sorting

mariadb-AlexeyAntipovsky:


drrtuy (approved):

drrtuy:
The failed MTR test is known to fail with the current code.

PR #1751, opened 2021-02-04
MCOL-3785 wrap names of db, table & columns in "`"

mariadb-AlexeyAntipovsky:


drrtuy (approved):

drrtuy:
The MTR failure caused by the fact that the pushed tree doesn't have a fix for a certain issue.

PR #1765, opened 2021-02-10
MCOL-4531 New string-to-decimal conversion implementation DRAFT

mariadb-RomanNavrotskiy:
This change fixes:

MCOL-4462 CAST(varchar_expr AS DECIMAL(M,N)) returns a wrong result
MCOL-4500 Bit functions processing throws internally trying to cast char into decimal representation
MCOL-4532 CAST(AS DECIMAL) returns a garbage for large values

Also, this change makes string-to-decimal conversion 5-10 times faster,
depending on exact data.
Performance implemenent is achieved by the fact that (unlike in the old
implementation), the new version does not do any "string" object copying.


drrtuy (approved):

drrtuy:
The failed MTR test isn't connected with the patch. We'll research it outside the scope of the patch.

PR #1768, opened 2021-02-15
A join patch for MCOL-4527 (a performance hack) and MCOL-4539 (a bug …

abarkov:
…fix)

- MCOL-4527 Simple query performace is degraded between 5.4 and 5.5
- MCOL-4539 WHERE short_char_column='literal' ignores the collation on a huge table

This patch also adds mtr/basic/suite.opt, so "mtr" can run without --extern.


drrtuy (approved):

drrtuy:
The opensuse pipeline failure caused by an external reason.

davidhall586 (approved):

PR #1771, opened 2021-02-16
Mcol 2044 update ranges during dml

mariadb-SergeyZefirov:
Relevant task: https://jira.mariadb.org/browse/MCOL-2044?focusedCommentId=178041&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel

### DIFF
--- a/dbcon/dmlpackageproc/commandpackageprocessor.cpp
+++ b/dbcon/dmlpackageproc/commandpackageprocessor.cpp
@@ -430,7 +430,14 @@ CommandPackageProcessor::processPackage(dmlpackage::CalpontDMLPackage& cpackage)
 
                 if (!cpInvalidated)
                 {
-                    fDbrm->invalidateUncommittedExtentLBIDs(0, &lbidList);
+		    // The code below assumes that in case of COMMIT all ranges for all touched LBIDs
+		    // are either correctly set or correctly reset.
+		    // It is also assumes that ROLLBACK or other operations but COMMIT may not return ranges
+		    // to state that is correct. This is why we invalidate extents when we are not committing.
+		    if (stmt != "COMMIT")
+		    {
+                        fDbrm->invalidateUncommittedExtentLBIDs(0, &lbidList);
+		    }
                 }
             }
         }
--- a/dbcon/dmlpackageproc/dmlpackageprocessor.cpp
+++ b/dbcon/dmlpackageproc/dmlpackageprocessor.cpp
@@ -485,37 +485,7 @@ int DMLPackageProcessor::commitBatchAutoOnTransaction(uint64_t uniqueId, BRM::Tx
     //set CP data before hwm.
 
     //cout << "setting hwm allHwm size " << allHwm.size() << endl;
-    vector<BRM::LBID_t> lbidList;
-
-    if (idbdatafile::IDBPolicy::useHdfs())
-    {
-        BRM::LBID_t startLbid;
-
-        for ( unsigned i = 0; i < allHwm.size(); i++)
-        {
-            rc = fDbrm->lookupLocalStartLbid(allHwm[i].oid, allHwm[i].partNum, allHwm[i].segNum, allHwm[i].hwm, startLbid);
-            lbidList.push_back(startLbid);
-        }
-    }
-    else
-        fDbrm->getUncommittedExtentLBIDs(static_cast<BRM::VER_t>(txnID.id), lbidList);
-
-    vector<BRM::LBID_t>::const_iterator iter = lbidList.begin();
-    vector<BRM::LBID_t>::const_iterator end = lbidList.end();
     BRM::CPInfoList_t cpInfos;
-    BRM::CPInfo aInfo;
-
-    while (iter != end)
-    {
-        aInfo.firstLbid = *iter;
-        aInfo.max = numeric_limits<int64_t>::min(); // Not used
-        aInfo.min = numeric_limits<int64_t>::max(); // Not used
-        utils::int128Min(aInfo.bigMax); // Not used
-        utils::int128Max(aInfo.bigMin); // Not used
-        aInfo.seqNum = -1;
-        cpInfos.push_back(aInfo);
-        ++iter;
-    }
 
     std::vector<BRM::CPInfoMerge>  mergeCPDataArgs;
     rc = fDbrm->bulkSetHWMAndCP(allHwm, cpInfos, mergeCPDataArgs, txnID.id);
--- a/mtr/basic/r/mcol2044_signed_bigint_delete-drop-max.result
+++ b/mtr/basic/r/mcol2044_signed_bigint_delete-drop-max.result
@@ -1,0 +1,13 @@
+create database if not exists test_ranges;
+use test_ranges;
+drop table if exists t;
+create table t(x bigint) engine=columnstore;
+insert into t(x) values (44),(55),(66);
+select max_value, min_value from information_schema.columnstore_extents;
+max_value	min_value
+66	44
+delete from t where x=44;
+select max_value, min_value from information_schema.columnstore_extents;
+max_value	min_value
+NULL	NULL
+drop database test_ranges;
--- a/versioning/BRM/brmtypes.h
+++ b/versioning/BRM/brmtypes.h
@@ -145,6 +145,10 @@ typedef struct _SIDTIDEntry SIDTIDEntry;
 // @bug 1970 - Added CPInfo and CPMaxMin structs used by new interface that allows setting the max and min CP data
 // for multiple extents.
 
+// Special seqNum field values.
+#define SEQNUM_MARK_INVALID (-1)
+#define SEQNUM_MARK_INVALID_SET_RANGE (-2)
+
 // Used in vectors.
 struct CPInfo
 {
@@ -162,7 +166,7 @@ struct CPInfo
         int128_t bigMin;
         int64_t min_;
     };
-    bool isBinaryColumn;
+    bool isBinaryColumn;   // XXX: we should remove these two fields and replace it with type handler.
 };
 typedef std::vector<CPInfo> CPInfoList_t;
 
@@ -182,7 +186,8 @@ struct CPMaxMin
         int128_t bigMin;
         int64_t min_;
     };
-    bool isBinaryColumn;
+    bool isBinaryColumn;   // XXX: these two fields should be replaced with type handler pointer.
+
 };
 typedef std::tr1::unordered_map<LBID_t, CPMaxMin> CPMaxMinMap_t;
 
--- a/writeengine/server/we_ddlcommandproc.cpp
+++ b/writeengine/server/we_ddlcommandproc.cpp
@@ -4998,7 +4998,6 @@ void WE_DDLCommandProc::purgeFDCache()
                 aFile.compType = aIt->compType;
                 files.push_back(aFile);
                 fDbrm.lookupLocalStartLbid(aFile.oid, aFile.partitionNum, aFile.segmentNum, aIt->hwm, startLbid);
-                lbidList.push_back(startLbid);
                 //cout <<"Added to files oid:dbroot:part:seg:compType = " << aFile.oid<<":"<<aFile.dbRoot<<":"<<aFile.partitionNum<<":"<<aFile.segmentNum
                 //<<":"<<aFile.compType <<endl;
                 aIt++;
--- a/writeengine/shared/we_brm.h
+++ b/writeengine/shared/we_brm.h
@@ -33,6 +33,7 @@
 #include "we_obj.h"
 #include<sys/time.h>
 #include "brmtypes.h"
+#include "mcs_datatype.h"
 #include "IDBDataFile.h"
 #include "IDBPolicy.h"
 
@@ -47,6 +48,42 @@ namespace WriteEngine
 {
 // forward reference
 class DbFileOp;
+
+/** @brief Extended CPInfo - with type handler for all type-related information */
+struct ExtCPInfo
+{
+    execplan::CalpontSystemCatalog::ColDataType fColType;
+    int                                         fColWidth;
+    BRM::CPInfo                                 fCPInfo;
+    ExtCPInfo(execplan::CalpontSystemCatalog::ColDataType colType, int colWidth)
+        : fColType(colType), fColWidth(colWidth)
+    {
+        fCPInfo.isBinaryColumn = colWidth > datatypes::MAXLEGACYWIDTH;
+    }
+    void toInvalid()
+    {
+        auto mm = datatypes::MinMaxInfo::invalidRange(fColType);
+        fCPInfo.max = mm.max;
+	fCPInfo.min = mm.min;
+	fCPInfo.bigMax = mm.int128Max;
+	fCPInfo.bigMin = mm.int128Min;
+    }
+
+    bool isInvalid()
+    {
+        datatypes::MinMaxInfo mm;
+	mm.max = fCPInfo.max;
+	mm.min = fCPInfo.min;
+	mm.int128Max = fCPInfo.bigMax;
+	mm.int128Min = fCPInfo.bigMin;
+	return datatypes::MinMaxInfo::isRangeInvalid(mm, fColType, fColWidth);
+    }
+    bool isBinaryColumn()
+    {
+        return fCPInfo.isBinaryColumn;
+    }
+};
+typedef std::vector<ExtCPInfo> ExtCPInfoList;
 
 /** Class BRMWrapper */
 class BRMWrapper : public WEObj
@@ -262,7 +299,7 @@ public:
     /**
      * @brief set extents CP min/max info into extent map
      */
-    int setExtentsMaxMin(const BRM::CPInfoList_t& cpinfoList);
+    int setExtentsMaxMin(const ExtCPInfoList& cpinfoList);
 
     /**
      * @brief Perform bulk rollback of any column extents that logically follow
@@ -473,6 +510,8 @@ public:
         m_useVb = val;
     }
 
+    int getExtentCPMaxMin(const BRM::LBID_t lbid, BRM::CPMaxMin& cpMaxMin);
+
 private:
     //--------------------------------------------------------------------------
     // Private methods
@@ -647,9 +686,15 @@ inline int BRMWrapper::bulkSetHWMAndCP(
     return getRC( rc, ERR_BRM_BULK_UPDATE );
 }
 
-inline int BRMWrapper::setExtentsMaxMin(const BRM::CPInfoList_t& cpinfoList)
+inline int BRMWrapper::setExtentsMaxMin(const ExtCPInfoList& extCPInfoList)
 {
-    int rc = blockRsltnMgrPtr->setExtentsMaxMin(cpinfoList);
+    BRM::CPInfoList_t toSet;
+    toSet.reserve(extCPInfoList.size());
+    for (const auto& extCPInfo : extCPInfoList)
+    {
+        toSet.push_back(extCPInfo.fCPInfo);
+    }
+    int rc = blockRsltnMgrPtr->setExtentsMaxMin(toSet);
     return getRC( rc, ERR_BRM_SET_EXTENTS_CP );
 }
 
--- a/writeengine/wrapper/we_colop.cpp
+++ b/writeengine/wrapper/we_colop.cpp
@@ -82,7 +82,8 @@ ColumnOp::~ColumnOp()
  *    tableFid - the file id for table bitmap file
  *    totalRow - the total number of rows need to be allocated
  *    useStartingExtent - Indicates whether rows can be added to an existing
-*               starting extent
+ *               starting extent
+ *    newExtents - where to write extents allocated for newColStructList, etc, structures.
  * RETURN:
  *    NO_ERROR if success
  *    rowIdArray - allocation of the row id left here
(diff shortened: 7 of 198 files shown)


dbcon/dmlpackageproc/commandpackageprocessor.cpp
```diff
@@ -430,7 +430,11 @@ CommandPackageProcessor::processPackage(dmlpackage::CalpontDMLPackage& cpackage)
 
                 if (!cpInvalidated)
                 {
-                    fDbrm->invalidateUncommittedExtentLBIDs(0, &lbidList);
+		    // XXX: HACK!!!HACK!!!HACK!!!
```
drrtuy: I like the comment but what it is about? Why do we need a hack and what would be a less hacky solution in your opinion?

dbcon/dmlpackageproc/dmlpackageprocessor.cpp
```diff
@@ -487,6 +487,7 @@ int DMLPackageProcessor::commitBatchAutoOnTransaction(uint64_t uniqueId, BRM::Tx
     //cout << "setting hwm allHwm size " << allHwm.size() << endl;
     vector<BRM::LBID_t> lbidList;
 
+#if 0 // XXX: can we do better?
```
drrtuy: Better than what?
Why do you comment out instead of deleting something useless?

mtr/basic/r/mcol2044_signed_bigint_delete-drop-max.result
```diff
@@ -0,0 +1,13 @@
+create database if not exists test_ranges;
+use test_ranges;
+drop table if exists t;
+create table t(x bigint) engine=columnstore;
+insert into t(x) values (44),(55),(66);
+select max_value, min_value from information_schema.columnstore_extents;
+max_value	min_value
+66	44
+delete from t where x=44;
+select max_value, min_value from information_schema.columnstore_extents;
```
drrtuy, line 10: This query expects columnstore_extent to be empty except the tables created. In real life testing MCS might contain a lot other unexpected extents so I suggest adding a filter.

writeengine/server/we_ddlcommandproc.cpp
```diff
@@ -4998,7 +4998,7 @@ void WE_DDLCommandProc::purgeFDCache()
                 aFile.compType = aIt->compType;
                 files.push_back(aFile);
                 fDbrm.lookupLocalStartLbid(aFile.oid, aFile.partitionNum, aFile.segmentNum, aIt->hwm, startLbid);
-                lbidList.push_back(startLbid);
+                //lbidList.push_back(startLbid);
```
drrtuy: If you don't need this plz remove it.

writeengine/wrapper/we_colop.cpp
```diff
@@ -1738,6 +1743,12 @@ int ColumnOp::writeRow(Column& curCol, uint64_t totalRow, const RID* rowIdArray,
         }
 
         // This is the write stuff
+        if (oldValArray)
```
drrtuy, line 1699: I would move this block into a separate inline function to reduce boilerplate but this is not mandatory.

writeengine/wrapper/writeengine.cpp
```diff
@@ -220,6 +218,178 @@ void WriteEngineWrapper::findSmallestColumn(uint32_t& colId, ColStructList colSt
...
+                fetchNewOldValues<uint64_t, uint64_t>(uvalue, oldUValue, valArrayVoid, oldValArrayVoid, i, totalNewRow);
+                break;
+            }
+            default:
+                idbassert(0 && "unknown WR type tag");
+                return;
+        }
+        if (isUnsigned)
+        {
+            if (!oldValArrayVoid)
```
drrtuy: These if blocks should be moved into a  parametrised function  to reduce boilerplate IMHO. 
Moreover I don't see support for wide-decimal here.

writeengine/wrapper/writeengine.cpp
```diff
@@ -760,6 +933,8 @@ int WriteEngineWrapper::deleteRow(const TxnID& txnid, const vector<CSCTypesList>
     string           tmpStr("");
     vector<DctnryStructList> dctnryExtentsStruct;
 
+    m_opType = DELETE;
```
drrtuy: How this assignment differs from the same from line 988?

writeengine/wrapper/writeengine.cpp
```diff
@@ -953,6 +1130,58 @@ static void log_this(const char *message,
...
+        case WriteEngine::WR_SHORT:
+        case WriteEngine::WR_USHORT:
+        case WriteEngine::WR_INT:
+        case WriteEngine::WR_UINT:
+        case WriteEngine::WR_LONGLONG:
+        case WriteEngine::WR_ULONGLONG:
+            return currentCPInfo;
+        // all unsupported types must not be supported.
+        default:
+            return NULL; // safe choice for everything we can't do.
```
drrtuy: nullptr plz.

writeengine/wrapper/writeengine.cpp
```diff
@@ -953,6 +1130,58 @@ static void log_this(const char *message,
...
+    }
+}
+
+/** @brief Check if range is valid.
+ */
+static bool
+validRange(const BRM::CPInfo& cpInfo)
+{
+    return false;
```
drrtuy: Really? This is all MCS does here ?

writeengine/wrapper/writeengine.cpp
```diff
@@ -953,6 +1130,58 @@ static void log_this(const char *message,
...
+ *
+ * There can be a case that we have computed invalid range while computing updated ranges.
+ *
+ * These invalid ranges should not have CP_VALID in the status code. So, they must not
+ * come into final call to setCPInfos or something.
+ *
+ * To achieve that, we filter these invalid ranges here.
+ */
+static void
+squeezeValidCPInfos(std::vector<BRM::CPInfo>& cpInfos)
```
drrtuy: That is a very interesting way to clean up a vector. That what a current implementation of  validRange does.

writeengine/wrapper/writeengine.cpp
```diff
@@ -1026,6 +1258,12 @@ int WriteEngineWrapper::insertColumnRecs(const TxnID& txnid,
     for (i = 0; i < colStructList.size(); i++)
         Convertor::convertColType(&colStructList[i]);
 
+    for (i = 0; i < colStructList.size(); i++)
```
drrtuy: I believe there is [vector ctor](https://en.cppreference.com/w/cpp/container/vector/vector) that allows to fill with a class instances when called.

writeengine/wrapper/writeengine.cpp
```diff
@@ -1690,34 +1929,70 @@ int WriteEngineWrapper::insertColumnRecs(const TxnID& txnid,
         int         curFbo = 0, curBio, lastFbo = -1;
 
         if (isFirstBatchPm && (totalRow == rowsLeft))
-        {}
+        {
```
drrtuy, line 1939: Could you ellaborate on this case? I don't see the meaning behind the argument sent to idbassert.
And what is the case we describe with the above if condition?

writeengine/wrapper/writeengine.cpp
```diff
@@ -1690,34 +1929,70 @@ int WriteEngineWrapper::insertColumnRecs(const TxnID& txnid,
...
+                        RETURN_ON_ERROR(GetLBIDRange(newExtentsStartingLbids[i], colStructList[i], *cpInfoP));
+		    }
+
+                    maxMins[i].fSplitMaxMinInfoPtrs[1] = cpInfoP;
+		}
             }
         }
 
-        markTxnExtentsAsInvalid(txnid);
+        markTxnExtentsAsInvalid(txnid); // moved here as it can be source of bug.
```
drrtuy: I didn't get the comment meaning b/c the placement of t he statement looks the same for me. Could you ellaborate on this?

writeengine/wrapper/writeengine.cpp
```diff
@@ -1690,34 +1929,70 @@ int WriteEngineWrapper::insertColumnRecs(const TxnID& txnid,
...
 
         //----------------------------------------------------------------------
         // Write row(s) to database file(s)
         //----------------------------------------------------------------------
-        rc = writeColumnRec(txnid, cscColTypeList, colStructList, colOldValueList, rowIdArray, newColStructList, colNewValueList, tableOid, useTmpSuffix); // @bug 5572 HDFS tmp file
+        std::vector<BRM::CPInfo> cpinfoList;
+        for (auto& splitCPInfo : maxMins)
```
drrtuy, line 1985: IMHO it would be nice to put the boilerplate used here and in the line 2013 into a separate function.

writeengine/wrapper/writeengine.cpp
```diff
@@ -1690,34 +1929,70 @@ int WriteEngineWrapper::insertColumnRecs(const TxnID& txnid,
...
+            {
+                BRM::CPInfo* cpInfo = splitCPInfo.fSplitMaxMinInfoPtrs[i];
+                if (cpInfo)
+                {
+                    cpinfoList.push_back(*cpInfo);
+                    cpinfoList[cpinfoList.size() - 1].seqNum = -1;
+                }
+            }
+        }
+        rc = BRMWrapper::getInstance()->setExtentsMaxMin(cpinfoList);
```
drrtuy, line 1997: I didn't get why do we call setExtentsMaxMin() twice. Here for every column and later at 2030 after writeColumnRec and squeezeValidCPInfos. Could you ellaborate on this?

writeengine/wrapper/writeengine.cpp
```diff
@@ -4280,27 +4590,47 @@ int WriteEngineWrapper::updateColumnRec(const TxnID& txnid,
...
+		cpInfoPtr->seqNum ++;
+	        infosToUpdate.push_back(*cpInfoPtr);
+	    }
+	}
+    }
+    markTxnExtentsAsInvalid(txnid);
+    if (rc == NO_ERROR)
+    {
+        std::vector<BRM::CPInfo> infosToDrop = infosToUpdate;
+	for (auto& cpInfo : infosToDrop)
```
drrtuy: Not sure why but some of the lines are totaly misaligned.

writeengine/wrapper/writeengine.cpp
```diff
@@ -4348,7 +4686,39 @@ int WriteEngineWrapper::updateColumnRecs(const TxnID& txnid,
...
+    for (auto cpInfoP : pointersToInfos)
+    {
+        if (cpInfoP)
+        {
+	    auto tmp = *cpInfoP;
+	    tmp.seqNum = -1;
+            bulkUpdateInfos.push_back(tmp);
+        }
+    }
+    if (bulkUpdateInfos.size() > 0)
```
drrtuy: Could you use ::empty() instead?

writeengine/wrapper/writeengine.cpp
```diff
@@ -4348,7 +4686,39 @@ int WriteEngineWrapper::updateColumnRecs(const TxnID& txnid,
     if (m_opType != DELETE)
         m_opType = UPDATE;
 
-    rc = writeColumnRecords(txnid, cscColTypeList, colExtentsStruct, colValueList, ridLists, tableOid);
+    for (auto cpInfoP : pointersToInfos)
```
drrtuy, line 4673: It looks like this double setExtentsMaxMin() call is a pattern that I don't get.
If it is really necessary then plz make it a function to reduce boilerplate.

writeengine/wrapper/writeengine.cpp
```diff
@@ -220,6 +218,178 @@ void WriteEngineWrapper::findSmallestColumn(uint32_t& colId, ColStructList colSt
...
+            }
+	}
+    }
+    size_t i;
+    for (i = 0; i < totalOldRow; i++) {
+	int64_t value, oldValue;
+        uint64_t uvalue, oldUValue;
+
+	// fetching. May not assign value or oldValue variables when array is not present.
+        switch (colType)
```
drrtuy, line 353: Despite the fact that CPU branch prediction should eliminate the negative effect of a switch called in a loop is it possible to move the switch above the loop?

drrtuy: I also don't see int128_t support here.

writeengine/wrapper/writeengine.cpp
```diff
@@ -6119,6 +6578,42 @@ void WriteEngineWrapper::AddDictToList(const TxnID txnid,
 
 }
 
+// Get CPInfo for given starting LBID and column description structure.
+int WriteEngineWrapper::GetLBIDRange(const BRM::LBID_t startingLBID, const ColStruct& colStruct, BRM::CPInfo& cpInfo)
+{
+    int rtn;
+    BRM::CPMaxMin maxMin;
+    bool isBinary = colStruct.colWidth > 8;
```
drrtuy: Plz use datatypes::MAXLEGACYWIDTH constant.

drrtuy (changes requested):
In general the approach is OK. There are some comments/suggestions that should be addressed or replied though.

writeengine/wrapper/writeengine.cpp
mariadb-SergeyZefirov: I deleted this one because it and assignment below are duplicates.

versioning/BRM/brmtypes.h
```diff
@@ -163,6 +163,36 @@ struct CPInfo
...
+	else
+	{
+            if (isColTypeUnsigned)
+            {
+                bigMax = 0;
+                bigMin = -1;
+            }
+            else
+            {
+                utils::int128Min(bigMax);
```
drrtuy: Better replace this with datatypes::numeric_limits<int128_t>max() taken from ./datatypes/mcs_numeric_limits.h

writeengine/wrapper/writeengine.cpp
mariadb-SergeyZefirov: I did just that.

writeengine/wrapper/writeengine.cpp
mariadb-SergeyZefirov: The logic has changed since. ;)

versioning/BRM/brmtypes.h
mariadb-SergeyZefirov: It was moved to we_brm.h (ExtCPInfo::toInvalid()) and now uses mcs_datatypes.h-defined ranges.

writeengine/wrapper/writeengine.cpp
mariadb-SergeyZefirov: Moved to we_brm.h, ExtCPInfo::isBinaryColumn().

writeengine/wrapper/writeengine.cpp
mariadb-SergeyZefirov: Now we do support wide decimals.

writeengine/wrapper/writeengine.cpp
mariadb-SergeyZefirov: Done.

writeengine/wrapper/writeengine.cpp
mariadb-SergeyZefirov: Forgot tabs->spaces.

dbcon/dmlpackageproc/commandpackageprocessor.cpp
mariadb-SergeyZefirov: I explained in the comment.

Basically, it is not clear to me is it good to dispatch on the command string.

writeengine/wrapper/writeengine.cpp
mariadb-SergeyZefirov: The habit of copying general approach of code around sometimes fails me.

The code needs a mapper (which I hadn't added yet) for new structure constructor. So, it is an iterator loop now.

writeengine/wrapper/writeengine.cpp
mariadb-SergeyZefirov: This is hard to test case. I guess it can happen when running in cluster and node just went up and it is asked to insert values into an extent which does not exists on it right now.

writeengine/wrapper/writeengine.cpp
mariadb-SergeyZefirov: I haven't figured out a good way to do that.

mtr/basic/r/mcol2044_signed_bigint_delete-drop-max.result
mariadb-SergeyZefirov: We have other extent-related test cases with same problem. ;)

mcs211_idbExtentId_function.test is one such test.

drrtuy: Thank you. GTK.

writeengine/wrapper/we_colop.cpp
```diff
@@ -1165,48 +1141,20 @@ int ColumnOp::fillColumn(const TxnID& txnid, Column& column, Column& refCol, voi
...
-                        utils::int128Max(cpInfo1.bigMin);
-                    }
-                }
-
-                cpInfo1.seqNum = -1;
-                cpInfo1.firstLbid = startLbid;
-                BRM::CPInfoList_t cpinfoList1;
+                ExtCPInfo cpInfo1(column.colDataType, column.colWidth);
+		cpInfo1.toInvalid();
+		cpInfo1.fCPInfo.seqNum = -1;
```
drrtuy: I think you should use here the constant you introduced with this patch.

writeengine/wrapper/writeengine.cpp
drrtuy: Plz put the switch statement outside of the loop using std::function wrapper to save needed fetchNewOldValues<T, T> and call it later in the loop.

writeengine/wrapper/writeengine.cpp
```diff
@@ -1045,7 +1317,8 @@ int WriteEngineWrapper::insertColumnRecs(const TxnID& txnid,
     IDBDataFile* pFile = NULL;
     std::vector<DBRootExtentInfo> extentInfo;
     int currentDBrootIdx = 0;
-    std::vector<BRM::CreateStripeColumnExtentsArgOut> extents;
+    std::vector<BRM::CreateStripeColumnExtentsArgOut> extents; // this structure valid only when isFirstOnBatchPm is true.
+    std::vector<BRM::LBID_t> newExtentsStartingLbids; // this structure
```
drrtuy: It would be great if you finish the commentary.

writeengine/wrapper/writeengine.cpp
```diff
@@ -220,6 +220,212 @@ void WriteEngineWrapper::findSmallestColumn(uint32_t& colId, ColStructList colSt
...
+                fetchNewOldValues<uint64_t, uint64_t>(uvalue, oldUValue, valArrayVoid, oldValArrayVoid, i, totalNewRow);
+                break;
+            }
+            case WR_BINARY:
+            {
+                fetchNewOldValues<int128_t, int128_t>(bvalue, oldBValue, valArrayVoid, oldValArrayVoid, i, totalNewRow);
+                break;
+            }
+            default:
+                idbassert(0 && "unknown WR type tag");
```
drrtuy: If you want a new error message you should add it accordingly. Now this looks like a hack to log a specific message.

writeengine/wrapper/writeengine.cpp
```diff
@@ -1102,39 +1375,12 @@ int WriteEngineWrapper::insertColumnRecs(const TxnID& txnid,
                 if (rc != NO_ERROR)
                     return rc;
 
-                cpInfo.isBinaryColumn = colStructList[i].colWidth > 8;
```
drrtuy, line 1105: Where do  we set cpInfo.isBinaryColumn now?

writeengine/shared/we_brm.h
```diff
@@ -48,6 +49,42 @@ namespace WriteEngine
...
+/** @brief Extended CPInfo - with type handler for all type-related information */
+struct ExtCPInfo
+{
+    execplan::CalpontSystemCatalog::ColDataType fColType;
+    int                                         fColWidth;
+    BRM::CPInfo                                 fCPInfo;
+    ExtCPInfo(execplan::CalpontSystemCatalog::ColDataType colType, int colWidth)
+        : fColType(colType), fColWidth(colWidth)
+    {
+        fCPInfo.isBinaryColumn = colWidth > 8;
```
drrtuy: Plz replace 8 with datatypes::MAXLEGACYWIDTH.

writeengine/wrapper/writeengine.cpp
```diff
@@ -4278,31 +4571,50 @@ int WriteEngineWrapper::updateColumnRec(const TxnID& txnid,
...
         if (m_opType != DELETE)
             m_opType = UPDATE;
 
-        rc = writeColumnRec(txnid, cscColTypeList, colStructList, colValueList, colOldValueList,
-                            ridLists[extent], tableOid, true, ridLists[extent].size());
+        rc = writeColumnRecUpdate(txnid, cscColTypeList, colStructList, colValueList, colOldValueList,
+                            ridLists[extent], tableOid, true, ridLists[extent].size(), &currentExtentRangesPtrs);
 
-        m_opType = NOOP;
+//        m_opType = NOOP; // XXX: WHY WAS IT HERE???!!!
```
drrtuy: Plz remove if you comment this out.

drrtuy (changes requested):
Plz address the comentaries.

mariadb-SergeyZefirov:
I cannot correctly answer the [concern about branch prediction and switch/case inside a loop](https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/1771#pullrequestreview-598370960) using github UI.

Here's my take. First, the code works alongside deserialisation of boost::any and conversion from boost::any with the same and even greater number of elements processed. The loop inside updateMaxMinRanges can bail out earlier than complete set of rows when range becomes invalid. Second, the approach with std::function wrapper would require two additional template functions (set or not  set int64_t and/or int128_t values) or use of lambda function and, subsequently, the approach will result in more source code and, possibly, even more confused compiler analysis.

I would prefer to leave what I did. And wait for strength reduction of switch/case target computation to be added to compilers.

drrtuy (approved):

PR #1774, opened 2021-02-18
Adding tests for MCOL-4531 New string-to-decimal conversion implement…

abarkov:
…ation

And also for:

- MCOL-4462 CAST(varchar_expr AS DECIMAL(M,N)) returns a wrong result
- MCOL-4500 Bit functions processing throws internally trying to cast char into decimal representation
- MCOL-4532 CAST(AS DECIMAL) returns a garbage for large values

(which were fixed by the same patch)


drrtuy (approved):
The tests looks OK. There are some unrelated issues in MTR cause failures.

PR #1785, opened 2021-02-26
MCOL-4543 Subquery optimization.

tntnatbry:
For a query of the form:

SELECT COUNT(c2) FROM (SELECT * FROM t1) q;

where t1 contains 10 columns c1, c2, ... , c10.

We currently create an intermediate RowGroup in ExeMgr with
a row of the form (1, c2_value1, 1, 1, 1, 1, 1, 1, 1, 1), i.e.
for all the columns of the subquery which are not referenced in
the outer query, we substitute a constant value, which is wasteful.

With this optimization, we are trimming the RowGroup to a row
of the form (1, c2_value1). This can have non-trivial query
execution time improvements if the subquery contains large number
of columns (such as a "select *" on a very wide table) and the outer
query is only referencing a subset of these columns with lower
index values from the subquery (as an example, c1 or c2 above).
That is, the current limitation of this optimization is we are not
removing those non-referenced subquery columns (c1 in the query above)
which are to the left of a referenced column.


tntnatbry:
Please don't merge this yet.

dbcon/mysql/ha_from_sub.cpp
```diff
@@ -93,27 +93,86 @@ void derivedTableOptimization(THD* thd, SCSEP& csep)
...
                 }
+                else if (lastNonConstIndex == -1)
+                {
+                    lastNonConstIndex = i;
+                }
+            }
+
+            if (lastNonConstIndex == -1)
+            {
+                // None of the subquery columns are referenced, just use the first one
```
drrtuy, line 154: We should pick the smallest column not the first IMHO.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -2906,15 +2906,36 @@ SimpleColumn* getSmallestColumn(boost::shared_ptr<CalpontSystemCatalog> csc,
...
+                    if ((*iter)->refCount() != 0)
+                    {
+                        rc = dynamic_cast<ReturnedColumn*>(iter->get());
+                        break;
+                    }
+                }
+
+                if (iter == cols.end())
+                {
+                    rc = dynamic_cast<ReturnedColumn*>(cols[0].get());
```
drrtuy, line 2938: Plz add a comment explains that cols[0] is a dummy ConstantColumn made in derivedTableOptimization in the context of this block.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -2906,15 +2906,36 @@ SimpleColumn* getSmallestColumn(boost::shared_ptr<CalpontSystemCatalog> csc,
 
             if (tan.alias == csep->derivedTbAlias())
             {
-                assert (!csep->returnedCols().empty());
-                ReturnedColumn* rc = dynamic_cast<ReturnedColumn*>(csep->returnedCols()[0].get());
+                const CalpontSelectExecutionPlan::ReturnedColumnList& cols = csep->returnedCols();
+
+                assert (!cols.empty());
```
drrtuy: I hope that we will run MTR using debug builds otherwise this assert effectively does nothing. IMHO better to check cols for emptiness before col[0] will be referenced in the line 2928.

drrtuy (changes requested):
Very minimal changes are requested but in general the patch is good and outlines the scope for future optimization.

dbcon/mysql/ha_mcs_execplan.cpp
tntnatbry: Good catch. I will move the assert later where we reference cols[0].

dbcon/mysql/ha_mcs_execplan.cpp
tntnatbry: Actually, cols[0] will not be a dummy ConstantColumn, since we increment its refCount here. An example query that triggers this codeblock:

```sql
SELECT "123" FROM (SELECT * FROM t1) q;
```

If table `t1` has two columns `a` and `b` in this order. And let's say `t1` has 10 records. Then rc here is a SimpleColumn referencing `a` for correct projection of 10 rows from the subquery.

drrtuy: It is worth explaining for what reason do you blindly take cols[0].

tntnatbry: We take `cols[0]` here due to the optimization happening in derivedTableOptimization. All cols with refCount 0 from the end of the cols list are optimized out, until the first column with non-zero refCount is encountered. So here, if instead of `cols[0]`, we take `cols[1]` (based on some logic) and increment it's refCount, then `cols[0]` is not optimized out in derivedTableOptimization and is added as a ConstantColumn to the derived table's returned column list. This later causes an ineffective row group with row of the form (1, cols[1]_value1) to be created in ExeMgr.

drrtuy: Very good. Could you put the same as the comment in the code?

tntnatbry: Sure, will add a comment here.

drrtuy (approved):

PR #1800, opened 2021-03-08
MCOL-4589 Follow up on MCOL-4543 to optimize out non-referenced

tntnatbry:
columns from a subquery involving UNIONs.


dbcon/mysql/ha_from_sub.cpp
```diff
@@ -139,8 +140,23 @@ void derivedTableOptimization(THD* thd, SCSEP& csep)
 
                     for (uint j = 0; j < unionColVec.size(); j++)
                     {
-                        unionColVec[j][i].reset(new ConstantColumn(val));
-                        (dynamic_cast<ConstantColumn*>(unionColVec[j][i].get()))->timeZone(thd->variables.time_zone->get_name()->ptr());
+                        if (lastNonConstIndex == -1)
+                        {
+                            CalpontSelectExecutionPlan* unionSubPlan =
+                                dynamic_cast<CalpontSelectExecutionPlan*>(plan->unionVec()[j].get());
```
drrtuy: I am curios what is the reason to use dynamic_cast if you don't check unionSubPlan after the cast? Better use reinterpret_cast instead.

tntnatbry: I agree, `reinterpret_cast` is sufficient here as well as at line 201. I have updated the PR.

drrtuy (approved):

PR #1804, opened 2021-03-09
MCOL-4566: Update on review.

denis0x0D:
* Use `literal::UnsignedInteger` instead of `atoi`.
* Combine common code for `_fromDir`, `_fromFile` to `_fromText`.
* Pass `dmFilePathArgs_t` as a const reference instead of pointer,
  don't write result codes to this struct.


denis0x0D:
@abarkov @drrtuy updated on review, can you please take a look. Thanks!

abarkov (approved):
Looks good for me! Thank you very much!

PR #1805, opened 2021-03-09
DO NOT MERGE: MCOL-4440 Related. Check if still connected to old master

jmrojas2332:
There is a case where engine plugin code is still connected to old controllernode, causing communication timeouts. Reconnect if the address of the msg client (possibly previous controllernode) does not match the address of controllernode


jmrojas2332:
Closing. PR should go into develop, not develop-5

PR #1806, opened 2021-03-09
MCOL-4597: fix upgrades 

benthompson15:
logic from postConfigure migrated to post-install to upgrade existing columnstore.cnf files if upgrading


drrtuy:
Do we really need a tool to upgrade a configuration? I would better use a combination of pre-install/post-install hooks of RPM/DEB.

benthompson15:
> Do we really need a tool to upgrade a configuration? I would better use a combination of pre-install/post-install hooks of RPM/DEB.

Tool has been previously verified for every release prior to removal of postConfigure. It is simplest solution to get this functionality back without significant retesting. As discussed offline fix removing this tool for develop branch is planned.

jmrojas2332 (approved):
LGTM

PR #1807, opened 2021-03-09
MCOL-4516 check for var_pop < 0

davidhall586:
In some cases, because of rounding error, var_pop will evaluate to some value just less than 0. We check for this and force to round to 0.

### DIFF
--- a/utils/regr/regr_syy.cpp
+++ b/utils/regr/regr_syy.cpp
@@ -129,6 +129,8 @@ mcsv1_UDAF::ReturnCode regr_syy::evaluate(mcsv1Context* context, static_any::any
     if (N > 0)
     {
         long double var_popy = (data->sumy2 - (data->sumy * data->sumy / N));
+        if (var_popy < 0)  // might be -0
+            var_popy = 0;
         valOut = static_cast<double>(var_popy);
     }
     return mcsv1_UDAF::SUCCESS;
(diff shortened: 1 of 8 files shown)


abarkov (approved):

utils/regr/regr_syy.cpp
```diff
@@ -129,6 +129,8 @@ mcsv1_UDAF::ReturnCode regr_syy::evaluate(mcsv1Context* context, static_any::any
     if (N > 0)
     {
         long double var_popy = (data->sumy2 - (data->sumy * data->sumy / N));
+        if (var_popy < 0)  // might be -0
```
drrtuy, line 132: Could I ask you @dhall-MariaDB to make the non-positive checks consistent? Previously you use LEQ and here strict L.

drrtuy (changes requested):
A tiny change needed.

utils/regr/regr_syy.cpp
davidhall586: Notice that when '<= 0' is used, it is to take an action (almost always return). This is the action defined for '== 0' and we're using '<= 0' to catch micro negative numbers. In these cases, any negative value should be 0 but rounding error in the intermediate may fuzz it. When we use '< 0', it is to convert all micro negative values to 0 and then we continue processing.. This is a different application so consistency exists separately for these two uses.

drrtuy (approved):

PR #1808, opened 2021-03-10
 MCOL-4566: Add rebuildEM tool support to work with compressed files.

denis0x0D:
* This patch adds rebuildEM tool support to work with compressed files.
* This patch adds test, currently test work only for users with `write` access
  to systemcat.

Note: Default version of the `rebuildEM` tool was using very old API,
those functions are not present currently. So `rebuildEM` will not work with
files created without compression, because we cannot deduce some info which are
needed to create column extent.


denis0x0D:
Fixes:
1. Fixed work with old header version.
2. Added support for dictionary segment files.

denis0x0D:
Fixes:
1. Add support of initializing system tables from binary blob.
2. Setting localHWM, make extent - available.
3. Check status of extent before creating.
4. Refactored logic: a) At first we go inside dbroot and collect all needed information (oid, partition, segment, width, coldataype isDict). Store collected data in sorted order via set<(extent info), oidComparator> 2) Than we create extents from collected data.

denis0x0D:
Fixes:
1. Fixed order of extent creation. Currently we keep it sorted by lbid and create in greedy way.

denis0x0D:
This tool is not compatible with `cpimport` tool, because `cpimport` uses `BulkLoad`. Investigating how big changes is needed to support it. That meas if we insert data with `$cpimport temp t1 temp.tbl` we cannot restore the extent map.

denis0x0D:
Added fixes:
1. Proper HWM recovery from segment file.
2. Added support for `bulk` insertion via `cpimport.`

Current limitations - it does not work with multiple extents per segment file.
We can simple detect those kind of files in case we recover `hwm >= (columnWidth * numExtentRows) / blockSizeInBytes` but there is no straight way to create an extent in the same order as it was created originally, because starting lbid is not known for each extent.
Test for different tables schemes, for example:
`create table t1 (a int, b varchar (255), c int, d varchar(255), e int, d varchar(255)) engine=columnstore;`
and insert 20M rows into table via `cpimport`.
In this case bulk will create 3 segment file for each `int` column and 6 for each `varchar` column (1 segment file with tokens and 1 dictionary file).

denis0x0D:
Fixes related to CI build.

denis0x0D:
Final fixes:
1. Proper calculation HWM for dictionary file.
2. Add support for 2 extents per segment files, could be increased if need. Actually bulk can create even more than 2, even if it set to 2 in config.

denis0x0D:
Rebase on develop and add review fixes.

drrtuy (changes requested):
I'd shared some aspects that should be addressed in Slack.

denis0x0D:
Initialize chunkManager with ColumnOpCompress1 and DctnryOpCompress1 instead of default. This does not affect current version of chunkManager but when we have more than 1 compression algo and will get compress interface by compression type, the compression type == 0 could cause an error.

drrtuy (approved):

PR #1810, opened 2021-03-11
Update `dmFilePathArgs_t` struct.

denis0x0D:
* This patch updates `dmFilePathArgs_t` struct to eliminate common code.
* This patch add `dmFilePathPart_t` which represents a part of the full path
  to a segment file.


denis0x0D:
Updated according to suggestions from @abarkov .
@abarkov @drrtuy can you please take a look. Thanks!

drrtuy (approved):

PR #1811, opened 2021-03-12
[MCOL-563] Disk-based aggregation

mariadb-AlexeyAntipovsky:


davidhall586:
Please never merge this. This feature should not be added to develop-5

tools/rgprint/rgprint.cpp
```diff
@@ -0,0 +1,81 @@
+//
```
drrtuy: License.

writeengine/splitter/we_splitterapp.cpp
```diff
@@ -486,6 +486,10 @@ void WESplitterApp::invokeCpimport()
     {
         //we need something that works on Windows as well as linux
         v2[i++] = arg;
+        if (i >= v2.size())
```
drrtuy: Could you pull this in a separate patch?

tools/rgprint/rgprint.cpp
mariadb-AlexeyAntipovsky: Forgot about it. Actually it's just debugging tool, so I've also removed it from tools/CMakeLists.txt

writeengine/splitter/we_splitterapp.cpp
mariadb-AlexeyAntipovsky: https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/1847
https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/1848

dbcon/joblist/tupleaggregatestep.h
```diff
@@ -156,14 +157,37 @@ class TupleAggregateStep : public JobStep, public TupleDeliveryStep
         {}
         void operator()()
         {
-            utils::setThreadName("TASThrAggr");
+            std::string t{"TASThrAggr"};
+            t.append(std::to_string(fThreadID));
+            utils::setThreadName(t.c_str());
+            //utils::setThreadName("TASThrAggr");
```
drrtuy: Plz remove the commented line.

dbcon/joblist/tupleaggregatestep.cpp
```diff
@@ -366,16 +366,18 @@ TupleAggregateStep::TupleAggregateStep(
...
+    fNumOfBuckets = fRm->aggNumBuckets();
+    fNumOfRowGroups = fRm->aggNumRowGroups();
+
     fAggregator->setInputOutput(fRowGroupIn, &fRowGroupOut);
 
     // decide if this needs to be multi-threaded
     RowAggregationDistinct* multiAgg = dynamic_cast<RowAggregationDistinct*>(fAggregator.get());
     fIsMultiThread = (multiAgg || fAggregator->aggMapKeyLength() > 0);
 
     // initialize multi-thread variables
```
drrtuy: Take a comment with the asignments you moved.

dbcon/joblist/tupleaggregatestep.cpp
```diff
@@ -5192,6 +5194,25 @@ void TupleAggregateStep::aggregateRowGroups()
     }
 }
 
+void TupleAggregateStep::threadedAggregateFinalize(uint32_t threadID)
+{
+  for (uint32_t i = 0; i < fNumOfBuckets; ++i)
+  {
+    if (fAgg_mutex[i]->try_lock())
```
drrtuy, line 5204: I am curios about the reasons why try_lock() and not lock() is used here?

utils/rowgroup/rowstorage.h
```diff
@@ -0,0 +1,329 @@
...
+  void freeData();
+
+  /** @brief Move internal data & row position inside [insIdx, startIdx] up by 1.
+   *
+   * @param startIdx(in) last element's index to move
+   * @param insIdx(in)   first element's index to move
+   */
+  void shiftUp(size_t startIdx, size_t insIdx);
+
+  inline static size_t mixHash(size_t x) noexcept
```
drrtuy: I am curios which hash function mixer did you re-use or it was your own and why did you choose uint64_t and not uint32_t as the hash output?

utils/rowgroup/rowstorage.cpp
```diff
@@ -0,0 +1,2139 @@
...
+      h.add(r.getCharset(i), r.getConstString(i));
+      break;
+
+    default:
+      ret = hashData(r.getData() + r.getOffset(i), r.getColumnWidth(i), ret);
+      break;
+    }
+  }
+
+  if (UNLIKELY(h.wasUsed()))
```
drrtuy: I don't think this condition is unlikely for a whole Row.

dbcon/joblist/tupleaggregatestep.h
mariadb-AlexeyAntipovsky: Oh, sorry, I missed this. Actually I added thr number to the thread name for debugging (so I was able to see exactly which thread was working on particular bucket). Do you thing it makes sense for production or should I rollback this?

dbcon/joblist/tupleaggregatestep.cpp
mariadb-AlexeyAntipovsky: This is accidentally left over from the experiments, so I'll rollback it.

dbcon/joblist/tupleaggregatestep.cpp
mariadb-AlexeyAntipovsky: We go through the bucket list and take the first one that isn't currently being finalized, so there is no need to wait for the processing to complete. The state of the race here is resolved by the flag in RowAggStorage.

utils/rowgroup/rowstorage.cpp
mariadb-AlexeyAntipovsky: Yes, you're right.

utils/rowgroup/rowaggregation.cpp
```diff
@@ -751,9 +680,47 @@ void RowAggregation::initialize()
...
+      {
+        allow_gen = false;
+        break;
+      }
+    }
+
+    config::Config* config = config::Config::makeConfig();
+    string tmpDir = config->getConfig("RowAggregation", "TempDir");
+    if (tmpDir.empty() || tmpDir == "/")
+      tmpDir.assign("/tmp/cs-agg");
```
drrtuy: A questionable default since we want to offload the data from RAM and /tmp in most cases is tmpfs. I think it would be better to check if disk-based aggregation is enable and check if RowAggregation.TempDir is set. If not we should throw an error.

utils/rowgroup/rowaggregation.cpp
```diff
@@ -4180,7 +4079,14 @@ void RowAggregationUMP2::doAvg(const Row& rowIn, int64_t colIn, int64_t colOut,
     if (cnt == 0)
     {
         fRow.setLongDoubleField(valIn, colOut);
-        fRow.setUintField(rowIn.getUintField(colIn + 1), colAux);
+        if (merge)
```
drrtuy: What is the difference b/w these two branches?

utils/rowgroup/rowstorage.h
```diff
@@ -0,0 +1,329 @@
...
+  std::string makeDumpFilename(int32_t gen = -1) const;
+
+private:
+  static constexpr size_t   INIT_SIZE{sizeof(uint64_t)};
+  static constexpr uint32_t INIT_INFO_BITS{5};
+  static constexpr uint8_t  INIT_INFO_INC{1U << INIT_INFO_BITS};
+  static constexpr size_t   INFO_MASK{INIT_INFO_INC - 1U};
+  static constexpr uint8_t  INIT_INFO_HASH_SHIFT{0};
+
+  RowPosHashStorage* fHashes{nullptr};
```
drrtuy, line 291: Just a note that this type of initialization breaks exuberant-ctags shipped with U20.

utils/loggingcpp/ErrorMessage.txt
```diff
@@ -100,6 +100,10 @@
 
 2053	ERR_FUNC_OUT_OF_RANGE_RESULT	The result is out of range for function %1% using value(s): %2% %3%
 
+2054	ERR_DISKAGG_ERROR	Unknown error while aggregation.
```
drrtuy, line 103: There is no way to detect which method throws this error. The thow should log at least a method name.

oam/etc/Columnstore.xml.singleserver
```diff
@@ -516,6 +516,7 @@
                <!-- <RowAggrThreads>4</RowAggrThreads> --> <!-- Default value is the number of cores -->
 		<!-- <RowAggrBuckets>32</RowAggrBuckets> --> <!-- Default value is number of cores * 4 -->
 		<!-- <RowAggrRowGroupsPerThread>20</RowAggrRowGroupsPerThread> --> <!-- Default value is 20 -->
+		<!-- <AllowDiskBasedAggregation>N</AllowDiskBasedAggregation> --> <!-- Default value is N -->
```
drrtuy, line 519: JFYI. This file is obsolete and should be removed but not with this patch.

drrtuy (commented):
I've no structural suggestions so far. My main question is why RowHashPos uses uint64_t for its hash and not uint32_t?

utils/rowgroup/rowstorage.h
mariadb-AlexeyAntipovsky: mixHash was left over after using the original Row::hash hasher, I removed it. robinhood hash looks more evenly distributed.
And 64 bits for hash was chosen simply because RH uses 64 bits.

utils/rowgroup/rowaggregation.cpp
mariadb-AlexeyAntipovsky: If it was used while update the aggregated row with the input row it just increments the aggregated value by 1, otherwise it sums the target aggregated value with the source aggregated value.

utils/rowgroup/rowstorage.h
mariadb-AlexeyAntipovsky: It seems that universal-ctags is able to work with modern syntax

drrtuy: Thx for the note.

utils/rowgroup/rowaggregation.cpp
drrtuy: In this particular case the statements in both scopes are identical. Or I am wrong?

drrtuy (approved):

PR #1820, opened 2021-03-23
MCOL-4633 Remove duplicate code for DECIMAL to int64_t rounding conve…

abarkov:
…rsion

Detailed change list:

- Splitting out the narrow part of "class Decimal" into a separate class TDecimal64

- Adding a method TDecimal64::toSInt64Round()

- Reusing the method TDecimal64::toSInt64Round() in:
   * Func_cast_signed::getIntVal()
   * Func_char::getStrVal()
   * Func_elt::getStrVal()
   * makedate()
   * Func_maketime::getStrVal()

   Note, reusing this method in Func_char::getStrVal() also fixed this bug:
     MCOL-4634 CHAR(negativeWideDecimal) is not like InnoDB
   because the old code handled negative wide decimal values
   in a wrong way.

- Adding a new class TDecimal128 for symmetry.
  Moving a few wide decimal methods and constexpr's from Decimal to TDecimal128.
  The new class TDecimal128 does not do much at this point yet.
  Later we should be able to use TDecimal128 vs TDecimal64 in templates.


datatypes/mcs_decimal.h
```diff
@@ -179,41 +179,89 @@ inline lldiv_t_128 lldiv128(const int128_t& dividend, const int128_t& divisor)
...
+    { }
+};
+
+
 // @brief The class for Decimal related operations
 // The class contains Decimal related operations are scale and
 // precision aware.
 // This class will inherit from:
 //      DecimalMeta class that stores scale and precision
 //      Storage classes TSInt128 and int64
```
tntnatbry: Since we don't directly inherit from the storage classes TSInt128 and TSInt64 (future), can we fix this comment? I think we can simple remove this above line, and just keep the "DecimalMeta" comment and change it to "In future, this class will inherit from DecimalMeta that stores scale and precision".

tntnatbry (changes requested):

tntnatbry (approved):

PR #1822, opened 2021-03-24
Fix for failure of unsigned_joins test under centos8

mariadb-SergeyZefirov:
The problem was that column width was not set during BRM reporting stage in cpimport processing. That made column width a garbage, prevented correct ranges to be set and as a result some extents were incorrectly skipped during SELECT operation.

The fixes are in we_brmupdater.cpp: a change to display proper "CP" line format in the comments in the BRM report file, a change to add print of column width into file and a change to parse column width.

### DIFF
--- a/writeengine/splitter/we_brmupdater.cpp
+++ b/writeengine/splitter/we_brmupdater.cpp
@@ -354,6 +354,15 @@ bool WEBrmUpdater::prepareCasualPartitionInfo()
             pTok = strtok(NULL, " ");
 
             if (pTok)
+                cpInfoMerge.colWidth = boost::lexical_cast<int32_t>(pTok);
+            else
+            {
+                //cout << "CP Entry : " << aEntry << endl;
+                throw (runtime_error("Bad column width in CP entry string"));
+            }
+            pTok = strtok(NULL, " ");
+
+            if (pTok)
                 cpInfoMerge.newExtent = (atoi(pTok) != 0);
             else
             {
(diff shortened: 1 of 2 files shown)


drrtuy (approved):

writeengine/splitter/we_brmupdater.cpp
```diff
@@ -353,6 +353,15 @@ bool WEBrmUpdater::prepareCasualPartitionInfo()
...
 
+            if (pTok)
+                cpInfoMerge.colWidth = boost::lexical_cast<int32_t>(pTok);
+            else
+            {
+                //cout << "CP Entry : " << aEntry << endl;
+                throw (runtime_error("Bad column width in CP entry string"));
+            }
+            pTok = strtok(NULL, " ");
+
```
tntnatbry, line 364: We need to add support for 128-bit min/max values as well here. The sending side `BRMReporter::sendCPToFile` either streams `CPInfoMerge::max/CPInfoMerge::min` or `CPInfoMerge::bigMax/CPInfoMerge::bigMin` based on `CPInfoMerge::type` and `CPInfoMerge::colWidth`. So the parser here should first parse out `type` and `colWidth` and then based on these 2 values, should either set the 128-bit or the 64-bit min/max values.

tntnatbry:
Looks like the PR is already merged before my review. I have requested changes in the comment above.

Commit message is also not correct.

I would also add the MCOL number as the first thing in the PR title.

writeengine/splitter/we_brmupdater.cpp
mariadb-SergeyZefirov: Thanks!

It is not possible to have tests for that functionality using joins. I've created task specifically to address your comment: https://jira.mariadb.org/browse/MCOL-4652

PR #1825, opened 2021-03-26
MCOL-4617 Move in-to-exists predicate creation and injection into the engine.

tntnatbry:
We earlier leveraged the server functionality provided by

Item_in_subselect::create_in_to_exists_cond and
Item_in_subselect::inject_in_to_exists_cond

to create and inject the in-to-exists predicate into an IN
subquery's JOIN struct. With this patch, we leave the IN subquery's
JOIN unaltered and instead directly perform this predicate creation
and injection into ColumnStore's select execution plan.

### DIFF
--- a/dbcon/mysql/ha_in_sub.cpp
+++ b/dbcon/mysql/ha_in_sub.cpp
@@ -41,6 +41,7 @@ using namespace std;
 #include "simplefilter.h"
 #include "predicateoperator.h"
 #include "rowcolumn.h"
+#include "constantcolumn.h"
 using namespace execplan;
 
 #include "errorids.h"
@@ -101,6 +102,17 @@ InSub::InSub(const InSub& rhs) :
 InSub::~InSub()
 {}
 
+inline void setCorrelatedFlag(execplan::ReturnedColumn* col, gp_walk_info* gwip)
+{
+    ConstantColumn* cc = dynamic_cast<ConstantColumn*>(col);
+
+    if (!cc)
+    {
+        col->joinInfo(col->joinInfo() | JOIN_CORRELATED | JOIN_NULLMATCH_CANDIDATE);
+        gwip->subQuery->correlated(true);
+    }
+}
+
 /** MySQL transform (NOT) IN subquery to (NOT) EXIST
  *
  */
@@ -125,10 +137,9 @@ execplan::ParseTree* InSub::transform()
     delete rhs;
     ReturnedColumn* lhs = fGwip.rcWorkStack.top();
     fGwip.rcWorkStack.pop();
-    delete lhs;
 
     fSub = (Item_subselect*)(fFunc->arguments()[1]);
-    idbassert(fSub && fFunc);
+    idbassert(fSub);
 
     SCSEP csep (new CalpontSelectExecutionPlan());
     csep->sessionID(fGwip.sessionid);
@@ -139,6 +150,24 @@ execplan::ParseTree* InSub::transform()
     gp_walk_info gwi;
     gwi.thd = fGwip.thd;
     gwi.subQuery = this;
+
+    // The below 2 fields are used later on in buildInToExistsFilter()
+    gwi.inSubQueryLHS = lhs;
+    gwi.inSubQueryLHSItem = fFunc->arguments()[0];
+
+    RowColumn* rlhs = dynamic_cast<RowColumn*>(lhs);
+
+    if (rlhs)
+    {
+        for (auto& col : rlhs->columnVec())
+        {
+            setCorrelatedFlag(col.get(), &gwi);
+        }
+    }
+    else
+    {
+        setCorrelatedFlag(lhs, &gwi);
+    }
 
     // @4827 merge table list to gwi in case there is FROM sub to be referenced
     // in the FROM sub
--- a/dbcon/mysql/ha_mcs_execplan.cpp
+++ b/dbcon/mysql/ha_mcs_execplan.cpp
@@ -1677,6 +1677,180 @@ bool buildRowColumnFilter(gp_walk_info* gwip, RowColumn* rhs, RowColumn* lhs, It
     return true;
 }
 
+void checkOuterTableColumn(gp_walk_info* gwip,
+    const CalpontSystemCatalog::TableAliasName& tan,
+    execplan::ReturnedColumn* col)
+{
+    set<CalpontSystemCatalog::TableAliasName>::const_iterator it;
+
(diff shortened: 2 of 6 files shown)


davidhall586 (commented):
In buildInToExistsFilter(), It calls buildEqualityPredicate(). It's possible for buildEqualityPredicate() to find an error and set gwip->fatalParseError. Should we check for this condition in the caller?

I may be missing something, but the code at line 1804 appears duplicated at line 1821. I realize this is old code, but does it do anything useful the second time?

drrtuy (commented):
Could you ellaborate on why there is no examples with correlated subquery in the test file?

dbcon/mysql/ha_in_sub.cpp
```diff
@@ -125,10 +137,9 @@ execplan::ParseTree* InSub::transform()
     delete rhs;
     ReturnedColumn* lhs = fGwip.rcWorkStack.top();
     fGwip.rcWorkStack.pop();
-    delete lhs;
 
     fSub = (Item_subselect*)(fFunc->arguments()[1]);
-    idbassert(fSub && fFunc);
+    idbassert(fSub);
```
drrtuy, line 142: This is a poor man way to send the related fFunc into the log so plz left it intact or suggest your way to relay this information.

dbcon/mysql/ha_mcs_impl_if.h
```diff
@@ -177,7 +183,8 @@ struct gp_walk_info
         inCaseStmt(false),
         cs_vtable_is_update_with_derive(false),
         cs_vtable_impossible_where_on_union(false),
-        isGroupByHandler(false)
+        isGroupByHandler(false),
+        inSubQueryLHS(0)
```
drrtuy: There should be a default for inSubQueryLHSItem

drrtuy: Plz use nullptr instead of 0 constant.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -1677,6 +1677,188 @@ bool buildRowColumnFilter(gp_walk_info* gwip, RowColumn* rhs, RowColumn* lhs, It
     return true;
 }
 
+bool buildEqualityPredicate(execplan::ReturnedColumn* lhs,
+    execplan::ReturnedColumn* rhs,
+    gp_walk_info* gwip,
+    boost::shared_ptr<Operator>& sop,
+    Item_func::Functype funcType,
```
drrtuy: I don't see that we change this argument so it should be const.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -6849,6 +6868,115 @@ int processLimitAndOffset(
...
+            (rlhs->columnVec().size() == gwi.returnedCols.size()) &&
+            row->cols() && (row->cols() == select_lex.item_list.elements) &&
+            (row->cols() == gwi.returnedCols.size()));
+
+        List_iterator_fast<Item> it(select_lex.item_list);
+        Item* item;
+
+        int i = 0;
+
+        ParseTree* rowFilter = 0;
```
drrtuy: nullptr plz.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -6849,6 +6868,115 @@ int processLimitAndOffset(
...
+        for (size_t i = 0; i < gwi.returnedCols.size(); i++)
+        {
+            gwi.returnedCols[i]->expressionId(
+                gwi.additionalRetCols[additionalRetColsBefore + i]->expressionId());
+            gwi.returnedCols[i]->colSource(
+                gwi.additionalRetCols[additionalRetColsBefore + i]->colSource());
+        }
+
+        // Delete the duplicate copy of the returned cols
+        auto iter = gwi.additionalRetCols.begin() + additionalRetColsBefore;
```
drrtuy: Plz use std::advance here.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -1677,6 +1677,188 @@ bool buildRowColumnFilter(gp_walk_info* gwip, RowColumn* rhs, RowColumn* lhs, It
...
+    const vector<Item*>& itemList,
+    bool isInSubs)
+{
+    cal_connection_info* ci = reinterpret_cast<cal_connection_info*>(get_fe_conn_info_ptr());
+
+    // push the column that is associated with the correlated column to the returned
+    // column list, so the materialized view have the complete projection list.
+    // e.g. tout.c1 in (select tin.c1 from tin where tin.c2=tout.c2);
+    // the projetion list of subquery will have tin.c1, tin.c2.
+    ReturnedColumn* correlatedCol = NULL;
```
drrtuy: plz use nullptr.

drrtuy:
I don't see any serious functional issue so plz address the comments or comment back so I can proceed.

drrtuy:
You should also rebase the patch on top of the current develop so that MTR failure goes away.

dbcon/mysql/ha_in_sub.cpp
tntnatbry: This is not the case. `idbassert(fSub && fFunc)` will print the string `"fsub && fFunc"` in the log, rather than the actual values of the arguments. Since we already check for `fFunc` at the start of the function, we don't need to add an assertion check on `fFunc` here again.

dbcon/mysql/ha_mcs_impl_if.h
tntnatbry: Done.

dbcon/mysql/ha_mcs_execplan.cpp

dbcon/mysql/ha_mcs_execplan.cpp

dbcon/mysql/ha_mcs_execplan.cpp

dbcon/mysql/ha_mcs_execplan.cpp

tntnatbry:
@dhall-MariaDB I have addressed both your comments, thanks for highlighting them. Please review again.

@drrtuy I have added test case for correlated subquery under the comment `### Correlated IN subquery`, as well, also addressed most of your additional comments. Please review again.

davidhall586 (approved):

dbcon/mysql/ha_in_sub.cpp
drrtuy: Agree.

PR #1837, opened 2021-03-31
MCOL-987 Refactor compression interface.

denis0x0D:
This patch changes compression interface. Motivation is to be able to add
new compression methods.

* All methods which use static data and do not modify any internal data - become `static`,
  so we can use them without creation of the specific object. This is possible, because
  the header specification has not been modified. We still use 2 sections in header, first
  one with file meta data, the second one with pointers for compressed chunks.

* Methods `compress`, `uncompress`, `maxCompressedSize`, `getUncompressedSize` - become
  pure virtual, so we can override them for the other compression algos.

* Adds method `getChunkMagicNumber`, so we can verify chunk magic number
  for each compression algo.

* Renames "s/IDBCompressInterface/CompressInterface/g" according to requirement.

### DIFF
--- a/dbcon/joblist/pcolstep.cpp
+++ b/dbcon/joblist/pcolstep.cpp
@@ -146,9 +146,7 @@ pColStep::pColStep(
     if (fOid < 1000)
         throw runtime_error("pColStep: invalid column");
 
-    compress::IDBCompressInterface cmpif;
-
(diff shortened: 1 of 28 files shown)


denis0x0D:
Moved to: https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/1842

PR #1839, opened 2021-04-01
Fixing DOUBLE-to-[U]INT conversion (MCOL-4649, MCOL-4631, MCOL-4647)

abarkov:
Bugs fixed:
- MCOL-4649 CAST(double AS UNSIGNED) returns 0
- MCOL-4631 CAST(double AS SIGNED) returns 0 or NULL
- MCOL-4647 SEC_TO_TIME(double_or_float) returns a wrong result

Problems:
- The code in Func_cast_unsigned::getUintVal() and
  Func_cast_signed::getIntVal() did not properly check
  the double value to fit inside a uint64_t/int64_t range.
  So the corner cases:
  - numeric_limits<uint64_t>::max()-2 for uint64_t
  - numeric_limits<int64_t>::max() for int64_t
  produced unexpected results.

  The problem was in tests like this:
    if (value > (double) numeric_limits<int64_t>::max())
  A correct test would be:
    if (value >= (double) numeric_limits<int64_t>::max())

- The code in Func_sec_to_time::getStrVal() searched for the decimal
  dot character, assuming that the next character after the dot
  was the leftmost fractional digit.
  This assumption was wrong because huge double numbers use
  scientific notation. So for example in "2.5e-40" the
  digit "5" following the dot is NOT the leftmost fractional digit.
  Also, the code in Func_sec_to_time::getStrVal() was slow
  because of using non necessary to-string and from-string
  data conversion.
  Also, the code in Func_sec_to_time::getStrVal() evaluated
  the argument two times: using getStrVal() then using getIntVal().

Solution:
- Adding new classes TDouble and TLongDouble.
- Adding a few function templates to reuse the code easier.
- Moving the conversion code inside TDouble and TLongDouble
  methods toMCSSInt64Round() and toMCSUInt64Round().
- Reusing new classes and their methods in func_cast.cc and
  func_sec_to_time.cc.


mtr/basic/r/func_cast.result
```diff
@@ -0,0 +1,20 @@
...
+INSERT INTO t1 VALUES (18446744073709551614,18446744073709551614);
+SELECT d1, CAST(d1 AS SIGNED), CAST(d2 AS SIGNED) FROM t1;
+d1	CAST(d1 AS SIGNED)	CAST(d2 AS SIGNED)
+9.223372036854776e18	9223372036854775807	9223372036854775807
+1.8446744073709552e19	9223372036854775807	9223372036854775807
+DROP TABLE t1;
+#
+# MCOL-4649 MCOL-4631 CAST(double AS UNSIGNED) returns 0
+#
+CREATE TABLE t1 (d1 DOUBLE, d2 DOUBLE NOT NULL) ENGINE=ColumnStore;
```
tntnatbry: I think we shouldn't add `ENGINE=ColumnStore` here.

tntnatbry (changes requested):

mtr/basic/r/func_cast.result
abarkov: Thanks for a good catch!

I erroneously copied this from the how-to-repeat script in the MCOL.
There should not be the ENGINE clause - this test is combination aware (it runs two times - with MyISAM and ColumnStore as a default engine).

tntnatbry (approved):

PR #1841, opened 2021-04-02
A joint patch for MCOL-4614 and MCOL-4615 (decimal to string conversion)

abarkov:
This patch fixes:
- MCOL-4614 calShowPartitions() precision loss for huge narrow decimal
- MCOL-4615 GROUP_CONCAT() precision loss for huge narrow decimal

Changes:
- Implementing Row::getDecimalField()

- Removing double arithmetic from the code printing DECIMAL values
  in TypeHandlerXDecimal::format64() and GroupConcator::outputRow().
  Using Decimal::toString() instead.

An additional cleanup:
- Removing the ENGINE=COLUMNSTORE clause from tests for MCOL-4532 and MCOL-4640
  type_decimal.test is combinations-aware. It's run two times with
  default_storage_engine=MyISAM and default_storage_engine=COLUMNSTORE.
  So the CREATE TABLE statements should not specify the engine explicitly.
- Adding --disable_warnings in the old fixed test.
  We needed to suppress warnings when the MyISAM combination is being run.
  Previously the table was erroneously created with ENGINE=COLUMNSTORE
  even with the MyISAM combination run. So warning were not generated.


dbcon/joblist/groupconcat.cpp
```diff
@@ -500,8 +476,8 @@ void GroupConcator::outputRow(std::ostringstream& oss, const rowgroup::Row& row)
                 }
                 else
                 {
-                    long double dblVal = uintVal / pow(10.0, (double)scale);
-                    oss << fixed << setprecision(scale) << dblVal;
+                    oss << fixed << datatypes::Decimal(datatypes::TSInt64((int64_t) uintVal),
+                                                       scale, row.getPrecision(*i));
```
tntnatbry: I think casting an `uint64_t` to a `int64_t` here can overflow, we should use the `TSInt128` version here.

tntnatbry (changes requested):

dbcon/joblist/groupconcat.cpp
abarkov: It should not overflow. scale is not zero. So the number is at least 10 times smaller. It fits into int64_t.
I think there should be a comment about this.

tntnatbry (approved):

PR #1842, opened 2021-04-02
MCOL-987 LZ4 compression support.

denis0x0D:
LZ4 compression support.


denis0x0D:
Note, I added LZ4 as default by separate commit on top on this pull request to trigger test suite under LZ4 compression, this patch should be removed, before merging.

utils/compress/idbcompress.cpp
```diff
@@ -107,53 +108,58 @@ namespace compress
...
-                                        unsigned int&  outLen) const
+int CompressInterface::compressBlock(const char* in, const size_t inLen,
+                                     unsigned char* out, size_t& outLen) const
 {
     size_t snaplen = 0;
     utils::Hasher128 hasher;
 
     // loose input checking.
-    if (outLen < snappy::MaxCompressedLength(inLen) + HEADER_SIZE)
+    if (outLen < maxCompressedSize(inLen) + HEADER_SIZE)
```
mariadb-AlexeyAntipovsky: HEADER_SIZE is added twice. Is that how it's meant to be?

utils/compress/idbcompress.cpp
```diff
@@ -180,19 +186,20 @@ int IDBCompressInterface::uncompressBlock(const char* in, const size_t inLen, un
     uint8_t storedMagic;
     utils::Hasher128 hasher;
 
-    outLen = 0;
 
     if (inLen < 1)
     {
+        outLen = 0;
```
mariadb-AlexeyAntipovsky: It seems to make sense to put the resetting of `outLen` at the beginning of the function and remove it from everywhere except line 227

utils/joiner/joinpartition.h
```diff
@@ -164,7 +164,7 @@ class JoinPartition
 
     /* Compression support */
     bool useCompression;
-    compress::IDBCompressInterface compressor;
+    std::shared_ptr<compress::CompressInterface> compressor;
```
mariadb-AlexeyAntipovsky, line 167: `std::unique_ptr` isn't enough here?

utils/messageqcpp/compressed_iss.h
```diff
@@ -54,7 +54,7 @@ class CompressedInetStreamSocket : public InetStreamSocket
     virtual const IOSocket accept(const struct timespec* timeout);
     virtual void connect(const sockaddr* addr);
 private:
-    compress::IDBCompressInterface alg;
+    std::shared_ptr<compress::CompressInterface> alg;
```
mariadb-AlexeyAntipovsky, line 57: Same question about `unique_ptr`

oam/etc/Columnstore.xml
```diff
@@ -494,6 +494,7 @@
 			files are left behind.
 		<TempFilePath>/tmp/cs-diskjoin</TempFilePath>  -->
 		<TempFileCompression>Y</TempFileCompression>
+    <TempFileCompressionType>2</TempFileCompressionType>
```
mariadb-AlexeyAntipovsky: Wouldn't it be better to use symbolic names?

utils/compress/idbcompress.h
```diff
@@ -276,112 +290,229 @@ class IDBCompressInterface
...
+EXPORT CompressInterface*
+getCompressInterfaceByType(uint32_t compressionType,
+                           uint32_t numUserPaddingBytes = 0);
+
+/**
+ *  Initializes a given `unordered_map` with all available compression
+ *  interfaces.
+ */
+EXPORT void initializeCompressorPool(
+    std::unordered_map<uint32_t, std::shared_ptr<CompressInterface>>&
```
mariadb-AlexeyAntipovsky: IMHO, we need a short alias for `std::unordered_map<uint32_t, std::shared_ptr<CompressInterface>>`

mariadb-AlexeyAntipovsky (changes requested):

denis0x0D:
@mariadb-AlexeyAntipovsky thanks for review! I'll address your comments.

utils/compress/idbcompress.cpp
denis0x0D: Thanks!

utils/compress/idbcompress.cpp
denis0x0D: Thanks for point on this, I see that the function signature is different than function semantics, which cause misunderstanding. Basically LZ4 1. requires a size of uncompressed buffer (the buffer where the decompressed data will be written) and 2. does not have `getUncompressedSize`. Originally `uncomressBlock` wrote the size of uncompressed buffer which was received from `getUncompressedSize` to `outLen`. Basically we don't need to write the size of uncompressed buffer, because CS operates on chunks (`uncompressed buffer`) with fixed size. For initial abbreviated extent - initial chunk is about 1m (depends on row width) and regular chunk is about 4m, so we know the size of uncompressed buffer in any time for segment files. (Does not work for socket and joinpartitions). So, the `outLen` actually contains the size of output buffer, which we cannot set to 0 at the beginning. Anyway, I think it was my mistake to keep old function signature, I'll update it, and make `outLen` param as a value, not a reference.

utils/joiner/joinpartition.h
denis0x0D: It has the copy constructor, https://github.com/mariadb-corporation/mariadb-columnstore-engine/blob/develop/utils/joiner/joinpartition.cpp#L109 as far as I know we cannot copy `unique_ptr`, please fix me if I'm wrong.

utils/messageqcpp/compressed_iss.h
denis0x0D: has default copy cosntructor https://github.com/mariadb-corporation/mariadb-columnstore-engine/blob/develop/utils/messageqcpp/compressed_iss.h#L45 , but I'm not an expert in smart pointers, just seems to me `unique_ptr` does not have copy constructor, please fix me if I'm wrong.

oam/etc/Columnstore.xml
denis0x0D: Sure, thanks!

utils/compress/idbcompress.h
denis0x0D: Sure, any suggestions?

utils/joiner/joinpartition.h
mariadb-AlexeyAntipovsky: Yes, you're right, I missed that point.

utils/compress/idbcompress.h
mariadb-AlexeyAntipovsky: `CompressorPool` maybe?

denis0x0D: sure, thanks!

utils/compress/idbcompress.cpp
mariadb-AlexeyAntipovsky: I'm a little worried about writeengine/bulk/we_colbufcompressed.cpp:766, where the outLen is saved for future use (?). Will everything be all right there?

denis0x0D: I think so, this function uncompress chunk based on the chunk index, to find the appropriate chunk index it uses `locateBlock`, if the chunk size is different against hardcoded (4m) it will locate the wrong chunk.
```

470 void CompressInterface::locateBlock(unsigned int block,
471                                     unsigned int& chunkIndex,
472                                     unsigned int& blockOffsetWithinChunk) const
473 {
474     const uint64_t BUFLEN  = UNCOMPRESSED_INBUF_LEN;
475 
476     uint64_t byteOffset    = (uint64_t)block * BLOCK_SIZE;
477     uint64_t chunk         = byteOffset / BUFLEN;
478     uint64_t blockInChunk  = (byteOffset % BUFLEN) / BLOCK_SIZE;
479 
480     chunkIndex             = chunk;
481     blockOffsetWithinChunk = blockInChunk;
482 }

```

denis0x0D: Anyway thanks for point on this, looking more detailed on the code, after the call of this function we have 
```
228     // Expand the compression buffer size if working with an abbrev extent, and
229     // the bytes we are about to add will overflow the abbreviated extent.
230     if ((fToBeCompressedCapacity < CompressInterface::UNCOMPRESSED_INBUF_LEN) &&
231             ((fNumBytes + writeSize + fillUpWEmptiesWriteSize) > fToBeCompressedCapacity) )
232     {
233         std::ostringstream oss;
234         oss << "Expanding abbrev to-be-compressed buffer for: OID-" <<
235             fColInfo->curCol.dataFile.fid <<
236             "; DBRoot-"   << fColInfo->curCol.dataFile.fDbRoot    <<
237             "; part-"     << fColInfo->curCol.dataFile.fPartition <<
238             "; seg-"      << fColInfo->curCol.dataFile.fSegment;
239         fLog->logMsg( oss.str(), MSGLVL_INFO2 );
240         fToBeCompressedCapacity = CompressInterface::UNCOMPRESSED_INBUF_LEN;
241     }
```

Probably locateBlock works, because block == 0 for abbreviated extend. But currently not sure, could it brake something or not.

mariadb-AlexeyAntipovsky: Thanks for explaining!

utils/compress/idbcompress.cpp
denis0x0D: oh, sorry, forgot to update it.

mariadb-AlexeyAntipovsky (approved):

utils/compress/idbcompress.cpp
denis0x0D: @drrtuy can you please comment on part of code which @mariadb-AlexeyAntipovsky found (thanks a lot for this!), does the new api break something or not, because it seems to me not, but I'm not sure.
The new API does not modify the `outLen` for `uncompressBlock` because LZ4 does not have the function `getUncompressedSize`, and it seems it all should work fine, because the chunk size is hardcoded.

The code in which I'm not super sure are:

bulk load.
```
742         // Uncompress the chunk into our 4MB buffer
743         size_t outLen = CompressInterface::UNCOMPRESSED_INBUF_LEN;
744         int rc = fCompressor->uncompressBlock(
745                      compressedOutBuf,
746                      fChunkPtrs[chunkIndex].second,
747                      fToBeCompressedBuffer,
748                      outLen);

768         fToBeCompressedCapacity = outLen;
769 
for the function
nt ColumnBufferCompressed::initToBeCompressedBuffer
```
it saves the `outLen` and stores it to `fToBeCompressedCapacity`. 

Then it checks for abbreviated extent:

```
228     // Expand the compression buffer size if working with an abbrev extent, and
229     // the bytes we are about to add will overflow the abbreviated extent.
230     if ((fToBeCompressedCapacity < CompressInterface::UNCOMPRESSED_INBUF_LEN) &&
231             ((fNumBytes + writeSize + fillUpWEmptiesWriteSize) > fToBeCompressedCapacity) )
232     {
233         std::ostringstream oss;
234         oss << "Expanding abbrev to-be-compressed buffer for: OID-" <<
235             fColInfo->curCol.dataFile.fid <<
236             "; DBRoot-"   << fColInfo->curCol.dataFile.fDbRoot    <<
237             "; part-"     << fColInfo->curCol.dataFile.fPartition <<
238             "; seg-"      << fColInfo->curCol.dataFile.fSegment;
239         fLog->logMsg( oss.str(), MSGLVL_INFO2 );
240         fToBeCompressedCapacity = CompressInterface::UNCOMPRESSED_INBUF_LEN;
241     `}`
```

As far as I understood the chunk size could be different than 4m only for abbreviated extent and after first write it expands to the full extent. (Updated: was wrong, it's abbreviated until it has empty rows in abbreviated size)
Also this part of code relies on function `locateBlock`, to find `chunk index`, not sure how it possible to find appropriate chunk, if written chunk has len different than 4m using this function

```
470 void CompressInterface::locateBlock(unsigned int block,
471                                     unsigned int& chunkIndex,
472                                     unsigned int& blockOffsetWithinChunk) const
473 {
474     const uint64_t BUFLEN  = UNCOMPRESSED_INBUF_LEN;
475 
476     uint64_t byteOffset    = (uint64_t)block * BLOCK_SIZE;
477     uint64_t chunk         = byteOffset / BUFLEN;
478     uint64_t blockInChunk  = (byteOffset % BUFLEN) / BLOCK_SIZE;
479 
480     chunkIndex             = chunk;
481     blockOffsetWithinChunk = blockInChunk;
482 }

```

Tried different cases to trigger this part of code using `develop` branch and this branch
and chunk size is 4194304
Insert 1 row using cpimort into empty column, or into not empty column, and all work fine.
Updated (was wrong, if we insert 1 row into empty column it will use abbreviated size)

Could you please comment on this change?
Thanks!

Updated:
only problem I found, after updating an API, we will store the chunk with full size instead of `abbreviated chunk` which is 1MB. May be it's not important, because `cpimport` is using for insertion million of rows.
Workaroud could be, to update a header for compressed chunk and store uncompressed size, but it will break a support for old files.

denis0x0D: @mariadb-AlexeyAntipovsky @drrtuy found how to fix it, @mariadb-AlexeyAntipovsky thanks for review!

dbcon/mysql/ha_mcs_ddl.cpp
```diff
@@ -981,7 +980,9 @@ int ProcessDDLStatement(string& ddlStatement, string& schema, const string& tabl
 
                     if (compressionType == 1) compressionType = 2;
```
mariadb-AlexeyAntipovsky, line 981: Can we use not magic numbers here and below in all similar places, but enum values, some function or something like that?

primitives/primproc/primitiveserver.cpp
```diff
@@ -696,13 +696,25 @@ void loadBlock (
                 i = fp->pread( &cmpHdrBuf[0], 0, 4096 * 3);
```
mariadb-AlexeyAntipovsky, line 696: 4096 here and below is some random number or HDR_BUF_LEN?

utils/compress/idbcompress.cpp
mariadb-AlexeyAntipovsky: I'm sorry for unresolving this but it seems that it would be better to store original `outLen` into a local variable, set `outLen` to 0 and remove zeroing every where. So the actual uncompress part (lines 222-229) will be something like
```
        auto rc = uncompress(&in[HEADER_SIZE], storedLen,
                             reinterpret_cast<char*>(out), &tmpOutLen);
        if (rc == ERR_OK)
        {
            outLen = tmpOutLen;
        }
        else
        {
            cerr << "uncompressBlock failed!" << endl;
            return ERR_DECOMPRESS;
        }
```

utils/messageqcpp/compressed_iss.cpp
```diff
@@ -75,6 +76,22 @@ CompressedInetStreamSocket::CompressedInetStreamSocket()
...
+    catch (...) { }
+
+    if (compressionType == "LZ4")
+    {
+        alg.reset(new compress::CompressInterfaceLZ4());
+    }
+    else
+    {
+        alg.reset(new compress::CompressInterfaceSnappy());
+    }
```
mariadb-AlexeyAntipovsky: Could you please make it a separate function (for example, in the utils/configcpp or utils/compress) that would also accept "off" and return an error if something unknown is written in the value? It would be useful for disk aggregation/joins/etc temporary files.

writeengine/bulk/we_colbufcompressed.cpp
```diff
@@ -377,21 +381,31 @@ int ColumnBufferCompressed::writeToFile(int startOffset, int writeSize,
 //------------------------------------------------------------------------------
 int ColumnBufferCompressed::compressAndFlush( bool bFinishingFile )
 {
-    const int OUTPUT_BUFFER_SIZE = IDBCompressInterface::maxCompressedSize(fToBeCompressedCapacity) +
-                                   fUserPaddingBytes;
+    auto fCompressor = compress::getCompressorByType(
```
mariadb-AlexeyAntipovsky: `fCompressor` is a bit confusing, since the prefix `f` is usually used for class/struct fields. Could you please rename it (just `compressor`?) here and below?

mariadb-AlexeyAntipovsky (changes requested):

dbcon/mysql/columnstore_info.sql
```diff
@@ -95,7 +95,11 @@ DROP PROCEDURE IF EXISTS `compression_ratio` //
 
 CREATE PROCEDURE compression_ratio() SQL SECURITY INVOKER
 BEGIN
-SELECT CONCAT((SELECT SUM(data_size) FROM information_schema.columnstore_extents ce left join information_schema.columnstore_columns cc on ce.object_id = cc.object_id where compression_type='Snappy') / (SELECT SUM(compressed_data_size) FROM information_schema.columnstore_files WHERE compressed_data_size IS NOT NULL), ':1') COMPRESSION_RATIO;
+
+SELECT CONCAT((SELECT SUM(data_size) FROM information_schema.columnstore_extents ce left join information_schema.columnstore_columns cc on ce.object_id = cc.object_id where compression_type='Snappy') / (SELECT SUM(compressed_data_size) FROM information_schema.columnstore_files co left join information_schema.columnstore_columns cc on (co.object_id = cc.object_id) left join information_schema.columnstore_extents ce on (ce.object_id = co.object_id) where compression_type='Snappy' and compressed_data_size IS NOT NULL /* could be a situation when compressed_data_size != NULL but data_size == 0, in this case we will get wrong ratio */ and data_size > 0), ':1') COMPRESSION_RATIO_SNAPPY;
```
drrtuy: I think we need to put a literal column to tell which compression type is this and combine two SELECTs into a single UNION if UNION doesn't crash with queries from I_S :)

dbcon/mysql/ha_mcs_ddl.cpp
denis0x0D: We can set a compression type by session variable "columnstore_compression_type=1" or "2" or whatever we have.

Current distribution of compression types:
`
SNAPPY = 0,
SNAPPY = 1,
SNAPPY = 2,
LZ4 = 4,
`
Not sure how to make an enum from this.
Do you mean like
`
SNAPPY0 = 0
SNAPPY1 = 1
SNAPPY2 =2
`

dbcon/mysql/columnstore_info.sql
denis0x0D: I don't know an SQL good enough, can you please show an example, thanks.

denis0x0D:
@mariadb-AlexeyAntipovsky @drrtuy thanks for review, updated. For some comments I've added a questions. Thanks.

dbcon/mysql/ha_mcs_ddl.cpp
mariadb-AlexeyAntipovsky: I mean a function that maps old bad int compression type to the enum value, for example. Or, if it's possible, do that mapping while setting/getting variable.

denis0x0D: Sorry If I didn't get it correctly, do you mean to change the `int type` compression type everywhere in source to `enum type`? 
If, so, may be we can separate this to another pull request, because the change could be a huge, the `int type` for compression is using in a lot of place in some structs, classes and so on.
BTW, I totally agree that a lot of places in CS should be rewritten, my opinion that it would be we better if we can separate this work from "feature" patches, but I might be wrong about it.
Thanks.

drrtuy: I told Alexey to address this. However you are right and this looks like a big patch that is outside the scope.

utils/compress/idbcompress.cpp
denis0x0D: Thanks, fixed.

utils/messageqcpp/compressed_iss.cpp
denis0x0D: thanks, fixed.

writeengine/bulk/we_colbufcompressed.cpp
denis0x0D: Thanks, fixed.

primitives/primproc/primitiveserver.cpp
denis0x0D: Seems like it's HDR_BUF_LEN, I'm not sure, why author wrote 4096, instead of HDR_BUF_LEN.

mariadb-AlexeyAntipovsky (approved):

drrtuy (approved):

PR #1846, opened 2021-04-03
Deb: Define Breaks/Replaces relationship on package libthrift-dev

ottok:
The package libthrift-dev contains a file with the same name and
path as a file in mariadb-plugin-columnstore. The two packages
cannot be installed at the same time and by defining this
relationship in the debian/control file, apt will handle the case
automatically.

Error detected with check_for_missing_breaks_replaces.py:
  [ERROR] mariadb-plugin-columnstore conflicts with libthrift-dev
  files: {'/usr/lib/x86_64-linux-gnu/libthrift.so'}

Also ran 'wrap-and-sort -av' to ensure all debian/* contents is in
a deterministic order (=alphabetic).

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -8,9 +8,11 @@ Depends: binutils,
          python3,
          ${misc:Depends},
          ${shlibs:Depends}
-Breaks: mariadb-columnstore-libs,
+Breaks: libthrift-dev,
+        mariadb-columnstore-libs,
         mariadb-columnstore-platform
-Replaces: mariadb-columnstore-libs,
+Replaces: libthrift-dev,
+          mariadb-columnstore-libs,
           mariadb-columnstore-platform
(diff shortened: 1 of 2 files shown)


ottok:
CI seems to fail on some unrelated issue?
image

drrtuy:
Thx for the notification about the libthrift. JFYI We don't use it now actually so we will remove the outdated libthrift source code we have in the repo. I'll share an issue to follow on this.

drrtuy:
> CI seems to fail on some unrelated issue?
> image

We'll address this.

drrtuy:
[Here](https://jira.mariadb.org/browse/MCOL-316) is the issue about the future of Thrift in the MCS code.

drrtuy (approved):

drrtuy:
Thank you for the patch @ottok !

PR #1849, opened 2021-04-05
MCOL-4666 Empty set when using BIT OR and BIT AND functions in WHERE

abarkov:


dbcon/execplan/functioncolumn.h
```diff
@@ -309,10 +309,9 @@ class FunctionColumn : public ReturnedColumn
         return fFunctor->getTimeIntVal(row, fFunctionParms, isNull, fOperationType);
     }
 
-    bool fixIfNeeded() override
+    bool fix()
     {
-        if (fFixed)
-            return false;
+        idbassert(!fFixed);
```
tntnatbry: 1. I think we can make this function a private member.
2. Return type can be set to void.
3. The `fFixed` member variable looks redundant.

tntnatbry (changes requested):

tntnatbry (approved):

PR #1852, opened 2021-04-07
Remove unneeded second call to isEmptyVal()

mariadb-AlexeyAntipovsky:


primitives/linux-port/column.cpp
```diff
@@ -778,7 +778,7 @@ inline int64_t nextColValue(int type,
 
         vp = &val8[*index * W];
         *isNull = isNullVal<W>(type, vp);
-        *isEmpty = isEmptyVal<W>(type, vp);
+        *isEmpty = (OutputType & OT_RID) ? false : isEmptyVal<W>(type, vp);
```
drrtuy, line 712: Could you explain me offline why can we blindly set isEmpty to false if OutputType doesn't have OT_RID set ?

drrtuy (approved):

drrtuy (approved):

PR #1856, opened 2021-04-09
MCOL-4361 Replace pow(10.0, (double)scale) expressions with a static …

abarkov:
…dictionary lookup.

### DIFF
--- a/dbcon/joblist/tupleunion.cpp
+++ b/dbcon/joblist/tupleunion.cpp
@@ -479,45 +479,22 @@ void TupleUnion::normalize(const Row& in, Row* out)
                     case CalpontSystemCatalog::FLOAT:
                     case CalpontSystemCatalog::UFLOAT:
                     {
-                        int scale = in.getScale(i);
-
-                        if (scale != 0)
-                        {
-                            float f = in.getIntField(i);
-                            f /= (uint64_t) pow(10.0, scale);
-                            out->setFloatField(f, i);
-                        }
-                        else
-                            out->setFloatField(in.getIntField(i), i);
-
+                        auto d = in.getScaledSInt64FieldAsXFloat<double>(i);
+                        out->setFloatField((float) d, i);
                         break;
                     }
 
                     case CalpontSystemCatalog::DOUBLE:
                     case CalpontSystemCatalog::UDOUBLE:
                     {
-                        int scale = in.getScale(i);
-
-                        if (scale != 0)
-                        {
-                            double d = in.getIntField(i);
-                            d /= (uint64_t) pow(10.0, scale);
-                            out->setDoubleField(d, i);
-                        }
-                        else
-                            out->setDoubleField(in.getIntField(i), i);
-
+                        auto d = in.getScaledSInt64FieldAsXFloat<double>(i);
+                        out->setDoubleField(d, i);
                         break;
                     }
 
                     case CalpontSystemCatalog::LONGDOUBLE:
                     {
-                        int scale = in.getScale(i);
-                        long double d = in.getIntField(i);
-                        if (scale != 0)
-                        {
-                            d /= (uint64_t) pow(10.0, scale);
-                        }
+                        auto d = in.getScaledSInt64FieldAsXFloat<long double>(i);
                         out->setLongDoubleField(d, i);
                         break;
                     }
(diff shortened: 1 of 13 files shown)


dbcon/joblist/tupleunion.cpp
```diff
@@ -479,45 +479,22 @@ void TupleUnion::normalize(const Row& in, Row* out)
...
-                        if (scale != 0)
-                        {
-                            float f = in.getIntField(i);
-                            f /= (uint64_t) pow(10.0, scale);
-                            out->setFloatField(f, i);
-                        }
-                        else
-                            out->setFloatField(in.getIntField(i), i);
-
+                        auto d = in.getScaledSInt64FieldAsXFloat<double>(i);
```
tntnatbry, line 482: Shouldn't the template parameter be `float` here?

dbcon/joblist/tupleunion.cpp
```diff
@@ -606,64 +600,40 @@ void TupleUnion::normalize(const Row& in, Row* out)
...
-                        if (scale != 0)
-                        {
-                            float f = in.getUintField(i);
-                            f /= (uint64_t) pow(10.0, scale);
-                            out->setFloatField(f, i);
-                        }
-                        else
-                            out->setFloatField(in.getUintField(i), i);
-
+                        auto d = in.getScaledUInt64FieldAsXFloat<double>(i);
```
tntnatbry, line 603: I think the template parameter should be changed to `float`.

utils/rowgroup/rowaggregation.cpp
```diff
@@ -1458,12 +1458,8 @@ void RowAggregation::doSum(const Row& rowIn, int64_t colIn, int64_t colOut, int
...
             else if (width <= datatypes::MAXLEGACYWIDTH)
             {
-                valIn = rowIn.getIntField(colIn);
-                double scale = (double)(fRowGroupIn.getScale())[colIn];
-                if (valIn != 0 && scale > 0)
-                {
-                    valIn /= pow(10.0, scale);
-                }
+                uint32_t scale = fRowGroupIn.getScale()[colIn];
+                valIn = rowIn.getScaledSInt64FieldAsXFloat<long double>(colIn, scale);
```
tntnatbry, line 1462: Similar to previous client calls, we can just use `getScaledSInt64FieldAsXFloat<long double>(colIn)` here.

utils/rowgroup/rowaggregation.cpp
```diff
@@ -1938,12 +1934,8 @@ void RowAggregation::doAvg(const Row& rowIn, int64_t colIn, int64_t colOut, int6
...
```
tntnatbry, line 1938: Similar to previous client calls, we can just use `getScaledSInt64FieldAsXFloat<long double>(colIn)` here.

utils/rowgroup/rowaggregation.cpp
```diff
@@ -4454,12 +4448,8 @@ void RowAggregationUMP2::doAvg(const Row& rowIn, int64_t colIn, int64_t colOut,
...
```
tntnatbry, line 4452: Similar to previous client calls, we can just use `getScaledSInt64FieldAsXFloat<long double>(colIn)` here.

utils/udfsdk/udfsdk.cpp
```diff
@@ -181,16 +181,16 @@ double MCS_add::getDoubleVal(Row& row,
...
             {
                 dec.scale = op2.scale;
-                op1.value *= (int64_t)pow((double)10, op1.scale - op2.scale);
+                op1.value *= datatypes::scaleDivisor<int64_t>((uint32_t) (op1.scale - op2.scale));
             }
             else
             {
                 dec.scale = op1.scale;
-                op2.value *= (int64_t)pow((double)10, op2.scale - op1.scale);
+                op2.value *= datatypes::scaleDivisor<int64_t>((uint32_t) (op2.scale - op1.scale));
```
tntnatbry, line 189: In the above two changes, if the subtraction of the scales results in a negative value, op1.value or op2.value is basically set to 0. The new code is changing this behaviour.

tntnatbry (changes requested):

dbcon/joblist/tupleunion.cpp
abarkov: As discussed on Slask, this expression:
```
(gdb) p (float) (89999999999/(float) 100000000000)
$4 = 0.900000036
```
would return a less precise result than
```
(gdb) p (float) (89999999999/(double) 100000000000)
$5 = 0.899999976
```

The problem is that conversion of the divisor (which is stored as integer 100000000000)  to FLOAT is lossy, while conversion to DOUBLE is precise.

dbcon/joblist/tupleunion.cpp
abarkov: same here

tntnatbry (approved):

PR #1857, opened 2021-04-09
Don't merge MCOL-4438 Add cmapi_server.log to columnstoreReport

jmrojas2332:


drrtuy:
This patch adds the output into a logReport file that is unfortunately mostly useless.

PR #1863, opened 2021-04-13
A plugable PoorManProfiler singleton

drrtuy:


primitives/primproc/primitiveserver.h
```diff
@@ -44,7 +44,7 @@
 
 #include "service.h"
 
-//#define PRIMPROC_STOPWATCH
+#define PRIMPROC_STOPWATCH
```
mariadb-AlexeyAntipovsky: Is it really needed to be set in production?

utils/loggingcpp/poormanprofiler.inc
```diff
@@ -0,0 +1,140 @@
+#ifndef POORMANPROFILER_
+#define POORMANPROFILER
```
mariadb-AlexeyAntipovsky: Guard name mismatch

utils/loggingcpp/poormanprofiler.inc
```diff
@@ -0,0 +1,140 @@
...
+            stopwatch = stopwatchMapIter->second;
+
+            // If any threads have been active in the last 5 seconds, kick out.
+            if (((tvCurrent.tv_sec - stopwatch->fTvLast.tv_sec) < 5) || stopwatch->isActive())
+            {
+                procIsIdle = false;
+                break;
+            }
+
+            stopwatchMapIter++;
```
mariadb-AlexeyAntipovsky, line 74: I'm afraid there may be race condition here (i.e. in the case when inserting into the unordered_map from another thread will cause rehashing)

utils/loggingcpp/poormanprofiler.inc
```diff
@@ -0,0 +1,140 @@
...
+        if (err)
+            std::cerr << "Error creating thread to complete Stopwatches." << endl;
+
+        stopwatchThreadCreated = true;
+    }
+}
+
+inline logging::StopWatch* PoorManProfiler::start()
+{
+    auto stopwatchMapIter = stopwatchMap.find(pthread_self());
```
mariadb-AlexeyAntipovsky, line 117: And here too...

mariadb-AlexeyAntipovsky (changes requested):

primitives/primproc/primitiveserver.h
drrtuy: Surely not.

mariadb-AlexeyAntipovsky (approved):

PR #1881, opened 2021-04-22
MCOL-4693 ColumnStore MTR tests: FUNCTION mcs192_db.CORR does not exist

abarkov:


davidhall586 (commented):
Would it not save execution time to put all these "exists" tests into one test? We could do all of these with one CREATE DATABASE and one CREATE TABLE -- the two most time consuming operations. I fear mtr for columnstore will eventually grow to an unwieldy time consumption.

abarkov:
I'm not sure we need to put all functions provided by libregr_mysql.so into the same test.
Every single function must be installed by its own CREATE FUNCTION statement anyway. It does not matter where we call ten CREATE FUNCTION statements:
- from the same test file
- or from separate ten test files

But I agree with you that we should try to keep all related tests together, to reduce the number of CREATE FUNCTION statements.

I propose to keep:
- all CORR() related tests inside one file
- all COVAR_POP() related tests inside another file
- and so on

The current "one test file per function" approach made by Bharath looks good for me.

davidhall586 (approved):

PR #1886, opened 2021-04-23
MCOL-4680 FROM subquery containing nested joins returns an error.

tntnatbry:
Main theme of the patch is to fix joins processing in the plugin
code. We now use SELECT_LEX::top_join_list and process the nested
joins recursively, instead of SELECT_LEX::table_list struct which
we earlier used to build the join filters. The earlier approach
did not process certain nested join ON expressions, causing certain
queries to incorrectly error out such as that described in MCOL-4680.

In addition, some legacy code is also removed.

### DIFF
--- a/dbcon/mysql/ha_mcs_execplan.cpp
+++ b/dbcon/mysql/ha_mcs_execplan.cpp
@@ -134,6 +134,15 @@ public:
 namespace cal_impl_if
 {
 
+CalpontSystemCatalog::TableAliasName makeTableAliasName(TABLE_LIST* table)
+{
+    return make_aliasview(
+        (table->db.length ? table->db.str : ""),
+        (table->table_name.length ? table->table_name.str : ""),
+        (table->alias.length ? table->alias.str : ""),
+        getViewName(table), true, lower_case_table_names);
+}
+
 //@bug5228. need to escape backtick `
 string escapeBackTick(const char* str)
 {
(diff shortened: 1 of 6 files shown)


dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -6620,26 +6473,31 @@ int processWhere(SELECT_LEX &select_lex,
...
 #endif
 
-    uint32_t failed = buildOuterJoin(gwi, select_lex);
+    uint32_t failed = 0;
 
-    if (failed) return failed;
+    // InfiniDB bug5764 requires outer joins to be appended to the
+    // end of the filter list. This causes outer join filters to
+    // have a higher join id than inner join filters.
+    // TODO MCOL-4680 Figure out why this is the case, and possibly
```
drrtuy, line 6481: I presume b/c this reduces the cardinality of the result set. Moreover if a query has an outer joins we process joins sequentially so if we process outer joins first then we loose extra records from outer side applying inner join predicate to an outer join product.

drrtuy (changes requested):
Put this patch on hold please b/c the change might be a bit radical to get into develop-5 in the first place and Greg was too optimistic to put 5.6.1 as the fix version.

drrtuy:
It seems this is really a regression comparing with 1.2 so it is worth to try to merge it and embrace the results.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -8594,15 +8494,45 @@ int getGroupPlan(gp_walk_info& gwi, SELECT_LEX& select_lex, SCSEP& csep, cal_gro
             break;
 
         ptp = new ParseTree(new LogicOperator("and"));
-        //ptp->left(filters);
-        ptp->right(filters);
-        lhs = gwi.ptWorkStack.top();
+        ptp->left(filters);
```
drrtuy, line 8497: Why do you decide to swap the branches here?

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -6660,13 +6518,45 @@ int processWhere(SELECT_LEX &select_lex,
             break;
 
         ptp = new ParseTree(new LogicOperator("and"));
-        ptp->right(filters);
-        lhs = gwi.ptWorkStack.top();
+        ptp->left(filters);
```
drrtuy, line 6521: Why do you decide to swap branches?

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -6620,26 +6473,31 @@ int processWhere(SELECT_LEX &select_lex,
...
-    // to the end of the filter list.
-    for (uint i = 0; i < gwi.viewList.size(); i++)
-    {
-        failed = gwi.viewList[i]->processOuterJoin(gwi);
+    if ((failed = buildJoin(gwi, select_lex.top_join_list, outerJoinStack)))
+        return failed;
 
-        if (failed)
-            break;
+    if (gwi.subQuery)
```
drrtuy, line 6488: Why do you decide to put if condition here? I don't see that we specifically process views in buildJoin() so the question where do we handle views w/o subquery naturally arises.

dbcon/mysql/ha_mcs_execplan.cpp
tntnatbry: Table ordering is reverse in SELECT_LEX::join_list (which is what we use in buildJoin() now) compared to SELECT_LEX::table_list (which we used earlier). So I needed to swap the branches here to keep the filter ordering same as before.

dbcon/mysql/ha_mcs_execplan.cpp
tntnatbry: Table ordering is reverse in SELECT_LEX::join_list (which is what we use in buildJoin() now) compared to SELECT_LEX::table_list (which we used earlier). So I needed to swap the branches here to keep the filter ordering same as before.

dbcon/mysql/ha_mcs_execplan.cpp
tntnatbry: Views without subqueries are handled by buildJoin(). If I don't put this if condition here, we will end up creating duplicate filters. For views inside subqueries, however, server does not flatten the join expressions of the views to the outer select and we need access to the subquery view's SELECT_LEX::join_list which is what we achieve with View::processJoin().

drrtuy (approved):

PR #1890, opened 2021-04-26
MCOL-4613 Garbage result of a union between huge narrow DECIMAL and BIGINT

tntnatbry:
We earlier set the unioned column type as DECIMAL/UDECIMAL if one
of the column types involved in the UNION operation was a DECIMAL/UDECIMAL.
However this approach does not work if a BIGINT/UBIGINT column is also
involved in the UNION operation, since a BIGINT/UBIGINT value has a wider range
than a DECIMAL(18)/UDECIMAL(18), causing some high positive/negative values
to overflow. In this patch, we instead set the unioned type to a DOUBLE
for the above case. With the new approach, we do lose some precision of very large
BIGINT values, but the BIGINT/UBIGINT values will no longer overflow.
This limitation, however, will be addressed when we add support for DECIMAL(38).

### DIFF
--- a/dbcon/joblist/tupleunion.cpp
+++ b/dbcon/joblist/tupleunion.cpp
@@ -1104,7 +1104,7 @@ dec4:					/* have to pick a scale to use for the double. using 5... */
                         if (out->getScale(i) == scale)
                             out->setIntField(val, i);
                         else if (out->getScale(i) > scale)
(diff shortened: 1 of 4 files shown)


abarkov (approved):
Looks OK to push. Thanks.

PR #1894, opened 2021-04-28
MCOL-1205 Support queries with circular joins

denis0x0D:
This patch adds support for queries with circular joins.
Currently support added for inner joins only.


denis0x0D:
UPD:
Sync with develop-5

denis0x0D:
Does not work correct.

PR #1914, opened 2021-05-03
MCOL-4626: Check memory > 3G before starting PrimProc.

benthompson15:
Fix how memory is checked for docker containers.


davidhall586 (approved):

drrtuy (changes requested):
Plz hold the patch.

PR #1917, opened 2021-05-04
MCOL-4410 SELECT bypasses select handler if @variables are involved in the query projection list.

tntnatbry:
With this patch, we enable SELECT/DERIVED handler execution for
queries involving 'get_user_var' function. Whereas for 'set_user_var'
function, the handlers are still disabled and the query execution
fallsback to the server, except for some edges case described in
the test file mcol-4410.test, in which case, an appropriate error
message is returned to the client.


drrtuy:
There are some MTR failures that should be addressed.

tntnatbry:
MTR failure is fixed. I added the test case from the issue verbatim, and it had one of the tables created with lower case name but the same table in the SELECT query was using a mix of upper and lower cases.

Please review.

drrtuy (approved):

PR #1918, opened 2021-05-04
MCOL-4440 Primary on S3 doesn't download BRM files anymore but waits …

drrtuy:
…till SM

starts up


drrtuy:
The main idea of the patch is to stop downloading and moving BRM files on the primary of a multi-node cluster.

mariadb-LeonidFedorov (approved):

PR #1922, opened 2021-05-06
MCOL-4685: Eliminate some irrelevant settings (uncompressed data and extents per file)

denis0x0D:
This patch:
1. Removes the option to declare uncompressed columns (set columnstore_compression_type = 0).
2. Ignores [COMMENT '[compression=0] option at table or column level (no error messages, just disregard).
3. Removes the option to set more than 2 extents per file (ExtentsPreSegmentFile).
4. Updates rebuildEM tool to support up to 10 dictionary extent per dictionary segment file.
5. Adds check for `DBRootStorageType` for rebuildEM tool.

### DIFF
--- a/dbcon/joblist/tuple-bps.cpp
+++ b/dbcon/joblist/tuple-bps.cpp
@@ -180,10 +180,8 @@ void TupleBPS::initializeConfigParms()
     fNumThreads = 0;
 
     config::Config* cf = config::Config::makeConfig();
-    string epsf = cf->getConfig("ExtentMap", "ExtentsPerSegmentFile");
 
-    if ( epsf.length() != 0 )
-        fExtentsPerSegFile = cf->uFromText(epsf);
+    fExtentsPerSegFile = DEFAULT_EXTENTS_PER_SEG_FILE;
 
     if (fRequestSize >= fMaxOutstandingRequests)
         fRequestSize = 1;
--- a/dbcon/mysql/ha_mcs_ddl.cpp
+++ b/dbcon/mysql/ha_mcs_ddl.cpp
@@ -189,6 +189,11 @@ int parseCompressionComment ( std::string comment )
     else
         compressiontype = MAX_INT;
 
+    // MCOL-4685: ignore [COMMENT '[compression=0] option at table or column level (no error
+    // messages, just disregard);
+    if (compressiontype == 0)
+      compressiontype = 2;
+
     return compressiontype;
 }
 
@@ -2321,7 +2326,10 @@ int ha_mcs_impl_create_(const char* name, TABLE* table_arg, HA_CREATE_INFO* crea
 
     int compressiontype = get_compression_type(thd);
 
-    if (compressiontype == 1) compressiontype = 2;
+    // MCOL-4685:
+    // remove the option to declare uncompressed columns (set infinidb_compression_type = 0).
+    if (compressiontype == 1 || compressiontype == 0)
+        compressiontype = 2;
 
     //string tablecomment;
     if (table_arg->s->comment.length > 0 )
@@ -2330,8 +2338,14 @@ int ha_mcs_impl_create_(const char* name, TABLE* table_arg, HA_CREATE_INFO* crea
         compressiontype = parseCompressionComment( tablecomment );
     }
 
-    if ( compressiontype == MAX_INT )
+    if (compressiontype == MAX_INT)
+    {
         compressiontype = get_compression_type(thd);
+        // MCOL-4685:
+        // remove the option to declare uncompressed columns (set infinidb_compression_type = 0).
+        if (compressiontype == 0)
+            compressiontype = 2;
+    }
     else if ( compressiontype < 0 )
     {
         string emsg = IDBErrorInfo::instance()->errorMsg(ERR_INVALID_COMPRESSION_TYPE);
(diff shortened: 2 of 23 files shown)


dbcon/joblist/tuple-bps.cpp
```diff
@@ -180,10 +180,8 @@ void TupleBPS::initializeConfigParms()
     fNumThreads = 0;
 
     config::Config* cf = config::Config::makeConfig();
-    string epsf = cf->getConfig("ExtentMap", "ExtentsPerSegmentFile");
 
-    if ( epsf.length() != 0 )
-        fExtentsPerSegFile = cf->uFromText(epsf);
+    fExtentsPerSegFile = 2;
```
drrtuy: Could you use a contstant you define previousl, namely DEFAULT_EXTENTS_PER_SEGMENT_FILE ?

dbcon/mysql/ha_mcs_ddl.cpp
```diff
@@ -2320,7 +2325,10 @@ int ha_mcs_impl_create_(const char* name, TABLE* table_arg, HA_CREATE_INFO* crea
 
     int compressiontype = get_compression_type(thd);
 
-    if (compressiontype == 1) compressiontype = 2;
+    // MCOL-4685:
+    // remove the option to declare uncompressed columns (set infinidb_compression_type = 0).
+    if (compressiontype == 1 || compressiontype == 0)
```
drrtuy, line 2331: Changing this line we should also rename the first element of the enum [here](https://github.com/mariadb-corporation/mariadb-columnstore-engine/blob/develop/dbcon/mysql/ha_mcs_sysvars.cpp#L24).

versioning/BRM/extentmap.cpp
```diff
@@ -96,7 +96,7 @@ namespace
 unsigned ExtentSize = 0; // dmc-need to deprecate
 unsigned ExtentRows              = 0;
 unsigned filesPerColumnPartition = 0;
-unsigned extentsPerSegmentFile   = 0;
+unsigned extentsPerSegmentFile   = 2;
```
drrtuy: We should use the constant defined in dbcon/joblist/columncommand-jl.h  if this doesn't make a circular dependency.

writeengine/server/we_ddlcommandproc.cpp
```diff
@@ -59,10 +59,8 @@ WE_DDLCommandProc::WE_DDLCommandProc()
     if (fpc.length() != 0)
         filesPerColumnPartition = cf->uFromText(fpc);
 
-    string epsf = cf->getConfig("ExtentMap", "ExtentsPerSegmentFile");
-
-    if (epsf.length() != 0)
-        extentsPerSegmentFile = cf->uFromText(epsf);
+    // MCOL-4685: remove the option to set more than 2 extents per file (ExtentsPreSegmentFile).
+    extentsPerSegmentFile = 2;
```
drrtuy: We should use the constant defined in dbcon/joblist/columncommand-jl.h if this doesn't make a circular dependency.

writeengine/server/we_dmlcommandproc.cpp
```diff
@@ -70,10 +70,8 @@ WE_DMLCommandProc::WE_DMLCommandProc()
     if (fpc.length() != 0)
         filesPerColumnPartition = cf->uFromText(fpc);
 
-    string epsf = cf->getConfig("ExtentMap", "ExtentsPerSegmentFile");
-
-    if (epsf.length() != 0)
-        extentsPerSegmentFile = cf->uFromText(epsf);
+    // MCOL-4685: remove the option to set more than 2 extents per file (ExtentsPreSegmentFile).
+    extentsPerSegmentFile = 2;
```
drrtuy: We should use the constant defined in dbcon/joblist/columncommand-jl.h if this doesn't make a circular dependency.

drrtuy (changes requested):
Plz take a look at the comments and make changes if possible.

writeengine/server/we_ddlcommandproc.cpp
denis0x0D: Since EXTENT_PER_SEGMENT_FILE has a storage type (not a macro) it requires to link semantically different parts of projects (ExeManager responsible for select (execplan)) and DDL (responsible for drop, create and so on)). I've added a variable with storage class to DDL, please let me know, if it's not right, and we need to link it against columncomand.

denis0x0D:
Thanks for review, updated.

davidhall586 (approved):

tools/rebuildEM/main.cpp
```diff
@@ -76,6 +76,12 @@ int main(int argc, char** argv)
         }
     }
 
+    // MCOL-4685
+    std::cout << "The launch of mcsRebuildEM tool must be sanctioned by MariaDB support. "
+              << std::endl;
+    std::cout << "Requirement: all DBRoots must be on this node. " << std::endl;
+    std::cout << "Note: that the launch can break the cluster." << std::endl;
```
drrtuy: @denis0x0D I might had missed the point but where does the tool asks for a confirmation before it begins?

drrtuy:
@denis0x0D Could you take a look at the last question I asked?

writeengine/server/we_ddlcommandproc.cpp
drrtuy: No, its OK.

dbcon/mysql/ha_mcs_ddl.cpp
drrtuy: Could you rename the mentioned enum element to UNUSED plz?

tools/rebuildEM/main.cpp
denis0x0D: Oh, my fault, forgot to add the question about confirmation. Thanks.

denis0x0D:
@drrtuy yep, fixed according to latest comments.

drrtuy (approved):

PR #1927, opened 2021-05-11
MCOL-4407 and condtion does not work when HWM > columnstore_string_san_threshold - 1

denis0x0D:


mysql-test/columnstore/basic/r/mcol-4407.result
```diff
@@ -0,0 +1,32 @@
+DROP DATABASE IF EXISTS mcol_4407;
```
drrtuy: Plz move the tests into bugfixes instead.

drrtuy (commented):
Plz move the test into bugfixes directory.

mysql-test/columnstore/basic/r/mcol-4407.result
denis0x0D: Sure. will fix it.

drrtuy (approved):

PR #1935, opened 2021-05-18
MCOL-4665 Move outer join to inner join conversion into the engine.

tntnatbry:
This is a subtask of MCOL-4525 Implement select_handler=AUTO.

Server performs outer join to inner join conversion using simplify_joins()
in sql/sql_select.cc, by updating the TABLE_LIST::outer_join variable.
In order to perform this conversion, permanent changes are made in some
cases to the SELECT_LEX::JOIN::conds and/or TABLE_LIST::on_expr.
This is undesirable for MCOL-4525 which will attemp to fallback and execute
the query inside the server, in case the query execution fails in ColumnStore
using the select_handler.

For a query such as:
  SELECT * FROM t1 LEFT JOIN t2 ON expr1 LEFT JOIN t3 ON expr2
In some cases, server can update the original SELECT_LEX::JOIN::conds
and/or TABLE_LIST::on_expr and create new Item_cond_and objects
(e.g. with 2 Item's expr1 and expr2 in Item_cond_and::list).
Instead of making changes to the original query structs, we use
gp_walk_info::tableOnExprList and gp_walk_info::condList. 2 Item's,
expr1 and expr2, in the condList, mean Item_cond_and(expr1, expr2), and
hence avoid permanent transformations to the SELECT_LEX.

We also define a new member variable
ha_columnstore_select_handler::tableOuterJoinMap
which saves the original TABLE_LIST::outer_join values before they are
updated. This member variable will be used later on to restore to the original
state of TABLE_LIST::outer_join in case of a query fallback to server execution.

The original simplify_joins() implementation in the server also performs a
flattening of the JOIN nest, however we don't perform this operation in
convertOuterJoinToInnerJoin() since it is not required for ColumnStore.


davidhall586 (approved):

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -134,6 +135,146 @@ class RecursionCounter
...
+            not_null_tables |= item->not_null_tables();
+        }
+    }
+}
+
+// Recursively iterate through the join_list and store all non-null
+// TABLE_LIST::on_expr items to a hash map keyed by the TABLE_LIST ptr.
+// This is then used by convertOuterJoinToInnerJoin().
+void buildTableOnExprList(List<TABLE_LIST>* join_list,
+    std::tr1::unordered_map<TABLE_LIST*, std::vector<COND*>>& tableOnExprList)
```
drrtuy: Plz use a type alias for the map.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -134,6 +135,146 @@ class RecursionCounter
...
+            //if (item->eval_const_cond() == is_and_cond && top_level())
+            if (item->eval_const_cond())
+            {
+            }
+            else
+            {
+                not_null_tables = (table_map) 0;
+            }
+
+//            if (thd->is_error())
```
drrtuy: Do you need this comments?

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -134,6 +135,146 @@ class RecursionCounter
...
+// This is taken from Item_cond::fix_fields in sql/item_cmpfunc.cc.
+void calculateNotNullTables(const std::vector<COND*>& condList,
+    table_map& not_null_tables)
+{
+    for (Item* item : condList)
+    {
+        if (item->const_item() && !item->with_param &&
+            !item->is_expensive() && !cond_has_datetime_is_null(item))
+        {
+            //if (item->eval_const_cond() == is_and_cond && top_level())
```
drrtuy: Do you need this commented line?

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -2665,6 +2829,24 @@ void setError(THD* thd, uint32_t errcode, string errmsg, gp_walk_info& gwi)
...
+int setErrorAndReturn(gp_walk_info &gwi)
+{
+    // if this is dervied table process phase, mysql may have not developed the plan
+    // completely. Do not error and eventually mysql will call JOIN::exec() again.
+    // related to bug 2922. Need to find a way to skip calling rnd_init for derived table
+    // processing.
+    if (gwi.thd->derived_tables_processing)
+    {
+        // MCOL-2178 isUnion member only assigned, never used
+        //MIGR::infinidb_vtable.isUnion = false;
```
drrtuy: Feel free to finally remove these two obsolete lines.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -134,6 +135,146 @@ class RecursionCounter
...
+            auto iter = tableOnExprList.find(table);
+
+            if (iter != tableOnExprList.end() && !iter->second.empty())
+            {
+                // The original implementation in simplify_joins() creates
+                // an Item_cond_and object here. Instead of doing so, we
+                // append the table->on_expr to the condList and hence avoid
+                // making any permanent changes to SELECT_LEX::JOIN::conds.
+                condList.insert(condList.end(),
+                    tableOnExprList[table].begin(), tableOnExprList[table].end());
```
drrtuy, line 269: IMHO this tabulation style is hard to read b/c I can't clearly see where the argument list begins.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -8370,9 +8537,20 @@ int cs_get_derived_plan(derived_handler* handler, THD* thd, SCSEP& csep, gp_walk
     return 0;
 }
 
-int cs_get_select_plan(select_handler* handler, THD* thd, SCSEP& csep, gp_walk_info& gwi)
+int cs_get_select_plan(ha_columnstore_select_handler* handler, THD* thd, SCSEP& csep, gp_walk_info& gwi)
 {
-    SELECT_LEX select_lex = *handler->select;
+    SELECT_LEX& select_lex = *handler->select;
```
drrtuy, line 8538: Why do you need this?
It looks more safe to work with a copy or make it a const ref to avoid SELECT_LEX contents changes.

drrtuy (changes requested):
The patch looks OK to me but I would like to see answers/small syntactical changes I mentioned throught the code.

dbcon/mysql/ha_mcs_execplan.cpp

dbcon/mysql/ha_mcs_execplan.cpp
tntnatbry: Removed it.

dbcon/mysql/ha_mcs_execplan.cpp
tntnatbry: Removed it.

dbcon/mysql/ha_mcs_execplan.cpp

dbcon/mysql/ha_mcs_execplan.cpp
tntnatbry: We have mixed use of this style in our code base:
```
some_func_name(some_var1,
    some_var2)
```
and style of the form:
```
some_func_name(some_var1,
               some_var2)
```
Not sure which one is more appropriate. I think use of one of these styles over the other would depend on how long the function name `some_func_name` is. Both should be OK depending on the context.

dbcon/mysql/ha_mcs_execplan.cpp
tntnatbry: I am not sure about this. I don't see definitions of the copy constructor for `SELECT_LEX` so I am not sure what the behaviour of using the compiler generated copy ctor here would be. `getGroupPlan` (although obsolete) also uses a reference instead of a copy. Use of const ref generates a lot of compiler errors.

dbcon/mysql/ha_mcs_execplan.cpp
drrtuy: Agree. I just personally prefer the 2nd style unless the initial indentation is too far from the left.

drrtuy (approved):

drrtuy:
Waiting for CI to converge.

PR #1937, opened 2021-05-21
Fix different signedness comparison warning

mariadb-SergeyZefirov:
self explanatory. oneliner fix for compiler warning.


drrtuy (approved):
Accepted as a temporary solution. 
That was an unfortunate decision to use int as a type to store a column width in writeengine code that must be addressed separately.

PR #1962, opened 2021-06-02
MCOL-4642 NOT IN subquery containing an isnull in the OR predicate crashes server.

tntnatbry:
InSub::handleFunc() was incorrectly exiting early for an IN subquery
containing an isnull predicate in the OR operation in the WHERE clause.
This patch properly handles the OR predicate containing an isnull/isnotnull
predicate in the WHERE clause. We don't remove the isnull operand from the
filter ParseTree in 6.x as the server no longer injects the isnull predicate
in the IN subquery due to MCOL-4617, where we moved the creation and injection
of in-to-exists predicate into the engine.


tntnatbry:
Opensuse build failed due to failure in the smoke step, not related to the PR.

davidhall586 (approved):

PR #1970, opened 2021-06-03
MCOL-4525 Implement columnstore_select_handler=AUTO.

tntnatbry:
This feature allows a query execution to fallback to the server,
in case query execution using the select_handler (SH) fails. In case
of fallback, a warning message containing the original reason for
query failure using SH is generated.

To accomplish this task, SH execution is moved to an earlier step when
we create the SH in create_columnstore_select_handler(), instead of the
previous call to SH execution in ha_columnstore_select_handler::init_scan().
This requires some pre-requisite steps that occur in the server in
JOIN::optimize() and JOIN::exec() to be performed before starting SH execution.

In addition, missing test cases from MCOL-424 are also added to the MTR suite,
and the corresponding fix using disable_indices_for_CEJ() is reverted back
since the original fix now appears to be redundant.

### DIFF
--- a/dbcon/mysql/ha_mcs_impl.cpp
+++ b/dbcon/mysql/ha_mcs_impl.cpp
@@ -76,6 +76,7 @@ using namespace dataconvert;
 
 #include "sm.h"
 #include "ha_mcs_pushdown.h"
+#include "ha_mcs_sysvars.h"
 
 #include "bytestream.h"
 #include "messagequeue.h"
@@ -2692,11 +2693,12 @@ int ha_mcs_impl_write_row(const uchar* buf, TABLE* table, uint64_t rows_changed)
     ha_rows rowsInserted = 0;
     int rc = 0;
 
-    // ci->useCpimport = 2 means ALWAYS use cpimport, whether it's in a
-    // transaction or not. User should use this option very carefully since
-    // cpimport currently does not support rollbacks
-    if (((ci->useCpimport == 2) ||
-         ((ci->useCpimport == 1) && (!(thd->variables.option_bits & (OPTION_NOT_AUTOCOMMIT | OPTION_BEGIN))))) &&
+    // ci->useCpimport = mcs_use_import_for_batchinsert_mode_t::ALWAYS means ALWAYS use
+    // cpimport, whether it's in a transaction or not. User should use this option
+    // very carefully since cpimport currently does not support rollbacks
+    if (((ci->useCpimport == mcs_use_import_for_batchinsert_mode_t::ALWAYS) ||
+         ((ci->useCpimport == mcs_use_import_for_batchinsert_mode_t::ON) &&
+          (!(thd->variables.option_bits & (OPTION_NOT_AUTOCOMMIT | OPTION_BEGIN))))) &&
         (!ci->singleInsert) &&
         ((ci->isLoaddataInfile) ||
          ((thd->lex)->sql_command == SQLCOM_INSERT) || ((thd->lex)->sql_command == SQLCOM_LOAD) ||
@@ -2832,20 +2834,21 @@ void ha_mcs_impl_start_bulk_insert(ha_rows rows, TABLE* table, bool is_cache_ins
             (thd->lex)->sql_command == SQLCOM_INSERT_SELECT ||
             ci->isCacheInsert) && !ci->singleInsert )
     {
-        ci->useCpimport = get_use_import_for_batchinsert(thd);
+        ci->useCpimport = get_use_import_for_batchinsert_mode(thd);
 
         if (((thd->lex)->sql_command == SQLCOM_INSERT) && (rows > 0))
-            ci->useCpimport = 0;
+            ci->useCpimport = mcs_use_import_for_batchinsert_mode_t::OFF;
 
         // For now, disable cpimport for cache inserts
         if (ci->isCacheInsert)
-            ci->useCpimport = 0;
+            ci->useCpimport = mcs_use_import_for_batchinsert_mode_t::OFF;
 
-        // ci->useCpimport = 2 means ALWAYS use cpimport, whether it's in a
-        // transaction or not. User should use this option very carefully since
-        // cpimport currently does not support rollbacks
-        if ((ci->useCpimport == 2) ||
-            ((ci->useCpimport == 1) && (!(thd->variables.option_bits & (OPTION_NOT_AUTOCOMMIT | OPTION_BEGIN))))) //If autocommit on batch insert will use cpimport to load data
+        // ci->useCpimport = mcs_use_import_for_batchinsert_mode_t::ALWAYS means ALWAYS use
+        // cpimport, whether it's in a transaction or not. User should use this option
+        // very carefully since cpimport currently does not support rollbacks
+        if ((ci->useCpimport == mcs_use_import_for_batchinsert_mode_t::ALWAYS) ||
+            ((ci->useCpimport == mcs_use_import_for_batchinsert_mode_t::ON) &&
+             (!(thd->variables.option_bits & (OPTION_NOT_AUTOCOMMIT | OPTION_BEGIN))))) //If autocommit on batch insert will use cpimport to load data
         {
             //store table info to connection info
             CalpontSystemCatalog::TableName tableName;
@@ -3315,8 +3318,9 @@ int ha_mcs_impl_end_bulk_insert(bool abort, TABLE* table)
     // @bug 2515. Check command intead of vtable state
     if ( ( ((thd->lex)->sql_command == SQLCOM_INSERT) ||  ((thd->lex)->sql_command == SQLCOM_LOAD) || (thd->lex)->sql_command == SQLCOM_INSERT_SELECT || ci->isCacheInsert) && !ci->singleInsert )
     {
-        if (((ci->useCpimport == 2) ||
-             ((ci->useCpimport == 1) && (!(thd->variables.option_bits & (OPTION_NOT_AUTOCOMMIT | OPTION_BEGIN))))) &&
+        if (((ci->useCpimport == mcs_use_import_for_batchinsert_mode_t::ALWAYS) ||
+             ((ci->useCpimport == mcs_use_import_for_batchinsert_mode_t::ON) &&
+              (!(thd->variables.option_bits & (OPTION_NOT_AUTOCOMMIT | OPTION_BEGIN))))) &&
             (!ci->singleInsert) &&
             ((ci->isLoaddataInfile) ||
              ((thd->lex)->sql_command == SQLCOM_INSERT) || ((thd->lex)->sql_command == SQLCOM_LOAD) ||
@@ -3526,7 +3530,7 @@ int ha_mcs_impl_end_bulk_insert(bool abort, TABLE* table)
     ci->isCacheInsert = false;
     ci->tableOid = 0;
     ci->rowsHaveInserted = 0;
-    ci->useCpimport = 1;
+    ci->useCpimport = mcs_use_import_for_batchinsert_mode_t::ON;
 
     return rc;
 }
--- a/dbcon/mysql/ha_mcs_opt_rewrites.cpp
+++ b/dbcon/mysql/ha_mcs_opt_rewrites.cpp
@@ -73,37 +73,51 @@ void in_subselect_rewrite_walk(const Item* item_arg, void* arg)
     }
 }
 
-/* @brief  opt_flag_unset_PS() - Unsets first_cond_optimization */
+// Sets SELECT_LEX::first_cond_optimization
+void first_cond_optimization_flag_set(SELECT_LEX* select_lex)
+{
+    select_lex->first_cond_optimization = true;
+}
+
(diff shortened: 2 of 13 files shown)


dbcon/mysql/ha_mcs_impl.cpp
```diff
@@ -2832,7 +2832,7 @@ void ha_mcs_impl_start_bulk_insert(ha_rows rows, TABLE* table, bool is_cache_ins
             (thd->lex)->sql_command == SQLCOM_INSERT_SELECT ||
             ci->isCacheInsert) && !ci->singleInsert )
     {
-        ci->useCpimport = get_use_import_for_batchinsert(thd);
```
drrtuy, line 2835: IMHO it is better to chage ci->useCpimport type than using the cast.

dbcon/mysql/ha_mcs_opt_rewrites.cpp
```diff
@@ -73,38 +73,52 @@ void in_subselect_rewrite_walk(const Item* item_arg, void* arg)
     }
 }
 
-/* @brief  opt_flag_unset_PS() - Unsets first_cond_optimization */
+// Sets SELECT_LEX::first_cond_optimization
+void flag_set(SELECT_LEX* select_lex)
```
drrtuy: Reading the name of the function I can't get what is the flag that we manipulate with. Could you somehow make the functions names more cluefull?

dbcon/mysql/ha_mcs_pushdown.cpp
```diff
@@ -792,71 +781,184 @@ create_columnstore_select_handler(THD* thd, SELECT_LEX* select_lex)
...
-            if (select_lex->first_cond_optimization)
+            // check fallback
+            if (select_handler_mode == 2) // == 2 means columnstore_select_handler=AUTO
             {
-                create_explain_query_if_not_exists(thd->lex, thd->mem_root);
-                arena = thd->activate_stmt_arena_if_needed(&backup);
-                select_lex->first_cond_optimization= false;
-                conds = join->conds;
-                select_lex->where = conds;
+                push_warning(thd, Sql_condition::WARN_LEVEL_WARN, 9999,
```
drrtuy, line 851: The warning here might produce an error to the client depending on the sql_mode session variable value. So the behavior here IMHO an open question whether to log it into our log or return a warning over a client proto.

dbcon/mysql/ha_mcs_pushdown.cpp
```diff
@@ -792,71 +781,184 @@ create_columnstore_select_handler(THD* thd, SELECT_LEX* select_lex)
...
+            if (!thd->is_error())
+            {
+                my_printf_error(ER_INTERNAL_ERROR, "%s", MYF(0),
+                    "Error occured in handler->prepare()");
+            }
+
+            return handler;
+        }
+
+        //if (join->have_query_plan != JOIN::QEP_DELETED && join->build_explain())
```
drrtuy: Do you need this?

dbcon/mysql/ha_mcs_pushdown.cpp
```diff
@@ -792,71 +781,184 @@ create_columnstore_select_handler(THD* thd, SELECT_LEX* select_lex)
...
+                    oss << " F";
+                }
+                else
+                {
+                    oss << ", f";
+                }
 
-                // Unset SL::first_cond_optimization
-                opt_flag_unset_PS(select_lex);
+                oss << "alling back to server execution";
```
drrtuy, line 940: Typo

dbcon/mysql/ha_mcs_pushdown.cpp
```diff
@@ -21,23 +21,12 @@
 
 void check_walk(const Item* item, void* arg);
 
-void disable_indices_for_CEJ(THD *thd_)
+void restoreQueryState(ha_columnstore_select_handler* handler)
```
drrtuy: I would prefer to stick with the consistent naming schema in this unit. Could you plz rename the function?

dbcon/mysql/idb_mysql.h
```diff
@@ -81,6 +81,7 @@ template <class T> bool isnan(T);
 #include "select_handler.h"
 #include "rpl_rli.h"
 #include "my_dbug.h"
+#include "sql_show.h"
```
drrtuy, line 84: Why do you include this header?

dbcon/mysql/ha_mcs_pushdown.cpp
```diff
@@ -407,7 +396,7 @@ create_columnstore_group_by_handler(THD* thd, Query* query)
     // MCOL-2178 Disable SP support in the group_by_handler for now
     // Check the session variable value to enable/disable use of
     // group_by_handler. There is no GBH if SH works for the query.
-    if (get_select_handler(thd) || !get_group_by_handler(thd) || (thd->lex)->sphead)
+    if (((int) get_select_handler(thd) == 1) || !get_group_by_handler(thd) || (thd->lex)->sphead)
```
drrtuy: Could you replace 1 with the enum value name? It will make the text more readable.

dbcon/mysql/ha_mcs_sysvars.h
```diff
@@ -34,12 +34,19 @@ enum mcs_compression_type_t {
...
 // use_import_for_batchinsert
-enum mcs_use_import_for_batchinsert_t {
+enum class mcs_use_import_for_batchinsert_t {
     OFF = 0,
     ON = 1,
     ALWAYS = 2
 };
 
+// select_handler
+enum class mcs_select_handler_t {
```
drrtuy: I would call it select_handler_mode_t to clearly outline what is this enum for.

dbcon/mysql/ha_mcs_sysvars.h
```diff
@@ -34,12 +34,19 @@ enum mcs_compression_type_t {
 };
 
 // use_import_for_batchinsert
-enum mcs_use_import_for_batchinsert_t {
+enum class mcs_use_import_for_batchinsert_t {
```
drrtuy: I would rename this one also but it goes outside of the scope.

dbcon/mysql/ha_mcs_pushdown.cpp
```diff
@@ -754,14 +743,14 @@ int ha_mcs_group_by_handler::end_scan()
 select_handler*
 create_columnstore_select_handler(THD* thd, SELECT_LEX* select_lex)
 {
-    ha_columnstore_select_handler* handler = NULL;
+    int select_handler_mode = (int) get_select_handler(thd);
```
drrtuy: Could you declare it like this ?
`select_handler_mode_t select_handler_mode = get_select_handler(thd) `

dbcon/mysql/ha_mcs_pushdown.cpp
drrtuy: Do you need this?

dbcon/mysql/ha_mcs_pushdown.cpp
```diff
@@ -792,71 +781,184 @@ create_columnstore_select_handler(THD* thd, SELECT_LEX* select_lex)
...
+
+        // Prepare query execution
+        // This is taken from JOIN::exec_inner()
+        if (!select_lex->outer_select() &&                            // (1)
+            select_lex != select_lex->master_unit()->fake_select_lex) // (2)
+            thd->lex->set_limit_rows_examined();
+
+        if (!join->tables_list && (join->table_count || !select_lex->with_sum_func) &&
+            !select_lex->have_window_funcs())
+            return handler;
```
tntnatbry, line 884: Comment from Roman: check if `return nullptr;` is more appropriate here instead of `return handler;`. This also applies to other scenarios where we `return handler;`

dbcon/mysql/ha_mcs_impl.cpp
tntnatbry: I agree. Done.

dbcon/mysql/ha_mcs_opt_rewrites.cpp
tntnatbry: Done.

dbcon/mysql/ha_mcs_pushdown.cpp
tntnatbry: It does not produce an error irrespective of the sql_mode since we have already called thd->clear_error().

dbcon/mysql/ha_mcs_pushdown.cpp
tntnatbry: Removed it.

dbcon/mysql/ha_mcs_pushdown.cpp
tntnatbry: I agree. Updated the name.

dbcon/mysql/idb_mysql.h
tntnatbry: This is needed by the call to get_schema_tables_result() in create_columnstore_select_handler().

dbcon/mysql/ha_mcs_pushdown.cpp
tntnatbry: I agree. Done.

dbcon/mysql/ha_mcs_sysvars.h
tntnatbry: Done.

dbcon/mysql/ha_mcs_sysvars.h
tntnatbry: Also updated this one.

dbcon/mysql/ha_mcs_pushdown.cpp
tntnatbry: Done.

dbcon/mysql/ha_mcs_pushdown.cpp
tntnatbry: Removed it.

dbcon/mysql/ha_mcs_pushdown.cpp
tntnatbry: I have updated this part to `return nullptr;` if thd is not in an error state. Same is done for the `if (join->zero_result_cause)` branch.

tntnatbry:
@drrtuy I have addressed the comments. Please review.

Looks like deb builds have failed due to an unrelated reason.

dbcon/mysql/ha_mcs_pushdown.cpp
drrtuy: My point is that certain sql_mode treats warning as an error so that this behavior might be not very MODE_STRICT_TABLE(sql_mode value) freiendly.

drrtuy (approved):

tntnatbry:
@drrtuy I ran the following test case (unsupported cartesian join scenario) with sql_mode=STRICT_TRANS_TABLES, setting the mode to AUTO does not result in an error:
```sql
MariaDB [test]> set default_storage_engine=columnstore;
Query OK, 0 rows affected (0.000 sec)

MariaDB [test]> show variables like 'sql_mode';
+---------------+-------------------------------------------------------------------------------------------+
| Variable_name | Value                                                                                     |
+---------------+-------------------------------------------------------------------------------------------+
| sql_mode      | STRICT_TRANS_TABLES,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION |
+---------------+-------------------------------------------------------------------------------------------+
1 row in set (0.001 sec)

MariaDB [test]> create table t1 (a int);
Query OK, 0 rows affected (0.376 sec)

MariaDB [test]> create table t2 (a int);
Query OK, 0 rows affected (0.372 sec)

MariaDB [test]> insert into t1 values (1), (2);
Query OK, 2 rows affected (0.239 sec)
Records: 2  Duplicates: 0  Warnings: 0

MariaDB [test]> insert into t2 values (1), (2);
Query OK, 2 rows affected (0.200 sec)
Records: 2  Duplicates: 0  Warnings: 0

MariaDB [test]> set columnstore_select_handler=ON;
Query OK, 0 rows affected (0.000 sec)

MariaDB [test]> select * from t1,t2;
ERROR 1815 (HY000): Internal error: IDB-1000: 't1' and 't2' are not joined.
MariaDB [test]> set columnstore_select_handler=AUTO;
Query OK, 0 rows affected (0.000 sec)

MariaDB [test]> select * from t1,t2;
+------+------+
| a    | a    |
+------+------+
|    1 |    1 |
|    2 |    1 |
|    1 |    2 |
|    2 |    2 |
+------+------+
4 rows in set, 1 warning (0.064 sec)

MariaDB [test]> show warnings;
+---------+------+--------------------------------------------------------------------------------------------------------------------------------------------+
| Level   | Code | Message                                                                                                                                    |
+---------+------+--------------------------------------------------------------------------------------------------------------------------------------------+
| Warning | 9999 | MCS select_handler execution failed due to: 1815: Internal error: IDB-1000: 't1' and 't2' are not joined. Falling back to server execution |
+---------+------+--------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.000 sec)
```

PR #1987, opened 2021-06-10
MCOL-4700 Wrong result of a UNION for INT and INT UNSIGNED

abarkov:


mysql-test/columnstore/basic/r/union.result
```diff
@@ -0,0 +1,90 @@
+SET default_storage_engine=ColumnStore;
+DROP DATABASE IF EXISTS mcs_union;
+CREATE DATABASE mcs_union;
+USE mcs_union;
+#
+# MCOL-4700 Wrong result of a UNION for INT and INT UNSIGNED
+#
+CREATE TABLE t1 (a INT, b INT UNSIGNED);
+INSERT INTO t1 VALUES (-1,1);
+SELECT * FROM (SELECT a FROM t1 UNION SELECT b FROM t1) tu ORDER BY a;
```
tntnatbry, line 10: If we inter-change the 2 columns, we get the same bad output:
```sql
MariaDB [test]> SELECT * FROM (SELECT b FROM t1 UNION SELECT a FROM t1) tu;
+------------+
| b          |
+------------+
|          1 |
| 4294967295 |
+------------+
```
Something like this for the if condition should work:
```
if (type.colWidth == unionedType.colWidth &&
    ((!isUnsigned(unionedType.colDataType) && isUnsigned(type.colDataType)) ||
     (!isUnsigned(type.colDataType) && isUnsigned(unionedType.colDataType))))
```

tntnatbry (changes requested):

tntnatbry (approved):

PR #1989, opened 2021-06-13
MCOL-4713 Analyze table implementation.

denis0x0D:


mariadb-DenisKhalikov:
This implementation is enough to solve circular join problem we have.
The full pipeline is implemented from the:
Plugin code (Handle analyze table command, and create a analyze table execution plan) -> ExeMan (Handle the analyze table exec plan and create a job steps) -> PrimProc (Execute commands)  -> ExeMan (Hande the result RowGroup)

dbcon/mysql/ha_mcs_impl.cpp
```diff
@@ -1898,8 +1899,228 @@ uint32_t doUpdateDelete(THD* thd, gp_walk_info& gwi, const std::vector<COND*>& c
...
+        simpleColumn->timeZone(thd->variables.time_zone->get_name()->ptr());
+
+        returnedColumn.reset(simpleColumn);
+        returnedColumnList.push_back(returnedColumn);
+        columnMap.insert(execplan::CalpontSelectExecutionPlan::ColumnMap::value_type(
+            simpleColumn->columnName(), returnedColumn));
+    }
+
+    // Create execution plan and initialize it with `returned columns` and `column map`.
+    execplan::MCSAnalyzeTableExecutionPlan* caep =
```
drrtuy: I missed where you delete this ptr. IMHO this pointer should be std::uniqie_ptr.

dbcon/execplan/mcsanalyzetableexecutionplan.h
```diff
@@ -0,0 +1,170 @@
...
+/**
+ * Namespace
+ */
+namespace execplan
+{
+
+class MCSAnalyzeTableExecutionPlan : public CalpontExecutionPlan
+{
+  public:
+    typedef std::vector<SRCP> ReturnedColumnList;
```
drrtuy, line 46: Could you make these members private.

dbcon/mysql/ha_mcs_impl.cpp
denis0x0D: Thanks, yes, I did introduce a memory leak here.

dbcon/execplan/mcsanalyzetableexecutionplan.h
denis0x0D: Sure.

denis0x0D: Oops, probably no, because I use them to define a type in plugin code, before creating exec plan, because exec plan is initialized by `ReturnedColumnList` and `ColumnMap`.

dbcon/mysql/ha_mcs_impl.cpp
```diff
@@ -1898,8 +1899,228 @@ uint32_t doUpdateDelete(THD* thd, gp_walk_info& gwi, const std::vector<COND*>& c
...
+
+    if (caep->traceOn())
+        std::cout << caep->toString() << std::endl;
+
+    {
+        ByteStream msg;
+        try
+        {
+            // We use 6 to indicate that we want to run `ANALYZE TABLE` command.
+            ByteStream::quadbyte qb = 6;
```
drrtuy: Could you make 6 and other values named constants?

dbcon/joblist/joblistfactory.cpp
```diff
@@ -1847,206 +1998,309 @@ SJLP makeJobList_(
...
-    *jobInfo.smallSideUsage = 0;
-
-    // set fifoSize to 1 for CalpontSystemCatalog query
-    if (csep->sessionID() & 0x80000000)
-        jobInfo.fifoSize = 1;
-    else if (csep->traceOn())
-        cout << (*csep) << endl;
-
-    try
+    if (csep != nullptr)
```
drrtuy, line 2039: There are lots of boilerplate code. Is it possible to reduce it?

dbcon/joblist/joblistfactory.cpp
```diff
@@ -1847,206 +1998,309 @@ SJLP makeJobList_(
...
+            auto stepNo = numberSteps(querySteps, 0, jobInfo.traceFlags);
+
+            jl->addQuery(querySteps);
+            jl->addDelivery(deliverySteps);
 
+            dynamic_cast<TupleJobList*>(jl)->setDeliveryFlag(true);
+        }
+        catch (IDBExcept& iex)
+        {
+            jobInfo.errorInfo->errCode = iex.errorCode();
```
drrtuy: Is there a way to reduce these sections boilerplate using JobStep::handleException()?

denis0x0D:
Added `saveToFile` after each analyze command and `loadFrom` file on ExeMgr startup.

denis0x0D:
Added a statistics distribution across all ExeMgr in cluster.

versioning/BRM/statistics.cpp
```diff
@@ -0,0 +1,372 @@
...
+    }
+}
+
+void StatisticsDistributor::countClients()
+{
+    std::lock_guard<std::mutex> lock(mut);
+#ifdef DEBUG_STATISTICS
+    std::cout << "count clients to distribute statistics " << std::endl;
+#endif
+    auto* config = config::Config::makeConfig();
```
drrtuy: JFYI Strictly speaking right now it is impossible to change EM configuration in the config w/o restart so the lock here overlaps with the lock config::Config::makeConfig(). 
What if you make exeMgrNumber atomic number, put  countClients() call into distributeStatistics() and redice the crit section of distributeStatistics() so the section only contains StatisticsManager inst serialization? It looks like this approach doesn't have races and its crit section doesn't contain network ops that might take a lot. What do you think?

versioning/BRM/statistics.cpp
```diff
@@ -0,0 +1,372 @@
...
+    if (!rowCount || !columnCount)
+        return;
+
+    auto& oids = rowGroup.getOIDs();
+
+    rowgroup::Row r;
+    rowGroup.initRow(&r);
+    rowGroup.getRow(0, &r);
+
+    std::vector<std::unordered_set<uint32_t>> columns(columnCount, std::unordered_set<uint32_t>());
```
drrtuy: Could you name this type to improve readability?

versioning/BRM/statistics.cpp
```diff
@@ -0,0 +1,372 @@
...
+    // Init key types.
+    for (uint32_t index = 0; index < columnCount; ++index)
+        keyTypes[oids[index]] = KeyType::PK;
+
+    const uint32_t maxRowCount = 4096;
+    // TODO: We should read just couple of blocks from columns, not all data, but this requires
```
drrtuy: Agree. The "generous" behavior of TupleBPS that reads 20 extents at once even if EM looks for a single record is a waste. I will create an issue for this.

versioning/BRM/statistics.cpp
```diff
@@ -0,0 +1,372 @@
...
+    // more deep refactoring of column commands.
+    rowCount = std::min(rowCount, maxRowCount);
+    // This is strange, it's a CS but I'm processing data as row by row, how to fix it?
```
drrtuy: Another good point. This is columnstore but in-memory repr is row-oriented for a questionable perf gain hashing for  GB and JOIN operators. 
The project on turning RowGroup 90 degrees is a big project that you will be involved for sure.

drrtuy: But here you have a limitation of a hash map used. Even if column values are continues vectors one can't put N values into the std::unordered_map at a time.  
There are couple ideas how to vectorize the section but let's discuss this offline.

versioning/BRM/statistics.cpp
```diff
@@ -0,0 +1,372 @@
...
+
+    const char* fileName = statsFile.c_str();
+    std::unique_ptr<IDBDataFile> out(
+        IDBDataFile::open(IDBPolicy::getType(fileName, IDBPolicy::WRITEENG), fileName, "wb", 1));
+
+    if (!out)
+    {
+        BRM::log_errno("StatisticsManager::saveToFile(): open");
+        throw ios_base::failure(
+            "StatisticsManager::saveToFile(): open failed. Check the error log.");
```
drrtuy: If I read IDBDataFile::open() correctly it writes into debug.log not error.log

versioning/BRM/statistics.cpp
```diff
@@ -0,0 +1,372 @@
...
+    }
+
+    // Number of pairs.
+    uint64_t count = keyTypes.size();
+    // count, [[uid, keyType], ... ]
+    const uint64_t dataStreamSize = sizeof(uint64_t) + count * (sizeof(uint32_t) + sizeof(KeyType));
+
+    // Allocate memory for data stream.
+    char* dataStream = new char[dataStreamSize];
```
drrtuy: IIRC there is a move ctor for unique_ptr so std::unique_ptr<char[]> dataStreamSmartPtr(new char[dataStreamSize]) should do the trick.

versioning/BRM/statistics.cpp
```diff
@@ -0,0 +1,372 @@
...
+    std::unique_ptr<char[]> dataStreamSmartPtr;
+    dataStreamSmartPtr.reset(dataStream);
+
+    // Initialize the data stream.
+    uint64_t offset = 0;
+    std::memcpy(dataStream, reinterpret_cast<char*>(&count), sizeof(uint64_t));
```
drrtuy: What do you think about re-using ByteStream here? The question I see is that it might need to add some BS methods to  allow to hash BS'es contents.

versioning/BRM/statistics.cpp
```diff
@@ -0,0 +1,372 @@
...
+        uint32_t oid = p.first;
+        std::memcpy(&dataStream[offset], reinterpret_cast<char*>(&oid), sizeof(uint32_t));
+        offset += sizeof(uint32_t);
+
+        KeyType keyType = p.second;
+        std::memcpy(&dataStream[offset], reinterpret_cast<char*>(&keyType), sizeof(KeyType));
+        offset += sizeof(KeyType);
+    }
+
+    utils::Hasher128 hasher;
```
drrtuy: As far as I understand the code serizalization of unordered_map might produce different byte sequences with different hash values. This difference makes it even harder to compare two StatisticsManager instances together. I will put another not where StatisticsManager's unordered_map is defined.

versioning/BRM/statistics.h
```diff
@@ -0,0 +1,118 @@
...
+    // Loads stats from the file.
+    void loadFromFile();
+    void incEpoch() { ++epoch; }
+    // Serialize stats to the given `bs`.
+    void serialize(messageqcpp::ByteStream& bs);
+    // Unserialize stats from the given `bs`.
+    void unserialize(messageqcpp::ByteStream& bs);
+
+  private:
+    std::unordered_map<uint32_t, KeyType> keyTypes;
```
drrtuy: As far as I understand the code serizalization of unordered_map might produce different byte sequences with different hash values. This difference makes it even harder to compare two StatisticsManager instances together. What if you make this member an ordered_map?
The original idea behind hash is to save some network and cpu cycles if an originator EM offers the same StatisticsManager instance to a destination EM.

versioning/BRM/statistics.h
```diff
@@ -0,0 +1,118 @@
...
+    StatisticsManager() : epoch(0), version(1) { IDBPolicy::init(true, false, "", 0); }
+    std::mutex mut;
+    uint32_t epoch;
+    uint32_t version;
+    // FIXME:
+    // Should we put it to the config file?
+    std::string statsFile = "/var/lib/columnstore/local/statistics";
```
drrtuy: Let' begin with the hardcoded value.

versioning/BRM/statistics.cpp
```diff
@@ -0,0 +1,372 @@
...
+        throw ios_base::failure("StatisticsManager::loadFromFile(): read failed. ");
+
+    // Initialize fields from the file header.
+    version = fileHeader.version;
+    epoch = fileHeader.epoch;
+    const auto fileHash = fileHeader.fileHash;
+    const auto dataStreamSize = fileHeader.dataSize;
+
+    // Allocate the memory for the file data.
+    char* dataStream = new char[dataStreamSize];
```
drrtuy: Plz take a look at the very similar comments in saveToFile() on unique_ptr ctor and ByteStream.

versioning/BRM/statistics.cpp
```diff
@@ -0,0 +1,372 @@
...
+                << std::endl;
+#endif
+            // Write a flag to client ExeMgr.
+            exemgrClient->write(msg);
+
+#ifdef DEBUG_STATISTICS
+            std::cout << "Write statistics from ExeMgr(Server) to ExeMgr(Clients) " << std::endl;
+#endif
+            // Write a statistics to client ExeMgr.
+            exemgrClient->write(statsBs);
```
drrtuy: Could you extend proto a bit. First a sender sends ANALYZE_TABLE_REC_STATS, then its SessionManager current hash. The receiver should calculate its own hash if it has anything and compare with the incoming. If they are the same then write a byte to the sender telling it is better to stop. The sender should read from the receiver at that point and reacts accordingly. I will send you the text description in Slack.

dbcon/execplan/mcsanalyzetableexecutionplan.cpp
```diff
@@ -0,0 +1,156 @@
...
+    fReturnedCols.clear();
+    fColumnMap.clear();
+    uint32_t size;
+
+    bs >> size;
+    for (uint32_t i = 0; i < size; ++i)
+    {
+        auto* returnedColumn = dynamic_cast<ReturnedColumn*>(ObjectReader::createTreeNode(bs));
+        SRCP srcp(returnedColumn);
+        fReturnedCols.push_back(srcp);
```
drrtuy, line 117: What do you think about expression fReturnedCols.emplace_back(SRCP(returnedColumn)) ?

dbcon/mysql/ha_mcs_impl.cpp
```diff
@@ -1898,8 +1900,228 @@ uint32_t doUpdateDelete(THD* thd, gp_walk_info& gwi, const std::vector<COND*>& c
...
+    {
+        execplan::SRCP returnedColumn;
+        const auto objNum = oidlist[i].objnum;
+        auto tableColName = csc->colName(objNum);
+        auto colType = csc->colType(objNum);
+
+        if (!isSupportedToAnalyze(colType))
+            continue;
+
+        execplan::SimpleColumn* simpleColumn = new execplan::SimpleColumn();
```
drrtuy, line 2041: What if you add SimpleColumn ctor that takes the list of the arguments.
This would save us some lines.
Another option is to leverage move ctor with emplace_back like this returnedColumnList.emplace_back(execplan::SRCP(simpleColumn))
What do you think?

dbcon/mysql/ha_mcs_impl.cpp
```diff
@@ -1898,8 +1900,228 @@ uint32_t doUpdateDelete(THD* thd, gp_walk_info& gwi, const std::vector<COND*>& c
...
+    {
+        caep->priority(ci->stats.userPriority(ci->stats.fHost, ci->stats.fUser));
+    }
+    catch (std::exception& e)
+    {
+        string msg = string("Columnstore User Priority - ") + e.what();
+        push_warning(thd, Sql_condition::WARN_LEVEL_WARN, 9999, msg.c_str());
+    }
+
+    // FIXME: Should we send a message to ExeMgr?
```
drrtuy: We send nothing to EM yet IIRC so the answer is no.

dbcon/mysql/ha_mcs_impl.cpp
```diff
@@ -1898,8 +1900,228 @@ uint32_t doUpdateDelete(THD* thd, gp_walk_info& gwi, const std::vector<COND*>& c
...
+    caep->traceFlags(ci->traceFlags);
+
+    cal_table_info ti;
+    sm::cpsm_conhdl_t* hndl;
+
+    bool localQuery = (get_local_query(thd) > 0 ? true : false);
+    caep->localQuery(localQuery);
+
+    {
+        ci->stats.reset();
```
drrtuy: We have 4 copies of this stats related block so if you make it a function we would re-use it in ha_mcs_impl.cpp unit to future reduce boilerplate code.

dbcon/mysql/ha_mcs_impl.cpp
```diff
@@ -1898,8 +1900,228 @@ uint32_t doUpdateDelete(THD* thd, gp_walk_info& gwi, const std::vector<COND*>& c
...
+    hndl = ci->cal_conn_hndl;
+
+    if (caep->traceOn())
+        std::cout << caep->toString() << std::endl;
+
+    {
+        ByteStream msg;
+        try
+        {
+            ByteStream::quadbyte qb = ANALYZE_TABLE_EXECUTE;
```
drrtuy: I think block can be also generalized as a function b/c we have 3 of these with a tiny modification in ha_mcs_impl.cpp.

exemgr/main.cpp
```diff
@@ -691,6 +692,112 @@ class SessionThread
...
+                        // Process `RowGroup`, increase an epoch and save statistics to the file.
+                        auto* statisticsManager = statistics::StatisticsManager::instance();
+                        statisticsManager->analyzeColumnKeyTypes(outRG, caep.traceOn());
+                        statisticsManager->incEpoch();
+                        statisticsManager->saveToFile();
+
+                        // Distribute statistics across all ExeMgr clients if possible.
+                        // FIXME: Should we count the number of clients only once at the beginning
+                        // or CMAPI can modify the spec file without ExeMgr restart?
+                        statistics::StatisticsDistributor::instance()->countClients();
```
drrtuy: What do you think if you put countClients() inside distributeStatistics(). There are some details on that as comments in distributeStatistics().

exemgr/main.cpp
```diff
@@ -691,6 +692,112 @@ class SessionThread
                         fIos.close();
                         break;
                     }
+                    else if (qb == ANALYZE_TABLE_EXECUTE)
```
drrtuy, line 824: Personally I would make the handlers for the new qbs as functions b/c it improves readability but it is up to you.

versioning/BRM/statistics.cpp
```diff
@@ -0,0 +1,372 @@
+/* Copyright (C) 2021 MariaDB Corporation
```
drrtuy: Since we agreed that we don't use BRM for the stats could you move it into ./utils/common?

drrtuy (changes requested):
Plz take a look at the review.

dbcon/mysql/ha_mcs_impl.cpp
```diff
@@ -1899,8 +1901,249 @@ uint32_t doUpdateDelete(THD* thd, gp_walk_info& gwi, const std::vector<COND*>& c
...
+    caep->traceFlags(ci->traceFlags);
+
+    cal_table_info ti;
+    sm::cpsm_conhdl_t* hndl;
+
+    bool localQuery = (get_local_query(thd) > 0 ? true : false);
+    caep->localQuery(localQuery);
+
+    // Try to initialize connection.
+    if (!initializeCalConnectionInfo(ci, thd, csc, sessionID, localQuery))
```
drrtuy, line 2116: Nice one

dbcon/mysql/ha_mcs_impl.cpp
```diff
@@ -1899,8 +1901,249 @@ uint32_t doUpdateDelete(THD* thd, gp_walk_info& gwi, const std::vector<COND*>& c
...
+        goto error;
+
+    hndl = ci->cal_conn_hndl;
+
+    if (caep->traceOn())
+        std::cout << caep->toString() << std::endl;
+    {
+        ByteStream::quadbyte qb = ANALYZE_TABLE_EXECUTE;
+        // Serialize and the send the `anlyze table` execution plan.
+        if (!sendExecutionPlanToExeMgr(hndl, qb, caep, ci, thd))
```
drrtuy, line 2126: Very good.

drrtuy (approved):

PR #1995, opened 2021-06-16
MCOL-4758 Limit LONGTEXT and LONGBLOB to 16MB

davidhall586:


dbcon/ddlpackage/ddl.y
```diff
@@ -89,8 +89,8 @@ void fix_column_length(SchemaObject* elem, const CHARSET_INFO* def_cs) {
             column->fType->fLength = 65535;
         else if (column->fType->fLength <= 16777215)
             column->fType->fLength = 16777215;
-        else if (column->fType->fLength <= 2100000000)
-            column->fType->fLength = 2100000000;
+        else 
+            column->fType->fLength = 16777215;
         // otherwise leave the decision to a caller code
```
tntnatbry: We don't need the previous `else if`. Also, since we now use an `else` block in the end instead of an `else if`, we should remove the comment as well.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -3157,15 +3157,15 @@ CalpontSystemCatalog::ColType colType_MysqlToIDB (const Item* item)
         case STRING_RESULT:
             ct.colDataType = CalpontSystemCatalog::VARCHAR;
 
-            // MCOL-697 the longest TEXT we deal with is 2100000000 so
+            // MCOL-697 the longest TEXT we deal with is 16777215 so
             // limit to that. Otherwise for LONGBLOB ct.colWidth ends
             // up being -1 and demons eat your data and first born
             // (or you just get truncated to 20 bytes with the 'if'
             // below)
```
tntnatbry: Is the full comment still relevant?

writeengine/dictionary/we_dctnry.cpp
```diff
@@ -67,7 +67,7 @@ const int m_bigSpace  =                    // free space in an empty block
 const int START_HDR1  =                    // start loc of 2nd offset (HDR1)
     HDR_UNIT_SIZE + NEXT_PTR_BYTES + HDR_UNIT_SIZE;
 const int PSEUDO_COL_WIDTH = DICT_COL_WIDTH; // used to convert row count to block count
-const int MAX_BLOB_SIZE = 2100000000; // for safety, we use an 18bit block count of 8KB blocks
+const int MAX_BLOB_SIZE = 16777215; // for safety, we use an 18bit block count of 8KB blocks
```
tntnatbry: Does this comment make sense for this const definition of `MAX_BLOB_SIZE`?

tntnatbry (changes requested):
@dhall-MariaDB I don't see any test cases for the changes. Can you add tests for CREATE and ALTER showing the new behaviour?

And as far as I understand, the crashing query from https://jira.mariadb.org/browse/MCOL-3879 is also now supposed to work. Can you please add that test case into the MTR suite as well?

mysql-test/columnstore/bugfixes/mcol-4758.test
```diff
@@ -0,0 +1,22 @@
...
+# To be uncommented when MCOL-4480 is fixed
+#ALTER TABLE src ADD COLUMN (cLT2 LONGTEXT);
+#UPDATE src SET cLT2="My Friday Night" where c0=1;
+#select * from src where c0=1 and substr(cLT, 1, 4)="This";
+
+# cleanup
+DROP DATABASE `mcol_4758`;
+
+
+
```
tntnatbry: There are 3 redundant linespaces here.

tntnatbry (changes requested):

tntnatbry:
@dhall-MariaDB Changes look good to me now. Some additional comments:

1. I left 1 comment in-line to fix the redundant linespaces.

2. Can you add the test case from https://jira.mariadb.org/browse/MCOL-3879 as well into a standalone test file at bugfixes/mcol-3879.test? Not the original test from the issue description but the simpler example you shared in a comment there.

3. Once the changes are done, please squash all commits into the original commit.

tntnatbry:
@dhall-MariaDB MTR failed for basic/mcol_2000.test. Looks like this test uses `information_schema` metadata tables that were outputting 2GB. It should be a simple fix to update the result file with 16MB.

tntnatbry (approved):

PR #1996, opened 2021-06-18
MCOL-4738 AVG() returns a wrong result.

tntnatbry:
Release/RelWithDebInfo build on some platforms (Ubuntu18/20, Fedora)
is returning wrong results for the AVG() distributed function.

In Row::setLongDoubleField(), we remove the code that masks higher
bits of a long double field (if sizeof(long double) == 16 on a
given platform) since this masking appears to be redundant and also
fixes the issue with the AVG() function.

### DIFF
--- a/utils/rowgroup/rowgroup.h
+++ b/utils/rowgroup/rowgroup.h
@@ -1306,12 +1306,7 @@ inline void Row::setFloatField(float val, uint32_t colIndex)
 inline void Row::setLongDoubleField(const long double& val, uint32_t colIndex)
 {
     uint8_t* p = &data[offsets[colIndex]];
-    *((long double*)p) = val;
-    if (sizeof(long double) == 16)
-    {
-        // zero out the unused portion as there may be garbage there.
-        *((uint64_t*)p+1) &= 0x000000000000FFFFULL;
-    }
+    *reinterpret_cast<long double*>(p) = val;
 }
 
 inline void Row::setInt128Field(const int128_t& val, uint32_t colIndex)


abarkov (approved):
Look fine for me. The old code looked strange. No idea why we need to clear the unused part.

If we always store values like this:

```((long double *) ptr)[0]= long_double_value;```

and retrieve like this:

```long_double_value= ((long double *) ptr)[0];```

then no zeroing is needed.

And it seems we always do like this.

**Comment by drrtuy** [comment]
> Look fine for me. The old code looked strange. No idea why we need to clear the unused part.
> 
> If we always store values like this:
> 
> `((long double *) ptr)[0]= long_double_value;`
> 
> and retrieve like this:
> 
> `long_double_value= ((long double *) ptr)[0];`
> 
> then no zeroing is needed.
> 
> And it seems we always do like this.

D.Hall added it assuming there are platforms where long double is 10 bits wide and there are others that uses 16 bits for long double.

**Comment by tntnatbry** [comment]
@drrtuy I think you meant 80 bits and 16 bytes respectively.

I added @dhall-MariaDB  as a reviewer as well since he made changes to this function last time.

**Review by davidhall586 — COMMENTED** [review]
In bool Row::equals(const Row& r2, uint32_t lastCol) const

We find

    // If there are no strings in the row, then we can just memcmp the whole row.
    // hasCollation is true if there is any column of type CHAR, VARCHAR or TEXT
    // useStringTable is true if any field declared > max inline field size, including BLOB
    // For memcmp to be correct, both must be false.
    if (!hasCollation && !useStringTable && !r2.hasCollation && !r2.useStringTable)
        return !(memcmp(&data[offsets[0]], &r2.data[offsets[0]], offsets[lastCol + 1] - offsets[0]));

The memcmp will fail on machines that use 80 bit long double unless those bits are zeroed.

I recommend rather than remove the code, we #ifdef it somehow.

I have verified that gcc  4.8.5 on CentOS 7 doesn't clear these bits automatically.

**Comment by drrtuy** [comment]
> In bool Row::equals(const Row& r2, uint32_t lastCol) const
> 
> We find
> 
> ```
> // If there are no strings in the row, then we can just memcmp the whole row.
> // hasCollation is true if there is any column of type CHAR, VARCHAR or TEXT
> // useStringTable is true if any field declared > max inline field size, including BLOB
> // For memcmp to be correct, both must be false.
> if (!hasCollation && !useStringTable && !r2.hasCollation && !r2.useStringTable)
>     return !(memcmp(&data[offsets[0]], &r2.data[offsets[0]], offsets[lastCol + 1] - offsets[0]));
> ```
> 
> The memcmp will fail on machines that use 80 bit long double unless those bits are zeroed.
> 
> I recommend rather than remove the code, we #ifdef it somehow.
> 
> I have verified that gcc 4.8.5 on CentOS 7 doesn't clear these bits automatically.

If it doesn't clear then how does the CI finish the MTR on centos 7?

File: utils/rowgroup/rowgroup.h [hunk 1]
```diff
@@ -1306,12 +1306,7 @@ inline void Row::setFloatField(float val, uint32_t colIndex)
 inline void Row::setLongDoubleField(const long double& val, uint32_t colIndex)
 {
     uint8_t* p = &data[offsets[colIndex]];
-    *((long double*)p) = val;
-    if (sizeof(long double) == 16)
-    {
-        // zero out the unused portion as there may be garbage there.
-        *((uint64_t*)p+1) &= 0x000000000000FFFFULL;
-    }
+    *reinterpret_cast<long double*>(p) = val;
```
drrtuy, line 1309: We need this change for ARM so the #ifdef must take this fact into account.

davidhall586:
The masking bits are required for AMD64 architectures. A new PR will be produced that is generic for 128 bit float types like ARM uses.

PR #1999, opened 2021-06-22
MCOL-4741 Fix extentmap handling of string prefixes encoded as 64-bit integers

mariadb-SergeyZefirov:
It so happened that string prefixes, encoded as integers for comparison, were compared in range merge operation (used primarily by cpimport) as signed integers, not unsigned ones (as memcmp/strcmp would).

This was the main reason of range values became messed up after cpimport of completely NULL column.


mariadb-SergeyZefirov:
Relevant task: https://jira.mariadb.org/browse/MCOL-4741

tntnatbry:
@mariadb-SergeyZefirov Code changes look good to me. However, we are missing test case for this. In https://jira.mariadb.org/browse/MCOL-4741 you have mentioned you were able to come up with a small reproducible test case. Can you please add it to MTR under mysql-test/columnstore/bugfixes ?

Also, once all changes are pushed, please also fix the original commit message. Right now, it contains some garbage text. As standard practice, we start the commit message with the MCOL number followed by the message. As reference, you can take a look at the existing commit history here: https://github.com/mariadb-corporation/mariadb-columnstore-engine/commits/develop

tntnatbry (approved):

PR #2018, opened 2021-06-30
Mcol 4779 invalid extent ranges for short character columns

mariadb-SergeyZefirov:
What is in pull request changes:

1. First, added a support for CHAR(n) columns. It was missing due to oversight.
2. Second, I changed logic of supported (in the "can keep ranges" sense) column type determination to be unified. We now switching on the internal Write Engine column type, not on the Calpont column types.
3. Third, extended CPInfo has a method for to/from string conversion. It is wrong (ranges should be kept differently) but it is in line with other parts of code.
4. This method is used before and after range updates to, again, be in line with other parts of code.
5. The values fetched from CHAR column are converted in same manner for them to be kept in a compatible format.

That's about it. I guess it warrants a test in the basic suite which I'll do shortly.

### DIFF
--- a/writeengine/wrapper/writeengine.cpp
+++ b/writeengine/wrapper/writeengine.cpp
@@ -331,6 +331,11 @@ void WriteEngineWrapper::updateMaxMinRange(const size_t totalNewRow, const size_
     {
         return ;
     }
+    if (colType == WR_CHAR)
+    {
+        maxMin->toInvalid(); // simple and wrong solution.
+	return;
+    }
     bool isUnsigned = false; // TODO: should change with type.
     switch (colType)
     {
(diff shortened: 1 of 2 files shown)


tntnatbry:
@drrtuy is reviewing this PR. I just noticed the original commit message which should be fixed before merging.

writeengine/wrapper/writeengine.cpp
```diff
@@ -321,13 +321,19 @@ void WriteEngineWrapper::updateMaxMinRange(const size_t totalNewRow, const size_
     {
         return ;
     }
+    if (colType == WR_CHAR)
+    {
+        maxMin->toInvalid(); // simple and wrong solution.
```
drrtuy, line 336: I would suggest to change the comment b/c I can see sarcasm but foreigner might misread this :)

drrtuy (approved):

PR #2019, opened 2021-06-30
MCOL-4738 Mask long doubles only for AMD64

davidhall586:
By using std::numeric_limits, we can determine if long double is 80 bits effective or not. Mask only if it is.

### DIFF
--- a/utils/rowgroup/rowgroup.h
+++ b/utils/rowgroup/rowgroup.h
@@ -1370,7 +1370,17 @@ inline void Row::setLongDoubleField(const long double& val, uint32_t colIndex)
 {
     uint8_t* p = &data[offsets[colIndex]];
     *((long double*)p) = val;
-    if (sizeof(long double) == 16)
+    
+    
+    int64_t d = std::numeric_limits<long double>::digits;  // Number of base 2 digits (64 for 80 bit fpu)
+    int64_t e = std::numeric_limits<long double>::max_exponent; // max base 2 exponent digits (16384 for 80-0 bit fpu)
+    
+    // If long double is 16 bytes and digits and exponent are 64 and 16384 respectively, then we need to mask out the 
+    // unused bits, as they contain garbage. There are times we test for equality by memcmp of a buffer containing,
+    // in part, the long double set here. Garbage bytes will adversly affect that compare.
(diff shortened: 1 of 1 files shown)


utils/rowgroup/rowgroup.h
```diff
@@ -1370,7 +1370,17 @@ inline void Row::setLongDoubleField(const long double& val, uint32_t colIndex)
...
+    
+    int64_t d = std::numeric_limits<long double>::digits;  // Number of base 2 digits (64 for 80 bit fpu)
+    int64_t e = std::numeric_limits<long double>::max_exponent; // max base 2 exponent  (16384 for 80 bit fpu)
+    
+    // If long double is 16 bytes and digits and exponent are 64 and 16384 respectively, then we need to mask out the 
+    // unused bits, as they contain garbage. There are times we test for equality by memcmp of a buffer containing,
+    // in part, the long double set here. Garbage bytes will adversly affect that compare.
+    // Note: There may be compilers that store 80 bit floats in 12 bytes. We do not account for that here. I don't believe
+    // there are any modern Linux compilers that do that as a default. Windows uses 64 bits, so no masking is needed.
+    if (d  == 64 && e == 16384 && sizeof(long double) == 16)
```
drrtuy, line 1383: IMHO These should become constants.

drrtuy: I would run this check only once b/c it depends on runtime env only so this might be a global bool flag to reduce and number of cmp instructions per Row. However 4603 should replace long double and reduce its importance.

utils/rowgroup/rowgroup.h
```diff
@@ -1370,7 +1370,17 @@ inline void Row::setLongDoubleField(const long double& val, uint32_t colIndex)
 {
     uint8_t* p = &data[offsets[colIndex]];
     *((long double*)p) = val;
-    if (sizeof(long double) == 16)
+    
+    
+    int64_t d = std::numeric_limits<long double>::digits;  // Number of base 2 digits (64 for 80 bit fpu)
```
drrtuy, line 1375: If we have a global bool we can calculate it once on startup and fill the global to leverage it later. This should save us 2 mov instr + 1 cmp perRow::setLongDoubleField call.

drrtuy (changes requested):

PR #2026, opened 2021-07-03
MCOL-1205 Remove old tests.

denis0x0D:
This patch removes some tests which check `circular join` error.


mariadb-DenisKhalikov:
Hey @tntnatbry after MCOL-1205 CS supports inner circular joins, I've just removed some tests which were introduced in automatic select handler, but I'm not familiar with automatic select handler logic, does it require to update something internally? Thanks.

drrtuy (approved):

tntnatbry:
@denis0x0D This looks good to me. As part of MCOL-4525, I added test cases which would error out. Now that MCOL-1205 is supported, we no longer need this failing case.

PR #2031, opened 2021-07-06
MCOL-4801 Replace Row methods getStringLength() and getStringPointer(…

abarkov:
…) to getConstString()

### DIFF
--- a/utils/common/conststring.h
+++ b/utils/common/conststring.h
@@ -38,9 +38,17 @@ public:
   const char *str() const { return mStr; }
   const char *end() const { return mStr + mLength; }
   size_t length() const { return mLength; }
+  std::string toString() const
+  {
+    return std::string(mStr, mLength);
+  }
   bool eq(char ch) const
   {
     return mLength == 1 && mStr[0] == ch;
+  }
+  bool eq(const ConstString &rhs) const
+  {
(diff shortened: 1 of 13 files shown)


utils/common/conststring.h
```diff
@@ -38,10 +38,18 @@ class ConstString
   const char *str() const { return mStr; }
   const char *end() const { return mStr + mLength; }
   size_t length() const { return mLength; }
```
drrtuy, line 40: IMHO it is worth to add ConstString::empty() to replace expressions like str.length() == 0
I would rename ConstString::str into ConstString::c_str. This improves readability and allow to use ConstString::str for std::string getter.

drrtuy (approved):
The patch is good. It is up to you how to deal with my comments to this patch.

utils/common/conststring.h
abarkov: Thanks for the review!

c_str() implies that the returned value is a NULL terminated C style string. It is not the case. The NULL terminator is not guaranteed to be there. Let's keep it str().  We could perhaps rename it to data(), for std::string() compatibility.

I'm not sure at this point that we should add empty() as an alias for length()==0.
It's very likely that in the future we'll return {0,0} to indicate SQL NULL value and return {"",0} to return MCS EMPTY value.
I'm asking to postpone adding empty() to a later time, when we're more certain with NULL vs EMPTY.

PR #2032, opened 2021-07-06
MCOL-4802 Removed ByteStream methods for bool and add some logging in…

drrtuy:
…to I_S.columnstore_files


mariadb-DenisKhalikov (approved):
Works for me, thanks.

PR #2040, opened 2021-07-08
MCOL-641 Move MTR tests from future/ to basic/

tntnatbry:


drrtuy (changes requested):
If you moving the tests then mcol641.functions should be changed. There is an outdated results for cos(). The current output is actual.

drrtuy (approved):

drrtuy:
@tntnatbry Could you rebase on top of the current HEAD and push it again?

PR #2041, opened 2021-07-08
Add LZ4 libs for debian depends

denis0x0D:


debian/control
```diff
@@ -3,6 +3,7 @@ Architecture: amd64
 Depends: binutils,
          libjemalloc1 | libjemalloc2,
          libsnappy1 | libsnappy1v5,
+         liblz4-1, liblz4-dev,
```
drrtuy: We don't need liblz4-dev in a user's runtime.

drrtuy (changes requested):

debian/control
denis0x0D: Fixed

drrtuy (approved):

PR #2042, opened 2021-07-08
MCOL-4786 Fix wrong comparison for the filters.

denis0x0D:
Fix wrong comparison for the filters while creating case function.


denis0x0D:
Will add tests as well.

drrtuy (approved):

PR #2045, opened 2021-07-09
MCOL-1175: add ability to encrypt CEJ password and use in Columnstore.xml

benthompson15:


drrtuy (dismissed):
At least DEB packages won't have the dependency.

davidhall586 (approved):

PR #2046, opened 2021-07-12
[MCOL-4786] Fix filter comparison.

denis0x0D:
Compare ParseTree by dereferencing pointers.


denis0x0D:
Another problem that CS merges filter which is out of "case function", if it equals to one from "case function".

davidhall586 (approved):

PR #2052, opened 2021-07-16
MCOL-4815 ColumnCommand was replaced with a set of derived classes sp…

drrtuy:
…ecified by

column width

RTSCommand was modified to use a fabric that produces CC class based on column width

NB this patch doesn't affect PseudoCC that also leverages ColumnCommand

### DIFF
--- a/primitives/primproc/columncommand.cpp
+++ b/primitives/primproc/columncommand.cpp
@@ -65,6 +65,12 @@ ColumnCommand::ColumnCommand() :
 {
 }
 
+ColumnCommand::ColumnCommand(execplan::CalpontSystemCatalog::ColType& aColType)
+: ColumnCommand()
+{
+    colType = aColType;
+}
+
 ColumnCommand::~ColumnCommand() { }
 
 void ColumnCommand::_execute()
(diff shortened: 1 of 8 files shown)


primitives/primproc/columncommand.h
```diff
@@ -175,6 +196,82 @@ class ColumnCommand : public Command
     friend class RTSCommand;
 };
 
+using ColumnCommandShPtr = std::unique_ptr<ColumnCommand>;
```
mariadb-AlexeyAntipovsky: What does `Sh` in `ColumnCommandShPtr` stand for?

primitives/primproc/columncommand.cpp
```diff
@@ -463,44 +544,20 @@ void ColumnCommand::processResult()
...
 }
 
+// Used by PseudoCC at least
 void ColumnCommand::createCommand(ByteStream& bs)
 {
     uint8_t tmp8;
-    uint32_t tmp32;
-
     bs.advance(1);
+    colType.unserialize(bs);
```
mariadb-AlexeyAntipovsky, line 552: Could you please move Command serialization/deserialization closer to each other? Maybe it makes sense to create separate functions for this purpose, and in ColumnCommand and ColumnCommandJL just call them.

mariadb-AlexeyAntipovsky (changes requested):

primitives/primproc/columncommand.h
drrtuy: Good point. The name is misleading.

primitives/primproc/columncommand.cpp
mariadb-AlexeyAntipovsky: Decided to do this refactoring separately.

mariadb-AlexeyAntipovsky (approved):

PR #2057, opened 2021-07-22
MCOL-4719 iterate into subquery looking for windowfunctions

davidhall586:
When an outer query filter accesses an subquery column that contains an aggregate or a window function, certain optimizations can't be performed. We had been looking at the surface of the returned column. We now iterate into any functions or operations looking for aggregates and window functions.

### DIFF
--- a/dbcon/execplan/simplecolumn.cpp
+++ b/dbcon/execplan/simplecolumn.cpp
@@ -471,21 +471,24 @@ bool SimpleColumn::sameColumn(const ReturnedColumn* rc) const
 
 void SimpleColumn::setDerivedTable()
 {
-    if (hasAggregate())
+    if (hasAggregate() || hasWindowFunc())
     {
         fDerivedTable = "";
         return;
     }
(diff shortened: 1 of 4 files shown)


dbcon/execplan/simplecolumn.cpp
```diff
@@ -471,22 +471,25 @@ bool SimpleColumn::sameColumn(const ReturnedColumn* rc) const
 
 void SimpleColumn::setDerivedTable()
 {
-    if (hasAggregate())
+    if (hasAggregate() || hasWindowFunc())
     {
         fDerivedTable = "";
         return;
     }
+    ReturnedColumn* rc = dynamic_cast<ReturnedColumn*>(fDerivedRefCol);
```
drrtuy, line 479: You don't need full-blown dynamic_cast here and simple typeid call would be enough. dynamic_cast traverses the whole tree of the classes hierarchy whilst typeid don't.

tntnatbry (approved):

PR #2063, opened 2021-08-03
[MCOL-4829] Compression for the temp disk-based aggregation files

mariadb-AlexeyAntipovsky:


utils/rowgroup/rowaggregation.cpp
```diff
@@ -684,6 +684,12 @@ void RowAggregation::initialize()
 
     config::Config* config = config::Config::makeConfig();
     string tmpDir = config->getTempFileDir(config::Config::TempDirPurpose::Aggregates);
+    bool compress = false;
+    string compStr = config->getConfig("RowAggregation", "Compression");
+    if (boost::iequals(compStr, "snappy"))
```
drrtuy: iequals internally takes a global kernel lock getting locale so plz don't use it.

oam/etc/Columnstore.xml
```diff
@@ -524,6 +524,7 @@
 		<!-- <RowAggrBuckets>32</RowAggrBuckets> --> <!-- Default value is number of cores * 4 -->
 		<!-- <RowAggrRowGroupsPerThread>20</RowAggrRowGroupsPerThread> --> <!-- Default value is 20 -->
 		<!-- <AllowDiskBasedAggregation>N</AllowDiskBasedAggregation> --> <!-- Default value is N -->
+		<!-- <Compression>SNAPPY</Compression> --> <!-- Disabled by default -->
```
drrtuy, line 526: This stanza is for default network compression method.

utils/rowgroup/rowaggregation.cpp
mariadb-AlexeyAntipovsky: Right, forgot about that. Fixed.

drrtuy (approved):

PR #2088, opened 2021-08-17
Wrong power of 2 in esimator`

mariadb-LeonidFedorov:
To calculate limits we should not xor 2 but to power


davidhall586:
Silliness. ^ is used for power of in math texts, so naturally someone would make this mistake.

davidhall586 (approved):

PR #2101, opened 2021-08-20
MCOL-4810 Redundant copying and wasting memory in PrimProc

denis0x0D:


utils/messageqcpp/inetstreamsocket.cpp
```diff
@@ -525,153 +504,141 @@ const SBS InetStreamSocket::read(const struct ::timespec* timeout, bool* isTimeO
...
+        // the following fills the warning log.
+        //		if (isTimeOut && *isTimeOut)
+        //		{
+        //			logIoError("InetStreamSocket::read: timeout during readToMagic", 0);
+        //		}
+        return SBS(new ByteStream(0));
+    }
 
-        if (timeout != NULL)
+    try
```
drrtuy: I am curios why do we need try here? The code rethrows all types of exceptions that must be captured by higher levels anyway.

utils/messageqcpp/inetstreamsocket.cpp
```diff
@@ -94,6 +96,8 @@ using boost::scoped_array;
 #include "logger.h"
 #include "loggingid.h"
 #include "idbcompress.h"
+#include "rowgroup.h"
+using namespace rowgroup;
```
drrtuy: IMHO this is not a good idea b/c 'using namespace' can get us into symbols collision later. Very unfortunately it is widely used in the code but we shouldn't add more havoc except in unit tests.

utils/messageqcpp/inetstreamsocket.cpp
```diff
@@ -525,153 +504,141 @@ const SBS InetStreamSocket::read(const struct ::timespec* timeout, bool* isTimeO
...
+
+        // Read the number of the `long strings`.
+        uint32_t longStringSize;
+        if (!readFixedSizeData(pfd, reinterpret_cast<uint8_t*>(&longStringSize),
+                               sizeof(longStringSize), timeout, isTimeOut, stats, msecs))
+            return SBS(new ByteStream(0));
+
+        // Read the actual data of the `ByteStream`.
+        SBS res(new ByteStream(msglen));
+        //        uint8_t* bufp = res->getInputPtr();
```
drrtuy: Do we need this?

utils/messageqcpp/inetstreamsocket.cpp
```diff
@@ -692,26 +659,40 @@ void InetStreamSocket::do_write(const ByteStream& msg, uint32_t whichMagic, Stat
...
-        written(fSocketParms.sd(), (const uint8_t*)realBuf, msglen + sizeof(msglen) + sizeof(magic));
+        auto bytesToWrite = sizeof(msglen) + sizeof(magic) + sizeof(uint32_t) + msglen;
+        written(fSocketParms.sd(), (const uint8_t*) realBuf, bytesToWrite);
+
+        for (const auto& longString : longStrings)
+        {
+            const StringStore::MemChunk* memChunk =
+                reinterpret_cast<StringStore::MemChunk*>(longString.get());
+            written(fSocketParms.sd(), (const uint8_t*) longString.get(),
+                    memChunk->currentSize + sizeof(StringStore::MemChunk));
```
drrtuy: We calculate the sum twice so we can put it into a variable on a stack.

utils/messageqcpp/inetstreamsocket.h
```diff
@@ -237,6 +237,9 @@ class InetStreamSocket : public Socket
 
     void do_write(const ByteStream& msg, uint32_t magic, Stats* stats = NULL) const;
     ssize_t written(int fd, const uint8_t* ptr, size_t nbytes) const;
+    bool readFixedSizeData(struct pollfd* pfd, uint8_t* buffer, size_t numberOfBytes,
```
drrtuy: I believe that numberOfBytes should have const.

utils/rowgroup/rowgroup.h
```diff
@@ -191,7 +184,14 @@ class StringStore
         uint8_t data[];
     };
 
-    std::vector<boost::shared_array<uint8_t> > mem;
+private:
+    std::string empty_str;
+
+    StringStore(const StringStore&);
+    StringStore& operator=(const StringStore&);
+    static const uint32_t CHUNK_SIZE = 64 * 1024;  // allocators like powers of 2
```
drrtuy: constexpr ?

drrtuy (commented):
Nicely done. Plz take a look at the comments I left.
This patch highlights couple inefficiencies of long strings serialization and StringStore format:
- IS::read() calls new N times(N = number of long strings sent). It would be better to send a cummulative buf size that StringStore needs for all strings. Surely StringStore needs a vector of offsets to navigate through the cummulative buffer.
- vector<shared_array<uint8_t>> doesn't look reasonable b/c of a multiple overhead of shared_array. std::unique_ptr looks more appropriate IMHO.

Could you estimate the effort for the first suggestion?

utils/messageqcpp/inetstreamsocket.cpp
denis0x0D: Thanks, fixed. I was mistaken.

utils/messageqcpp/inetstreamsocket.cpp
denis0x0D: Fixed.

utils/messageqcpp/inetstreamsocket.cpp
denis0x0D: Thanks, fixed.

utils/messageqcpp/inetstreamsocket.cpp
denis0x0D: Fixed.

utils/rowgroup/rowgroup.h
denis0x0D: Fixed.

drrtuy (approved):

PR #2105, opened 2021-08-27
MCOL-4566 Fixes after review.

denis0x0D:
Added:
1. More output logs.
2. Check for CS service state.
3. Save EM to file after rebuilding.

### DIFF
--- a/tools/rebuildEM/main.cpp
+++ b/tools/rebuildEM/main.cpp
@@ -89,8 +89,7 @@ int main(int argc, char** argv)
     std::cout << "The launch of mcsRebuildEM tool must be sanctioned by MariaDB support. "
               << std::endl;
     std::cout << "Requirement: all DBRoots must be on this node. " << std::endl;
-    std::cout << "Note: that the launch can break the cluster." << std::endl;
-    std::cout << "Do you want to continue Y/N? " << std::endl;
+    std::cout << "Do you want to continue Y/N? ";
     std::string confirmation;
     cin >> confirmation;
     if (confirmation.size() == 0)
@@ -124,6 +123,7 @@ int main(int argc, char** argv)
         std::cout << "If `BRM_saves_em` "
                      "exists extent map will be restored from it. "
                   << std::endl;
+        std::cout << "Exiting. " << std::endl;
         return 0;
     }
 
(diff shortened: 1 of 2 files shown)


tools/rebuildEM/main.cpp
```diff
@@ -77,6 +78,22 @@ int main(int argc, char** argv)
         }
     }
 
+    // Check that `mariadb-columstore` is in inactive state.
+    const std::string checkColumnstoreService =
+        "if systemctl status mariadb-columnstore | grep -q "
+        "\"inactive\"\n then\n exit 1\n else\n exit 0\n fi\n";
```
grooverdan: There's no need to do this grep and shell. Look at man systemctl and the exit status is:

```
       ┌──────┬──────────────────────────────────────────┬─────────────────────────────────────┐
       │Value │ Description in LSB                       │ Use in systemd                      │
       ├──────┼──────────────────────────────────────────┼─────────────────────────────────────┤
       │0     │ "program is running or service is OK"    │ unit is active                      │
       ├──────┼──────────────────────────────────────────┼─────────────────────────────────────┤
       │1     │ "program is dead and /var/run pid file   │ unit not failed (used by is-failed) │
       │      │ exists"                                  │                                     │
       ├──────┼──────────────────────────────────────────┼─────────────────────────────────────┤
       │2     │ "program is dead and /var/lock lock file │ unused                              │
       │      │ exists"                                  │                                     │
       ├──────┼──────────────────────────────────────────┼─────────────────────────────────────┤
       │3     │ "program is not running"                 │ unit is not active                  │
       ├──────┼──────────────────────────────────────────┼─────────────────────────────────────┤
       │4     │ "program or service status is unknown"   │ no such unit                        │
       └──────┴──────────────────────────────────────────┴─────────────────────────────────────┘
```

So you can use that exit status directly.

As a better invocation without output and minimal examination of unneeded things:`systemctl is-active --quiet mariadb-columnstore.service`

drrtuy: Thx for the note @grooverdan

denis0x0D: @grooverdan thanks!

drrtuy (approved):

PR #2107, opened 2021-08-31
This patch adds ORDER BY to make the tests outputs deterministic

drrtuy:


mysql-test/columnstore/basic/t/mcs47_cpimport_central_loc_sin_source.test
```diff
@@ -25,7 +25,7 @@ CREATE TABLE t1(col1 INT, col2 INT, col3 CHAR(8)) ENGINE=Columnstore;
 --enable_result_log
 
 #Validate data
-SELECT * FROM t1;
+SELECT * FROM t1 ORDER BY col1, col3;
```
mariadb-AlexeyAntipovsky: Wouldn't it be better to use `sorted_result` pragma?

mariadb-AlexeyAntipovsky (changes requested):

mysql-test/columnstore/basic/t/mcs47_cpimport_central_loc_sin_source.test
drrtuy: Fixed.

mariadb-AlexeyAntipovsky (approved):

PR #2108, opened 2021-09-05
MCOL-4858 This patch fixes HWM comparison for 16 columns and reduces …

drrtuy:
…boilerplate code for other column widths

### DIFF
--- a/writeengine/bulk/we_tableinfo.cpp
+++ b/writeengine/bulk/we_tableinfo.cpp
@@ -72,7 +72,8 @@ namespace WriteEngine
 // Helpers
 int TableInfo::compareHWMs(const int smallestColumnId,
     const int widerColumnId,
-    const size_t widerColumnWidth,
+    const uint32_t smallerColumnWidth,
+    const uint32_t widerColumnWidth,
     const std::vector<DBRootExtentInfo>& segFileInfo,
     int& colIdx)
 {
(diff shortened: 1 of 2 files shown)


writeengine/bulk/we_tableinfo.cpp
```diff
@@ -2232,78 +2234,31 @@ int TableInfo::validateColumnHWMs(
...
-    //  "; hwm-" << segFileInfo[byte8First].fLocalHwm << std::endl;
-
     // Validate/compare HWMs given a 1-byte column as a starting point
     if (byte1First >= 0)
     {
         refCol = byte1First;
 
-        if (byte2First >= 0)
+        if ((rc = compareHWMs(byte1First, byte2First, 1, 2,
+             segFileInfo, colIdx) > 0))
```
mariadb-AlexeyAntipovsky: Looks like using `!= NO_ERROR` instead of `> 0` would be more understandable

mariadb-AlexeyAntipovsky (changes requested):

writeengine/bulk/we_tableinfo.cpp
drrtuy: Good point indeed.

mariadb-AlexeyAntipovsky (approved):

PR #2114, opened 2021-09-09
MCOL-4852: Eliminate build race condition, messageids is generated

cvicentiu:
This ensures that building loggingcpp has messageids.h generated beforehand, just like errorids.h .


cvicentiu:
Hi @drrtuy ! This patch will fix a sporadic failure when building columnstore. We've identified the problem in the new buildbot. See https://jira.mariadb.org/browse/MCOL-4852

drrtuy:
Thx for the patch @cvicentiu. Could I ask you to make PR for develop-6 branch also b/c it should affect 10.6 builds.

drrtuy (approved):

PR #2116, opened 2021-09-16
MCOL-4728 Query with unusual use of aggregate functions on ColumnStor…

abarkov:
…e table crashes MariaDB Server


mariadb-DenisKhalikov (approved):
LGTM

PR #2118, opened 2021-09-28
MCOL-4871 Extent elimination for short CHAR does not work well

abarkov:


mariadb-SergeyZefirov (approved):

primitives/linux-port/column.cpp
```diff
@@ -1240,6 +1240,14 @@ void filterColumnData(
     T Min = datatypes::numeric_limits<T>::max();
     T Max = (KIND == KIND_UNSIGNED) ? 0 : datatypes::numeric_limits<T>::min();
 
+   if (KIND == KIND_TEXT)
```
drrtuy, line 1243: It would be better to have a template function that assigns/returns Min/Max having <T, KIND> as its params and write couple SFINAE versions of the function.

drrtuy (changes requested):

primitives/linux-port/column.cpp
drrtuy: After some consideration I arrived to the conclusion I can fix this myself later.

PR #2124, opened 2021-10-14
MCOL-4810 Add support for missed operation.

denis0x0D:


utils/messageqcpp/bytestream.cpp
```diff
@@ -564,7 +565,10 @@ bool ByteStream::operator==(const ByteStream& b) const
     if (b.length() != length())
         return false;
 
-    return (memcmp(fCurOutPtr, b.fCurOutPtr, length()) == 0);
+    if (memcmp(fCurOutPtr, b.fCurOutPtr, length()) != 0)
+      return false;
+
+    return longStrings == b.longStrings;
```
drrtuy: Here is an interesting question about semantics of BS::operator==. If we have two equal BS do they need to have the same longStrings ptrs or the contents of longStrings buffers must be equal? With this approach we compare the uint8_t pointers. What do you think @denis0x0D ?

drrtuy: JFYI  [shared_array::operator==](https://www.boost.org/doc/libs/1_64_0/boost/smart_ptr/shared_array.hpp)

drrtuy: JFYI  [shared_array::operator==](https://www.boost.org/doc/libs/1_64_0/boost/smart_ptr/shared_array.hpp)

denis0x0D: @drrtuy you are right, we are comparing pointers, I was originally thinking that it's a right approach, since we are sharing data for long strings, but seems that I was wrong, we have to make a deep cmp operations. Thanks for looking on it.

drrtuy (approved):

drrtuy:
Could you rebase in top of the current develop HEAD and run the patch through the ci?

denis0x0D:
It seems like the patch is applied to HEAD on CLI
5	+ git fetch '--depth=10' origin +refs/heads/develop:
8s
6	From https://github.com/mariadb-corporation/mariadb-columnstore-engine
11s
7	 * branch            develop    -> FETCH_HEAD
11s
8	 * [new branch]      develop    -> origin/develop
11s
9	+ git checkout develop
11s
10	Branch 'develop' set up to track remote branch 'develop' from 'origin'.
11s
11	Switched to a new branch 'develop'
11s
12	+ git fetch origin refs/pull/2124/head:
11s
13	From https://github.com/mariadb-corporation/mariadb-columnstore-engine
12s
14	 * branch            refs/pull/2124/head -> FETCH_HEAD
12s
15	+ git merge 867650d5c98b8f25f648fadc9c05921fd036ce68
12s
16	Updating 07f2e59..867650d

I'll trigger rebuild, not sure what errors are meaning.

PR #2131, opened 2021-10-25
MCOL-4841 dev-6 Handle large join results

davidhall586:
Use a loop to shrink the number of results reported per message to something manageable.
See MCOL-4841 for more information.

Also in this PR is the fix for MCOL-4846 Switch to disk based join causes segv

Added thread names to some unnamed threads for easier debugging.

### DIFF
--- a/dbcon/joblist/distributedenginecomm.cpp
+++ b/dbcon/joblist/distributedenginecomm.cpp
@@ -308,6 +308,12 @@ void DistributedEngineComm::Setup()
 
             writeToLog(__FILE__, __LINE__, "Could not connect to PMS" + std::to_string(connectionId) + ": " + ex.what(), LOG_TYPE_ERROR);
             cerr << "Could not connect to PMS" << std::to_string(connectionId) << ": " << ex.what() << endl;
+            
+            if (newPmCount == 0)
+            {
+                writeToLog(__FILE__, __LINE__, "No more PMs to try to connect to", LOG_TYPE_ERROR);
+                break;
+            }
         }
         catch (...)
         {
@@ -315,6 +321,11 @@ void DistributedEngineComm::Setup()
                 newPmCount--;
 
             writeToLog(__FILE__, __LINE__, "Could not connect to PMS" + std::to_string(connectionId), LOG_TYPE_ERROR);
+            if (newPmCount == 0)
+            {
+                writeToLog(__FILE__, __LINE__, "No more PMs to try to connect to", LOG_TYPE_ERROR);
+                                break;
+            }
         }
     }
 
--- a/dbcon/joblist/primitivestep.h
+++ b/dbcon/joblist/primitivestep.h
@@ -1421,7 +1421,8 @@ private:
     bool BPPIsAllocated;
     uint32_t uniqueID;
     ResourceManager* fRm;
-
+    boost::shared_ptr<int64_t> fJoinMemLimit;
+    
     /* HashJoin support */
 
     void serializeJoiner();
@@ -1430,8 +1431,10 @@ private:
     void generateJoinResultSet(const std::vector<std::vector<rowgroup::Row::Pointer> >& joinerOutput,
                                rowgroup::Row& baseRow, const std::vector<boost::shared_array<int> >& mappings,
                                const uint32_t depth, rowgroup::RowGroup& outputRG, rowgroup::RGData& rgData,
-                               std::vector<rowgroup::RGData>* outputData,
-                               const boost::scoped_array<rowgroup::Row>& smallRows, rowgroup::Row& joinedRow);
+                               std::vector<rowgroup::RGData>& outputData,
+                               const boost::scoped_array<rowgroup::Row>& smallRows, rowgroup::Row& joinedRow,
+                               RowGroupDL* dlp, rowgroup::RowGroup& fe2OutputRG, rowgroup::Row& fe2OutRow,
+                               funcexp::FuncExpWrapper& fe2, uint64_t memSizeForOutputRG);
 
     std::vector<boost::shared_ptr<joiner::TupleJoiner> > tjoiners;
     bool doJoin, hasPMJoin, hasUMJoin;
--- a/dbcon/joblist/primitivestep.h
+++ b/dbcon/joblist/primitivestep.h
@@ -1428,13 +1428,13 @@ private:
     void serializeJoiner();
     void serializeJoiner(uint32_t connectionNumber);
 
-    void generateJoinResultSet(const std::vector<std::vector<rowgroup::Row::Pointer> >& joinerOutput,
+    uint64_t generateJoinResultSet(const std::vector<std::vector<rowgroup::Row::Pointer> >& joinerOutput,
                                rowgroup::Row& baseRow, const std::vector<boost::shared_array<int> >& mappings,
                                const uint32_t depth, rowgroup::RowGroup& outputRG, rowgroup::RGData& rgData,
                                std::vector<rowgroup::RGData>& outputData,
                                const boost::scoped_array<rowgroup::Row>& smallRows, rowgroup::Row& joinedRow,
                                RowGroupDL* dlp, rowgroup::RowGroup& fe2OutputRG, rowgroup::Row& fe2OutRow,
-                               funcexp::FuncExpWrapper& fe2, uint64_t memSizeForOutputRG);
+                               funcexp::FuncExpWrapper& fe2);
 
     std::vector<boost::shared_ptr<joiner::TupleJoiner> > tjoiners;
     bool doJoin, hasPMJoin, hasUMJoin;
--- a/dbcon/joblist/resourcemanager.cpp
+++ b/dbcon/joblist/resourcemanager.cpp
@@ -414,7 +414,7 @@ bool ResourceManager::userPriorityEnabled() const
 // Counts memory. This funtion doesn't actually malloc, just counts against two limits
 // totalUmMemLimit for overall UM counting and (optional) sessionLimit for a single session.
 // If both have space, return true.
-bool ResourceManager::getMemory(int64_t amount, boost::shared_ptr<int64_t> sessionLimit, bool patience)
+bool ResourceManager::getMemory(int64_t amount, boost::shared_ptr<int64_t>& sessionLimit, bool patience)
 {
     bool ret1 = (atomicops::atomicSub(&totalUmMemLimit, amount) >= 0);
     bool ret2 = sessionLimit ? (atomicops::atomicSub(sessionLimit.get(), amount) >= 0) : ret1;
@@ -434,11 +434,6 @@ bool ResourceManager::getMemory(int64_t amount, boost::shared_ptr<int64_t> sessi
         // If  we  didn't  get any memory, restore the counters.
         atomicops::atomicAdd(&totalUmMemLimit, amount);
         sessionLimit ? atomicops::atomicAdd(sessionLimit.get(), amount) : 0;
-    }
-    // Debug: a place to put a breakpoint if totalUMMemLimit drops below 0. It should enver do that.
-    if  (totalUmMemLimit < 0)
-    {
-        return (ret1 && ret2);
     }
     return (ret1 && ret2);
 }
--- a/dbcon/joblist/resourcemanager.cpp
+++ b/dbcon/joblist/resourcemanager.cpp
@@ -411,22 +411,35 @@ bool ResourceManager::userPriorityEnabled() const
     return "Y" == val;
 }
 
+// Counts memory. This funtion doesn't actually malloc, just counts against two limits
+// totalUmMemLimit for overall UM counting and (optional) sessionLimit for a single session.
+// If both have space, return true.
 bool ResourceManager::getMemory(int64_t amount, boost::shared_ptr<int64_t> sessionLimit, bool patience)
 {
     bool ret1 = (atomicops::atomicSub(&totalUmMemLimit, amount) >= 0);
-    bool ret2 = (atomicops::atomicSub(sessionLimit.get(), amount) >= 0);
+    bool ret2 = sessionLimit ? (atomicops::atomicSub(sessionLimit.get(), amount) >= 0) : ret1;
 
     uint32_t retryCounter = 0, maxRetries = 20;   // 10s delay
 
     while (patience && !(ret1 && ret2) && retryCounter++ < maxRetries)
     {
         atomicops::atomicAdd(&totalUmMemLimit, amount);
-        atomicops::atomicAdd(sessionLimit.get(), amount);
+        sessionLimit ? atomicops::atomicAdd(sessionLimit.get(), amount) : 0;
         usleep(500000);
         ret1 = (atomicops::atomicSub(&totalUmMemLimit, amount) >= 0);
-        ret2 = (atomicops::atomicSub(sessionLimit.get(), amount) >= 0);
+        ret2 = sessionLimit ? (atomicops::atomicSub(sessionLimit.get(), amount) >= 0) : ret1;
     }
-
+    if  (!(ret1 && ret2))
+    {
+        // If  we  didn't  get any memory, restore the counters.
+        atomicops::atomicAdd(&totalUmMemLimit, amount);
+        sessionLimit ? atomicops::atomicAdd(sessionLimit.get(), amount) : 0;
+    }
+    // Debug: a place to put a breakpoint if totalUMMemLimit drops below 0. It should enver do that.
+    if  (totalUmMemLimit < 0)
+    {
+        return (ret1 && ret2);
+    }
     return (ret1 && ret2);
 }
 
--- a/dbcon/joblist/resourcemanager.h
+++ b/dbcon/joblist/resourcemanager.h
@@ -82,6 +82,7 @@ const uint32_t defaultRequestSize  = 1;
 const uint32_t defaultMaxOutstandingRequests = 20;
 const uint32_t defaultProcessorThreadsPerScan = 16;
 const uint32_t defaultJoinerChunkSize = 16 * 1024 * 1024;
+const uint64_t defaultMaxBPPSendQueue = 250000000; // 250MB
 
 //bucketreuse
 const std::string defaultTempDiskPath = "/tmp";
@@ -381,6 +382,11 @@ public:
         return getUintVal(fJobListStr, "DECThrottleThreshold", defaultDECThrottleThreshold);
     }
 
+    uint64_t    getMaxBPPSendQueue() const
+    {
+        return getUintVal(fPrimitiveServersStr, "MaxBPPSendQueue", defaultMaxBPPSendQueue);
+    }
+    
     EXPORT void  emServerThreads();
     EXPORT void  emServerQueueSize();
     EXPORT void  emSecondsBetweenMemChecks();
@@ -402,7 +408,7 @@ public:
     inline void returnMemory(int64_t amount, boost::shared_ptr<int64_t> sessionLimit)
     {
         atomicops::atomicAdd(&totalUmMemLimit, amount);
-        atomicops::atomicAdd(sessionLimit.get(), amount);
+        sessionLimit? atomicops::atomicAdd(sessionLimit.get(), amount): 0;
     }
     inline int64_t availableMemory() const
     {
@@ -673,3 +679,4 @@ inline bool ResourceManager::getBoolVal(const std::string& section, const std::s
 #undef EXPORT
 
 #endif
+
--- a/dbcon/joblist/resourcemanager.h
+++ b/dbcon/joblist/resourcemanager.h
@@ -404,11 +404,11 @@ public:
     /* sessionLimit is a pointer to the var holding the session-scope limit, should be JobInfo.umMemLimit
        for the query. */
     /* Temporary parameter 'patience', will wait for up to 10s to get the memory. */
-    EXPORT bool getMemory(int64_t amount, boost::shared_ptr<int64_t> sessionLimit, bool patience = true);
-    inline void returnMemory(int64_t amount, boost::shared_ptr<int64_t> sessionLimit)
+    EXPORT bool getMemory(int64_t amount, boost::shared_ptr<int64_t>& sessionLimit, bool patience = true);
+    inline void returnMemory(int64_t amount, boost::shared_ptr<int64_t>& sessionLimit)
     {
         atomicops::atomicAdd(&totalUmMemLimit, amount);
-        sessionLimit? atomicops::atomicAdd(sessionLimit.get(), amount): 0;
+        sessionLimit ? atomicops::atomicAdd(sessionLimit.get(), amount): 0;
     }
     inline int64_t availableMemory() const
     {
--- a/dbcon/joblist/tuple-bps.cpp
+++ b/dbcon/joblist/tuple-bps.cpp
@@ -272,11 +272,13 @@ TupleBPS::TupleBPS(const pColStep& rhs, const JobInfo& jobInfo) :
     fDelivery = false;
     fExtendedInfo = "TBPS: ";
     fQtc.stepParms().stepType = StepTeleStats::T_BPS;
+    fJoinMemLimit.reset();
+#if 0    
     if (jobInfo.umMemLimit && *jobInfo.umMemLimit > 0)
         fJoinMemLimit.reset(new int64_t(min((uint64_t)*jobInfo.umMemLimit, fRm->getConfiguredUMMemLimit() / 4)));
     else
         fJoinMemLimit.reset(new int64_t(fRm->getConfiguredUMMemLimit() / 4));
-    
+#endif    
     hasPCFilter = hasPMFilter = hasRIDFilter = hasSegmentFilter = hasDBRootFilter = hasSegmentDirFilter =
                                     hasPartitionFilter = hasMaxFilter = hasMinFilter = hasLBIDFilter = hasExtentIDFilter = false;
 }
@@ -360,10 +362,7 @@ TupleBPS::TupleBPS(const pColScanStep& rhs, const JobInfo& jobInfo) :
 
     initExtentMarkers();
     fQtc.stepParms().stepType = StepTeleStats::T_BPS;
-    if (jobInfo.umMemLimit && *jobInfo.umMemLimit > 0)
-        fJoinMemLimit.reset(new int64_t(min((uint64_t)*jobInfo.umMemLimit, fRm->getConfiguredUMMemLimit() / 4)));
-    else
-        fJoinMemLimit.reset(new int64_t(fRm->getConfiguredUMMemLimit() / 4));
+    fJoinMemLimit.reset();
     
     hasPCFilter = hasPMFilter = hasRIDFilter = hasSegmentFilter = hasDBRootFilter = hasSegmentDirFilter =
                                     hasPartitionFilter = hasMaxFilter = hasMinFilter = hasLBIDFilter = hasExtentIDFilter = false;
@@ -431,10 +430,7 @@ TupleBPS::TupleBPS(const PassThruStep& rhs, const JobInfo& jobInfo) :
     fDelivery = false;
     fExtendedInfo = "TBPS: ";
     fQtc.stepParms().stepType = StepTeleStats::T_BPS;
-    if (jobInfo.umMemLimit && *jobInfo.umMemLimit > 0)
-        fJoinMemLimit.reset(new int64_t(min((uint64_t)*jobInfo.umMemLimit, fRm->getConfiguredUMMemLimit() / 4)));
-    else
-        fJoinMemLimit.reset(new int64_t(fRm->getConfiguredUMMemLimit() / 4));
+    fJoinMemLimit.reset();
     
     hasPCFilter = hasPMFilter = hasRIDFilter = hasSegmentFilter = hasDBRootFilter = hasSegmentDirFilter =
                                     hasPartitionFilter = hasMaxFilter = hasMinFilter = hasLBIDFilter = hasExtentIDFilter = false;
@@ -500,10 +496,7 @@ TupleBPS::TupleBPS(const pDictionaryStep& rhs, const JobInfo& jobInfo) :
     fDelivery = false;
     fExtendedInfo = "TBPS: ";
     fQtc.stepParms().stepType = StepTeleStats::T_BPS;
-    if (jobInfo.umMemLimit && *jobInfo.umMemLimit > 0)
-        fJoinMemLimit.reset(new int64_t(min((uint64_t)*jobInfo.umMemLimit, fRm->getConfiguredUMMemLimit() / 4)));
-    else
-        fJoinMemLimit.reset(new int64_t(fRm->getConfiguredUMMemLimit() / 4));
+    fJoinMemLimit.reset();
     
     hasPCFilter = hasPMFilter = hasRIDFilter = hasSegmentFilter = hasDBRootFilter = hasSegmentDirFilter =
                                     hasPartitionFilter = hasMaxFilter = hasMinFilter = hasLBIDFilter = hasExtentIDFilter = false;
@@ -2877,7 +2870,7 @@ uint64_t TupleBPS::generateJoinResultSet(const vector<vector<Row::Pointer> >& jo
     
     if (depth < smallSideCount - 1)
     {
-        for (i = 0; i < joinerOutput[depth].size(); i++)
+        for (i = 0; i < joinerOutput[depth].size() && !fDie; i++)
         {
             smallRow.setPointer(joinerOutput[depth][i]);
             applyMapping(mappings[depth], smallRow, &baseRow);
@@ -2890,7 +2883,7 @@ uint64_t TupleBPS::generateJoinResultSet(const vector<vector<Row::Pointer> >& jo
     {
         outputRG.getRow(outputRG.getRowCount(), &joinedRow);
 
-        for (i = 0; i < joinerOutput[depth].size(); i++, joinedRow.nextRow(),
+        for (i = 0; i < joinerOutput[depth].size() && !fDie; i++, joinedRow.nextRow(),
                 outputRG.incRowCount())
         {
             smallRow.setPointer(joinerOutput[depth][i]);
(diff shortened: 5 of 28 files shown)


drrtuy:
@dhall-MariaDB MTR results suggest me a couple crashes were introduced in ExeMgr. We surely need to address these.

dbcon/joblist/distributedenginecomm.cpp
```diff
@@ -308,13 +308,24 @@ void DistributedEngineComm::Setup()
 
             writeToLog(__FILE__, __LINE__, "Could not connect to PMS" + std::to_string(connectionId) + ": " + ex.what(), LOG_TYPE_ERROR);
             cerr << "Could not connect to PMS" << std::to_string(connectionId) << ": " << ex.what() << endl;
+
```
drrtuy, line 311: These two exception handlers are almost the same(the message differs) so it would be better to reduce boilerplate and put the handlers into a function.

dbcon/joblist/primitivestep.h
```diff
@@ -1421,7 +1421,8 @@ class TupleBPS : public BatchPrimitive, public TupleDeliveryStep
     bool BPPIsAllocated;
     uint32_t uniqueID;
     ResourceManager* fRm;
-
+    boost::shared_ptr<int64_t> fJoinMemLimit;
```
drrtuy, line 1424: It seems that it would be better to introduce a new RM::returnMemory overload b/c shared_ptr is an overkill for a simple atomic counter.

dbcon/joblist/resourcemanager.cpp
```diff
@@ -411,22 +411,35 @@ bool ResourceManager::userPriorityEnabled() const
     return "Y" == val;
 }
 
+// Counts memory. This funtion doesn't actually malloc, just counts against two limits
+// totalUmMemLimit for overall UM counting and (optional) sessionLimit for a single session.
+// If both have space, return true.
 bool ResourceManager::getMemory(int64_t amount, boost::shared_ptr<int64_t> sessionLimit, bool patience)
```
drrtuy: Moreover since we send down boost::shared_ptr<int64_t> and not its ref the code runs shared_ptr ctor that is an extra overhead. I don't see we really need this for the semantics but this particular note goes outside the scope of the patch.

dbcon/joblist/resourcemanager.cpp
```diff
@@ -411,22 +411,35 @@ bool ResourceManager::userPriorityEnabled() const
...
-        ret2 = (atomicops::atomicSub(sessionLimit.get(), amount) >= 0);
+        ret2 = sessionLimit ? (atomicops::atomicSub(sessionLimit.get(), amount) >= 0) : ret1;
+    }
+    if  (!(ret1 && ret2))
+    {
+        // If  we  didn't  get any memory, restore the counters.
+        atomicops::atomicAdd(&totalUmMemLimit, amount);
+        sessionLimit ? atomicops::atomicAdd(sessionLimit.get(), amount) : 0;
+    }
+    // Debug: a place to put a breakpoint if totalUMMemLimit drops below 0. It should enver do that.
```
drrtuy: What do you think about replacing this if with assert or just removing the block?

dbcon/joblist/resourcemanager.cpp
```diff
@@ -411,22 +411,35 @@ bool ResourceManager::userPriorityEnabled() const
...
         usleep(500000);
         ret1 = (atomicops::atomicSub(&totalUmMemLimit, amount) >= 0);
-        ret2 = (atomicops::atomicSub(sessionLimit.get(), amount) >= 0);
+        ret2 = sessionLimit ? (atomicops::atomicSub(sessionLimit.get(), amount) >= 0) : ret1;
+    }
+    if  (!(ret1 && ret2))
+    {
+        // If  we  didn't  get any memory, restore the counters.
+        atomicops::atomicAdd(&totalUmMemLimit, amount);
+        sessionLimit ? atomicops::atomicAdd(sessionLimit.get(), amount) : 0;
```
drrtuy, line 436: In most cases we don't need the second atomic add b/c the query execution ends up here. It might worth to put an explicit comment about this and remove the actual op call.

dbcon/joblist/resourcemanager.h
```diff
@@ -82,6 +82,7 @@ const uint32_t defaultRequestSize  = 1;
 const uint32_t defaultMaxOutstandingRequests = 20;
 const uint32_t defaultProcessorThreadsPerScan = 16;
 const uint32_t defaultJoinerChunkSize = 16 * 1024 * 1024;
+const uint64_t defaultMaxBPPSendQueue = 250000000; // 250MB
```
drrtuy, line 85: What is the reasoning behind this value? Could you share it?

dbcon/joblist/resourcemanager.h
```diff
@@ -402,7 +408,7 @@ class ResourceManager
     inline void returnMemory(int64_t amount, boost::shared_ptr<int64_t> sessionLimit)
     {
         atomicops::atomicAdd(&totalUmMemLimit, amount);
-        atomicops::atomicAdd(sessionLimit.get(), amount);
+        sessionLimit? atomicops::atomicAdd(sessionLimit.get(), amount): 0;
```
drrtuy: It is harder to read and as I mentioned earlier it is better to introduce RM::returnMemory to avoid using shared_ptr.

dbcon/joblist/tuple-bps.cpp
```diff
@@ -273,7 +272,7 @@ TupleBPS::TupleBPS(const pColStep& rhs, const JobInfo& jobInfo) :
     fDelivery = false;
     fExtendedInfo = "TBPS: ";
     fQtc.stepParms().stepType = StepTeleStats::T_BPS;
-
+    fJoinMemLimit.reset(new int64_t(*jobInfo.umMemLimit == 0 ? fRm->getConfiguredUMMemLimit() / 4 : min((uint64_t)*jobInfo.umMemLimit, fRm->getConfiguredUMMemLimit() / 4)));
```
drrtuy: Not sure we need shared_ptr for the purpose.

dbcon/joblist/tuple-bps.cpp
```diff
@@ -358,7 +357,8 @@ TupleBPS::TupleBPS(const pColScanStep& rhs, const JobInfo& jobInfo) :
 
     initExtentMarkers();
     fQtc.stepParms().stepType = StepTeleStats::T_BPS;
-
+    fJoinMemLimit.reset(new int64_t(*jobInfo.umMemLimit == 0 ? fRm->getConfiguredUMMemLimit() / 4 : min((uint64_t)*jobInfo.umMemLimit, fRm->getConfiguredUMMemLimit() / 4)));
```
drrtuy: Not sure we need shared_ptr for the purpose.

dbcon/joblist/tuple-bps.cpp
```diff
@@ -425,7 +425,8 @@ TupleBPS::TupleBPS(const PassThruStep& rhs, const JobInfo& jobInfo) :
     fDelivery = false;
     fExtendedInfo = "TBPS: ";
     fQtc.stepParms().stepType = StepTeleStats::T_BPS;
-
+    fJoinMemLimit.reset(new int64_t(*jobInfo.umMemLimit == 0 ? fRm->getConfiguredUMMemLimit() / 4 : min((uint64_t)*jobInfo.umMemLimit, fRm->getConfiguredUMMemLimit() / 4)));
```
drrtuy: Not sure we need shared_ptr for the purpose.

dbcon/joblist/tuple-bps.cpp
drrtuy: What is the semantics of "/4" op ?

dbcon/joblist/tuple-bps.cpp
```diff
@@ -1933,7 +1934,8 @@ void TupleBPS::receiveMultiPrimitiveMessages(uint32_t threadID)
     bool didEOF = false;
     bool unused;
 
-    /* Join vars */
+    // Join vars
+    // These can't be members of tupleBPS because this* may be used by multiple threads
```
drrtuy, line 1943: Denis rewrote the TBPS::RMPM() so now there is a very limited part of TBPS::RMPM()  logic where multiple threads are used.

drrtuy: And they don't share crit section execution in the most heavy part.

dbcon/joblist/tuple-bps.cpp
```diff
@@ -2312,42 +2315,29 @@ void TupleBPS::receiveMultiPrimitiveMessages(uint32_t threadID)
...
                         rgDatav.push_back(rgData);
                     }
-
+                                            
                     /* Execute UM F & E group 2 on rgDatav */
                     if (fe2 && !runFEonPM && rgDatav.size() > 0 && !cancelled())
                     {
                         processFE2(local_outputRG, local_fe2Output, postJoinRow, local_fe2OutRow, &rgDatav, &local_fe2);
                         rgDataVecToDl(rgDatav, local_fe2Output, dlp);
+                        fRm->returnMemory(memSizeForOutputRG, fJoinMemLimit);
```
drrtuy: I would suggest to change TBPS::generateJoinResultSet return type so that it returns another portion to add to memSizeForOutputRG if needed. Otherwise memSizeForOutputRG change looks like a hard-to-track side-effect that happens somewhere.

dbcon/joblist/tuple-bps.cpp
```diff
@@ -2678,7 +2672,12 @@ void TupleBPS::receiveMultiPrimitiveMessages(uint32_t threadID)
         }
     }
 
-    // Bug 3136, let mini stats to be formatted if traceOn.
+    if (memSizeForOutputRG)
+    {
+        fRm->returnMemory(memSizeForOutputRG, fJoinMemLimit);
```
drrtuy: You have at least 3 places where these two statements are called. It is worth to make it a method or a function.

utils/threadpool/threadpool.cpp
```diff
@@ -89,6 +89,7 @@ void ThreadPool::setQueueSize(size_t queueSize)
 void ThreadPool::pruneThread()
 {
     boost::unique_lock<boost::mutex> lock2(fPruneMutex);
+    utils::setThreadName("pruneThread");
```
drrtuy, line 91: We don't need to extend crit section so setThreadName should be put in front of the mutex lock.

dbcon/joblist/tuple-bps.cpp
```diff
@@ -2884,8 +2886,26 @@ void TupleBPS::generateJoinResultSet(const vector<vector<Row::Pointer> >& joiner
             {
                 uint32_t dbRoot = outputRG.getDBRoot();
                 uint64_t baseRid = outputRG.getBaseRid();
-                outputData->push_back(rgData);
-                rgData = RGData(outputRG);
+                outputData.push_back(rgData);
```
drrtuy, line 2895: I didn't get why did you decide to move this processing here and why didn't you remove it from TBPS::RMPM?

dbcon/joblist/tuple-bps.cpp
```diff
@@ -3382,7 +3402,8 @@ void TupleBPS::abort_nolock()
         }
 
         BPPIsAllocated = false;
-        fDec->shutdownQueue(uniqueID);
+//        fDec->shutdownQueue(uniqueID);
```
drrtuy: Comment here?

dbcon/joblist/tuplehashjoin.cpp
```diff
@@ -407,7 +406,8 @@ void TupleHashJoinStep::smallRunnerFcn(uint32_t index, uint threadID, uint64_t *
             utils::releaseSpinlock(rgdLock);
 
             rgSize = smallRG.getSizeWithStrings();
-            atomicops::atomicAdd(&memUsedByEachJoin[index], rgSize);
+            memUsedByThisJoin += rgSize;
+//            atomicops::atomicAdd(&memUsedByEachJoin[index], rgSize);
```
drrtuy: Comment?

dbcon/joblist/tuplehashjoin.cpp
```diff
@@ -1549,15 +1548,16 @@ void TupleHashJoinStep::joinRunnerFcn(uint32_t threadID)
             if (local_inputRG.getRowCount() == 0)
                 continue;
 
-            joinOneRG(threadID, &joinedRowData, local_inputRG, local_outputRG, largeRow,
-                      joinFERow, joinedRow, baseRow, joinMatches, smallRowTemplates);
+            joinOneRG(threadID, joinedRowData, local_inputRG, local_outputRG, largeRow,
```
drrtuy, line 1550: I strongly disagree that we need to future encourage puting one RGData at a time into upstream DL. It is very, very inefficient. We won't address this with this patch but should have this in mind burrowing this operation deeper down the call stack.

dbcon/joblist/tuplehashjoin.cpp
```diff
@@ -1863,19 +1863,25 @@ void TupleHashJoinStep::generateJoinResultSet(const vector<vector<Row::Pointer>
             {
                 uint32_t dbRoot = l_outputRG.getDBRoot();
                 uint64_t baseRid = l_outputRG.getBaseRid();
-                outputData->push_back(rgData);
+                outputData.push_back(rgData);
+                // Count the memory
+                if (UNLIKELY(!getMemory(l_outputRG.getMaxDataSize())))
```
drrtuy, line 1867: Sorry what do we do here? If i read the semantics right we just drop some of the data w/o an explicit exception.

dbcon/joblist/tuplehashjoin.h
```diff
@@ -427,7 +441,9 @@ class TupleHashJoinStep : public JobStep, public TupleDeliveryStep
     std::vector<std::vector<uint32_t> > smallSideKeys;
 
     ResourceManager* resourceManager;
-
+    uint64_t fMemSizeForOutputRG;
+    boost::shared_ptr<int64_t> fJoinMemLimit;
```
drrtuy, line 445: I disagree we need shared_ptr just b/c RM:: methods provide such API.

dbcon/mysql/ha_mcs_dml.cpp
```diff
@@ -758,7 +758,7 @@ int ha_mcs_impl_write_batch_row_(const uchar* buf, TABLE* table, cal_impl_if::ca
     rc = fprintf(ci.filePtr, "\n"); //@bug 6077 check whether thhe pipe is still open
 
     if ( rc < 0)
-        rc = -1;
+        rc = errno;
```
drrtuy, line 761: Really? IIRC we return this value up the call stack back to MDB that knows nothing regarding errno values after fprintf errors.

dbcon/mysql/ha_mcs_impl.cpp
```diff
@@ -3086,6 +3086,10 @@ void ha_mcs_impl_start_bulk_insert(ha_rows rows, TABLE* table, bool is_cache_ins
             //@bug 6122 Check how many columns have not null constraint. columnn with not null constraint will not show up in header.
             unsigned int numberNotNull = 0;
 
+            // THere's a chance that a previously aborted query left garbage in ci->columnTypes
```
drrtuy: Good catch. Unfortunately we can't just put a partial cleanup here. The clean up must happen inside ha_mcs::impl_external_lock(). If it is impossible we need to find out why it isn't fired up for aborted queries. This change calls for a separate issue IMHO.

oam/etc/Columnstore.xml
```diff
@@ -486,10 +487,10 @@
 		<MaxBuckets>128</MaxBuckets>
 		<MaxElems>128K</MaxElems>  <!-- 128 buckets * 128K * 16 = 256 MB -->
 		<FifoSizeLargeSide>64</FifoSizeLargeSide>
-		<PmMaxMemorySmallSide>1G</PmMaxMemorySmallSide>
+		<PmMaxMemorySmallSide>100M</PmMaxMemorySmallSide>
```
drrtuy: I disagree with this default. If the cloud needs it then put it into init scripts for the cloud image.

oam/etc/Columnstore.xml
```diff
@@ -486,10 +487,10 @@
...
 		<TotalUmMemory>25%</TotalUmMemory>
 		<CPUniqueLimit>100</CPUniqueLimit>
-		<AllowDiskBasedJoin>N</AllowDiskBasedJoin>
+		<AllowDiskBasedJoin>Y</AllowDiskBasedJoin>
```
drrtuy: I disagree with this default either. This slows down JOINs w/o an explicit notification. A user must explicitly enables this IMHO so he knows this .

utils/common/atomicops.h
```diff
@@ -165,6 +165,28 @@ inline bool atomicCAS(volatile T* mem, T comp, T swap)
 #endif
 }
 
+// implements a zero out of a variable
+template <typename T>
+inline void atomicZero(volatile T* mem)
+{
+#ifdef _MSC_VER
```
drrtuy, line 172: I doubt we will have MCS on Windows again.

utils/joiner/joinpartition.cpp
```diff
@@ -122,6 +119,11 @@ JoinPartition::JoinPartition(const RowGroup& lRG,
     {
         compressor.reset(new compress::CompressInterfaceSnappy());
     }
+
+    for (int i = 0; i < (int) bucketCount; i++)
```
drrtuy: Could you share the semantics of this particular change? Why did you move this block? And if you are going to touch this could you remove an explicit int cast?

primitives/primproc/bppsendthread.h
```diff
@@ -100,7 +110,11 @@ class BPPSendThread
     bool waiting;
     boost::mutex ackLock;
     boost::condition okToSend;
-
+    // Condition to prevent run away queue
+    bool respondWait;
+    boost::mutex respondLock;
```
drrtuy, line 118: I am highly against introducing another mutex b/c they introduce a lot of overhead that makes our software less efficient with multiple queries. 
Did I get it right that you are trying to add another flowcontrol b/w PP and EM. What are the triggers that forces PP to wait?

primitives/primproc/batchprimitiveprocessor.cpp
```diff
@@ -1384,8 +1392,18 @@ void BatchPrimitiveProcessor::executeTupleJoin()
             //else
             // cout << "j != joinerCount\n";
         }
+        // If we've accumulated more than 1048576 (2^20)_ of resultCounts, cut off processing. The caller will
+        // restart to continue where we left off.
+        if (resultCount >= 1048576)
```
drrtuy, line 1397: Why did we add this condition and what is the rationale behind the literal?

primitives/primproc/batchprimitiveprocessor.cpp
```diff
@@ -1384,8 +1392,18 @@ void BatchPrimitiveProcessor::executeTupleJoin()
...
+        {
+            newStartRid += newRowCount;
+            break;
+        }
     }
 
+    if (resultCount < 1048576)
```
drrtuy, line 1404: Why did we add this condition and what is the rationale behind the literal?

drrtuy (changes requested):
This patch has its meaning however it introduces another mutex in BPP runtime. This should be future discussed. Maybe together we can find ways to avoid it.
Moreover this functionality must be merged after Denis'es patch of TBPS methods.
And the last thing. This patch introduces some changes that should be separately tested and filed, e.g. changes in the plugin code.

drrtuy:
I forgot to mention the rationale behind BPP:: methods changes aren't clear to me. According to MTR the codebase doesn't have much degradations however I didn't get how the changes are important to the goal of the patch.

dbcon/joblist/primitivestep.h
davidhall586: There's a shared pointer called fSessionMemLimit that is used in calls to
fRm->getMemory(newSize, fSessionMemLimit)
fSessionMemLimit is optionally set by the .cnf variable columnstore_um_mem_limit. It's passed around as a shared ptr so various parts of the query can access it. It controls a potentially smaller amount of memory that may trigger "UM limit exceeded" for a single session while still maintaining the larger xml setting for overall memory limit.

The reason that it's a shared pointer is that getMemory decrements it and we want all threads to see the decrement and similarly with releaseMemory.

In the case here, I want to ignore fSessionMemLimit, but the getMemory function demands a variable (shared ptr), so I set fJoinMemLimit to the larger um mem limit and and pass that.

dbcon/joblist/resourcemanager.cpp
davidhall586: I agree on both counts. It probably could be improved with a ref and it is outside the scope. However, adding a "&" to the call to improve performance is not much of a pain, so I have done so.

dbcon/joblist/resourcemanager.cpp
davidhall586: Ah. I was intending to remove that.

dbcon/joblist/resourcemanager.cpp
davidhall586: We do need it: we need to add it back regardless of the value of patience.
When we enter the function, we subtract from both totalUmMemLimit and sessionLimit. These may start with different values. If either goes below zero, we return false.

When we return false, we want both values to be unchanged from before the function call, hence the add to add back in amount.

When patience is true, we try again and again by adding amount back, wait a bit and the then subtract amount. If we time out, this loop leaves the values uncorrected for a false return since subtract is the last operation performed. So we need one last add whether patience is true or false to get the values back to their starting points.

dbcon/joblist/resourcemanager.h
davidhall586: I tried various values on my 16GB VM. This appeared a good compromise. It may not be optimal. The thought process is fuzzy.

dbcon/joblist/tuple-bps.cpp
davidhall586: The getMemory calls that use this value are determining if we should continue building a buffer or release the buffer to the next jobstep and start a new buffer. I have no idea what is a good value to set here so I used 25% of totalUMMemory. I would love to hear your thoughts on this.
Note that we are still limited to totalUMMemory. Also, fJoinMemLimit doesn't subtract from sessionLimit.

Before this patch, we didn't count this buffer memory, but rather filled until the loop finished: 99% of the time, this is OK, as the loop, for most JOINs, will be 8192 rows. But in the case of multiple matches, it can grow much larger. Hence the limiter.

dbcon/joblist/tuple-bps.cpp
davidhall586: When we merge the two, I will see if we can reduce the arg count. It is excessive.

dbcon/joblist/tuple-bps.cpp
davidhall586: That removes one of the too many args in that function call. Done.

utils/threadpool/threadpool.cpp
davidhall586: I don't see as it makes much difference ... but done.

dbcon/joblist/tuple-bps.cpp
davidhall586: I move it here to get it inside the loop. In the case of a large result set due to many matches, the output buffers can grow to enormous size. By moving everything to inside of generateJoinResultSet, we can release smaller and more manageable rowgroup sizes to the next jobstep.

However, the way generateJoinResultSet is written, it may leave good data in rgDatav that still needs to be processed after return. That same code also processes returned data that is not part of a join and would never enter generateJoinResultSet.

dbcon/joblist/tuple-bps.cpp
davidhall586: You will notice that the next line calls
    fdec->removeQueue(), which includes shutdownQueue functionality -- it's redundant. I removed the comment code.

dbcon/joblist/tuplehashjoin.cpp
davidhall586: This is questionable and I am of two minds about it. But inside this function is the logic to determine if a disk join should be entered. Since other threads are also using memory, the switch to diskjoin was undeterministic.

The switch to diskjoin is supposed to be determined by the amount of memory required for the small side dataset. By detaching the count of this memory from the overall query memory count, it becomes deterministic.

If you prefer, I'll switch it back to the way it was.

dbcon/joblist/tuplehashjoin.cpp
davidhall586: In this case, we're building a vector of RGData, max 8192 rows each RGData. All we do here is put a throttle on how big that vector can grow based on memory usage.
The full text is 
```
                if (UNLIKELY(!getMemory(l_outputRG.getMaxDataSize())))
                {
                    // Don't let the join results buffer get out of control.
                    RowGroup out(outputRG); // Useless line. Going away.
                    sendResult(outputData);
                    outputData.clear();
                    returnMemory();
                }

```
So if we fail to get memory, sendResult, clear the buffer and return the memory. The next time 'round starts with an empty vector. SendResult() iterates the vector and inserts into DL. No loss of data.

oam/etc/Columnstore.xml
davidhall586: A mistake. For testing purposes only.

utils/common/atomicops.h
davidhall586: Nevertheless, it doesn't hurt to keep the easy stuff compatible. You never know.

utils/joiner/joinpartition.cpp
davidhall586: This is the change for  MCOL-4846 Switch to disk based join causes segv.
The call to "new JoinPartition(*this, false)" attempts to copy the compressor, but in the code's old position, compressor hadn't been set yet. This potentially caused segv. I just moved the code. I can remove the (int).

primitives/primproc/bppsendthread.h
davidhall586: I know you hate mutexes. Not fond of them myself. But I need flow control between BPP and SendThread. Sendthread already has flow control between it and ExeMgr. So it can back up depending on how fast ExeMgr can consume. When it backs up to much, Sendthread queues up until OOM. 
So, in the normal small query case, send thread is not utilized. It doesn't come into play until ExeMgr triggers flowcontrol which happens when it's receive buffer backs up.

Once ExeMgr initiates flow control, all further calls by BPP to sendResponse() goes to the Queue rather than directly to TCP and hence to ExeMgr.

BPPSendThread doesn't use flow control between itself and BPP until it also backs up -- if (sizeTooBig();

So, while we are introducing a mutex, it doesn't affect other queries, and it only affects the current query iff it gets too backed up. Most queries will never notice.

There is a side effect. In the past, BPP flow was restricted in BatchPrimitiveProcessor::operator()()
```
            if (!sendThread->okToProceed())
            {
                freeLargeBuffers();
                return -1; // the reschedule error code
            }

```
okToProceed() basically checked BPPSendThread's queue size. but rather than traditional blocking flow control, it attempted to just reschedule the job. Unfortunately, in the test query, the queue overflowed and OOM occurred before the one block that a jobstep works on was processed.

Anyway, the old way (still there) freed up the BPP instance to continue on another jobstep. But if the new flow control kicks in, the BPP will be blocked and not released to work another job step. I don't believe this is an issue as another BPP instance will be available.

primitives/primproc/batchprimitiveprocessor.cpp
davidhall586: This can get complicated.
Each time Execute() is called, it passed a bloc of up to 8192 rows of a table. Somewhere in there executeTupleJoin() is called. It looks for all the join matches for each row and builds a data set in tSmallSideMatches. This little multi-dimensional array has a first dimension as the number of joins in the query (joinerCount) and the second dimension is a vector of match results for each row up 8192 for the first join and up to sum of matchCount for each of those 8192 for the second join, compounded for the third, etc. Each of these have unlimited number of entries for the final join.. In the test case, in the hundreds of thousands. The result can get extremely large.

After tSmallSideMatches is built, it either sends it back to ExeMgr along with the raw data, or if there's funcexp or aggregation, materializes the rows and performs the operations and sends it ExeMgr. In the later case, there was logic to limit package size, but in the former, there was no limit to package size. I was seeing packets in range of 2-3GB.

So, my changes limit the size of tSmallSideMatches, and thus the size of everything else being done. resultCount reflects the size of tSmallSideMatches. This limiting number has to be smaller than OOM and larger than 8192. I chose 2^10 as a good compromise. The condition says we have enough data. Return what we have so far.

The code in Execute() has been modified to handle this smaller count and loops to call executeTupleJoin() again with the new starting position until all 8192 initial rows are processed.

primitives/primproc/batchprimitiveprocessor.cpp
davidhall586: newStartRid is returned to Execute() to tell where to pick up next time around. If resultCount < {our arbitrary cutoff count}, then return 0 to tell Execute() that we have processed all there is.

The literal must be the same as the one in the above write up. It could be a const if that matters to you. But it is arbitrary and chosen as a "reasonable" data set size.

dbcon/mysql/ha_mcs_dml.cpp
davidhall586: I can't recall why I had to do this. I do know it became an issue.

dbcon/joblist/resourcemanager.h
davidhall586: I added sessionLimit ? so I could pass NULL if I didn't want to pay attention to sessionLimit. 

The problem is that RM is a singleton, but SessionLimit is not. I'm not sure what RM::returnMemory would get you.

dbcon/joblist/distributedenginecomm.cpp
davidhall586: Maybe, maybe not. To create a new function could de-clutter this code. but it would also be a bit klutzy itself. I'm not sure why the newPmCount == 0 code was added. Must of had a good reason.

oam/etc/Columnstore.xml
davidhall586: A mistake. For testing purposes only. Corrected.

dbcon/joblist/distributedenginecomm.cpp
drrtuy: This can be a crash when for some reason there is no more PM to consider.

drrtuy: I personally strongly disagree adding a function to reduce boilerplate code shouldn't be done here. However I can tolerate this.

dbcon/joblist/primitivestep.h
drrtuy: Thank you for the detailed explanation.
JFYI This approach do atomic ops twice that is why I don't like it. First shared_ptr atomically increments its internal counter and secondly RM atomically increments/decrements the value. It would be better to just do it once using std::atomic variable using memory_order::acquire/release modes IMHO.

dbcon/joblist/resourcemanager.cpp
drrtuy: Agree.

dbcon/joblist/resourcemanager.h
drrtuy: In most cases I've seen non-cloud clients uses hardware that has more than 16GB RAM that is why I am skeptical on this value. 
Could you share the reasons you used to arrive at this compromise?

dbcon/joblist/tuple-bps.cpp
drrtuy: I think you will need to reconsider this after I merge Denis'es patch.

dbcon/joblist/tuplehashjoin.cpp
drrtuy: My point is it doesn't look appropriate to have this unused statement as the comment in  production code.

dbcon/mysql/ha_mcs_dml.cpp
drrtuy: We can't let this pass with this patch. It looks unrelated.

dbcon/joblist/primitivestep.h
davidhall586: by changing the getMemory function to (adding '&')
`bool ResourceManager::getMemory(int64_t amount, boost::shared_ptr<int64_t>& sessionLimit, bool patience)
`
It should remove the shared_ptr increment.

dbcon/joblist/resourcemanager.h
davidhall586: I estimate that the average non-cloud columnstore node has 64GB. I've seen from 16GB to 256GB. Cloud can be as low as 4GB (some people are too cheap for their own good). 

However, ExeMgr has a targetRecvQueueSize hardcoded to 50,000,000, so some number greater than this makes sense. Seriously greater doesn't make sense, so I went with 5x. If there are a number of simultaneous queries that return giant result sets, then 0.25 GB each seems reasonable. This is only for the return queue. We still need room for all the processing, and if a single node system, for ExeMgr as well. On small systems, I recommend we use a smaller value. 

I believe a larger value will not improve anything since at this point, we're just filling a queue much faster than it can be emptied. Even if we make this default larger, giant results will still eventually block. Just with less memory available for other processing.

dbcon/joblist/tuple-bps.cpp
davidhall586: We do, because the default for calls to getMemory use a shared_ptr. We can probably do a reset() rather than all that size figuring. Since coding this, I've added the ability to accept a NULL shared_ptr in getMemory().

dbcon/joblist/tuple-bps.cpp
davidhall586: or add
```
if (!amount)
    return;

```
To returnMemory

dbcon/joblist/tuple-bps.cpp
davidhall586: Maybe

dbcon/joblist/tuplehashjoin.cpp
davidhall586: OK, I'll remove it.

dbcon/joblist/tuplehashjoin.cpp
davidhall586: In most cases I would agree with you. But in the case of Cartesian join, or near Cartesian join or even somewhere in between (join on column with lots of non-unique values), joinOneRG can produce any number of output rows. It could be in the hundreds of thousands or millions, as is the test case here. When attempting to not do one RG at a time, just keep this in mind.

davidhall586:
This PR is superseded by a new PR to account for the changes that occurred in develop-6 while this was in progress

PR #2136, opened 2021-10-26
This patch introduces support for scanning/filtering vectorized execu…

drrtuy:
…tion for numeric-based

data types
TEXT, CHAR, VARCHAR, FLOAT and DOUBLE are not yet supported by vectorized path
This patch introduces an example for Google benchmarking suite to measure a perf diff
b/w legacy scan/filtering code and the templated version


mariadb-LeonidFedorov (approved):

mariadb-AlexeyAntipovsky (approved):

drrtuy:
Merging b/c the only MTR failure occasionally happens and wasn't caused by the patch.

PR #2137, opened 2021-10-27
Introduced clang-format

mariadb-LeonidFedorov:
The purpose of this PR is manifestation of Columnstore C++ Codestyle.
After review, I want to force clang-format with this config on whole repo.
   Additionally:
- I want to clarify naming of class members, constants and enums
- replace ifdef header guards with pragma once
- replace NULL with nullptr

This PR is pointed for automatic code procession after discussion


mariadb-SergeyZefirov:
I agree. I see no problems here.

denis0x0D:
+1 nice to have this.

davidhall586 (approved):
I like the idea. This led me down the rabbit hole of studying google style which led me to abseil which got me thinking if maybe some of those hash containers would be more efficient for our join operations. Just musing.

drrtuy:
Good initiative since the styling is for us programmers and not for some abstract good thing.
I agree with some bits regarding the style. We already have two different indentation styles: 2 spaces(most code in datatypes) and 4 spaces(almost everywhere except the datatypes). The 2space style is suggested as a primary one to reduce a writing space wasted.
Another bit is that all files must have a vim trailer.

drrtuy:
@mariadb-LeonidFedorov I suggest we don't change the camel style and we need future discuss the preferable way to mark class members in the code. As of now we have to options, either mClassMember or _ClassMember. The first one is preferable IMHO b/c it brings the semantics whilst the second uses implicit semantics most newcomers don't have.

drrtuy:
@dhall-MariaDB You know we already changed our hash man in the group by code and the only thing that stopped me from changing all other hash maps throughout the code is time. This must be done systematically doing measurements. I am not 100% sure blind change will make processing reasonably better. But this discussion goes outside the scope.

drrtuy (approved):

drrtuy:
@mariadb-SergeyZefirov @denis0x0D Plz give your approval if you are agreed.

mariadb-LeonidFedorov:
@drrtuy my proposal for private members was _member not _Member, and semantics takes place because of arrangements, not by itself. Using trailing underscore is very popular across modern cpp projects, we can follow this style also, instead of leading underscore

mariadb-AlexeyAntipovsky (approved):

mariadb-SergeyZefirov (approved):

.clang-format
```diff
@@ -0,0 +1,22 @@
...
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+BreakConstructorInitializersBeforeComma: true
+ConstructorInitializerIndentWidth: 1
+DerivePointerAlignment: false
+IndentCaseLabels: false
+NamespaceIndentation: None
+PointerAlignment: Left
+SortIncludes: true
+Standard: Cpp11
```
drrtuy: It is c++17 already.

PR #2138, opened 2021-10-27
[MCOL-4849] Parallelize the processing of the bytestream vector.

denis0x0D:
Mostly to check MTR, also need to think about locks inside ExeMgr::RMPM. How we can communicate with thread which sends jobs for BPPs without locsk/cond_vars, the locks a still there.

### DIFF
--- a/dbcon/joblist/primitivestep.h
+++ b/dbcon/joblist/primitivestep.h
@@ -1075,6 +1075,24 @@ public:
     virtual void setFE23Output(const rowgroup::RowGroup& rg) = 0;
 };
 
+struct _CPInfo
+{
+    _CPInfo(int64_t MIN, int64_t MAX, uint64_t l, bool val) : min(MIN), max(MAX), LBID(l), valid(val){};
+    _CPInfo(int128_t BIGMIN, int128_t BIGMAX, uint64_t l, bool val)
+        : bigMin(BIGMIN), bigMax(BIGMAX), LBID(l), valid(val){};
+    union
+    {
+        int128_t bigMin;
+        int64_t min;
+    };
+    union
+    {
+        int128_t bigMax;
+        int64_t max;
+    };
+    uint64_t LBID;
+    bool valid;
+};
 
 /** @brief class TupleBPS
  *
@@ -1113,7 +1131,12 @@ public:
      *
      * The main loop for the receive-side thread.  Don't call it directly.
      */
-    void receiveMultiPrimitiveMessages(uint32_t threadID);
+    void receiveMultiPrimitiveMessages();
+
+    // Processes the vector of `bytestream` starting from `begin` index to the `end` index, non inclusive.
+    void processByteStreamVector(vector<boost::shared_ptr<messageqcpp::ByteStream>>& bsv,
+                                 const uint32_t begin, const uint32_t end, vector<_CPInfo>& cpv,
+                                 RowGroupDL* dlp, const uint32_t threadID);
 
     /** @brief Add a filter when the column is anything but a 4-byte float type.
      *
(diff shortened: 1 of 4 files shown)


dbcon/joblist/tuple-bps.cpp
```diff
@@ -2672,15 +2682,15 @@ void TupleBPS::receiveMultiPrimitiveMessages(uint32_t threadID)
...
             lbidList->UpdateAllPartitionInfo(fColType);
-            cpMutex.unlock();
+            // cpMutex.unlock();
         }
     }
 
     // Bug 3136, let mini stats to be formatted if traceOn.
-    if (lastThread && !didEOF)
-        dlp->endOfInput();
+    // How the comment above is related to the next line of code?
```
drrtuy: Feel free to remove the comment. It has no relation.

dbcon/joblist/tuple-bps.cpp
```diff
@@ -2672,15 +2682,15 @@ void TupleBPS::receiveMultiPrimitiveMessages(uint32_t threadID)
 
         if (ffirstStepType == SCAN && bop == BOP_AND && !cancelled())
         {
-            cpMutex.lock();
+            //cpMutex.lock();
             lbidList->UpdateAllPartitionInfo(fColType);
-            cpMutex.unlock();
+            // cpMutex.unlock();
```
drrtuy: Do we need these commented out lines?

dbcon/joblist/tuple-bps.cpp
```diff
@@ -2620,44 +2633,41 @@ void TupleBPS::receiveMultiPrimitiveMessages(uint32_t threadID)
...
+                          totalBlockedReadCount,             // blocked datalist input
+                          totalBlockedWriteCount,            // blocked datalist output
+                          fMsgBytesIn,                       // incoming msg byte count
+                          fMsgBytesOut);                     // outgoing msg byte count
             fExtendedInfo += toString() + logStr.str();
             formatMiniStats();
         }
 
-        if (lastThread && fOid >= 3000)
+        // What does this branch do? we already checked the table oid for this path?
```
drrtuy: Just an extra(and needless) check :)

dbcon/joblist/tuple-bps.cpp
```diff
@@ -2672,15 +2682,15 @@ void TupleBPS::receiveMultiPrimitiveMessages(uint32_t threadID)
...
-            cpMutex.unlock();
+            // cpMutex.unlock();
         }
     }
 
     // Bug 3136, let mini stats to be formatted if traceOn.
-    if (lastThread && !didEOF)
-        dlp->endOfInput();
+    // How the comment above is related to the next line of code?
+    dlp->endOfInput();
```
drrtuy: We should move this line up the code into 2621 so we notify the higher SQL operators earlier. After 2621 there are only stats related ops.

dbcon/joblist/tuple-bps.cpp
```diff
@@ -2103,25 +2296,6 @@ void TupleBPS::receiveMultiPrimitiveMessages(uint32_t threadID)
                 postStepStartTele(sts);
```
drrtuy, line 2296: It looks like the threadID value is always == 0 in the condition predicate for this block.

dbcon/joblist/tuple-bps.cpp
```diff
@@ -2157,256 +2332,93 @@ void TupleBPS::receiveMultiPrimitiveMessages(uint32_t threadID)
 
             tplLock.unlock();
 
-            for (i = 0; i < size && !cancelled(); i++)
+            vector<vector<_CPInfo>> cpInfos;
+            // Calculate the work sizes.
+            uint32_t currentThreadPoolSize = jobstepThreadPool.getThreadCount();
```
drrtuy: Could you change a logic a bit here. Now the code takes into account a number of threads in the thread pool but we need to take into account a number of threads free in the pool, namely  max(ThreadPool::fMaxThreads - ThreadPool::fIssued (running functors) , 1). 
It would be good to take waitingFunctorsSize - fIssued(waiting functors) into account also b/c there might be cases when ThreadPool::fIssued is small but number of waiting functors is large. But we can just ignore this corner case now.

dbcon/joblist/tuple-bps.cpp
```diff
@@ -2157,256 +2332,93 @@ void TupleBPS::receiveMultiPrimitiveMessages(uint32_t threadID)
...
+            uint32_t numOfThreads = 1;
+            // FIXME: How to choose the right size of thread pool?
```
drrtuy: A complicated thing taking a nature of moderns CPUs into account. There are at least 26 phases in instructions execution pipeline so the number can be as big as 26 * number of system threads available. As of now thread pool size it is 4 * system threads. This 4 is based on a very simple and old description of CPU instructions execution pipeline dated eary 00s.

drrtuy: We will get back to this after EM and PP runtimes converge.

dbcon/joblist/tuple-bps.cpp
```diff
@@ -2157,256 +2332,93 @@ void TupleBPS::receiveMultiPrimitiveMessages(uint32_t threadID)
...
             {
-                cpMutex.lock();
-
-                for (i = 0; i < size; i++)
+                size = cpv.size();
+                if (size > 0 && !cancelled())
                 {
-                    if (fColType.colWidth > 8)
+                    // TODO: Do we really need mutex here?
+                    //cpMutex.lock();
```
drrtuy: It seems we don't so it is better to remove the commented out lines.

dbcon/joblist/tuple-bps.cpp
drrtuy: Or even to 2579.

dbcon/joblist/tuple-bps.cpp
```diff
@@ -2157,256 +2332,93 @@ void TupleBPS::receiveMultiPrimitiveMessages(uint32_t threadID)
...
-                }
 
-                cpMutex.unlock();
+                    // cpMutex.unlock();
+                }
             }
 
-            cpv.clear();
-
             tplLock.lock();
```
drrtuy, line 2420: We are in a single thread here. Do we need mutex locks here and down the text?

dbcon/joblist/tuple-bps.cpp
```diff
@@ -2427,93 +2439,96 @@ void TupleBPS::receiveMultiPrimitiveMessages(uint32_t threadID)
...
-    if (++recvExited == fNumThreads)
+    // Take just the first one.
+    auto data = getJoinLocalDataByIndex(0);
+    // We have only one thread here.
     {
-
         if (doJoin && smallOuterJoiner != -1 && !cancelled())
         {
+            vector<rowgroup::RGData> rgDatav;
             tplLock.unlock();
```
drrtuy: If the lock is really needed(plz take a look at the prev comment) we better move vector ctor outside the crit section.

drrtuy (changes requested):
Thx for the patch. Could you address the comments?

denis0x0D:
@drrtuy thanks for review, fixed.

utils/threadpool/threadpool.h
```diff
@@ -225,6 +225,8 @@ class ThreadPool
         return fMaxThreads;
     }
 
+    inline size_t getIssuiedThreads() { return fIssued; }
```
drrtuy: A typo to fix if you are going to change the code.

dbcon/joblist/primitivestep.h
```diff
@@ -1348,158 +1370,243 @@ class TupleBPS : public BatchPrimitive, public TupleDeliveryStep
...
+  struct JoinLocalData
+  {
+      JoinLocalData() = delete;
+      JoinLocalData(const JoinLocalData&) = delete;
+      JoinLocalData(JoinLocalData&&) = delete;
+      JoinLocalData& operator=(const JoinLocalData&) = delete;
+      JoinLocalData& operator=(JoinLocalData&&) = delete;
+      ~JoinLocalData() = default;
+
+      JoinLocalData(rowgroup::RowGroup primRowGroup, rowgroup::RowGroup outputRowGroup,
```
drrtuy: What do you think if we send RG ref for the two args that come first?

dbcon/joblist/tuple-bps.cpp
```diff
@@ -156,6 +153,109 @@ struct TupleBPSAggregators
...
+    vector<boost::shared_ptr<messageqcpp::ByteStream>>& bsv;
+    uint32_t begin;
+    uint32_t end;
+    vector<_CPInfo>& cpv;
+    RowGroupDL* dlp;
+    uint32_t threadID;
+
+    void operator()()
+    {
+        utils::setThreadName("ByteStreamProcessor");
```
drrtuy, line 254: Could you change the label so that it contains TBPS, e.g. TbpsBsProcessor, so that we know where BS processing takes place?

dbcon/joblist/tuple-bps.cpp
```diff
@@ -1883,196 +1984,282 @@ void TupleBPS::sendPrimitiveMessages()
...
-        int128_t bigMax;
-        int64_t max;
-    };
-    uint64_t LBID;
-    bool valid;
-};
-
-void TupleBPS::receiveMultiPrimitiveMessages(uint32_t threadID)
+void TupleBPS::processByteStreamVector(vector<boost::shared_ptr<messageqcpp::ByteStream>>& bsv,
+                                       uint32_t begin, uint32_t end, vector<_CPInfo>& cpv, RowGroupDL* dlp,
```
drrtuy: I would suggest to make these variables constant if the function semantics allows this.

utils/threadpool/threadpool.h
denis0x0D: Oh, sorry, thanks!

drrtuy (approved):

PR #2139, opened 2021-10-28
MCOL-4910: Turning on C++ 20 for Columnstore

mariadb-LeonidFedorov:
Turning on C++ 20
No diffs in build environment except
Ubuntu 18.04 support removed
- volatile keyword removed
- some warnings suppressed 
- STLAllocator changed for new std::allocator signatures
- boost::Format removed and replaced with dummy handwritten function
- boost::RegEx replaced with std::regex with handmade extention
- boost::property_tree and boost::json_parsed replaced with json header only parser
[nlochman ](https://github.com/nlohmann/json) not fastest, but very useful, simple and good tested
[see benchmark](https://github.com/miloyip/nativejson-benchmark/blob/master/sample/performance_Corei7-4980HQ@2.80GHz_mac64_clang7.0_1._Parse_Memory_(byte).png)
Can be switched to [new boost.json](http://vinniefalco.github.io/doc/json/) 
Or [simjson](https://github.com/simdjson/simdjson) when we need perfomance

PR is passing test-001

PS. There is ```if constexpr``` here in code


utils/loggingcpp/format.h
```diff
@@ -0,0 +1,94 @@
+/*
+   Copyright (C) 2014 InfiniDB, Inc.
```
drrtuy: I doubt that InfiniDB adds something to this file except maybe a header.

utils/loggingcpp/format.h
```diff
@@ -0,0 +1,94 @@
...
+            throw std::logic_error("logggin::format: unexpected type in argslist");
+        }
+        ++iter;
+        ++position;
+    }
+    static std::regex restToken("%[0-9]%");
+    errMsg = std::regex_replace(errMsg, restToken, "");
+}
+
+} // namespace logging
```
drrtuy: Can you follow the style that you suggested?
I would appreciate if you put vim trailer here also.

drrtuy (commented):
Could you ask Daniel to compare the performance of this build with his latest measurements of develop-6 before we merge?

utils/loggingcpp/format.h
mariadb-LeonidFedorov: I have formatted new file like this, others will be better to format once in one big PR

utils/loggingcpp/format.h
mariadb-LeonidFedorov: :)

PR #2149, opened 2021-11-04
MCOL-4769 Fix cache bugs.

tntnatbry:
This patch fixes the following cache bugs:

1. If a table is created with the write cache disabled (i.e. when
columnstore_cache_inserts=OFF), make it accessible when the
cache feature is enabled (columnstore_cache_inserts=ON).

2. Do not replay INSERTs and LDIs on the replica nodes when
the write cache is enabled.

### DIFF
--- a/dbcon/mysql/ha_mcs.cpp
+++ b/dbcon/mysql/ha_mcs.cpp
@@ -1401,7 +1401,7 @@ void ha_mcs_cache_share::close()
 static plugin_ref plugin_maria = NULL;
 
 ha_mcs_cache::ha_mcs_cache(handlerton *hton, TABLE_SHARE *table_arg, MEM_ROOT *mem_root)
-  :ha_mcs(mcs_hton, table_arg), isSysCatTable(false)
+  :ha_mcs(mcs_hton, table_arg), isSysCatTable(false), isCacheDisabled(false)
 {
   if (table_arg && table_arg->db.str &&
       !strcasecmp(table_arg->db.str, "calpontsys") &&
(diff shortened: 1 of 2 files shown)


dbcon/mysql/ha_mcs.cpp
```diff
@@ -1451,7 +1454,7 @@ int ha_mcs_cache::create(const char *name, TABLE *table_arg,
   char cache_name[FN_REFLEN+8];
   DBUG_ENTER("ha_mcs_cache::create");
 
-  if (get_cache_inserts(current_thd) && !isSysCatTable)
+  if (isCacheEnabled())
   {
     create_cache_name(cache_name, name);
```
davidhall586, line 1459: Why open a block here? It seems redundant since it's inside the "if" block anyway.

dbcon/mysql/ha_mcs.cpp
```diff
@@ -1492,46 +1495,52 @@ int ha_mcs_cache::open(const char *name, int mode, uint open_flags)
...
-          lock->check_status=   &cache_check_status;
-        if (lock->restore_status)
-          lock->restore_status=  &cache_restore_status;
-        lock->get_status=       &get_status_and_flush_cache;
+        mysql_mutex_lock(&cache_handler->file->s->intern_lock);
+
+        /* The following lock is here just to establish mutex locking order */
+        mysql_mutex_lock(&lock->mutex);
+        mysql_mutex_unlock(&lock->mutex);
+
```
davidhall586, line 1514: Not sure what this is about. A little more about mutex locking order and why it's important and locking and unlocking this establishes order.

dbcon/mysql/ha_mcs.cpp
```diff
@@ -1492,46 +1495,52 @@ int ha_mcs_cache::open(const char *name, int mode, uint open_flags)
...
+        mysql_mutex_unlock(&cache_handler->file->s->intern_lock);
       }
-      mysql_mutex_unlock(&cache_handler->file->s->intern_lock);
+      cache_handler->file->lock.status_param= (void*) this;
+    }
+    else if (cache_error == ENOENT && !(error= parent::open(name, mode, open_flags)))
+    {
+        isCacheDisabled = true;
+        DBUG_RETURN(0);
     }
```
davidhall586, line 1552: What if cache_error is something other than ENOENT?

dbcon/mysql/ha_mcs.cpp
tntnatbry: Good point, this is fixed.

dbcon/mysql/ha_mcs.cpp
tntnatbry: I am not sure about this, this is Monty's work.

dbcon/mysql/ha_mcs.cpp
tntnatbry: Good point, I added additional if else conditions to cover the case when cache_error is not ENOENT.

davidhall586 (approved):

PR #2150, opened 2021-11-04
mcs50_cpimport_stdin results are not deterministic

davidhall586:
The query "SELECT * FROM t1 ORDER BY col1" does not return a deterministic order. col1 does not contain unique values, so ordering by it alone does not result in the same order consistently.
Added col2 to the ORDER BY to get a unique value.


drrtuy:
I would suggest to use MTR's --sorted_result option like [here.](https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/2107/files)

PR #2155, opened 2021-11-16
Mcol 4841 dev6 Handle large joins without OOM

davidhall586:
I moved generateJoinResultSet() and processFE2() into class JoinLocalData. This reduces the calling signature significantly since most of the items it uses for computation exist there.


dbcon/joblist/diskjoinstep.cpp
```diff
@@ -484,6 +482,21 @@ void DiskJoinStep::joinFcn()
                         {
                             outputDL->insert(rgData);
                             //cout << "inserting a full RG" << endl;
+                            if (thjs)
+                            {
+                                if (!thjs->getMemory(l_outputRG.getMaxDataSize()))
+                                {
+                                    // calculate guess of size required for error message
+                                    uint64_t memReqd = (unmatched.size() * outputRG.getDataSize(1)) / 1048576;;
```
mariadb-DenisKhalikov: Typo: extra `;`

dbcon/joblist/tuple-bps.cpp
```diff
@@ -2383,10 +2486,7 @@ void TupleBPS::receiveMultiPrimitiveMessages()
                 start = end;
             }
 
-            // Join threads.
-            for (uint32_t i = 0, e = fProcessorThreads.size(); i < e; ++i)
```
mariadb-DenisKhalikov, line 2386: I see, thanks for the fix =)

dbcon/joblist/tuple-bps.cpp
```diff
@@ -2495,9 +2593,8 @@ void TupleBPS::receiveMultiPrimitiveMessages()
                     if (fe2)
                     {
                         rgDatav.push_back(data->joinedData);
-                        processFE2(data->local_outputRG, data->local_fe2Output, data->postJoinRow,
-                                   data->local_fe2OutRow, &rgDatav, &data->local_fe2);
-
+                        data->processFE2(rgDatav);
```
mariadb-DenisKhalikov, line 2603: This looks very nice comparing to what it was. Thanks.

primitives/primproc/batchprimitiveprocessor.cpp
```diff
@@ -1384,8 +1392,18 @@ void BatchPrimitiveProcessor::executeTupleJoin()
             //else
             // cout << "j != joinerCount\n";
         }
+        // If we've accumulated more than 1048576 (2^20)_ of resultCounts, cut off processing. The caller will
+        // restart to continue where we left off.
+        if (resultCount >= 1048576)
```
mariadb-DenisKhalikov: Nit: It would be nice to have a named constant for this constant 1048576. And in other places in this code.

mariadb-DenisKhalikov (commented):
@dhall-MariaDB thanks for the patch!
The changes for the `TupleBPS` and `tuplehashjoin` LGTM, but for other parts like priority thread pool - I don't have enough understanding for all corner cases, I thinks @drrtuy should take a look on that parts.
Thanks!

dbcon/mysql/ha_mcs_impl.cpp
```diff
@@ -3086,6 +3086,10 @@ void ha_mcs_impl_start_bulk_insert(ha_rows rows, TABLE* table, bool is_cache_ins
             //@bug 6122 Check how many columns have not null constraint. columnn with not null constraint will not show up in header.
             unsigned int numberNotNull = 0;
 
+            // There's a chance that a previously aborted query left garbage in ci->columnTypes
```
drrtuy: IMHO This change is outside the scope of the patch and musn't get into this patch. Moreover the clean-up made in the middle of an operation looks like a side-effect nobody knows about so we better put it where all clean-up happens, e.g. impl_external_lock().

oam/etc/Columnstore.xml
```diff
@@ -96,6 +96,7 @@
 		<!-- <MediumPriorityPercentage>30</MediumPriorityPercentage> -->
 		<!-- <LowPriorityPercentage>10</LowPriorityPercentage> -->
 		<DirectIO>y</DirectIO>
+        <MaxBPPSendQueue>250M</MaxBPPSendQueue>
```
drrtuy: As you know we can't deliver Columnstore.xml changes until we deliver the MCOL-4928.

primitives/primproc/bppsendthread.h
```diff
@@ -124,7 +141,10 @@ class BPPSendThread
 
     /* secondary queue size restriction based on byte size */
     volatile uint64_t currentByteSize;
-    uint64_t maxByteSize;
+    static uint64_t maxByteSize;
+    // Used to tell the PriorityThreadPool It should consider additional threads because a 
+    // queue full event has happened and a thread has been blocked.
+    boost::shared_ptr<threadpool::PriorityThreadPool> fProcessorPool;
```
drrtuy: Could you elaborate on why do we need smart pointer here?

primitives/primproc/bppsendthread.h
```diff
@@ -124,7 +141,10 @@ class BPPSendThread
 
     /* secondary queue size restriction based on byte size */
     volatile uint64_t currentByteSize;
-    uint64_t maxByteSize;
+    static uint64_t maxByteSize;
```
drrtuy: Sorry but static member looks a bit hack-ish here.
AFAIK the only place where PP creates BPPSendThread is BPPV ctor(primproc/primitiveserver.cpp) and it is possible to ship the value inside PrimitiveServer which ptr is available in the BPPV ctor.

primitives/primproc/bppsendthread.h
```diff
@@ -63,7 +63,15 @@ class BPPSendThread
             msg(m), sock(so), sockLock(sl), sockIndex(si) { }
     };
 
-    bool okToProceed();
+    bool sizeTooBig()
+    {
+      // keep the queue size below the 100 msg threshold & below the 250MB mark,
+      // but at least 2 msgs so there is always 1 ready to be sent.
```
drrtuy: at least 3 messages according with the condition below.

primitives/primproc/bppsendthread.cpp
```diff
@@ -53,29 +51,38 @@ BPPSendThread::~BPPSendThread()
...
-            || msgQueue.size() < 3) && !die;
-}
-
 void BPPSendThread::sendResult(const Msg_t& msg, bool newConnection)
 {
+    // Wait for the queue to empty out a bit if it's stuffed full
+    if (sizeTooBig())
+    {
+        boost::mutex::scoped_lock sl1(respondLock);
+        while (currentByteSize >= maxByteSize && msgQueue.size() > 3 && !die)
```
drrtuy, line 62: We have this condition in multiple places. Changing one of them it is easy to forget the others.

primitives/primproc/bppsendthread.cpp
```diff
@@ -53,29 +51,38 @@ BPPSendThread::~BPPSendThread()
...
-    return ((msgQueue.size() < sizeThreshold && currentByteSize < maxByteSize)
-            || msgQueue.size() < 3) && !die;
-}
-
 void BPPSendThread::sendResult(const Msg_t& msg, bool newConnection)
 {
+    // Wait for the queue to empty out a bit if it's stuffed full
+    if (sizeTooBig())
+    {
+        boost::mutex::scoped_lock sl1(respondLock);
```
drrtuy: I personally don't like using boost where std:: elements can be used.

primitives/primproc/bppsendthread.cpp
```diff
@@ -53,29 +51,38 @@ BPPSendThread::~BPPSendThread()
...
+        while (currentByteSize >= maxByteSize && msgQueue.size() > 3 && !die)
+        {
+            respondWait = true;
+            fProcessorPool->incBlockedThreads();
+            okToRespond.wait(sl1);
+            fProcessorPool->decBlockedThreads();
+            respondWait = false;
+        }
+        sl1.unlock();
```
drrtuy: This call doesn't look necessary b/c scoped_lock dtor releases the mutex exiting this scope.

utils/threadpool/prioritythreadpool.cpp
```diff
@@ -158,6 +179,14 @@ void PriorityThreadPool::threadFcn(const Priority preferredQueue) throw()
 
             if (jobQueues[queue].empty())
             {
+                // If this is an EXTRA thread due toother threads blocking, and all blockers are unblocked, 
+                // we don't want this one any more.
+                if (preferredQueue == EXTRA && stopExtra)
+                {
+                    extraThreads--;
```
drrtuy, line 186: I am not clear on stopExtra semantics. It looks like this flag signals when there are no blocking threads. If so I can't get why don't we set extraThreads to zero instead of decrementing it?

primitives/primproc/bppsendthread.cpp
drrtuy: Moreover we have the same scope in the parenthesis in two BPPSendThread methods so it is better to wrap it up in a function.

primitives/primproc/batchprimitiveprocessor.cpp
drrtuy: +1

dbcon/joblist/primitivestep.h
```diff
@@ -1515,21 +1512,31 @@ class TupleBPS : public BatchPrimitive, public TupleDeliveryStep
...
       ~JoinLocalData() = default;
 
-      JoinLocalData(rowgroup::RowGroup& primRowGroup, rowgroup::RowGroup& outputRowGroup,
+      JoinLocalData(TupleBPS* pTupleBPS, rowgroup::RowGroup& primRowGroup, rowgroup::RowGroup& outputRowGroup,
                     boost::shared_ptr<funcexp::FuncExpWrapper>& fe2, rowgroup::RowGroup& fe2Output,
                     std::vector<rowgroup::RowGroup>& joinerMatchesRGs, rowgroup::RowGroup& joinFERG,
                     std::vector<boost::shared_ptr<joiner::TupleJoiner>>& tjoiners, uint32_t smallSideCount,
                     bool doJoin);
 
+      friend class TupleBPS;
```
drrtuy, line 1531: If you want to make it proper you need to use getter methods for JoinLocalData attributes are accessed in the code. Friend used here gives me an impression that original idea was partially implemented.

dbcon/joblist/primitivestep.h
```diff
@@ -1366,7 +1366,15 @@ class TupleBPS : public BatchPrimitive, public TupleDeliveryStep
     {
         return true;
     }
-
+    ResourceManager* resourceManager() const
+    {
+      return fRm;
+    }
+    bool runFEonPM()
```
drrtuy: Can't we use const specifier for this method?

dbcon/joblist/primitivestep.h
```diff
@@ -1564,7 +1571,9 @@ class TupleBPS : public BatchPrimitive, public TupleDeliveryStep
       boost::scoped_array<uint8_t> largeNullMemory;
       boost::scoped_array<boost::shared_array<uint8_t>> smallNullMemory;
       uint32_t matchCount;
-
+      uint64_t memUsedByOutputRG = 0;
```
drrtuy: Why don't we assign this member in the ctor?

dbcon/joblist/distributedenginecomm.cpp
```diff
@@ -308,13 +308,24 @@ void DistributedEngineComm::Setup()
 
             writeToLog(__FILE__, __LINE__, "Could not connect to PMS" + std::to_string(connectionId) + ": " + ex.what(), LOG_TYPE_ERROR);
             cerr << "Could not connect to PMS" << std::to_string(connectionId) << ": " << ex.what() << endl;
+
```
drrtuy: We have almost identical exceptions handlers and I suggest you to put the common code into a function parametrizing the messages if needed.

dbcon/joblist/resourcemanager.h
```diff
@@ -82,6 +82,7 @@ const uint32_t defaultRequestSize  = 1;
 const uint32_t defaultMaxOutstandingRequests = 20;
 const uint32_t defaultProcessorThreadsPerScan = 16;
 const uint32_t defaultJoinerChunkSize = 16 * 1024 * 1024;
+const uint64_t defaultMaxBPPSendQueue = 250000000; // 250MB
```
drrtuy: Plz put your considerations on this value that you once shared with me in a comment here.

utils/common/atomicops.h
```diff
@@ -165,6 +165,28 @@ inline bool atomicCAS(volatile T* mem, T comp, T swap)
 #endif
 }
 
+// implements a zero out of a variable
+template <typename T>
+inline void atomicZero(volatile T* mem)
+{
+#ifdef _MSC_VER
```
drrtuy, line 172: I am still sure we won't support windows anymore so it is better to avoid adding new preprocessor Windows related conditions.

primitives/primproc/bppsendthread.h
drrtuy: There can be only one processor pool so it must be safe to use plain PriorityThreadPool pointer.

primitives/primproc/bppsendthread.cpp
drrtuy: Can you replace boost::mutex with std::mutex?

oam/etc/Columnstore.xml
davidhall586: This is the default, so we can leave it out of the xml for now.

primitives/primproc/bppsendthread.cpp
davidhall586: NP

primitives/primproc/bppsendthread.cpp
davidhall586: It is redundant. removed.

utils/threadpool/prioritythreadpool.cpp
davidhall586: extraThreads counts how many extra threads are currently executing. When stopExtra is set, it effectively tells extra threads to go away when finished with whatever they're working on. When a thread goes away, it decrements extraThreads to reflect the current count of extra threads.

If there are blocked threads and extra threads, then this keeps a cap on how many extra threads will be created -- never more than the number of blocked threads.

However, as written, the number of extra threads may grow (if needed) as the number of blocked threads grows. The number of extra threads will not reduce until the number of blocked threads reaches 0 (thus setting stopExtra = true). I consider this to be OK, as blocked threads are a rarity. But if they do happen, they probably happen in bunches.

We may think to set extraThreads to zero in addJob() at the same time we set stopExtra to true. However, there will still be extra threads around until they finish whatever they were working on. During this period, extraThreads would be inaccurate. Meantime If threads block again, there could be confusion. stopExtra will be set to false and new extra threads will be created. There would be more extra threads in existence than extraThreads reflects. This would allow for more extraThreads than blocked threads. While not a killer, it's not what is intended.

dbcon/joblist/primitivestep.h
davidhall586: Absolutely true. Friend here is to alleviate having to implement a pile of setter/getters. JoinLocalData was originally a struct as written by Denis and all members were public. I changed it to a class making all the member variables private and added friend so I didn't have to go about the code to change all access to setter/getter. An option is to leave all members public as they were before.

dbcon/mysql/ha_mcs_impl.cpp
davidhall586: I disagree. This is something that belongs here. There is a clear() at the tail of the function, but errors may cause early exit, bypassing the clear(). It is common to initialize variables before use. In this case, it's a good safe way to do it.
As for scope, that can be debated, but uninitialized ci->columnTypes caused a bit of pain while testing this patch. Thus this change. To just remove it as not in the scope of the problem will cause unnecessary headache. We'd end up putting it back in at a later time. Or perhaps find every branch that exits the function and add cleanup code? Nah.

primitives/primproc/bppsendthread.h
davidhall586: It's a shared pointer in PrimitiveServer is why it became a shared pointer in BPPSendThread. Since PrimitiveServer is a singleton whose life, once created, lasts until process termination, and fProcessorPool, is a singleton within it, I don't see why it needs to be a shared pointer.

primitives/primproc/bppsendthread.h
davidhall586: Yes indeed we could do that. My intent was to not have to deal with (not spend cpu cycles on) that. It's not horribly expensive, but I am trying to shave where I can.
The drawback of my approach is that a restart is definitely required to notice a change to the xml setting.

primitives/primproc/bppsendthread.cpp
davidhall586: It could be even harder to read if all this is in a seperate function, though I could do that. However, after a code search. the function sendResults() appears to no longer be used anywhere. I don't know if we would ever want to resurrect it. Perhaps if we just delete the unreferenced function?

primitives/primproc/bppsendthread.h
davidhall586: changed the comment

primitives/primproc/bppsendthread.cpp
davidhall586: As for the two remaining places where sizeTooBig is called, in sendResult() this is blocking the thread since it's in the middle of a block. In BatchPrimitiveProcessor::operator(), a positive too big will result in the BPP being returned to the pool and the job rescheduled.

dbcon/joblist/primitivestep.h
davidhall586: const added

dbcon/joblist/primitivestep.h
davidhall586: An argument can be made, but it's moot, since the variable is no longer used. I removed it.

utils/common/atomicops.h
davidhall586: I added this as following the pattern of all the other functions in the lib. Don't assume never. Never tends to byte programmers in the butt.

dbcon/joblist/resourcemanager.h
davidhall586: Done

dbcon/joblist/distributedenginecomm.cpp
davidhall586: I thought about this a bit and even started writing the function, but it turns out to not save a lot. Sure I could pass in a ref to newPmCount and the ex.what(). This could standardize the log messages in one place, but I'd still have to test newPmCount ==0 for the break after return. Seems a lot of cpu for very little savings in line count.

primitives/primproc/batchprimitiveprocessor.cpp
davidhall586: Done

dbcon/mysql/ha_mcs_impl.cpp
drrtuy: No query passes impl_external_lock() and adding uncorrelated side effects makes it hard to debug/maintain the code IMHO. That is why I am strongly against putting it in the middle of  ha_mcs_impl_start_bulk_insert

primitives/primproc/bppsendthread.h
drrtuy: I will take another look at this to convince myself static fits here.

dbcon/joblist/primitivestep.h
drrtuy: Agree with you on that. The most appropriate way is to add setters/getters but it is up to you whether to stick with this friend approach or to add methods.

utils/common/atomicops.h
drrtuy: Hm, all others are removing these blocks but I won't insist on removing this if you don't want.

primitives/primproc/bppsendthread.cpp
drrtuy: I think it is a good idea to comment out/remove unreferenced function.

dbcon/mysql/ha_mcs_impl.cpp
davidhall586: I removed the offensive code for now and opened MCOL-4968 to address the issue.

primitives/primproc/bppsendthread.h
davidhall586: I don't suffer from a prejudice against static members:)

primitives/primproc/bppsendthread.cpp
davidhall586: I created MCOL-4969 to remove SendResults.

primitives/primproc/bppsendthread.cpp
```diff
@@ -254,19 +265,26 @@ void BPPSendThread::mainLoop()
...
 {
-    boost::mutex::scoped_lock sl(msgQueueLock);
-    boost::mutex::scoped_lock sl2(ackLock);
-    die = true;
-    queueNotEmpty.notify_one();
-    okToSend.notify_one();
-    sl.unlock();
-    sl2.unlock();
+    {
+        std::lock_guard<std::mutex> sl(msgQueueLock);
```
drrtuy: I was curios about the lifetime of these lock_guards and it seems their dtors are called exiting the scope at line 284 not in the end of the function. [Here](http://cpp.sh/2xqv3) is a fiddler to see and run the code. Is this an expected behavior?

primitives/primproc/bppsendthread.h
drrtuy: I do b/c it represents a state variable that lives outside the scope of the code that uses it. Nobody guarantees anything about such states. It is hard to maintain, hard to follow, hard to debug.

primitives/primproc/bppsendthread.cpp
davidhall586: This is the expected behavior. The lock is only taken if SizeTooBig -- a rare case. It's there to protect the condition variable. Once we're done with the condition, then we don't need or want the lock.

primitives/primproc/bppsendthread.h
davidhall586: So the options are to create a unscoped global, or to spend cpu to unnecessarily re-initialize in the constructor or to call joblist::ResourceManager::instance()->getMaxBPPSendQueue() each time we need the value.

davidhall586: I find it easier to maintain and follow static than global variables. At least as a static, it's usage is clearly (without an accesor) for use by class members.

davidhall586: I could add const, if it helps.
If I just remove it from the class declaration, it still works as a global. The reason I put it in the class is for documentation (to make it clear it's use is for and by the class.
Not sure why you find this confusing.

drrtuy: I don't see any reason against moving config access from getMaxBPPSendQueue() code into RM ctor, so that RM::getMaxBPPSendQueue() just reads the value from RM singleton. Finally you can reference RM::getMaxBPPSendQueue() in the BPPSendThread ctor code in bppsendthread.cpp. What do you think about this approach?

davidhall586: Changed to your recommendation, though I still feel (even with the change to ResourceManager) that a static makes more sense. But for the sake of moving forward, I put it in the sendthread ctor.

drrtuy (approved):

PR #2169, opened 2021-12-07
MCOL-4874 Crossengine JOIN involving a ColumnStore table and a

tntnatbry:
wide decimal column in a non-ColumnStore table throws an exception.

ROW::getSignedNullValue() method does not support wide decimal fields
yet. To fix this exception, we remove the call to this method from
CrossEngineStep::setField().


dbcon/joblist/crossenginestep.cpp
```diff
@@ -359,27 +368,26 @@ int64_t CrossEngineStep::convertValueNum(
...
-                rv = boost::any_cast<char>(anyVal);
-            else if (ct.colWidth == CalpontSystemCatalog::TWO_BYTE)
-                rv = boost::any_cast<int16_t>(anyVal);
-            else if (ct.colWidth == CalpontSystemCatalog::FOUR_BYTE)
+            if (LIKELY(ct.colWidth == datatypes::MAXDECIMALWIDTH))
+                rv = boost::any_cast<int128_t>(anyVal);
+            else if (ct.colWidth == execplan::CalpontSystemCatalog::EIGHT_BYTE)
+                rv = boost::any_cast<long long>(anyVal);
+            else if (ct.colWidth == execplan::CalpontSystemCatalog::FOUR_BYTE)
 #ifdef _MSC_VER
```
drrtuy, line 374: Feel free to remove windows specific macros ifs.

dbcon/joblist/crossenginestep.cpp
```diff
@@ -229,37 +240,31 @@ inline void CrossEngineStep::addRow(RGData& data)
...
 
-    // non-null value
     switch (ct.colDataType)
     {
         case CalpontSystemCatalog::BIT:
             rv = boost::any_cast<bool>(anyVal);
             break;
 
         case CalpontSystemCatalog::TINYINT:
+#if BOOST_VERSION >= 105200
```
drrtuy, line 261: What is the boost version that satisfies the condition?

drrtuy: Will we ever see it in the wild?

drrtuy (changes requested):
Plz address the comments and convertValueNum() return type.

tntnatbry:
convertValueNum() return type has been updated. Please review. Regarding the if defs for Boost and Windows, I am reusing the code in dbcon/joblist/jlf_execplantojoblist.cpp::convertValueNum, these 2 functions should be consolidated to avoid code duplication.

drrtuy (approved):

PR #2192, opened 2021-12-15
MCOL-4868 UPDATE on a ColumnStore table containing an IN-subquery

tntnatbry:
on a non-ColumnStore table does not work.

As part of MCOL-4617, we moved the in-to-exists predicate creation
and injection from the server into the engine. However, when query
with an IN Subquery contains a non-ColumnStore table, the server
still performs the in-to-exists predicate transformation for the
foreign engine table. This caused ColumnStore's execution plan to
contain incorrect WHERE predicates. As a fix, we call
mutate_optimizer_flags() for the WRITE lock, in addition to the READ
table lock. And in mutate_optimizer_flags(), we change the optimizer
flag from OPTIMIZER_SWITCH_IN_TO_EXISTS to OPTIMIZER_SWITCH_MATERIALIZATION.

### DIFF
--- a/dbcon/mysql/ha_mcs_execplan.cpp
+++ b/dbcon/mysql/ha_mcs_execplan.cpp
@@ -6546,7 +6546,7 @@ bool isForeignTableUpdate(THD* thd)
 {
     LEX* lex = thd->lex;
 
-    if (lex->sql_command != SQLCOM_UPDATE_MULTI)
+    if (!isUpdateStatement(lex->sql_command))
         return false;
 
     Item_field* item;
@@ -6561,6 +6561,40 @@ bool isForeignTableUpdate(THD* thd)
     return false;
 }
 
+bool isMCSTableUpdate(THD* thd)
+{
+    LEX* lex = thd->lex;
+
+    if (!isUpdateStatement(lex->sql_command))
+        return false;
+
+    Item_field* item;
+    List_iterator_fast<Item> field_it(lex->first_select_lex()->item_list);
+
+    while ((item = (Item_field*) field_it++))
+    {
+        if (item->field && item->field->table && isMCSTable(item->field->table))
+            return true;
+    }
+
+    return false;
+}
+
+bool isMCSTableDelete(THD* thd)
+{
+    LEX* lex = thd->lex;
+
+    if (!isDeleteStatement(lex->sql_command))
+        return false;
+
+    TABLE_LIST* table_ptr = lex->first_select_lex()->get_table_list();
+
+    if (table_ptr && table_ptr->table && isMCSTable(table_ptr->table))
+        return true;
+
+    return false;
+}
+
 // This function is different from isForeignTableUpdate()
 // above as it only checks if any of the tables involved
 // in the multi-table update statement is a foreign table,
@@ -6570,7 +6604,7 @@ bool isUpdateHasForeignTable(THD* thd)
 {
     LEX* lex = thd->lex;
 
-    if (lex->sql_command != SQLCOM_UPDATE_MULTI)
+    if (!isUpdateStatement(lex->sql_command))
         return false;
 
     TABLE_LIST* table_ptr = lex->first_select_lex()->get_table_list();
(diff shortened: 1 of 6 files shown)


drrtuy:
Is it ready for review @tntnatbry ?

tntnatbry:
@drrtuy PR is now ready for review.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -6561,6 +6562,42 @@ bool isForeignTableUpdate(THD* thd)
     return false;
 }
 
+bool isMCSTableUpdate(THD* thd)
+{
+    LEX* lex = thd->lex;
+
+    if ((lex->sql_command != SQLCOM_UPDATE_MULTI) &&
```
drrtuy: What if we use isUpdateStatement() and isDeleteStatement here and in  isMCSTableDelete()?

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -6561,6 +6562,42 @@ bool isForeignTableUpdate(THD* thd)
...
+        (lex->sql_command != SQLCOM_UPDATE))
+        return false;
+
+    Item_field* item;
+    List_iterator_fast<Item> field_it(lex->first_select_lex()->item_list);
```
drrtuy, line 6572: It is worth to at least put an assert before this ctor, namely assert(lex->first_select_lex()->item_list);

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -6561,6 +6562,42 @@ bool isForeignTableUpdate(THD* thd)
...
+
+bool isMCSTableDelete(THD* thd)
+{
+    LEX* lex = thd->lex;
+
+    if ((lex->sql_command != SQLCOM_DELETE_MULTI) &&
+        (lex->sql_command != SQLCOM_DELETE))
+        return false;
+
+    TABLE_LIST* table_ptr = lex->first_select_lex()->get_table_list();
```
drrtuy, line 6590: Same things as in isMCSTableUpdate are relevant here.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -6570,7 +6607,8 @@ bool isUpdateHasForeignTable(THD* thd)
 {
     LEX* lex = thd->lex;
 
-    if (lex->sql_command != SQLCOM_UPDATE_MULTI)
+    if ((lex->sql_command != SQLCOM_UPDATE_MULTI) &&
```
drrtuy: What if we use isUpdateStatement() here?

drrtuy (changes requested):
Looks good to me except minor changes I commented about. Plz create a copy for develop branch.

drrtuy:
It is worth to take a look at isForeignTableUpdate() used in dbcon/mysql/ha_mcs_pushdown.cpp. Maybe we should replace isForeignTableUpdate with isMCSTableUpdate and drop isForeignTableUpdate. What do you think @tntnatbry ?

dbcon/mysql/ha_mcs_execplan.cpp
tntnatbry: `lex->first_select_lex()->item_list` is a `List<Item>` type. I don't think putting an assert on this makes much sense.

dbcon/mysql/ha_mcs_execplan.cpp

dbcon/mysql/ha_mcs_execplan.cpp

tntnatbry:
@drrtuy `isForeignTableUpdate()` is not the same as `!isMCSTableUpdate()`, so we need to keep this function as is.

tntnatbry:
I have addressed some of the comments/added follow-up comments above, please review.

drrtuy:
These two are different but their semantics are very close and isForeignTableUpdate is used only in create_SH() to detect foreign table update IMHO it is safe to replace isForeignTableUpdate with isMCSTableUpdate b/c there can be no single update for non-mcs table in create_SH().

drrtuy (approved):

drrtuy (approved):

PR #2193, opened 2021-12-15
This patch enables loop auto-vectorization

drrtuy:


mariadb-LeonidFedorov:
As I suppose, option works with clang only.
This template could be useful
```
if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
    ..... do clang options
elseif (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
   ... do gcc options
endif()
```

drrtuy:
It works for gcc as well.

PR #2194, opened 2021-12-15
[MCOL-4927] Fix DDLProc connection processing

mariadb-AlexeyAntipovsky:


ddlproc/ddlprocessor.cpp
```diff
@@ -725,23 +414,393 @@ void DDLProcessor::process()
...
-                handler.fDbrm = &dbrm;
-                handler.fQtc = fQtc;
-                handler.fOamCache = oam::OamCache::makeOamCache();
-                fDdlPackagepool.invoke(handler);
+                processStatement();
+            }
+        }
+        catch (const std::exception &ex)
+        {
+            fIos.close();
```
drrtuy: Shouldn't we re-throw so that the upper level catch was able to try to send a result. It might be dumb but at least we don't change the behavior of DMLProc towards external entities.

drrtuy (approved):

PR #2197, opened 2021-12-17
Mcol 4871 dev6

drrtuy:


abarkov (approved):
Hi Roman. The patch looks  OK for me. Thanks for adding templates.

PR #2208, opened 2021-12-30
MCOL-4899 MCS now applies a correct collation running IN for characte…

drrtuy:
…r data types


mysql-test/columnstore/bugfixes/mcol-4899.result
```diff
@@ -0,0 +1,46 @@
...
+a
+A
+SELECT * FROM collation_test WHERE LETTERS IN ('a','B');
+LETTERS
+a
+A
+b
+B
+DROP DATABASE `mcol_4899`;
+SET character_set_connection=@cs_conn;
```
tntnatbry, line 65: Original test case in the issue description has the CREATE TABLE statement without default values for CHARSET and COLLATE:
```sql
CREATE TABLE collation_test (LETTERS TEXT NULL) ENGINE=ColumnStore;
```
Test cases should also be added for this.

tntnatbry (changes requested):

tntnatbry (approved):

PR #2212, opened 2022-01-04
MCOL-4936 Disable binlog for DML statements.

tntnatbry:
DML statements executed on the primary node in a ColumnStore
cluster do not need to be written to the primary's binlog. This
is due to ColumnStore's distributed storage architecture.

With this patch, we disable writing to binlog when a DML statement
(INSERT/DELETE/UPDATE/LDI/INSERT..SELECT) is performed on a ColumnStore
table. HANDLER::external_lock() calls are used to
  1. Turn OFF the OPTION_BIN_LOG flag
  2. Turn ON the OPTION_BIN_TMP_LOG_OFF flag
in THD::variables.option_bits during a WRITE lock call.

THD::variables.option_bits is restored back to the original state
during the UNLOCK call in HANDLER::external_lock().

Further, isDMLStatement() function is added to reduce code verbosity
to check if a given statement is a DML statement.

Note that with this patch, not writing to primary's binlog means
DML replication from a ColumnStore cluster to another ColumnStore
cluster or to another foreign engine will not work.

### DIFF
--- a/dbcon/mysql/ha_mcs_impl.cpp
+++ b/dbcon/mysql/ha_mcs_impl.cpp
@@ -2152,15 +2152,8 @@ int ha_mcs_impl_direct_update_delete_rows(bool execute, ha_rows *affected_rows, 
     gwi.thd = thd;
     int rc = 0;
 
-    if (thd->slave_thread && !get_replication_slave(thd) && (
-                thd->lex->sql_command == SQLCOM_INSERT ||
-                thd->lex->sql_command == SQLCOM_INSERT_SELECT ||
-                thd->lex->sql_command == SQLCOM_UPDATE ||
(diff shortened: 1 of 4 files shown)


dbcon/mysql/ha_mcs_impl.cpp
```diff
@@ -3942,6 +3914,31 @@ COND* ha_mcs_impl_cond_push(COND* cond, TABLE* table, std::vector<COND*>& condSt
     return cond;
 }
 
+inline void disableBinlogForDML(THD* thd)
+{
+    if (isDMLStatement(thd->lex->sql_command) &&
```
drrtuy, line 3919: I would put assert(thd->lex) but this is optional.

drrtuy (approved):

PR #2216, opened 2022-01-17
CRC32 call upon Columnstore table returns a wrong value

dr-m:
[MDEV-27208](https://jira.mariadb.org/browse/MDEV-27208) extends the built-in `CRC32()` function by introducing a 2-ary variant that allows the piecewise calculation of an ISO 3309 checksum.

This fix aims to address [MDEV-27519](https://jira.mariadb.org/browse/MDEV-27519): ColumnStore would accept the 2-ary variant of the function, but return incompatible results.

Note: I did not test this fix, and I do not think that a normal CI build of MariaDB/server@f86e1fd81c4ff4e0253b3843a041efe6fafc7b47 would run any ColumnStore tests.

I hope that you can create a test case for this, also covering the original unary variant with an input buffer that includes NUL bytes.


dr-m:
I see that this failed the ColumnStore CI check due to the following:
```
git merge 40b495b05ae79aeda44635611b7c62ed5dae3e69
fatal: refusing to merge unrelated histories
```
It also [failed to build|http://buildbot.askmonty.org/buildbot/builders/kvm-tarbake-jaunty-x86/builds/45214/steps/compile/logs/stdio] on the old MariaDB Server CI system because the submitted branch is in a different repository:
```
Unable to checkout '40b495b05ae79aeda44635611b7c62ed5dae3e69' in submodule path 'storage/columnstore/columnstore'
```
Based on the change history, I understood that ColumnStore changes are supposed to be done in personal repositories. I hope that @drrtuy or someone can resolve this as appropriate.

dr-m:
@drrtuy posted a rebased version of this in #2219.

PR #2225, opened 2022-01-21
MCOL-4912 This patch introduces Extent Map index to speed up operatio…

drrtuy:
…ns that

translates <dbroot, oid, partition> tuple into LBID, e.g. most bulk insertion meta
info operations


dbcon/mysql/ha_mcs_client_udfs.cpp
```diff
@@ -875,5 +877,67 @@ extern "C"
...
+    }
+
+#ifdef _MSC_VER
+    __declspec(dllexport)
+#endif
+    my_bool mcs_emindex_size_init(UDF_INIT* initid, UDF_ARGS* args, char* message)
+    {
+        if (args->arg_count != 0)
+        {
+            strcpy(message, "mcs_emindex_size() takes no arguments");
```
mariadb-AlexeyAntipovsky: `MsgEMIndexSizeInitErrMsg` ?

versioning/BRM/brmshmimpl.cpp
```diff
@@ -232,7 +240,163 @@ void BRMShmImpl::destroy()
...
+
+    for(uint32_t tries = 0; fSize == 0 && tries <= ShmCreateMaxRetries; ++tries)
+    {
+        try
+        {
+            auto* shmSegment = new boost::interprocess::managed_shared_memory(bi::open_only, keyName.c_str());
+            curSize = shmSegment->get_size();
+
+            if (curSize == 0)
+                throw bi::interprocess_exception("shared memory segment size is 0.");
```
mariadb-AlexeyAntipovsky, line 259: Looks like a possible memory leak.

versioning/BRM/extentmap.cpp
```diff
@@ -269,19 +270,306 @@ FreeListImpl::FreeListImpl(unsigned key, off_t size, bool readOnly) :
...
+        const size_t memNeeded = (oids.size() + extraUnits_) * oidContainerUnitSize_;
+        if (oids.load_factor() >= oids.max_load_factor() ||
+            freeShmem <= freeSpaceThreshold_)
+        {
+            // Need to refresh all refs and iterators b/c the local address range changed.
+            shmemHasGrown = growIfNeeded(memNeeded);
+            auto* extMapIndexPtr = get();
+            assert(extMapIndexPtr);
+            auto& extMapIndex = *extMapIndexPtr;
+            shmemHasGrown = std::max(shmemHasGrown, aShmemHasGrown);
```
mariadb-AlexeyAntipovsky: Why not just `shmemHasGrown || aShmemHasGrown` ?

mariadb-AlexeyAntipovsky (changes requested):

versioning/BRM/extentmap.cpp
drrtuy: Good point.

versioning/BRM/brmshmimpl.cpp
drrtuy: I think execution never gets here b/c line 253 throws if the size is 0 but I will add a delete nevertheless.

versioning/BRM/extentmap.cpp
```diff
@@ -269,19 +270,306 @@ FreeListImpl::FreeListImpl(unsigned key, off_t size, bool readOnly) :
...
+}
+
+bool ExtentMapIndexImpl::growIfNeeded(const size_t memoryNeeded)
+{
+    auto freeShmem = getShmemFree();
+    // Worst case managed segment can't get continues buffer with len = memoryNeeded
+    if (freeShmem < memoryNeeded)
+    {
+        const size_t currentShmemSize = getShmemSize();
+        constexpr static const size_t minAllowance = 16 * 1024 * 1024;
```
denis0x0D, line 332: Nit, could we use named constant here like `InitEMIndexSize_`

versioning/BRM/extentmap.cpp
```diff
@@ -269,19 +270,306 @@ FreeListImpl::FreeListImpl(unsigned key, off_t size, bool readOnly) :
...
 }
 
+/*static*/
+std::mutex ExtentMapIndexImpl::fInstanceMutex_;
+
+/*static*/
+ExtentMapIndexImpl* ExtentMapIndexImpl::fInstance_ = nullptr;
+
+/*static*/
+ExtentMapIndexImpl* ExtentMapIndexImpl::makeExtentMapIndexImpl(unsigned key, off_t size, bool readOnly)
```
denis0x0D, line 281: Can you please clarify, why do we need a keys, since we identify that segment was grown by other process in the current process using size.

versioning/BRM/extentmap.cpp
```diff
@@ -269,19 +270,306 @@ FreeListImpl::FreeListImpl(unsigned key, off_t size, bool readOnly) :
...
+ExtentMapIndexImpl::ExtentMapIndexImpl(unsigned key, off_t size, bool readOnly) :
+    fBRMManagedShmMemImpl_(key, size, readOnly)
+{
+}
+
+void ExtentMapIndexImpl::createExtentMapIndexIfNeeded()
+{
+    // pair<T*, size>
+    // Hide fShmSegment
+    auto managedShmemSearchPair = fBRMManagedShmMemImpl_.getManagedSegment()->find<ExtentMapIndex>(EmIndexObjectName);
```
denis0x0D, line 309: Nit, `find_or_construct` here?

versioning/BRM/extentmap.cpp
```diff
@@ -269,19 +270,306 @@ FreeListImpl::FreeListImpl(unsigned key, off_t size, bool readOnly) :
...
+    }
+    return false;
+}
+
+InsertUpdateShmemKeyPair ExtentMapIndexImpl::insert(const EMEntry& emEntry, const size_t emIdx)
+{
+    auto dbRoot = emEntry.dbRoot;
+    assert(dbRoot > 0 && dbRoot <= numeric_limits<uint64_t>::max());
+    auto* extentMapIndexPtr = get();
+    assert(extentMapIndexPtr);
```
denis0x0D: Nit, seems like we have an assert in `get` method.

versioning/BRM/extentmap.cpp
drrtuy: We don't need them anymore however I preserve them to use parent class ctor BRMShmImplParent. It is pushed down  makeExtentMapIndexImpl -> ExtentMapIndexImpl -> BRMManagedShmImpl -> BRMShmImplParent

versioning/BRM/extentmap.cpp
drrtuy: Looks more natural indeed.
I have two concerns though: what if ExtentMapIndex instance is 0 and this is a major change. I don't want to have reasons to go though another round of full QA.

versioning/BRM/extentmap.cpp
drrtuy: Good point.

versioning/BRM/extentmap.cpp
drrtuy: I can't tell why but the linker complains when I make this ExtentMap::minAllowance_: 
`/usr/bin/ld: ../../lib/libbrm.so: undefined reference to `BRM::ExtentMap::minAllowance_'`
I would appreciate any suggestion.

versioning/BRM/extentmap.cpp
```diff
@@ -269,19 +270,306 @@ FreeListImpl::FreeListImpl(unsigned key, off_t size, bool readOnly) :
...
+        const size_t memNeeded = (partitions.size() + extraUnits_) * partitionContainerUnitSize_ + emIdentUnitSize_;
+        if (partitions.load_factor() >= partitions.max_load_factor() ||
+            freeShmem <= freeSpaceThreshold_)
+        {
+            // Need to refresh all refs and iterators b/c the local address range changed.
+            shmemHasGrown = growIfNeeded(memNeeded);
+            auto* extMapIndexPtr = get();
+            assert(extMapIndexPtr);
+            auto& extMapIndex = *extMapIndexPtr;
+            shmemHasGrown = std::max(shmemHasGrown, aShmemHasGrown);
```
mariadb-AlexeyAntipovsky: `shmemHasGrown || aShmemHasGrown` ?

versioning/BRM/extentmap.cpp
```diff
@@ -1247,13 +1559,15 @@ void ExtentMap::load(const string& filename, bool fixFL)
 #endif
 
     grabEMEntryTable(WRITE);
+    grabEMIndex(WRITE);
```
mariadb-AlexeyAntipovsky, line 1560: Won't it be possible to use RAII here  and below so as not to repeat releaseSomething in the right order? Can grabEMIndex throw an exception?

mariadb-AlexeyAntipovsky (changes requested):

versioning/BRM/extentmap.cpp
drrtuy: Sorry. Lost the fix by the way.

versioning/BRM/extentmap.cpp
drrtuy: Good idea indeed.
JFYI grabEMEntryTable, grabEMIndex, grabFreeList can throw with the same probability so everything must be wrapped up in number of try-catch blocks.

versioning/BRM/brmshmimpl.cpp
```diff
@@ -232,7 +240,166 @@ void BRMShmImpl::destroy()
...
+        off_t curSize = shmSegment->get_size();
+
+        idbassert(curSize > 0);
+        idbassert(curSize >= fSize);
+        fShmSegment = shmSegment;
+        fSize = curSize;
+    }
+}
+
+int BRMManagedShmImpl::grow(off_t newSize)
```
denis0x0D, line 327: Overall it's LGTM, I have only one concern about the grow function we implement -  regarding to docs we have to call grow after all other processes are released shared segment (call syscall `umap`), in our case we grow the segment in one process and other processes will know about that segment was grown when they try to grab the `lock` and check segment size, which is kinda breaking the official way of doing this, this potentially could cause an UB, but may be not. It's seems like now it works, but some day kernel people (since boost lib is a just wrapper around linux syscalls) can change something and it will not work anymore, because we use some non documented features, and they are in power to change non documented behavior.

versioning/BRM/extentmap.cpp
denis0x0D: Just something come to my mind, could we have a situation when we call `grow` for the segment with size 0, so we increase it on 0 bytes, with following sequence of syscalls `munmap`, `fruncate`, `mmap` and finally got the situation when one segment is `remaped` without the size increased and we cannot identify it, because the size is the same.

versioning/BRM/extentmap.cpp
```diff
@@ -2290,68 +2762,76 @@ LBID_t ExtentMap::_createColumnExtent_DBroot(uint32_t size, int OID,
...
-                            else
-                            {
-                                if (fExtentMap[i].blockOffset > iter->second)
-                                {
-                                    iter->second = fExtentMap[i].blockOffset;
-                                }
-                            }
-                        }
+                // 5. Track hi seg for hwm partition
+                auto emIdents = fPExtMapIndexImpl_->find(dbRootFromList, OID, targetDbRootPart);
```
denis0x0D, line 2825: We got `emIdents` here and iterating over `emIdentsNext` on the next line, is it correct?

drrtuy: Not at all. Looks like a bug. Thx a lot!

drrtuy: Fixed it in a separate PR.

versioning/BRM/extentmap.cpp
drrtuy: Only if there will be no grabEMIndex or grow called in the code. This can happen only if the programmer violates a "proto" to operate ManagedShmMem

drrtuy:
Closing this one b/c it was already released with a bug found by Denis and w/o RAII.

PR #2247, opened 2022-02-03
build dir clean-up and initial MCS bootstrap script version to simpli…

drrtuy:
…fy onboarding


mariadb-LeonidFedorov:
Why do we need bat files here, if we do not support windows?

mariadb-LeonidFedorov:
If we have build scripts in repo, much more useful not to force users to edit paths to cloned repo, but use script location
as ```dirname "$0"```, so user could just fire the script

build/randomSleep.sh
```diff
@@ -1,9 +0,0 @@
-#!/bin/bash
-
-# This a tool for buildbot to keep the nightlies from swamping svn
```
mariadb-LeonidFedorov, line 3: Are you sure we need it?

davidhall586 (approved):

build/bootstrap_mcs.sh
```diff
@@ -0,0 +1,183 @@
...
+MCS_SCRIPTS_REPO_PREFIX=/data/cs-docker-tools
+MDB_SOURCE_PATH=/data/mdb-server
+MCS_SOURCE_PATH=$MDB_SOURCE_PATH/storage/columnstore
+MCS_CONFIG_DIR=/etc/columnstore
+# Needs systemd to be installed obviously.
+# Feel free to ask in MariaDB Zulip how to bootstrap MCS in containers or other systemd-free environments.
+systemctl stop mariadb-columnstore
+MDB_GIT_URL=https://github.com/MariaDB/server.git
+MCS_GIT_URL=https://github.com/mariadb-corporation/mariadb-columnstore-engine.git
+MDB_GIT_TAG=10.5
```
mariadb-LeonidFedorov: can be parsed from .drone.jsonnet inside 
```
local server_ref_map = {
  develop: '10.7',
  'develop-6': '10.6-enterprise',
  'develop-5': '10.5',
  '**': '10.7',
}
```

drrtuy: I doubt it can given the private enterprise repo mentioned at least.

mariadb-LeonidFedorov (approved):

PR #2251, opened 2022-02-07
MCOL-4957 Fix performance slowdown for processing TIMESTAMP columns.

tntnatbry:


dbcon/execplan/treenode.h
```diff
@@ -1127,6 +1127,17 @@ inline int64_t TreeNode::getDatetimeIntVal()
...
         return (fResult.intVal);
+    else if (fResultType.colDataType == execplan::CalpontSystemCatalog::TIMESTAMP)
+    {
+        dataconvert::TimeStamp timestamp(fResult.intVal);
+        int64_t seconds = timestamp.second;
+        dataconvert::MySQLTime m_time;
+        dataconvert::gmtSecToMySQLTime(seconds, m_time, timeZone);
+        dataconvert::DateTime dt(m_time.year, m_time.month, m_time.day, m_time.hour,
+                                 m_time.minute, m_time.second, timestamp.msecond);
+        memcpy(&fResult.intVal, &dt, 8);
```
drrtuy, line 1138: What do you think about adding static constexpr size_t member to dataconvert::DateTime class and use it in memcpy.
I would better replace memcpy statement with an assignment expression, e.g. fResult.intVal = dt.someMethod()

dbcon/execplan/functioncolumn.h
```diff
@@ -219,31 +219,38 @@ class FunctionColumn : public ReturnedColumn
 public:
     virtual const std::string& getStrVal(rowgroup::Row& row, bool& isNull)
     {
+        fOperationType.timeZone = fTimeZone;
```
drrtuy, line 222: What do you think if we stick to OOP and use fOperationType::method(fTimeZone) to set the timezeone?

dbcon/mysql/ha_mcs_ddl.cpp
```diff
@@ -658,11 +658,14 @@ bool anyNullInTheColumn (THD* thd, string& schema, string& table, string& column
     csep.returnedCols(returnedColumnList);
 
     SimpleFilter* sf = new SimpleFilter();
-    sf->timeZone(thd->variables.time_zone->get_name()->ptr());
+    const char* timeZone = thd->variables.time_zone->get_name()->ptr();
+    long timeZoneOffset;
+    dataconvert::timeZoneToOffset(timeZone, strlen(timeZone), &timeZoneOffset);
```
drrtuy, line 663: We need to discuss this part.
But I suggest to change the API for timeZoneToOffset() and SimpleFilter ctor slightly to reduce side-effects. auto tzOffset = dataconvert::timeZoneToOffset(timeZone, len)); auto* sf = new SimpleFilter(tzOffset, sop);

dbcon/mysql/ha_mcs_ddl.cpp
```diff
@@ -1127,7 +1134,7 @@ int ProcessDDLStatement(string& ddlStatement, string& schema, const string& tabl
                 algorithm::to_lower(alterTable->fTableName->fName);
             }
 
-            alterTable->fTimeZone.assign(thd->variables.time_zone->get_name()->ptr());
+            alterTable->fTimeZone = timeZoneOffset;
```
drrtuy, line 1137: It is worth to discuss to use method to set alterTable::fTimeZone

dbcon/execplan/treenode.h
tntnatbry: `DateTime`, `Time` and `TimeStamp` are 64-bit structs, we cannot add a member variable to them.

dbcon/execplan/functioncolumn.h
tntnatbry: `fOperationType` here is of type `struct execplan::CalpontSystemCatalog::ColType`. Existing members of this struct are all public and there are no accessor/mutator methods. Changing the struct interface would be outside the scope of this patch.

dbcon/mysql/ha_mcs_ddl.cpp
tntnatbry: `timeZoneToOffset` returns a bool value indicating whether the conversion passed or failed. This return value is used in cpimport while parsing command line arguments, that's why we pass the offset by reference as a function argument. Changing `SimpleFilter` API would involve changing code at other places as well where we call this ctor, this would be outside the scope of this patch.

dbcon/mysql/ha_mcs_ddl.cpp
tntnatbry: All existing members of `struct AlterTableStatement` are public.

dbcon/execplan/treenode.h
drrtuy: Very good. Then Result should have a setter to assign a int value. And this is a very good example why I would like to see setters/getters used instead of direct attributes manipulations. Let me explain. You are setting Result::intVal. Imagine this Result needs to be used as an unsigned int(I am aware about TreeNode used in such context but for the sake of the explanation I will ignore this). Some Result::uIntValue will be used instead. And this potentially gives us hard to detect issues. With a setter you can set a value properly though.

dbcon/execplan/functioncolumn.h
drrtuy: I am not super crazy about this but I personally think this would allow us to reduce side-effects in the code though. If you don't want to plz resolve this convo. Feel free to introduce setters instead adding a new code. JFYI Bar left them public to reduce his patch.

dbcon/mysql/ha_mcs_ddl.cpp
drrtuy: Agree.

dbcon/mysql/ha_mcs_ddl.cpp
drrtuy: Plz see my prev comments regarding direct assignments to public attributes.

tntnatbry:
@drrtuy I have created a new PR for develop-6 to rebase over the clang-format commit, that addresses the comments above: https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/2262. So I am closing this PR.

PR #2254, opened 2022-02-09
New warnfixes

mariadb-LeonidFedorov:
- Fix new clang warnings
- One of them is definitely a bug: see 
```
int64_t x = atoll(value.c_str());
```
pattern, that shadows outer x variable and never uses
- removing vim tab guides

better to review by commits, second one is trivial

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -223,7 +223,7 @@ IF (MASK_LONGDOUBLE)
 ENDIF()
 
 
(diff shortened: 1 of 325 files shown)


drrtuy (commented):
The only thing that bothers me is the 'int64_t x' fix that might implictly change the behavior of the code. One needs to run test000, test001 with this patch.

drrtuy (approved):

PR #2262, opened 2022-02-14
MCOL-4957 Fix performance slowdown for processing TIMESTAMP columns.

tntnatbry:
Part 1:
 As part of MCOL-3776 to address synchronization issue while accessing
 the fTimeZone member of the Func class, mutex locks were added to the
 accessor and mutator methods. However, this slows down processing
 of TIMESTAMP columns in PrimProc significantly as all threads across
 all concurrently running queries would serialize on the mutex. This
 is because PrimProc only has a single global object for the functor
 class (class derived from Func in utils/funcexp/functor.h) for a given
 function name. To fix this problem:

   (1) We remove the fTimeZone as a member of the Func derived classes
   (hence removing the mutexes) and instead use the fOperationType
   member of the FunctionColumn class to propagate the timezone values
   down to the individual functor processing functions such as
   FunctionColumn::getStrVal(), FunctionColumn::getIntVal(), etc.

   (2) To achieve (1), a timezone member is added to the
   execplan::CalpontSystemCatalog::ColType class.

Part 2:
 Several functors in the Funcexp code call dataconvert::gmtSecToMySQLTime()
 and dataconvert::mySQLTimeToGmtSec() functions for conversion between seconds
 since unix epoch and broken-down representation. These functions in turn call
 the C library function localtime_r() which currently has a known bug of holding
 a global lock via a call to __tz_convert. This significantly reduces performance
 in multi-threaded applications where multiple threads concurrently call
 localtime_r(). More details on the bug:
   https://sourceware.org/bugzilla/show_bug.cgi?id=16145

 This bug in localtime_r() caused processing of the Functors in PrimProc to
 slowdown significantly since a query execution causes Functors code to be
 processed in a multi-threaded manner.

 As a fix, we remove the calls to localtime_r() from gmtSecToMySQLTime()
 and mySQLTimeToGmtSec() by performing the timezone-to-offset conversion
 (done in dataconvert::timeZoneToOffset()) during the execution plan
 creation in the plugin. Note that localtime_r() is only called when the
 time_zone system variable is set to "SYSTEM".

 This fix also required changing the timezone type from a std::string to
 a long across the system.


dbcon/execplan/calpontsystemcatalog.h
```diff
@@ -217,11 +216,25 @@ class CalpontSystemCatalog : public datatypes::SystemCatalog
     uint64_t nextvalue;  // next autoincrement value
     uint32_t charsetNumber;
     const CHARSET_INFO* cs;
+   private:
+    long timeZone;
```
drrtuy, line 220: The class attributes naming convention(that comes with the styling) tells that class attributes must be _-prefixed, e.g. timeZone_

drrtuy: e.g. [here](https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/2265/files#diff-5b944cc3f6f2c05bb660440855808abb59a7887266837d03800b4c750920c85dR138)

drrtuy (approved):

PR #2292, opened 2022-02-24
This patch now uses cmake3 on centos 7 and disables embeded libs comp…

drrtuy:
…ilation to reduce compilation time


mariadb-LeonidFedorov:
Decided to run detect_distro conditionaly for Centos and as it was, if manual OS is not in list

PR #2303, opened 2022-03-05
MCOL-4809 This patch introduces vectorized scanning/filtering for sho…

drrtuy:
…rt CHAR/VARCHAR columns

    Short CHAR/VARCHAR column values contain integer-encoded strings.
    After certain manipulations(orderSwap(strnxfrm(str))) the values
    become integers that preserve original strings order relation
    according to a certain translation rules(collation). Prepared
    values are ready to be SIMD-processed.

### DIFF
--- a/primitives/linux-port/column.cpp
+++ b/primitives/linux-port/column.cpp
@@ -1,5 +1,5 @@
 /* Copyright (C) 2014 InfiniDB, Inc.
-   Copyright (C) 2016-2021 MariaDB Corporation
+   Copyright (C) 2016-2022 MariaDB Corporation
 
    This program is free software; you can redistribute it and/or
    modify it under the terms of the GNU General Public License
@@ -22,6 +22,7 @@
 #include <cassert>
 #include <cmath>
 #include <functional>
+#include <type_traits>
 #ifndef _MSC_VER
 #include <pthread.h>
 #else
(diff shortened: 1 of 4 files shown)


primitives/linux-port/column.cpp
```diff
@@ -62,6 +63,55 @@ inline uint64_t order_swap(uint64_t x)
...
+inline T orderSwap(T x)
+{
+    return x;
+}
+
+template<typename T,
+        typename std::enable_if<sizeof(T) == sizeof(int64_t), T>::type* = nullptr>
+inline T orderSwap(T x)
+{
+    T ret = (x >> 56) |
```
mariadb-AlexeyAntipovsky, line 78: Maybe it's better to use `bswap_*` or even `htobe*`?

drrtuy: Any obvious reasons to use them ?

drrtuy: Moreover there will be no 128 bit version if we later need it.

mariadb-LeonidFedorov: Maybe this could be done with constexpr ?
https://stackoverflow.com/questions/36936584/how-to-write-constexpr-swap-function-to-change-endianess-of-an-integer

denis0x0D:
LGTM.

primitives/linux-port/column.cpp
drrtuy: I would like such a thing but imagine how complex it might be to support. IMHO this is an overkill.

mariadb-LeonidFedorov (approved):

PR #2305, opened 2022-03-11
[MCOL-4917] Update internal representation of ExtentMap.

denis0x0D:


tools/clearShm/main.cpp
```diff
@@ -79,6 +102,29 @@ void shmDoit(key_t shm_key, const string& label)
     }
 }
 
+void semDoit(std::string segmentName, const string& label)
```
drrtuy: What is the diff b/w shmDoit and semDoit? They looks almost identical and it is better to reduce boilerplate code.

tools/clearShm/main.cpp
```diff
@@ -213,6 +259,10 @@ int main(int argc, char** argv)
     semDoit(BrmKeys.KEYRANGE_VSS_BASE,        "VSS        ");
     semDoit(BrmKeys.KEYRANGE_EXTENTMAP_INDEX_BASE, "EXTMAP_INDX");
     semDoit(BrmKeys.MST_SYSVKEY,              "MST        ");
+    // We have only one segment, and grow it on request.
+    // FIXME: outline to constant.
+    semDoit("InfiniDB-shm-shared_managed_segment_00", "EXTENT_MAP_RB_TREE");
+    shmDoit("InfiniDB-shm-shared_managed_segment_00", "EXTENT_MAP_RB_TREE");
```
drrtuy: It is better to change the suffix from InfiniDB to MCS. Moreover it would be nice if you add the corresponding constants into shmemkeys.h

versioning/BRM/brmshmimpl.h
```diff
@@ -109,4 +109,29 @@ class BRMManagedShmImpl : public BRMShmImplParent
...
+  void reMapSegment();
+
+  boost::interprocess::managed_shared_memory* fShmSegment;
+
+private:
+  unsigned fKey;
+  off_t fSize;
+  bool fReadOnly;
+  // The `segment` name is fixed.
+  const char* segmentName = "MCS-shm-00020001";
```
drrtuy, line 128: const static char* ?

versioning/BRM/brmshmimpl.h
```diff
@@ -109,4 +109,29 @@ class BRMManagedShmImpl : public BRMShmImplParent
     bi::managed_shared_memory* fShmSegment;
 };
 
+class BRMManagedShmImplRBTree
```
drrtuy: Can you inherit from the BRMShmImplParent?

versioning/BRM/slavecomm.cpp
```diff
@@ -45,6 +45,8 @@
 
 #include "installdir.h"
 
+#define BRM_VERBOSE
```
drrtuy: I don't think this macros will activate logging on its own but it is better to remove it.

versioning/BRM/extentmap.h
```diff
@@ -187,49 +232,29 @@ class ExtentMapImpl
...
 
-    inline EMEntry* get() const
+    inline unsigned key() const { return fManagedShm.key(); }
+
+    inline ExtentMapRBTree* get() const
     {
-        return reinterpret_cast<EMEntry*>(fExtMap.fMapreg.get_address());
+        VoidAllocator allocator(fManagedShm.fShmSegment->get_segment_manager());
+        return fManagedShm.fShmSegment->find_or_construct<ExtentMapRBTree>("EmMapRBTree")(
+            std::less<int64_t>(), allocator);
```
drrtuy: What do you think if you replace int64_t with typeof(LBID_t) ?

versioning/BRM/extentmap.h
```diff
@@ -1037,109 +1028,140 @@ class ExtentMap : public Undoable
...
+
+  bool fDebug;
+
+  int _markInvalid(const LBID_t lbid, const execplan::CalpontSystemCatalog::ColDataType colDataType);
+
+  void loadVersion4(std::ifstream& in);
+  void loadVersion4(idbdatafile::IDBDataFile* in);
+
+  ExtentMapRBTree::iterator findByLBID(const LBID_t lbid);
+  std::vector<std::pair<UndoRecordType, EMEntry>> undoRecordsRBTree;
```
drrtuy: I would suggest to name the pair so that everybody else knows its semantics.

versioning/BRM/extentmap.h
```diff
@@ -1037,109 +1028,140 @@ class ExtentMap : public Undoable
...
+  void reserveLBIDRange(LBID_t start, uint8_t size); // used by load() to allocate pre-existing LBIDs
+  std::vector<EMEntry> getEmIdentsByLbids(const std::vector<LBID_t>& lbids);
+  std::vector<ExtentMapRBTree::iterator> getEmIteratorsByLbids(const std::vector<LBID_t>& lbids);
+
+  // Choose keys.
+  key_t chooseEMShmkey();
+  key_t chooseFLShmkey();
+  key_t chooseEMIndexShmkey();
+
+  // TODO: Add comments.
```
drrtuy: There are two options: remove the comment or add more comments b/c this TODO will last forever.

versioning/BRM/extentmap.cpp
```diff
@@ -204,37 +204,31 @@ bool EMEntry::operator< (const EMEntry& e) const
...
 
 /*static*/
-boost::mutex ExtentMapImpl::fInstanceMutex;
 boost::mutex ExtentMap::mutex;
 boost::mutex ExtentMap::emIndexMutex;
 
-/*static*/
-ExtentMapImpl* ExtentMapImpl::fInstance = nullptr;
+boost::mutex ExtentMapRBTreeImpl::fInstanceMutex;
+ExtentMapRBTreeImpl* ExtentMapRBTreeImpl::fInstance = 0;
```
drrtuy: 0 -> nullptr ?

versioning/BRM/extentmap.cpp
```diff
@@ -6164,6 +5894,43 @@ void ExtentMap::finishChanges()
...
+            auto emIt = findByLBID(key);
+            if (emIt != fExtentMapRBTree->end())
+            {
+                fExtentMapRBTree->erase(emIt);
+            }
+        }
+        else if (undoPair.first == UndoRecordType::DELETE)
+        {
+            const auto& emEntry = undoPair.second;
+            fExtentMapRBTree->insert(make_pair(emEntry.range.start, emEntry));
```
drrtuy, line 5918: What about emindex or it was not modified yet?

versioning/BRM/extentmap.cpp
```diff
@@ -6164,6 +5894,43 @@ void ExtentMap::finishChanges()
         releaseEMEntryTable(WRITE);
 }
 
+void ExtentMap::makeUndoRecordRBTree(UndoRecordType type, const EMEntry& emEntry)
+{
+    undoRecordsRBTree.push_back(make_pair(type, emEntry));
+}
+
+void ExtentMap::undoChangesRBTree()
```
drrtuy, line 5805: I am curios if undo code takes emindex into account? What happens if one removes/adds EMEntry so that it is removed/added from/into the index and rollback then?

tools/clearShm/main.cpp
denis0x0D: They are different, the original one has a input parameter as a key, transforms the key to name, in my implementation the key is not a name, the name is fixed.

tools/clearShm/main.cpp
denis0x0D: I see, that a bug, thanks.

versioning/BRM/extentmap.cpp
denis0x0D: Does it have support in original patch?

versioning/BRM/slavecomm.cpp
denis0x0D: Yes, thanks.

versioning/BRM/extentmap.h
denis0x0D: Sure, thanks.

tools/clearShm/main.cpp
denis0x0D: UPD, I'fixed that, thanks.

versioning/BRM/extentmap.cpp
drrtuy: No, it doesn't.

versioning/BRM/extentmap.cpp
```diff
@@ -604,46 +640,35 @@ ExtentMap::~ExtentMap()
...
+    auto& emEntry = emIt->second;
     {
-        lastBlock = fExtentMap[i].range.start +
-                    (static_cast<LBID_t>(fExtentMap[i].range.size) * 1024) - 1;
+        lastBlock = emEntry.range.start + (static_cast<LBID_t>(emEntry.range.size) * 1024) - 1;
 
-        if (fExtentMap[i].range.size != 0)
         {
-            if (lbid >= fExtentMap[i].range.start && lbid <= lastBlock)
+            // FIXME: Remove this check, since we use tree.
```
drrtuy: Should we just remove this and re-test or postpone this?

versioning/BRM/extentmap.cpp
```diff
@@ -782,68 +803,68 @@ int ExtentMap::setMaxMin(const LBID_t lbid,
...
-        {
-            lastBlock = fExtentMap[i].range.start +
-                        (static_cast<LBID_t>(fExtentMap[i].range.size) * 1024) - 1;
-            curSequence = fExtentMap[i].partition.cprange.sequenceNum;
+        lastBlock = emEntry.range.start + (static_cast<LBID_t>(emEntry.range.size) * 1024) - 1;
+        curSequence = emEntry.partition.cprange.sequenceNum;
 
-            if (lbid >= fExtentMap[i].range.start && lbid <= lastBlock)
-            {
+        if (lbid >= emEntry.range.start && lbid <= lastBlock)
```
drrtuy: This check might be also redundant after findByLBID().

versioning/BRM/extentmap.cpp
```diff
@@ -782,68 +803,68 @@ int ExtentMap::setMaxMin(const LBID_t lbid,
...
         }
     }
 
-    if (emLocked)
-        releaseEMEntryTable(WRITE);
-
     if (emIndexLocked)
         releaseEMIndex(WRITE);
 
+    if (emLocked)
```
drrtuy, line 843: You are changing the order here. IMHO this opens a tiny window for a deadlock.

versioning/BRM/extentmap.cpp
```diff
@@ -905,63 +924,46 @@ void ExtentMap::setExtentsMaxMin(const CPMaxMinMap_t& cpMap, bool firstNode, boo
         grabEMIndex(WRITE);
     }
 
-    entries = fEMShminfo->allocdSize / sizeof(struct EMEntry);
-
-    for (i = 0; i < entries; i++)
+    for (auto emIt = fExtentMapRBTree->begin(), end = fExtentMapRBTree->end(); emIt != end; ++emIt)
```
drrtuy: This setExtentsMaxMin() method' loop should iterate over cpMap and use findByLBID() to reduce computation complexity.

versioning/BRM/extentmap.cpp
```diff
@@ -1060,160 +1062,127 @@ void ExtentMap::mergeExtentsMaxMin(CPMaxMinMergeMap_t& cpMap, bool useLock)
         grabEMIndex(WRITE);
     }
 
-    int entries = fEMShminfo->allocdSize / sizeof(struct EMEntry);
-
-    for (int i = 0; i < entries; i++)  			// loop through all extents
+    for (auto emIt = fExtentMapRBTree->begin(), end = fExtentMapRBTree->end(); emIt != end; ++emIt)
```
drrtuy: This mergeExtentsMaxMin() method' loop should iterate over cpMap and use findByLBID() to reduce computation complexity.

versioning/BRM/extentmap.cpp
```diff
@@ -1060,160 +1062,127 @@ void ExtentMap::mergeExtentsMaxMin(CPMaxMinMergeMap_t& cpMap, bool useLock)
...
 
-                        if (isValidCPRange(
-                                    fExtentMap[i].partition.cprange.hi_val,
-                                    fExtentMap[i].partition.cprange.lo_val,
-                                    it->second.type))
+                    if (isValidCPRange(emEntry.partition.cprange.hi_val, emEntry.partition.cprange.lo_val,
+                                       it->second.type))
+                    {
+                        // Swap byte order to do binary string comparison
+                        if (isCharType(it->second.type))
```
drrtuy, line 1103: This is an interesting part. Where did this come from? Do you recall the case that causes you to add this if branch?

versioning/BRM/extentmap.cpp
```diff
@@ -1434,110 +1429,111 @@ void ExtentMap::reserveLBIDRange(LBID_t start, uint8_t size)
...
         log_errno("ExtentMap::loadVersion4(): read ");
         throw runtime_error("ExtentMap::loadVersion4(): read failed. Check the error log.");
     }
 
-    void *fExtentMapPtr = static_cast<void*>(fExtentMap);
-    memset(fExtentMapPtr, 0, fEMShminfo->allocdSize);
-    fEMShminfo->currentSize = 0;
-    // init the free list
+    // Clear the extent map.
+    fExtentMapRBTree->clear();
```
drrtuy, line 1396: This is a very good idea.

versioning/BRM/extentmap.cpp
```diff
@@ -1434,110 +1429,111 @@ void ExtentMap::reserveLBIDRange(LBID_t start, uint8_t size)
...
-            nrows *= EM_INCREMENT_ROWS;
-        }
+    // Calculate how much memory we need.
+    const uint32_t memorySizeNeeded = (emNumElements * EM_RB_TREE_NODE_SIZE) + EM_RB_TREE_EMPTY_SIZE;
 
-        growEMShmseg(nrows);
-        growEMIndexShmseg(ExtentMapIndexImpl::estimateEMIndexSize(emNumElements));
-    }
+    growEMShmseg(memorySizeNeeded);
+//    growEMIndexShmseg(ExtentMapIndexImpl::estimateEMIndexSize(emNumElements));
```
drrtuy: Plz remove the commented line.

versioning/BRM/extentmap.cpp
```diff
@@ -1434,110 +1429,111 @@ void ExtentMap::reserveLBIDRange(LBID_t start, uint8_t size)
...
+            progress += (uint) err;
         }
-        progress += (uint) err;
+
+        std::pair<int64_t, EMEntry> lbidEMEntryPair = make_pair(emEntry.range.start, emEntry);
+        fExtentMapRBTree->insert(lbidEMEntryPair);
     }
 
-    for (size_t i = 0; i < emNumElements; ++i)
+    for (auto& lbidEMEntryPair : *fExtentMapRBTree)
```
drrtuy, line 1435: I wonder if we can just iterate over all entries once in the previous loop.

versioning/BRM/extentmap.cpp
```diff
@@ -1946,22 +1899,12 @@ void ExtentMap::grabEMIndex(OPS op)
...
 void ExtentMap::releaseEMEntryTable(OPS op)
 {
     if (op == READ)
-    {
         fMST.releaseTable_read(MasterSegmentTable::EMTable);
-    }
     else
     {
-        /*
-           Note: Technically we should mark it unlocked after it's unlocked,
```
drrtuy, line 1959: Why do you think this comment is irrelevant now?

versioning/BRM/extentmap.cpp
```diff
@@ -2339,24 +2303,20 @@ int ExtentMap::lookupLocal_DBroot(int OID, uint16_t dbroot, uint32_t partitionNu
     grabEMEntryTable(READ);
     grabEMIndex(READ);
 
-    entries = fEMShminfo->allocdSize / sizeof(struct EMEntry);
-
-    for (i = 0; i < entries; i++)
+    for (auto emIt = fExtentMapRBTree->begin(), end = fExtentMapRBTree->end(); emIt != end; ++emIt)
```
drrtuy, line 2223: This is my bad that I didn't fix this routine.
Could I ask you for a favor to apply index search here?

versioning/BRM/extentmap.cpp
```diff
@@ -2406,20 +2363,20 @@ int ExtentMap::lookupLocalStartLbid(int      OID,
     grabEMEntryTable(READ);
     grabEMIndex(READ);
 
-    DBRootVec dbRootVec(std::move(getAllDbRoots()));
+    DBRootVec dbRootVec(getAllDbRoots());
```
drrtuy, line 2283: Nice one. This std::move prevents copy elision according with clangd suggestion.

versioning/BRM/extentmap.cpp
```diff
@@ -2659,87 +2586,47 @@ LBID_t ExtentMap::_createColumnExtent_DBroot(uint32_t size, int OID,
...
-                highestSegNum   = fExtentMap[i].segmentNum;
-                highestOffset   = fExtentMap[i].blockOffset;
-            }
-        }             // found valid extentmap entry
-
-        // 3. Find first available extent map entry that can be reused
-        else if (emptyEMEntry < 0)
-            emptyEMEntry = i;
-    } // Loop through extent map entries
+    LBID_t startLBID = getLBIDsFromFreeList(size);
```
drrtuy, line 2507: This is a widely used method and you should get back index search header IMHO.

versioning/BRM/extentmap.cpp
```diff
@@ -2759,80 +2646,69 @@ LBID_t ExtentMap::_createColumnExtent_DBroot(uint32_t size, int OID,
...
-                                     fExtentMap[lastExtentIndex].segmentNum,
-                                     fExtentMap[lastExtentIndex].blockOffset) );
+        uint32_t targetDbRootPart = lastExtent->partitionNum;
+        uint32_t targetDbRootPartNext = targetDbRootPart + 1;
+        partHighSeg = lastExtent->segmentNum;
+        targetDbRootSegs.insert(
+            TargetDbRootSegsMap::value_type(lastExtent->segmentNum, lastExtent->blockOffset));
 
-        for (auto dbRootFromList : dbRootVec)
+        for (auto emIt = fExtentMapRBTree->begin(), end = fExtentMapRBTree->end(); emIt != end;
```
drrtuy: This is a widely used method and you should get back index search header IMHO.

versioning/BRM/extentmap.cpp
```diff
@@ -3040,62 +2913,59 @@ LBID_t ExtentMap::_createColumnExtent_DBroot(uint32_t size, int OID,
...
+        e.partition.cprange.isValid = CP_INVALID;
 
+    partitionNum = e.partitionNum;
+    segmentNum = e.segmentNum;
+    startBlockOffset = e.blockOffset;
+
+    makeUndoRecordRBTree(UndoRecordType::INSERT, e);
+    makeUndoRecord(fEMRBTreeShminfo, sizeof(MSTEntry));
+    // Insert into RBTree.
+    std::pair<int64_t, EMEntry> lbidEmEntryPair = make_pair(startLBID, e);
```
drrtuy, line 2884: Use LBID_T here or even name std::pair<int64_t, EMEntry> as a type.

versioning/BRM/extentmap.cpp
```diff
@@ -3253,97 +3071,87 @@ void ExtentMap::logAndSetEMIndexReadOnly(const std::string& funcName)
...
-    LastExtentIndexT lastExtentIndex = lastIndEmptyIndEmptyInd.first;
-    EmptyEMEntry emptyEMEntry = lastIndEmptyIndEmptyInd.second;
+    uint32_t highestOffset = 0;
+    LBID_t startLBID = getLBIDsFromFreeList(size);
+    EMEntry* lastEmEntry = nullptr;
 
-    if (emptyEMEntry == -1)
+    ASSERT(fExtentMapRBTree);
+
+    for (auto& emEntryPair : *fExtentMapRBTree)
```
drrtuy: IMHO we should revert this chunk and apply Index search here b/c we increase the complexity from O(1) to O(N) in the worst case.

versioning/BRM/extentmap.cpp
```diff
@@ -3436,73 +3234,71 @@ void ExtentMap::createDictStoreExtent(int OID,
...
+LBID_t ExtentMap::_createDictStoreExtent(uint32_t size, int OID, uint16_t dbRoot, uint32_t partitionNum,
+                                         uint16_t segmentNum)
 {
-    auto lastIndEmptyIndEmptyInd = _createExtentCommonSearch(OID, dbRoot,
-        partitionNum, segmentNum);
-    LastExtentIndexT lastExtentIndex = lastIndEmptyIndEmptyInd.first;
-    EmptyEMEntry emptyEMEntry = lastIndEmptyIndEmptyInd.second;
 
-    if (emptyEMEntry == -1)
+    uint32_t highestOffset = 0;
```
drrtuy: IMHO we should revert this chunk and apply Index search here b/c we increase the complexity from O(1) to O(N) in the worst case.

versioning/BRM/extentmap.cpp
```diff
@@ -3689,33 +3478,33 @@ void ExtentMap::rollbackColumnExtents_DBroot ( int oid,
     grabEMIndex(WRITE);
     grabFreeList(WRITE);
 
-    int emEntries = fEMShminfo->allocdSize / sizeof(struct EMEntry);
+    auto emIt = fExtentMapRBTree->begin();
```
drrtuy, line 3391: EMIndex should  be used here instead IMHO.

versioning/BRM/extentmap.cpp
```diff
@@ -3895,42 +3673,42 @@ void ExtentMap::rollbackDictStoreExtents_DBroot ( int oid,
...
```
drrtuy, line 3586: Index search should be used wherever there is a <dbroot,oid,partition> tuple.

versioning/BRM/extentmap.cpp
```diff
@@ -4029,120 +3807,113 @@ void ExtentMap::deleteEmptyColExtents(const ExtentsInfoMap_t& extentsInfo)
     uint32_t fboHi = 0;
     uint32_t fboLoPreviousStripe = 0;
 
-    int emEntries = fEMShminfo->allocdSize / sizeof(struct EMEntry);
     ExtentsInfoMap_t::const_iterator it;
 
-    for (int i = 0; i < emEntries; i++)
+    auto emIt = fExtentMapRBTree->begin();
```
drrtuy, line 3722: ExtentsInfoMap_t contains <dbroot,oid,partition> so Index should be used here IMHO

versioning/BRM/extentmap.cpp
```diff
@@ -4161,106 +3932,108 @@ void ExtentMap::deleteEmptyDictStoreExtents(const ExtentsInfoMap_t& extentsInfo)
...
-    ExtentsInfoMap_t::const_iterator it;
-
     uint32_t fboLo = 0;
     uint32_t fboHi = 0;
 
-    int emEntries = fEMShminfo->allocdSize / sizeof(struct EMEntry);
-    it  = extentsInfo.begin();
+    auto it = extentsInfo.begin();
+    auto emIt = fExtentMapRBTree->begin();
+    auto end = fExtentMapRBTree->end();
```
drrtuy, line 3850: ExtentsInfoMap_t contains <dbroot,oid,partition> so Index should be used here IMHO

versioning/BRM/extentmap.cpp
```diff
@@ -4341,71 +4114,64 @@ void ExtentMap::deleteOIDs(const OidsMap_t& OIDs)
     grabEMIndex(WRITE);
     grabFreeList(WRITE);
 
-    OidsMap_t::const_iterator it;
-    int emEntries = fEMShminfo->allocdSize / sizeof(struct EMEntry);
-
-    const bool clearEMIndex = false;
     DBRootVec dbRootVec(std::move(getAllDbRoots()));
```
drrtuy: Feel free to remove std::move() call.

versioning/BRM/extentmap.cpp
```diff
@@ -4341,71 +4114,64 @@ void ExtentMap::deleteOIDs(const OidsMap_t& OIDs)
...
             fPExtMapIndexImpl_->deleteOID(dbRoot, oidOidPair.first);
     }
 
-    for (int emIndex = 0; emIndex < emEntries; emIndex++)
-    {
-        if (fExtentMap[emIndex].range.size > 0 )
-        {
-            it = OIDs.find ( fExtentMap[emIndex].fileID );
+    auto it = fExtentMapRBTree->begin();
+    auto end = fExtentMapRBTree->end();
```
drrtuy, line 4035: We need to apply index here.

versioning/BRM/extentmap.cpp
```diff
@@ -4530,12 +4299,15 @@ HWM_t ExtentMap::getLastHWM_DBroot(int OID, uint16_t dbRoot,
...
     uint32_t lastExtent = 0;
-    int  lastExtentIndex = -1;
+    ExtentMapRBTree::iterator lastIt = fExtentMapRBTree->end();
     partitionNum = 0;
-    segmentNum   = 0;
-    HWM_t hwm    = 0;
-    bFound       = false;
+    segmentNum = 0;
+    HWM_t hwm = 0;
+    bFound = false;
```
drrtuy, line 4220: EM Index should be used here IMHO.

versioning/BRM/extentmap.cpp
```diff
@@ -4781,17 +4534,14 @@ void ExtentMap::getExtentState(int OID, uint32_t partitionNum,
...
-    emEntries = fEMShminfo->allocdSize / sizeof(struct EMEntry);
-
-    for (i = 0; i < emEntries; i++)
+    for (auto emIt = fExtentMapRBTree->begin(), end = fExtentMapRBTree->end(); emIt != end; ++emIt)
     {
-        if ((fExtentMap[i].range.size  != 0) &&
-                (fExtentMap[i].fileID      == OID) &&
-                (fExtentMap[i].partitionNum == partitionNum) &&
-                (fExtentMap[i].segmentNum  == segmentNum))
+        const auto& emEntry = emIt->second;
```
drrtuy, line 4449: Index should be used here IMHO.

versioning/BRM/extentmap.cpp
```diff
@@ -5062,23 +4804,21 @@ void ExtentMap::bulkUpdateDBRoot(const vector<BulkUpdateDBRootArg>& args)
...
     for (uint32_t i = 0; i < args.size(); i++)
         sArgs.insert(args[i]);
 
     grabEMEntryTable(WRITE);
     grabEMIndex(WRITE);
 
-    emEntries = fEMShminfo->allocdSize / sizeof(struct EMEntry);
-
-    for (int i = 0; i < emEntries; i++)
+    for (auto emIt = fExtentMapRBTree->begin(), end = fExtentMapRBTree->end(); emIt != end; ++emIt)
```
drrtuy, line 4724: Index should be used here IMHO.

versioning/BRM/extentmap.cpp
```diff
@@ -5109,23 +4848,27 @@ void ExtentMap::getExtents(int OID, vector<struct EMEntry>& entries,
...
-    entries.reserve(emEntries);
+    entries.reserve(fExtentMapRBTree->size());
+
     if (incOutOfService)
     {
-        for (i = 0 ; i < emEntries; i++)
-            if ((fExtentMap[i].fileID == OID) &&
-                    (fExtentMap[i].range.size != 0))
-                entries.push_back(fExtentMap[i]);
+        for (auto emIt = fExtentMapRBTree->begin(), end = fExtentMapRBTree->end(); emIt != end;
```
drrtuy, line 4765: This method is an interesting one so let's discuss how to leverage index here.
The most straightforward way is to find all lbids that belongs to the target OID iterating over all dbroots available. And then iterate over lbids running findByLBID() in a loop. However this approach will be slow given a big table with lots of extents. It would be better to find the topmost tree iterator and start iterating from there but I am not sure how to find a suitable first iterator to be honest.

versioning/BRM/extentmap.cpp
```diff
@@ -5219,28 +4957,25 @@ void ExtentMap::getExtentCount_dbroot(int OID, uint16_t dbroot,
...
     grabEMEntryTable(READ);
     grabEMIndex(READ);
-    emEntries = fEMShminfo->allocdSize / sizeof(struct EMEntry);
-
-    numExtents = 0;
 
     if (incOutOfService)
     {
-        for (i = 0 ; i < emEntries; i++)
+        for (auto emIt = fExtentMapRBTree->begin(), end = fExtentMapRBTree->end(); emIt != end;
```
drrtuy, line 4873: This method surely needs index IMHO.

versioning/BRM/extentmap.cpp
```diff
@@ -5272,19 +5007,18 @@ void ExtentMap::getSysCatDBRoot(OID_t oid, uint16_t& dbRoot)
     bool bFound = false;
     grabEMEntryTable(READ);
     grabEMIndex(READ);
-    int emEntries = fEMShminfo->allocdSize / sizeof(struct EMEntry);
 
-    for (int i = 0 ; i < emEntries; i++)
+    for (auto emIt = fExtentMapRBTree->begin(), end = fExtentMapRBTree->end(); emIt != end; ++emIt)
```
drrtuy, line 4921: I presume that syscat will be in the beginning of the tree so it is safe to left this iteration loop.

versioning/BRM/extentmap.cpp
```diff
@@ -5338,45 +5072,49 @@ void ExtentMap::deletePartition(const set<OID_t>& oids,
...
-    // First: validate against referencing non-existent logical partitions
-    std::set<OID_t>::const_iterator it;
+    std::set<LogicalPartition> foundPartitions;
+    std::vector<ExtentMapRBTree::iterator> extents;
+
+    auto it = fExtentMapRBTree->begin();
+    auto end = fExtentMapRBTree->end();
 
-    for (int i = 0; i < emEntries; i++)
+    while (it != end)
```
drrtuy, line 4992: This method surely needs index IMHO.

versioning/BRM/extentmap.cpp
```diff
@@ -5442,53 +5173,47 @@ void ExtentMap::markPartitionForDeletion(const set<OID_t>& oids,
...
     set<LogicalPartition> foundPartitions;
-    vector<uint32_t> extents;
+    vector<ExtentMapRBTree::iterator> extents;
     bool partitionAlreadyDisabled = false;
 
     // Identify not exists partition first. Then mark disable.
     std::set<OID_t>::const_iterator it;
-
-    for (int i = 0; i < emEntries; i++)
+    for (auto emIt = fExtentMapRBTree->begin(), end = fExtentMapRBTree->end(); emIt != end; ++emIt)
```
drrtuy, line 5093: This method surely needs index IMHO.

versioning/BRM/extentmap.cpp
```diff
@@ -5555,19 +5280,16 @@ void ExtentMap::markAllPartitionForDeletion(const set<OID_t>& oids)
 
     grabEMEntryTable(WRITE);
     grabEMIndex(WRITE);
-    int emEntries = fEMShminfo->allocdSize / sizeof(struct EMEntry);
 
-    for (int i = 0; i < emEntries; i++)
+    for (auto emIt = fExtentMapRBTree->begin(), end = fExtentMapRBTree->end(); emIt != end; ++emIt)
```
drrtuy, line 5194: This method surely needs index IMHO.

versioning/BRM/extentmap.cpp
```diff
@@ -5610,39 +5332,38 @@ void ExtentMap::restorePartition(const set<OID_t>& oids,
...
     grabEMIndex(WRITE);
 
-    int emEntries = fEMShminfo->allocdSize / sizeof(struct EMEntry);
-    vector<uint32_t> extents;
+    vector<ExtentMapRBTree::iterator> extents;
     set<LogicalPartition> foundPartitions;
     bool partitionAlreadyEnabled = false;
 
-    for (int i = 0; i < emEntries; i++)
+    for (auto emIt = fExtentMapRBTree->begin(), end = fExtentMapRBTree->end(); emIt != end; ++emIt)
```
drrtuy, line 5249: This method surely needs index IMHO.

versioning/BRM/extentmap.cpp
```diff
@@ -5702,19 +5422,14 @@ void ExtentMap::getOutOfServicePartitions(OID_t oid,
 
     grabEMEntryTable(READ);
     grabEMIndex(READ);
-    int emEntries = fEMShminfo->allocdSize / sizeof(struct EMEntry);
 
-    for (int i = 0; i < emEntries; i++)
+    for (auto emIt = fExtentMapRBTree->begin(), end = fExtentMapRBTree->end(); emIt != end; ++emIt)
```
drrtuy, line 5336: This method surely needs index IMHO.

versioning/BRM/extentmap.cpp
```diff
@@ -5768,21 +5491,18 @@ bool ExtentMap::isDBRootEmpty(uint16_t dbroot)
...
 
-    if (fEMShminfo->currentSize == 0)
+    if (fEMRBTreeShminfo->currentSize == 0)
     {
         throw runtime_error(
             "ExtentMap::isDBRootEmpty() shared memory not loaded");
     }
 
-    for (i = 0; i < emEntries; i++)
+    for (auto it = fExtentMapRBTree->begin(), end = fExtentMapRBTree->end(); it != end; ++it)
```
drrtuy, line 5413: This method surely needs index IMHO.

versioning/BRM/extentmap.cpp
```diff
@@ -5850,23 +5568,22 @@ void ExtentMap::lookup(OID_t OID, LBIDRange_v& ranges)
 
     grabEMEntryTable(READ);
     grabEMIndex(READ);
-    emEntries = fEMShminfo->allocdSize / sizeof(struct EMEntry);
 
-    for (i = 0 ; i < emEntries; i++)
-        if ((fExtentMap[i].fileID     == OID) &&
-                (fExtentMap[i].range.size != 0) &&
-                (fExtentMap[i].status     != EXTENTOUTOFSERVICE))
+    for (auto it = fExtentMapRBTree->begin(), end = fExtentMapRBTree->end(); it != end; ++it)
```
drrtuy, line 5482: This method surely needs index IMHO.

versioning/BRM/extentmap.cpp
denis0x0D: It seems like it was a bug, because we do lock in following order:
1. lockEmEntryTable
2. lockEmIndex

So the release should be
1. releaseEmIndex
2. releaseEmEntryTable

versioning/BRM/extentmap.cpp
denis0x0D: Removed by accident, thanks.

versioning/BRM/extentmap.cpp
drrtuy: Correct. Thx for fixing my mistake.

versioning/BRM/extentmap.cpp
```diff
@@ -3253,101 +3006,85 @@ void ExtentMap::logAndSetEMIndexReadOnly(const std::string& funcName)
...
+{
+  uint32_t highestOffset = 0;
+  LBID_t startLBID = getLBIDsFromFreeList(size);
+  EMEntry* lastEmEntry = nullptr;
+
+  const auto lbids = fPExtMapIndexImpl_->find(dbRoot, OID);
+  auto emIdents = getEmIdentsByLbids(lbids);
+  for (auto& emEntry : emIdents)
+  {
+    if ((emEntry.fileID == OID) && (emEntry.segmentNum == segmentNum) &&
```
drrtuy: OID and segment num comparisons looks redundant. IMHO You either need to remove them or add an error processing block that throws if OID and segmentNum doesn't match.

drrtuy (approved):

PR #2318, opened 2022-03-23
Renamed cal-prefixed UDFs to mcs-prefixed

commander-trashdin:
Issue https://jira.mariadb.org/browse/MCOL-4984.

### DIFF
--- a/mysql-test/columnstore/basic/r/mcs_named_udfs.result
+++ b/mysql-test/columnstore/basic/r/mcs_named_udfs.result
@@ -1,0 +1,33 @@
+DROP DATABASE IF EXISTS cal_test_db;
+CREATE DATABASE cal_test_db;
+USE cal_test_db;
+create table t1(a int, b int comment 'autoincrement=1') engine=columnstore;
+select mcsflushcache();
+mcsflushcache()
+0
+select mcssettrace(0);
+mcssettrace(0)
+0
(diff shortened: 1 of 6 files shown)


mysql-test/columnstore/basic/r/mcs_named_udfs.result
```diff
@@ -0,0 +1,33 @@
...
+0
+select mcssetparms("pmmaxmemorysmallside","2048000000");
+mcssetparms("pmmaxmemorysmallside","2048000000")
+Updated pmmaxmemorysmallside 2048000000
+select mcsgettrace();
+mcsgettrace()
+NULL
+select mcsgetversion();
+mcsgetversion()
+8.1.1
```
drrtuy: This result will surely change so we need to test the shape of what mcsgetversion() cal returns.

drrtuy:
As Leonid mentioned there must be only one copy of actual code not two, e.g. there are two separate implementations for callastinsertid_init and mcslastinsertid_init that are identical except some string value.

mysql-test/columnstore/basic/r/mcs_named_udfs.result
mariadb-LeonidFedorov: We can use test query select ```calgetversion()=mcsgetversion()``` here

mariadb-LeonidFedorov (approved):

PR #2319, opened 2022-03-23
Renamed cal-prefixed UDFs to mcs-prefixed

mariadb-LeonidFedorov:
Issue https://jira.mariadb.org/browse/MCOL-4984.


drrtuy:
As Leonid mentioned there must be only one copy of actual code not two, e.g. there are two separate implementations for  callastinsertid_init and mcslastinsertid_init that are identical except some string value.

PR #2327, opened 2022-03-31
MCOL-4912 This patch introduces Extent Map index to improve EM scalea…

drrtuy:
…bility

        EM scaleability project has two parts: phase1 and phase2.
        This is phase1 that brings EM index to speed up(from O(n) down
        to the speed of boost::unordered_map) EM lookups looking for
        <dbroot, oid, partition> tuple to turn it into LBID,
        e.g. most bulk insertion meta info operations.
        The basis is boost::shared_managed_object where EMIndex is
        stored. Whilst it is not debug-friendly it allows to put a
        nested structs into shmem. EMIndex has 3 tiers. Top down description:
        vector of dbroots, map of oids to partition vectors, partition
        vectors that have EM indices.
        Separate EM methods now queries index before they do EM run.
        EMIndex has a separate shmem file with the fixed id
        MCS-shm-00060001.


mariadb-DenisKhalikov (approved):

drrtuy:
Can't merge into develop right now b/c of multiple boost issues. We either use legacy boost 1.55 in Centos 7 and somehow solve outdated(according to C++20) interface to access allocator or update boost to 1.69 and fix explicit server warnings produced by -no-virtual-dtor switch. Both ways aren't feasible so we decide to introduce standalone boost build into develop.

PR #2343, opened 2022-04-21
MCOL-5060: Remove old S3 configuration from post-install.

benthompson15:


drrtuy:
I am against such changes so plz don't merge the patch until we have some words with Todd.

drrtuy:
Plz take a look at my PR.

PR #2358, opened 2022-04-29
fix issue MCOL-4423

NTH19:
pionter conservation error,it is wrong for a parent class pointer to become a child class


mariadb-LeonidFedorov:
Deleting assert is not a problem solver. You should figure out, why we have wrong type in this cast, and what to do there.

NTH19:
> Deleting assert is not a problem solver. You should figure out, why we have wrong type in this cast, and what to do there. 

Oh,thanks.I just thought that was a line of useless code before.

I notice the Field class has a function named "is_unsigned".Does it has the same meaning as field_num::unsigned_flag?

PR #2372, opened 2022-05-13
Statistics Aggregate Functions Improvements

commander-trashdin:
Updated STD with Welford's parallel online algorithm (https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance). Doesn't work for decimals (yet). 
Note - -does division on each step.


utils/rowgroup/rowaggregation.cpp
```diff
@@ -3156,31 +3164,27 @@ void RowAggregationUM::calculateStatisticsFunctions()
         }
         else  // count > 1
         {
-          long double sum1 = fRow.getLongDoubleField(colAux);
-          long double sum2 = fRow.getLongDoubleField(colAux + 1);
-
+          //long double mean = fRow.getLongDoubleField(colAux);
```
drrtuy: Remove or uncomment.

PR #2373, opened 2022-05-16
MCOL-4912-MCOL-4917 The patch replaces RBTree-based EM traversal with a

drrtuy:
combination of EM Index lookup + targeted RBTree search in EM routines
 This combination reduces avg search complexity from O(N) down to 0(1) * O(logN)
 that is crucial for EMs > 50MB.
 This patch also reduces RBTree-based EM footprint on shmem.

### DIFF
--- a/versioning/BRM/extentmap.cpp
+++ b/versioning/BRM/extentmap.cpp
@@ -558,6 +558,14 @@ LBID_tFindResult ExtentMapIndexImpl::search3dLayer(PartitionIndexContainerT& par
     return result;
 }
 
+bool ExtentMapIndexImpl::isDBRootEmpty(const DBRootT dbroot)
+{
+    ExtentMapIndex& extMapIndex = *get();
+    if (dbroot >= extMapIndex.size())
+        return true;
+    return extMapIndex[dbroot].empty();
+}
+
 void ExtentMapIndexImpl::deleteDbRoot(const DBRootT dbroot)
 {
     auto& extMapIndex = *get();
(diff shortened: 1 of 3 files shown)


versioning/BRM/extentmap.cpp
```diff
@@ -4034,23 +4001,18 @@ void ExtentMap::deleteOID(int OID)
...
-    while (it != end)
+    for (auto dbRoot: dbRootVec)
     {
-        if (it->second.fileID == OID)
+        const auto lbids = fPExtMapIndexImpl_->find(dbRoot, OID);
+        auto emIdents = getEmIteratorsByLbids(lbids);
+        OIDExists = (!emIdents.empty());
+        for (auto& emIt : emIdents)
         {
             OIDExists = true;
```
mariadb-DenisKhalikov: Seems like this assignment is not needed anymore.

versioning/BRM/extentmap.cpp
```diff
@@ -4924,30 +4871,19 @@ void ExtentMap::getExtentCount_dbroot(int OID, uint16_t dbroot, bool incOutOfSer
...
-        for (auto emIt = fExtentMapRBTree->begin(), end = fExtentMapRBTree->end(); emIt != end;
-             ++emIt)
+        const auto emIdents = getEmIdentsByLbids(lbids);
+        for (auto& emEntry : emIdents)
         {
-            const auto& emEntry = emIt->second;
-            if ((emEntry.fileID == OID) && (emEntry.dbRoot == dbroot) &&
-                (emEntry.status != EXTENTOUTOFSERVICE))
-                numExtents++;
+            numExtents = (emEntry.status != EXTENTOUTOFSERVICE) ? numExtents + 1 : numExtents;
```
mariadb-DenisKhalikov: Nit: seems like this part will be complilled to something like this
```
if ()
  numExtents = numExtents + 1
else
  numExtents = numExtents
```
So, we will have an assignment on each branch, would it be better have only one branch and one assignment
```
if ()
  numExtents =  numExtents + 1
```
At least I'm not aware about ternary assembly command for x64

mariadb-DenisKhalikov (approved):
LGTM, I've added some NITS, but they are not blocking this PR.

mariadb-LeonidFedorov (approved):

versioning/BRM/extentmap.cpp
drrtuy: Done.

versioning/BRM/extentmap.cpp
drrtuy: Agree.
Fixed.

PR #2374, opened 2022-05-16

mariadb-SergeyZefirov:


datatypes/mcs_decimal.cpp
```diff
@@ -606,27 +606,46 @@ std::string Decimal::toStringTSInt64() const
...
-  // TSInt64 Decimal
-  if (scale)
+  else if (scale) // TSInt64 Decimal
+  {
+    v = toStringTSInt64();
+  }
+  else
   {
-    return toStringTSInt64();
+    v = std::to_string(value);
```
drrtuy, line 645: I won't let this pass w/o an explanation.

PR #2380, opened 2022-05-20
MCOL-5044 Initial version of a fair thread pool

drrtuy:
PP now uses PriorityThreadPool that arbitrary picks another jobs pack
    to run. This scheduling discipline tend to run portions of a single query
    forcing other simultaneous queries to wait. In result parallel queries
    timings variance is high. The FairThreadPool picks the job with the smallest
    amount of work done so far(see the code for details)

### DIFF
--- a/utils/threadpool/fair_threadpool.cpp
+++ b/utils/threadpool/fair_threadpool.cpp
@@ -1,0 +1,302 @@
+/* Copyright (c) 2022 MariaDB Corporation
+
+   This program is free software; you can redistribute it and/or
+   modify it under the terms of the GNU General Public License
+   as published by the Free Software Foundation; version 2 of
+   the License.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software
+   Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+   MA 02110-1301, USA. */
+
+#include <stdexcept>
+#include <unistd.h>
+#include <exception>
+using namespace std;
+
+#include "messageobj.h"
+#include "messagelog.h"
+#include "threadnaming.h"
+using namespace logging;
+
+#include "fair_threadpool.h"
+using namespace boost;
+
+#include "dbcon/joblist/primitivemsg.h"
+
+namespace threadpool
+{
+FairThreadPool::FairThreadPool(uint targetWeightPerRun, uint highThreads, uint midThreads, uint lowThreads,
+                               uint ID)
+ : _stop(false), weightPerRun(targetWeightPerRun), id(ID), blockedThreads(0), extraThreads(0), stopExtra(true)
+{
+  boost::thread* newThread;
+  size_t numberOfThreads = highThreads + midThreads + lowThreads;
+  for (uint32_t i = 0; i < numberOfThreads; ++i)
+  {
+    newThread = threads.create_thread(ThreadHelper(this, PriorityThreadPool::Priority::HIGH));
+    newThread->detach();
+  }
+  cout << "FairThreadPool started " << numberOfThreads << " thread/-s.\n";
+  defaultThreadCounts = threadCounts = numberOfThreads;
+}
+
+FairThreadPool::~FairThreadPool()
+{
+  stop();
+}
+
+void FairThreadPool::addJob(const Job& job)
+{
+  addJob_(job);
+}
+
+void FairThreadPool::addJob_(const Job& job, bool useLock)
+{
+  boost::thread* newThread;
+  std::unique_lock<std::mutex> lk(mutex, std::defer_lock_t());
+
+  if (useLock)
+    lk.lock();
+
+  // Create any missing threads
+  if (defaultThreadCounts != threadCounts)
+  {
+    newThread = threads.create_thread(ThreadHelper(this, PriorityThreadPool::Priority::HIGH));
+    newThread->detach();
+    ++threadCounts;
+  }
+
+  // If some threads have blocked (because of output queue full)
+  // Temporarily add some extra worker threads to make up for the blocked threads.
+  if (blockedThreads > extraThreads)
+  {
+    stopExtra = false;
+    newThread = threads.create_thread(ThreadHelper(this, PriorityThreadPool::Priority::EXTRA));
+    newThread->detach();
+    extraThreads++;
+  }
+  else if (blockedThreads == 0)
+  {
+    // Release the temporary threads -- some threads have become unblocked.
+    stopExtra = true;
+  }
+
+  auto jobsListMapIter = txn2JobsListMap_.find(job.txnIdx_);
+  if (jobsListMapIter == txn2JobsListMap_.end())
+  {
+    ThreadPoolJobsList* jobsList = new ThreadPoolJobsList;
+    jobsList->push_back(job);
+    txn2JobsListMap_[job.txnIdx_] = jobsList;
+    WeightT currentTopWeight = weightedTxnsQueue_.empty() ? 0 : weightedTxnsQueue_.top().first;
+    weightedTxnsQueue_.push({currentTopWeight, job.txnIdx_});
+  }
+  else
+  {
+    jobsListMapIter->second->push_back(job);
+  }
+
+  if (useLock)
+    newJob.notify_one();
+}
+
+void FairThreadPool::removeJobs(uint32_t id)
+{
+  std::unique_lock<std::mutex> lk(mutex);
+
+  for (auto& txnJobsMapPair : txn2JobsListMap_)
+  {
+    ThreadPoolJobsList* txnJobsList = txnJobsMapPair.second;
+    auto job = txnJobsList->begin();
+    while (job != txnJobsList->end())
+    {
+      if (job->id_ == id)
+      {
+        job = txnJobsList->erase(job); // update the job iter
+        if (txnJobsList->empty())
+        {
+          txn2JobsListMap_.erase(txnJobsMapPair.first);
+          delete txnJobsList;
+          break;
+          // There is no clean-up for PQ. It will happen later in threadFcn
+        }
+        continue; // go-on skiping job iter increment
+      }
+      ++job;
+    }
+  }
+}
+
+void FairThreadPool::threadFcn(const PriorityThreadPool::Priority preferredQueue)
+{
+  if (preferredQueue == PriorityThreadPool::Priority::EXTRA)
+    utils::setThreadName("Extra");
+  else
+    utils::setThreadName("Idle");
+  RunListT runList; // This is a vector to allow to grab multiple jobs
+  RescheduleVecType reschedule;
+  bool running = false;
+  bool rescheduleJob = false;
+
+  try
+  {
+    while (!_stop)
+    {
+      runList.clear(); // remove the job
+      std::unique_lock<std::mutex> lk(mutex);
+
+      if (preferredQueue == PriorityThreadPool::Priority::EXTRA && stopExtra)
+      {
+        --extraThreads;
+        return;
+      }
+
+      if (weightedTxnsQueue_.empty())
+      {
+        newJob.wait(lk);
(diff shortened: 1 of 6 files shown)


utils/threadpool/fair_threadpool.cpp
```diff
@@ -0,0 +1,295 @@
...
+using namespace logging;
+
+#include "fair_threadpool.h"
+using namespace boost;
+
+#include "dbcon/joblist/primitivemsg.h"
+
+namespace threadpool
+{
+FairThreadPool::FairThreadPool(uint targetWeightPerRun, uint highThreads, uint midThreads, uint lowThreads,
```
mariadb-DenisKhalikov, line 35: Does `high`, `mid`, `low` mean priority of the threads?

utils/threadpool/fair_threadpool.cpp
```diff
@@ -0,0 +1,295 @@
...
+                               uint ID)
+ : _stop(false), weightPerRun(targetWeightPerRun), id(ID), blockedThreads(0), extraThreads(0), stopExtra(true)
+{
+  boost::thread* newThread;
+  for (uint32_t i = 0; i < highThreads + midThreads +lowThreads; ++i)
```
mariadb-DenisKhalikov: Nit: would it be better to compute a loop condition (` i < highThreads + midThreads +lowThreads` ) when we initialize a loop, instead of computing it every iteration:
from:
`for (uint32_t i = 0; i < highThreads + midThreads +lowThreads; ++i)`
to:
`for (uint32_t i = 0, e = highThreads + midThreads +lowThreads; i < e; ++i)`
This is completely optional i just took it from llvm coding standards: https://llvm.org/docs/CodingStandards.html#turn-predicate-loops-into-predicate-functions

utils/threadpool/fair_threadpool.cpp
```diff
@@ -0,0 +1,295 @@
...
+  }
+}
+
+void FairThreadPool::threadFcn(const Priority preferredQueue)
+{
+  if (preferredQueue == EXTRA)
+    utils::setThreadName("Extra");
+  else
+    utils::setThreadName("Idle");
+  RunListT runList; // This is a vector to allow to grab multiple jobs
```
mariadb-DenisKhalikov, line 142: Just wondering, why do we use a vector, since it's for a single job?

utils/threadpool/fair_threadpool.cpp
```diff
@@ -0,0 +1,295 @@
...
+
+      if (weightedTxnsQueue_.empty())
+      {
+        newJob.wait(lk); // might need a lock here
+        continue;
+      }
+
+      // We have non-empty jobsList at this point.
+      // Remove the txn from a queue first to add it later
+      weightedTxnsQueue_.pop(); // remove the empty
```
mariadb-DenisKhalikov: Can you please clarify the comment `remove the empty`, seems like at this point `joblist` for the `top` transaction is not empty.

utils/threadpool/fair_threadpool.cpp
```diff
@@ -0,0 +1,295 @@
...
+      ThreadPoolJobsList* jobsList = txnAndJobListPair->second;
+      // Job& job = jobsList->front();
+      runList.push_back(jobsList->front());
+
+      jobsList->pop_front();
+      // Add the jobList back into the PQ adding some weight to it
+      // Current algo doesn't reduce total txn weight if the job is rescheduled.
+      if (!jobsList->empty())
+      {
+        weightedTxnsQueue_.push({weightedTxn.first + runList[0].weight_, txnIdx});
```
mariadb-DenisKhalikov, line 207: Just for my education, why do we choose a new weight for transaction a sum of current weight + job weight?

utils/threadpool/fair_threadpool.cpp
```diff
@@ -0,0 +1,295 @@
...
+        continue; // just go on w/o re-taking the lock
+      }
+
+      WeightedTxnT weightedTxn = weightedTxnsQueue_.top();
+      auto txnAndJobListPair = txn2JobsListMap_.find(weightedTxn.second);
+      // Looking for non-empty jobsList in a loop
+      // Waiting on cond_var if PQ is empty(no jobs in this thread pool)
+      while (txnAndJobListPair == txn2JobsListMap_.end() || txnAndJobListPair->second->empty())
+      {
+        if (txnAndJobListPair != txn2JobsListMap_.end()) // JobList is empty
```
mariadb-DenisKhalikov: Can you please clarify, how does `joblist` could be an empty at this point.
When we remove a `job` by id, if the `joblist` is empty it also will be removed:
```
        job = txnJobsList->erase(job); // update the job iter
        if (txnJobsList->empty())
        {
          txn2JobsListMap_.erase(txnJobsMapPair.first);
          delete txnJobsList;
          break;
```

utils/threadpool/fair_threadpool.cpp
drrtuy: Yes, it does. However these priorities are legacy inherited to be compatible with PriorityThreadPool.

utils/threadpool/fair_threadpool.cpp
drrtuy: Good catch.

utils/threadpool/fair_threadpool.cpp
drrtuy: In my next patch I will benchmark FairThreadPool grabbing multiply jobs to reduce TP scheduling algo overhead so this jobs collection will be either a vector or a list.

utils/threadpool/fair_threadpool.cpp
drrtuy: The comment is misguiding so I will remove it.

utils/threadpool/fair_threadpool.cpp
drrtuy: This allows to reduce a priority of the txn that owns the morsel(PP primitive job in the context) comparing with other txns so that FairThreadPool scheduling will pick another txn's morsel to run. In other words every time when txn runs a morsel in FTP it looses its priority to be scheduled next. However this algo doesn't affect single-query performance but adds some(presumably neglectable) overhead.

mariadb-DenisKhalikov: Ok, thanks. In this case an important part is to calculate a weight for the specific job, the calculation algo is not covered in this part, as far as I see.

utils/threadpool/fair_threadpool.cpp
drrtuy: threadFcn doesn't remove an empty jobsList(look at the code after the line jobsList->pop_front();). The reason why it isn't removed is that the job taken can be re-scheduled. I might reconsider this when I start testing the TP.

utils/threadpool/fair_threadpool.cpp
```diff
@@ -0,0 +1,295 @@
...
+    stopExtra = true;
+  }
+
+  auto jobsListMapIter = txn2JobsListMap_.find(job.txnIdx_);
+  if (jobsListMapIter == txn2JobsListMap_.end())
+  {
+    ThreadPoolJobsList* jobsList = new ThreadPoolJobsList;
+    jobsList->push_back(job);
+    txn2JobsListMap_[job.txnIdx_] = jobsList;
+    WeightT currentTopWeight = weightedTxnsQueue_.empty() ? 0 : weightedTxnsQueue_.top().first;
```
mariadb-DenisKhalikov, line 97: So, the weight for the new transaction does not depend on the weight of the job for this transaction, every new transaction will be on top of the priority queue, is it correct?

utils/threadpool/fair_threadpool.cpp
mariadb-DenisKhalikov: Oh, I see, thanks.

utils/threadpool/fair_threadpool.cpp
drrtuy: Ohh yes, the morsel job weight/cost calculation is a huge topic. To streamline the testing I will ignore the fact that jobs have different costs however I mind this fact.
As a simple cost model I consider a sum of different PrimitiveJob parts, e.g. filter, join, aggregate taking a number of columns into account.
As I said the equal cost fair TP should demonstrate timings variance reduction for multiple parallel queries.

utils/threadpool/fair_threadpool.h
```diff
@@ -0,0 +1,173 @@
...
+
+  using TransactionIdxT = uint32_t;
+  struct Job
+  {
+    Job() : weight_(1), priority_(0), id_(0)
+    {
+    }
+    Job(const uint32_t uniqueID, const uint32_t stepID, const TransactionIdxT txnIdx,
+        const boost::shared_ptr<Functor>& functor, const primitiveprocessor::SP_UM_IOSOCK& sock,
+        const uint32_t weight = 0, const uint32_t priority = 0, const uint32_t id = 0)
```
mariadb-DenisKhalikov: Nit: default constructor specifies a weight for the job as 1, this constructor specifies a default weight as 0, should it be consistent?

utils/threadpool/fair_threadpool.h
```diff
@@ -0,0 +1,173 @@
...
+  bool _stop;
+  uint32_t weightPerRun;
+  volatile uint id;  // prevent it from being optimized out
+
+  using WeightT = uint32_t;
+  using WeightedTxnT = std::pair<WeightT, TransactionIdxT>;
+  using WeightedTxnVec = std::vector<WeightedTxnT>;
+  struct PrioQueueCmp
+  {
+    bool operator()(WeightedTxnT lhs, WeightedTxnT rhs)
```
mariadb-DenisKhalikov, line 154: I was thinking that the transaction with lowest weight should be on top of queue, but this comparator (greater) makes a highest priority for the transaction with highest weight? Please fix me if I'm wrong.

drrtuy: Good point but [PrioQueue](https://en.cppreference.com/w/cpp/container/priority_queue) comparator works the other way around.

utils/threadpool/fair_threadpool.h
drrtuy: Good catch. Fixed.

utils/threadpool/fair_threadpool.h
mariadb-DenisKhalikov: I see, thanks!

utils/threadpool/fair_threadpool.cpp
```diff
@@ -0,0 +1,297 @@
...
+    // Release the temporary threads -- some threads have become unblocked.
+    stopExtra = true;
+  }
+
+  auto jobsListMapIter = txn2JobsListMap_.find(job.txnIdx_);
+  if (jobsListMapIter == txn2JobsListMap_.end())
+  {
+    ThreadPoolJobsList* jobsList = new ThreadPoolJobsList;
+    jobsList->push_back(job);
+    txn2JobsListMap_[job.txnIdx_] = jobsList;
```
mariadb-AlexeyAntipovsky, line 96: Shouldn't this call be protected with a mutex since insertion can lead to rehashing and invalidation of all iterators (for example, in a simultaneous removeJobs call)?

mariadb-AlexeyAntipovsky (changes requested):

utils/threadpool/fair_threadpool.cpp
drrtuy: The is a common mutex that protects all attributes of the class instance. The lock is taken at line 58 or outside this function.

mariadb-AlexeyAntipovsky: Sorry, I missed this moment. Then maybe it makes sense to make `addJob(Job, bool)` a private member, and add a public `addJob(Job)` as a proxy to `addJob(Job, true)` for more foolproof?

mariadb-DenisKhalikov (approved):

utils/threadpool/fair_threadpool.cpp
drrtuy: I personally disagree this is an important part for the initial version.

utils/threadpool/fair_threadpool.cpp
drrtuy: It will be on top only for the first morsel after that its weight  increases.

utils/threadpool/fair_threadpool.cpp
drrtuy: It makes sense to me. Done.

mariadb-AlexeyAntipovsky (approved):

PR #2389, opened 2022-05-27
MCOL-4912 This patch adds an additional check populating EM and incre…

drrtuy:
…ases EM increment step size

### DIFF
--- a/versioning/BRM/extentmap.h
+++ b/versioning/BRM/extentmap.h
@@ -1048,11 +1048,10 @@ private:
   static const constexpr size_t EM_FREELIST_INITIAL_SIZE = 50 * sizeof(InlineLBIDRange);
   static const constexpr size_t EM_FREELIST_INCREMENT = 50 * sizeof(InlineLBIDRange);
   // RBTree constants.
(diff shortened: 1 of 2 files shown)


denis0x0D (approved):

versioning/BRM/extentmap.h
```diff
@@ -1048,11 +1048,10 @@ class ExtentMap : public Undoable
   static const constexpr size_t EM_FREELIST_INITIAL_SIZE = 50 * sizeof(InlineLBIDRange);
   static const constexpr size_t EM_FREELIST_INCREMENT = 50 * sizeof(InlineLBIDRange);
   // RBTree constants.
-  static const size_t EM_RB_TREE_NODE_SIZE = 2 * (sizeof(EMEntry) + 8 * sizeof(uint64_t));
+  static const size_t EM_RB_TREE_NODE_SIZE = sizeof(EMEntry) + 8 * sizeof(uint64_t);
```
denis0x0D, line 1051: The node size was taken from the node size on my machine, I think to keep it safe we could add some padding, not sure it's needed currently, but how knows.

drrtuy:
This "2 *" ends up allocating 2x shmem on startup. Here is an example, 285MB EM results in 510MB EM shmem file if used w/o '2 *' and 1000MB if used with '2 *'.

PR #2401, opened 2022-05-31
StatisticsManager initialize all plugins.

denis0x0D:
This patch adds support for initializing all plugins in the system.


denis0x0D:
Statistics collection will work with S3.

drrtuy (approved):

PR #2406, opened 2022-06-06
MCOL-5021 AUX column implementation to improve DELETE performance.

tntnatbry:


tntnatbry:
This is a WIP PR, please don't merge.

tntnatbry:
This PR is now ready for review.

There is a corresponding PR in the regression repo which should be merged alongside this PR: https://github.com/mariadb-corporation/mariadb-columnstore-regression-test/pull/302

dbcon/dmlpackageproc/commandpackageprocessor.cpp
```diff
@@ -374,6 +374,9 @@ DMLPackageProcessor::DMLResult CommandPackageProcessor::processPackage(
 
           if (weRc == 0)
           {
+            // MCOL-5021
```
drrtuy, line 377: Do you recall by any chance @tntnatbry what it is for?

dbcon/execplan/calpontsystemcatalog.cpp
```diff
@@ -3546,13 +3554,13 @@ const CalpontSystemCatalog::ROPair CalpontSystemCatalog::tableRID(const TableNam
...
       << aTableName.table << "' --tableRID/";
 
-  csep.data(oss.str());  //@bug 6078. Log the statement
-
   if (fIdentity == EC)
     oss << "EC";
   else
     oss << "FE";
 
+  csep.data(oss.str());  //@bug 6078. Log the statement
```
drrtuy, line 3562: It might change the output. Did it break any tests?

oam/etc/Columnstore.xml
```diff
@@ -137,6 +137,7 @@
 		<BulkRollbackDir>/var/lib/columnstore/data1/systemFiles/bulkRollback</BulkRollbackDir>
 		<MaxFileSystemDiskUsagePct>98</MaxFileSystemDiskUsagePct>
 		<CompressedPaddingBlocks>1</CompressedPaddingBlocks> <!-- Number of blocks used to pad compressed chunks -->
+        <FastDelete>false</FastDelete>
```
drrtuy: JFYI Some time ago there was an issue updating in deb-based distros. Basically dpkg thinks the Columnstore.xml was manually changed(even if it didn't) thus adding new options into the config forces automatic(that is w/o human intervention)dpkg upgrades to fail waiting for a confirmation. We can take the risk and see if it fail or not.

primitives/primproc/columncommand.cpp
```diff
@@ -202,6 +202,20 @@ void ColumnCommand::_loadData()
   bpp->cachedIO += wasCached;
   bpp->physIO += blocksRead;
   bpp->touchedBlocks += blocksToLoad;
+
+  if (_hasAuxCol)
```
drrtuy: JFYI The naming conventions adds underscore as a suffix not as a prefix.

primitives/primproc/primitiveserver.cpp
```diff
@@ -1147,6 +1147,7 @@ int DictScanJob::operator()()
     {
       loadBlock(cmd->LBID, verInfo, cmd->Hdr.TransactionID, cmd->CompType, data, &wasBlockInCache,
                 &blocksRead, fLBIDTraceOn, session);
+      // TODO MCOL-5021
```
drrtuy: DictScan always has a ColumnCommand in front for its token column of it so you don't need to add support here.

oam/etc/Columnstore.xml
davidhall586: I thought we had resolved that issue

dbcon/execplan/calpontsystemcatalog.cpp
tntnatbry: No this change did not break any tests in MTR or test001.

primitives/primproc/primitiveserver.cpp
tntnatbry: This makes sense. I have removed this TODO.

oam/etc/Columnstore.xml
tntnatbry: I see. I was not aware of this issue of package install failing if we add a new option to Columnstore.xml. I can remove this option from the .xml. In which case, the customer will have to add this option back in on all nodes in the cluster and restart, if they want to enable `FastDelete`. By default, if this option is not present in the .xml file, it will be treated as false.

primitives/primproc/columncommand.cpp
tntnatbry: I used the existing naming convention for the other member variables in this file, such as `_isScan`.

dbcon/ddlpackageproc/createtableprocessor.cpp
```diff
@@ -267,12 +267,17 @@ CreateTableProcessor::DDLResult CreateTableProcessor::processPackage(
...
-                                                 1);  // include column, oids,dictionary oids and tableoid
+    // Include column oids, dictionary oids, tableoid, and
+    // also include AUX oid as of MCOL-5021
+    fStartingColOID = fObjectIDManager.allocOIDs(numColumns + numDictCols + 2);
 #ifdef IDB_DDL_DEBUG
     cout << fTxnid.id << " Create table allocOIDs got the starting oid " << fStartingColOID << endl;
 #endif
 
+    uint32_t size = numColumns + numDictCols;
+    idbassert(size > 0);
```
drrtuy: You think it is really possible?

dbcon/ddlpackageproc/createtableprocessor.cpp
```diff
@@ -267,12 +267,17 @@ CreateTableProcessor::DDLResult CreateTableProcessor::processPackage(
...
-    fStartingColOID = fObjectIDManager.allocOIDs(numColumns + numDictCols +
-                                                 1);  // include column, oids,dictionary oids and tableoid
+    // Include column oids, dictionary oids, tableoid, and
+    // also include AUX oid as of MCOL-5021
+    fStartingColOID = fObjectIDManager.allocOIDs(numColumns + numDictCols + 2);
 #ifdef IDB_DDL_DEBUG
     cout << fTxnid.id << " Create table allocOIDs got the starting oid " << fStartingColOID << endl;
 #endif
 
+    uint32_t size = numColumns + numDictCols;
```
drrtuy: IMHO the name size is too generic and it doesn't convey the fact that this is a number of columns involved.

dbcon/execplan/calpontsystemcatalog.h
```diff
@@ -908,6 +927,15 @@ class CalpontSystemCatalog : public datatypes::SystemCatalog
   static uint32_t fModuleID;
 };
 
+// MCOL-5021
+const datatypes::SystemCatalog::ColDataType AUX_COL_DATATYPE = datatypes::SystemCatalog::UTINYINT;
+const int32_t AUX_COL_WIDTH = 1;
+// TODO MCOL-5021 compressionType is hardcoded to 2 (SNAPPY)
+const CalpontSystemCatalog::CompressionType AUX_COL_COMPRESSION_TYPE = CalpontSystemCatalog::COMPRESSION2;
```
denis0x0D, line 934: Can you please clarify on "compression type"? You create an aux column in system cat with "compression type = 0", that means this column file is not compressed, why do you put compression type to 2 here?

dbcon/dmlpackageproc/commandpackageprocessor.cpp
```diff
@@ -374,6 +374,9 @@ DMLPackageProcessor::DMLResult CommandPackageProcessor::processPackage(
...
+            fDbrm->addToLBIDList(fSessionID, lbidList);
```
drrtuy, line 378: This is ad-hoc addition IMHO. Why does lbidList have no LBID's related with the fSessionID added in line 126, namely: fDbrm->getUncommittedExtentLBIDs()? How does it happen that aux column relates to fSessionID and not to txnid.id used in the line 126?

dbcon/joblist/tuple-bps.cpp
```diff
@@ -528,6 +529,63 @@ TupleBPS::TupleBPS(const pColScanStep& rhs, const JobInfo& jobInfo) : BatchPrimi
...
+    if (fOidAux > 3000)
+    {
+      hasAuxCol = true;
+
+      if (dbrm.getExtents(fOidAux, extentsAux))
+        throw runtime_error("TupleBPS::TupleBPS BRM extent lookup failure (1)");
+
+      sort(extentsAux.begin(), extentsAux.end(), BRM::ExtentSorter());
+
+      extentsMap[fOidAux] = tr1::unordered_map<int64_t, EMEntry>();
```
denis0x0D: I see the same initializing of unordered_map below as well, but what the idea of this, according to the spec
https://en.cppreference.com/w/cpp/container/unordered_map/operator_at:
`Returns a reference to the value that is mapped to a key equivalent to key, performing an insertion if such key does not already exist.`

dbcon/dmlpackageproc/commandpackageprocessor.cpp
```diff
@@ -433,6 +436,12 @@ DMLPackageProcessor::DMLResult CommandPackageProcessor::processPackage(
 
         if (!cpInvalidated)
         {
+          // MCOL-5021
+          if (stmt == "ROLLBACK")
+          {
+            fDbrm->addToLBIDList(fSessionID, lbidList);
```
drrtuy, line 442: See the prev question plz.

dbcon/execplan/calpontsystemcatalog.cpp
```diff
@@ -3607,6 +3615,237 @@ const CalpontSystemCatalog::ROPair CalpontSystemCatalog::tableRID(const TableNam
...
+// AUX column for the table
+CalpontSystemCatalog::OID CalpontSystemCatalog::tableAUXColumnOID(const TableName& tableName,
+                                                                  int lower_case_table_names)
+{
+  TableName aTableName;
+  aTableName.schema = tableName.schema;
+  aTableName.table = tableName.table;
+  if (lower_case_table_names)
+  {
+    boost::algorithm::to_lower(aTableName.schema);
```
drrtuy, line 3629: IIRC all boost::to_lower() invocations share the same mutex when they access system's locale. What is inside aTableName.schema or aTableName.table when non-ASCII is used? Maybe we can just use ASCII only tolower converter here?

dbcon/execplan/calpontsystemcatalog.cpp
```diff
@@ -3607,6 +3615,237 @@ const CalpontSystemCatalog::ROPair CalpontSystemCatalog::tableRID(const TableNam
...
+  lk1.unlock();
+
+  // select auxcolumnoid from systable where schema = tableName.schema and tablename = tableName.table;
+  CalpontSelectExecutionPlan csep;
+  CalpontSelectExecutionPlan::ReturnedColumnList returnedColumnList;
+  CalpontSelectExecutionPlan::FilterTokenList filterTokenList;
+  CalpontSelectExecutionPlan::ColumnMap colMap;
+
+  SimpleColumn* c1 =
+    new SimpleColumn(CALPONT_SCHEMA + "." + SYSTABLE_TABLE + "." + AUXCOLUMNOID_COL, fSessionID);
```
drrtuy: Not a big deal but we can have a constexpr string saving 12 string additions in the CALPONT_SCHEMA + "." + SYSTABLE_TABLE + "." expression.

drrtuy: Or even replace all those expressions with a static constexpr string.

versioning/BRM/dbrm.cpp
```diff
@@ -4453,6 +4454,97 @@ void DBRM::deleteAISequence(uint32_t OID)
...
+      std::vector<struct BRM::EMEntry>>> extentMap;
+
+  int err = 0;
+
+  std::unordered_set<LBID_t> lbidSet;
+  std::unordered_set<execplan::CalpontSystemCatalog::OID> tableOidSet;
+
+  for (const auto i : lbidList)
+  {
+    lbidSet.insert(i);
```
denis0x0D, line 4474: I do not really get the idea of adding the processing `lbid` from `lbidList`to `lbidSet`, it seems like you can process that list and adding `new lbid` to the set and than append that `lbidSet` to `lbidList` without clearing it at line 4537. Do I miss somethin? Also do not get why not to use vector for the `lbidSet`.

writeengine/bulk/we_bulkload.cpp
```diff
@@ -314,6 +314,84 @@ int BulkLoad::loadJobInfo(const string& fullName, bool bUseTempJobFile, int argc
...
       return rc;
     }
+
+    // MCOL-5021
+    execplan::CalpontSystemCatalog::OID tableAUXColOid;
+    std::string tblName;
+    std::string curTblName = curJob.jobTableList[i].tblName;
+    string::size_type startName = curTblName.rfind('.');
+
+    if (startName == std::string::npos)
```
denis0x0D, line 326: Could you please add comment what is happening in the next few lines?

writeengine/wrapper/we_colop.cpp
```diff
@@ -1791,6 +1800,59 @@ int ColumnOp::writeRows(Column& curCol, uint64_t totalRow, const RIDList& ridLis
...
+    // Read the old value of the record
+    if (oldValArray)
+    {
+      uint8_t* p = static_cast<uint8_t*>(oldValArray);
+      memcpy(p + i * curCol.colWidth, dataBuf + dataBio, curCol.colWidth);
+    }
+
+    i++;
+
+    if (i >= totalRow)
```
denis0x0D: Do not get how `i` could be greater than `totalRow`, you increment `i` at the line 1847, the `totalRow` has const semantics in this function.

dbcon/joblist/columncommand-jl.cpp
```diff
@@ -247,11 +261,26 @@ void ColumnCommandJL::setLBID(uint64_t rid, uint32_t dbRoot)
...
+  {
+    if (extentsAux[j].dbRoot == dbRoot && extentsAux[j].partitionNum == partNum &&
+        extentsAux[j].segmentNum == segNum && extentsAux[j].blockOffset == (extentNum * 1 * 1024))
+    {
+      lbidAux = extentsAux[j].range.start + (blockNum * 1);
+      break;
+    }
+  }
+
+  if (i == extents.size() || (hasAuxCol && j == extentsAux.size()))
```
drrtuy: I would separate these conditions. This will simplify tshooting.

oam/etc/Columnstore.xml
drrtuy: Let it be here and see what the server's CI tells us.

versioning/BRM/dbrm.cpp
```diff
@@ -4453,6 +4454,97 @@ void DBRM::deleteAISequence(uint32_t OID)
   }
 }
 
+void DBRM::addToLBIDList(uint32_t sessionID, vector<LBID_t>& lbidList)
```
drrtuy, line 4458: To be honest I didn't get the semantics of addToLBIDList(). Why do you expect lbidList to be different after rollBackTransaction() call in CommandPackageProcessor::processPackage(). 
My concern is that with a relatively huge EM addToLBIDList will become another perf killer for DML  IMHO b/c of multiple lookupLocal() and getExtents()(both have O(N) complexity where N is the EM size). Let's discuss why do we need this method im DBRM and how we can make it better.

primitives/linux-port/column.cpp
```diff
@@ -895,6 +976,75 @@ inline bool nextColValue(
   return true;
 }
 
+template <typename T, int COL_WIDTH>
```
drrtuy: Could you elaborate why did you make this nextColValueAux that is a very similar copy of nextColValue?

primitives/linux-port/column.cpp
```diff
@@ -1420,67 +1576,181 @@ void vectorizedFiltering(NewColRequestHeader* in, ColResultHeader* out, const T*
     }
   }
 
-  // main loop
-  // writeMask tells which values must get into the result. Includes values that matches filters. Can have
-  // NULLs. nonEmptyMask tells which vector coords are not EMPTY magics. nonNullMask tells which vector coords
-  // are not NULL magics.
-  for (uint16_t i = 0; i < iterNumber; ++i)
+  if (hasAuxCol)
```
drrtuy: Could you elaborate on why did you decide to make a very close copy of the main loop for the scope when hasAuxCol is true?

primitives/linux-port/column.cpp
```diff
@@ -1609,6 +1881,7 @@ void filterColumnData(NewColRequestHeader* in, ColResultHeader* out, uint16_t* r
   // Bit patterns in srcArray[i] representing EMPTY and NULL values
   T emptyValue = getEmptyValue<T>(dataType);
   T nullValue = getNullValue<T>(dataType);
+  uint8_t emptyValueAux = getEmptyValue<uint8_t>(execplan::AUX_COL_DATATYPE);
```
drrtuy, line 1911: We should use a decltype here, e.g. getEmptyValue<decltype(execplan::someAuxVarPlaceHolderVar)>(execplan::AUX_COL_DATATYPE) so that we don't go over the place if AUX Column will become uint16.

primitives/linux-port/column.cpp
```diff
@@ -1694,10 +1975,10 @@ void PrimitiveProcessor::scanAndFilterTypeDispatcher(NewColRequestHeader* in, Co
     const uint16_t ridSize = in->NVALS;
     uint16_t* ridArray = in->getRIDArrayPtr(W);
     const uint32_t itemsPerBlock = logicalBlockMode ? BLOCK_SIZE : BLOCK_SIZE / W;
-    filterColumnData<T, KIND_FLOAT>(in, out, ridArray, ridSize, block, itemsPerBlock, parsedColumnFilter);
+    filterColumnData<T, KIND_FLOAT>(in, out, ridArray, ridSize, block, itemsPerBlock, parsedColumnFilter, hasAuxCol, blockAux);
     return;
   }
-  _scanAndFilterTypeDispatcher<T>(in, out);
+  _scanAndFilterTypeDispatcher<T>(in, out, hasAuxCol);
```
drrtuy: What is the semantics of hasAuxCol exactly? If it is true when we process the aux column for a table or it is true for ColumnCommand for any column of that table?

drrtuy: Could you put this variable into in message plz? It just doesn't belong here: there are in and out messages and out of a sudden hasAuxCol. I mean it is important for sure but just doesn't look appropriate IMHO.

primitives/primproc/batchprimitiveprocessor.cpp
```diff
@@ -599,8 +601,8 @@ void BatchPrimitiveProcessor::resetBPP(ByteStream& bs, const SP_UM_MUTEX& w, con
 
   /* init vars not part of the BS */
   currentBlockOffset = 0;
-  memset(relLBID.get(), 0, sizeof(uint64_t) * (projectCount + 1));
-  memset(asyncLoaded.get(), 0, sizeof(bool) * (projectCount + 1));
+  memset(relLBID.get(), 0, sizeof(uint64_t) * (projectCount + 2));
```
drrtuy, line 605: What if we set this constant somewhere to be available globally b/c there are some places where you add this value in EM code also? So in future we don't need to search for these magic constants that takes some aspects into account and increase the size of the columns expected but change it in a single place.

primitives/primproc/columncommand.cpp
drrtuy: We have three different existing class members naming styles(or even 4) and we also decided to go with the one whatsoever. I understand that it might contradict with the style in this particular file but if we follow the style that exists(the three of them) we never reach a single style, right?

primitives/linux-port/column.cpp
```diff
@@ -1420,67 +1576,181 @@ void vectorizedFiltering(NewColRequestHeader* in, ColResultHeader* out, const T*
...
+      ridArray += VECTOR_SIZE_AUX;
+    }
+
+    ridArray = origRidArray;
+
+    MT (*getNonEmptyMaskPtr)(MT*, uint16_t);
+
+    switch(WIDTH)
+    {
+      case 1:
```
drrtuy: We shoud try do this w/o switch but using a mapping struct, e.g. MT (*getNonEmptyMaskPtr)(MT*, uint16_t) getNonEmptyMaskPtr = getMask<WIDTH>().
template<int W>
consteval MT (*getMask)(MT*, uint16_t)() { return nullptr};
template<>
consteval MT (*getMask)(MT*, uint16_t)()<1> { return getNonEmptyMask1Byte; };
etc

dbcon/dmlpackageproc/commandpackageprocessor.cpp
tntnatbry: No this is not an ad-hoc change. The DELETE code has changed now and we only perform version buffering on the aux column and not on the actual column. Due to this, `fDbrm->getUncommittedExtentLBIDs()` only returns the lbids for the aux column. Consider the following scenario:

```sql
create table (a int)engine=columnstore;
insert into t1 values (1), (2), (3);
select a from t1; -- a's extent min/max will be 1/3 respectively.
set autocommit=off;
delete from t1 where a in (1, 3);
select a from t1; -- a's extent min/max will be set to 2/2 now.
rollback;
select a from t1; -- this returns the correct 3 rows but the extent min/max for column a is incorrectly set to 2/2.
```

Without the above `fDbrm->addToLBIDList()` call, after the rollback, the extent min/max value for column a will not be invalidated and will still have the values 2 and 2 both for the min and the max, even though the column values are (1, 2, 3). This will cause the queries such as this one returning incorrect results after the rollback: `select a from t1 where a < 2;`

This was a test001 regression which is fixed with the `addToLBIDList()` change.

`fSessionID` passed in the call to `fDbrm->addToLBIDList()` is merely to get the singleton syscat object:

```
void DBRM::addToLBIDList(uint32_t sessionID, vector<LBID_t>& lbidList)                                                                       
{                                                                                                                                            
  boost::shared_ptr<execplan::CalpontSystemCatalog> systemCatalogPtr =                                                                       
    execplan::CalpontSystemCatalog::makeCalpontSystemCatalog(sessionID);                                                                     
```

primitives/primproc/batchprimitiveprocessor.cpp
```diff
@@ -1467,6 +1471,19 @@ void BatchPrimitiveProcessor::execute()
           asyncLoaded[p] = true;
         }
 
+        if (col->hasAuxCol())
```
drrtuy, line 1475: I am curios why did we add this loadBLockAsync if filterCount == 1 && hasScan? I got the impression that aux column must be loaded in any case.

primitives/primproc/columncommand.h
```diff
@@ -163,6 +173,8 @@ class ColumnCommand : public Command
   uint32_t baseMsgLength;
 
   uint64_t lbid;
+  bool _hasAuxCol;
```
drrtuy: Naming style calls for an underscore for the class members.

writeengine/server/we_dmlcommandproc.h
```diff
@@ -78,8 +78,8 @@ class WE_DMLCommandProc
   EXPORT uint8_t rollbackVersion(messageqcpp::ByteStream& bs, std::string& err);
   EXPORT uint8_t processBatchInsert(messageqcpp::ByteStream& bs, std::string& err,
                                     ByteStream::quadbyte& PMId);
-  EXPORT uint8_t processBatchInsertBinary(messageqcpp::ByteStream& bs, std::string& err,
-                                          ByteStream::quadbyte& PMId);
+  //EXPORT uint8_t processBatchInsertBinary(messageqcpp::ByteStream& bs, std::string& err,
```
drrtuy: We need this? If not plz remove.

writeengine/server/we_readthread.cpp
```diff
@@ -156,11 +156,13 @@ void DmlReadThread::operator()()
           break;
         }
 
+#if 0
```
drrtuy: This should be a remnant of bulk insertion API.

versioning/BRM/dbrm.cpp
tntnatbry: Refer to my reply in dbcon/dmlpackageproc/commandpackageprocessor.cpp where I explain why we need `addToLBIDList()`. Please note that this function is only invoked in case of a rollback so I am a bit less concerned about the performance implications, but sure let's discuss if there are ways to make it more efficient.

writeengine/server/we_readthread.cpp
tntnatbry: I did not find any references to `WE_SVR_BATCH_INSERT_BINARY` in the engine code, so I commented this out. Is the bulk insertion API or any external library still dependent on this code? If not, I can remove this.

dbcon/execplan/calpontsystemcatalog.cpp
tntnatbry: I am not sure if using the ASCII only `tolower` would be correct if `aTableName.schema` and `aTableName.table` contain non-ASCII strings. @dhall-MariaDB can probably answer why we use `boost::algorithm::to_lower()` here for converting to lower case.

dbcon/execplan/calpontsystemcatalog.h
tntnatbry: This compression type (which is set to 2) is for the AUX column in an actual user table, such as `test.t1`. While the `auxcolumnoid` is a column in the `calpontsys.systable` syscat table, where all columns are uncompressed.

versioning/BRM/dbrm.cpp
```diff
@@ -4453,6 +4454,97 @@ void DBRM::deleteAISequence(uint32_t OID)
...
+          uint32_t extentNum = extent.blockOffset / (extent.range.size * 1024);
+
+          if (dbRoot == extent.dbRoot && partitionNum == extent.partitionNum &&
+              segmentNum == extent.segmentNum && extentNumAux == extentNum)
+          {
+            lbidSet.insert(extent.range.start);
+            break;
+          }
+        }
+      }
```
tntnatbry, line 4534: @denis0x0D This nested for loop above where we add `extent.range.start` to `lbidSet` can potentially already exist in `lbidList`. So if I just create the `lbidSet` only for new lbids and then append those lbids to `lbidList`, there is a possibility of `lbidList`containing duplicate values.

dbcon/execplan/calpontsystemcatalog.h
denis0x0D: Ok, thanks, can you please clarify how much new columns your
 patch is adding by default, because I see in JIRA only one:
"This will be accomplished by implementing a hidden 1-byte boolean column with 0 indicating the column value is disabled (deleted) and 1 indicating the column value is enabled (active)."

versioning/BRM/dbrm.cpp
denis0x0D: ok, thanks got it, clear now.

dbcon/execplan/calpontsystemcatalog.h
tntnatbry: We are creating 1 AUX (hidden) column per user table.

denis0x0D: Ok, thanks, can you please share a link to the design doc (white paper) for this feature, I cannot find it in a Jira ticket, may be I'm missing something, but I only can see a "benchmarking file".

tntnatbry: I do not have a design doc for this feature yet, but I can write one.

denis0x0D: Yes, please, it would be nice to have one, because I need to update a "rebuildEM tool" based on your feature, but I have not been involved in design discussion and it's hard to understand a design based on just a code.

primitives/primproc/batchprimitiveprocessor.cpp
tntnatbry: AUX column is tightly coupled with the original scan column. So I am guarding it under the same condition under which we perform `loadBlockAsync` on the scan column, which is `if (filterCount == 1 && hasScan)`. There is a comment at the start of this `if` block: "`// if only one scan step which has no predicate, async load all columns`". I am not fully sure the reasoning behind this comment.

primitives/primproc/batchprimitiveprocessor.cpp
tntnatbry: I can't recall if I added/incremented any constant in the EM code for the AUX column. In the EM code, in addition to passing the lbid for the scan column, we now also pass the lbid for the AUX column.

primitives/linux-port/column.cpp
tntnatbry: This is to avoid the `hasAuxCol` based if branching inside the for loop in `scalarFiltering()`:

```c++
  auto nextColValuePtr = hasAuxCol ? nextColValueAux<T, WIDTH> : nextColValue<T, WIDTH>;                                                    
  
  // Loop over the column values, storing those matching the filter, and updating the min..max range                                        
  for (uint32_t i = initialRID; (*nextColValuePtr)(curValue, &isEmpty, &i, &rid, srcArray, srcSize,                                         
                                                   ridArray, ridSize, outputType, emptyValue,                                               
                                                   blockAux, emptyValueAux);)                                                               
  {
    // loop code
  }
```

dbcon/execplan/calpontsystemcatalog.cpp
drrtuy: I think the answer would be "b/c it was here since day 0".

dbcon/dmlpackageproc/commandpackageprocessor.cpp
drrtuy: Thx for the explanation. We need to discuss DBRM::addToLBIDList() calculation complexity though.

dbcon/execplan/calpontsystemcatalog.cpp
drrtuy: There is no way locale can affect aTableName.schema so the common mutex all to_lower() invocations share is a needless overhead. I don't have a workaround except writing our own converter that processes only upper-case ASCII alphabet symbols with a mapping skipping all other bytes.

writeengine/bulk/we_bulkload.cpp
drrtuy: @tntnatbry I am not sure if curJob.jobTableList[i].tblName supports the real FQN or not but FQN might look like database.schema.table structure having 2 '.'

writeengine/bulk/we_bulkload.cpp
```diff
@@ -314,6 +314,86 @@ int BulkLoad::loadJobInfo(const string& fullName, bool bUseTempJobFile, int argc
...
+      fLog.logMsg(oss.str(), rc, MSGLVL_ERROR);
+      return rc;
+    }
+
+    // MCOL-5021 Valid AUX column OID for a table is > 3000
+    // Tables that were created before this feature was added will have
+    // tableAUXColOid = 0
+    if (tableAUXColOid > 3000)
+    {
+      JobColumn curColumn;
```
drrtuy: I would add a ctor b/c IMHO it is an error prone approach to assign class instance attributes outside the scope of a setter or ctor.

writeengine/server/we_readthread.cpp
drrtuy: It does depend and I am not sure if we are going to revive the bulk load API or not so better leave it as it is IMHO.

writeengine/shared/we_config.cpp
```diff
@@ -185,6 +186,17 @@ void Config::checkReload()
     m_BulkRollbackDir += "/rollback";
   }
 
+  const std::string fastDeleteTemp = cf->getConfig("WriteEngine", "FastDelete");
+
+  if (fastDeleteTemp.length() != 0 && boost::iequals(fastDeleteTemp, "true"))
```
drrtuy: JFYI iequals also acquires the same mutex accessing locale and it is wasteful to use it checking for an ASCII value. BTW you can replace the true with a single letter 'y'.

writeengine/bulk/we_bulkload.cpp
```diff
@@ -265,7 +265,7 @@ int BulkLoad::loadJobInfo(const string& fullName, bool bUseTempJobFile, int argc
     return rc;
   }
 
-  const Job& curJob = fJobInfo.getJob();
+  Job& curJob = const_cast<Job&>(fJobInfo.getJob());
```
drrtuy: IMHO it is better to use a separate getter then using a const_cast that looks like a side effect hard to search for.

writeengine/server/we_dmlcommandproc.cpp
```diff
@@ -411,6 +440,27 @@ uint8_t WE_DMLCommandProc::processSingleInsert(messageqcpp::ByteStream& bs, std:
             dicStringList.push_back(dicStrings);
           }
 
+          // MCOL-5021
+          if ((i == numcols - 1) && (tableAUXColOid > 3000))
```
drrtuy, line 465: There are two identical blocks. IMHO it would be good to deduplicate them.

writeengine/wrapper/we_colop.cpp
```diff
@@ -1406,15 +1406,23 @@ bool ColumnOp::isValid(Column& column) const
...
  ***********************************************************/
 // @bug 5572 - HDFS usage: add *.tmp file backup flag
-int ColumnOp::openColumnFile(Column& column, std::string& segFile, bool useTmpSuffix, int ioBuffSize) const
+int ColumnOp::openColumnFile(Column& column, std::string& segFile, bool useTmpSuffix,
+                             int ioBuffSize, bool isReadOnly) const
 {
   if (!isValid(column))
     return ERR_INVALID_PARAM;
 
+  std::string mode;
```
drrtuy: What if do it like this:
std::string mode{"r"};
if (!isReadOnly)
{
  mode = "r+b";
}

primitives/linux-port/column.cpp
drrtuy: You can do it a bit shorter like [this](https://godbolt.org/z/K8xW19q89). The example isn't complete but consider if constexpr usage.

primitives/linux-port/column.cpp
```diff
@@ -1420,67 +1614,162 @@ void vectorizedFiltering(NewColRequestHeader* in, ColResultHeader* out, const T*
     }
   }
 
-  // main loop
-  // writeMask tells which values must get into the result. Includes values that matches filters. Can have
-  // NULLs. nonEmptyMask tells which vector coords are not EMPTY magics. nonNullMask tells which vector coords
-  // are not NULL magics.
-  for (uint16_t i = 0; i < iterNumber; ++i)
+  if (in->hasAuxCol)
```
drrtuy: Mb you will find some inspiration from the goldbot example I shared previously on how to make this part better.

drrtuy (changes requested):

primitives/linux-port/column.cpp
tntnatbry: Function calls inside `filterColumnData` such as `vectorizedFilteringDispatcher` and `scalarFiltering` will still have the `emptyValueAux` argument defined as an `uint8_t`. I am not sure how we can fix that.

writeengine/bulk/we_bulkload.cpp
tntnatbry: @drrtuy I don't think any of the CS code assumes the FQN could be `database.schema.table`. I am reusing the same code here as in `XMLJob::fillInXMLDataAsLoaded`.

primitives/primproc/columncommand.cpp
tntnatbry: Done. I have changed this member var with underscore as a suffix instead of a prefix.

dbcon/ddlpackageproc/createtableprocessor.cpp
tntnatbry: Not possible, I have removed this assert, as well as renamed `size` to something more meaningful.

dbcon/execplan/calpontsystemcatalog.cpp
tntnatbry: Good point, I have replaced these additions with a const string.

dbcon/joblist/tuple-bps.cpp
tntnatbry: Thanks for pointing this out. Yes this line doesn't make sense so I have removed it.

writeengine/wrapper/we_colop.cpp
tntnatbry: Good catch, the boolean var to check the loop condition was not required, so I have removed it.

dbcon/joblist/columncommand-jl.cpp
tntnatbry: Done.

primitives/linux-port/column.cpp
tntnatbry: Thanks for showing the usage of `if constexpr`, I was not aware of this. I have made this change.

primitives/linux-port/column.cpp
tntnatbry: Yes I agree, `hasAuxCol` does not belong here as an additional function argument. So I have moved it inside the `in` message.

primitives/linux-port/column.cpp
tntnatbry: I have made this change, thanks for this feedback.

primitives/primproc/columncommand.h
tntnatbry: Done.

writeengine/bulk/we_bulkload.cpp
tntnatbry: Done.

writeengine/shared/we_config.cpp
tntnatbry: I have removed the `boost::iequals()` call and now I am only performing a string comparison for either `y` or `Y`. I have also updated the .xml with the default value `n` instead of `false`.

writeengine/wrapper/we_colop.cpp
tntnatbry: Done.

writeengine/bulk/we_bulkload.cpp
tntnatbry: Done.

primitives/linux-port/column.cpp
tntnatbry: Done. I have changed this to use `if constexpr` semantics here and removed the code duplication in the if-else branches.

writeengine/server/we_dmlcommandproc.cpp
tntnatbry: Hmm, I don't see identical blocks here. Here we check if the table has AUX column and append the AUX column attributes to the vector.

tntnatbry:
@drrtuy Changes requested by you and Denis have been incorporated, please review. Once Daniel shares benchmarking numbers on this PR, please merge if ok. Also, there is a test001 PR that also needs to be merged alongside this engine PR: https://github.com/mariadb-corporation/mariadb-columnstore-regression-test/pull/302

primitives/linux-port/column.cpp
drrtuy: The proper way is to add another typename param for both scalarFiltering and vectorizedFilteringDispatcher to describe a AUX column integral type.

primitives/linux-port/column.cpp
```diff
@@ -1499,7 +1689,33 @@ void vectorizedFiltering(NewColRequestHeader* in, ColResultHeader* out, const T*
   scalarFiltering<T, FT, ST, KIND>(in, out, columnFilterMode, filterSet, filterCount, filterCOPs,
                                    filterValues, filterRFs, in->colType, origSrcArray, srcSize, origRidArray,
                                    ridSize, processedSoFar, outputType, validMinMax, emptyValue, nullValue,
-                                   Min, Max, isNullValueMatches);
+                                   Min, Max, isNullValueMatches, blockAux, emptyValueAux);
+}
+
+template<typename T, typename VT, bool HAS_INPUT_RIDS, int OUTPUT_TYPE,
+         ENUM_KIND KIND, typename FT, typename ST>
+void vectorizedFiltering(NewColRequestHeader* in, ColResultHeader* out, const T* srcArray,
```
drrtuy, line 1778: Nice move.

primitives/linux-port/column.cpp
```diff
@@ -1185,17 +1331,22 @@ void scalarFiltering(
...
+    const uint8_t* blockAux,
+    uint8_t emptyValueAux)
 {
   constexpr int WIDTH = sizeof(T);
   // Loop-local variables
   T curValue = 0;
   primitives::RIDType rid = 0;
   bool isEmpty = false;
 
+  auto nextColValuePtr = in->hasAuxCol ? nextColValue<T, WIDTH, true> : nextColValue<T, WIDTH, false>;
```
drrtuy, line 1338: What if you move this one level up so that scalarFiltering doesn't check in->hasAuxCol for every value? You did the similar thing for the vectorized processing path.

drrtuy: My bad I should have told you this right away.

primitives/linux-port/column.cpp
```diff
@@ -1420,6 +1572,38 @@ void vectorizedFiltering(NewColRequestHeader* in, ColResultHeader* out, const T*
     }
   }
 
+  MT* nonEmptyMaskAux;
+
+  if constexpr (IS_AUX_COLUMN)
```
drrtuy, line 1666: Did I get it right that if IS_AUX_COLUMN is true we process aux column buffer prior to the actual column buffer inside a single vectorizedFiltering_ call?
Is it possible to move the block into a separate template function which is empty for IS_AUX_COLUMN == false?

writeengine/bulk/we_bulkload.cpp
drrtuy: Agree.

writeengine/server/we_dmlcommandproc.cpp
drrtuy: Here and the block that starts at the line 1386 look very similar.

writeengine/server/we_dmlcommandproc.cpp
```diff
@@ -914,34 +993,38 @@ uint8_t WE_DMLCommandProc::processBatchInsert(messageqcpp::ByteStream& bs, std::
...
+          aColumn.compressionType = colType.compressionType;
+
+          if ((colType.compressionType > 0) && (colType.ddn.dictOID > 0))
+          {
+            DctnryStruct aDctnry;
+            aDctnry.dctnryOid = colType.ddn.dictOID;
+            aDctnry.fColPartition = dbRootExtent.fPartition;
+            aDctnry.fColSegment = dbRootExtent.fSegment;
+            aDctnry.fColDbRoot = dbRootExtent.fDbRoot;
+            dctnryList.push_back(aDctnry);
```
drrtuy, line 1024: I would replace this block with a single line dctnryList.emplace_back({colType.ddn.dictOID...}); 
But this is up to you to decide.

writeengine/server/we_dmlcommandproc.cpp
```diff
@@ -1062,6 +1146,31 @@ uint8_t WE_DMLCommandProc::processBatchInsert(messageqcpp::ByteStream& bs, std::
...
+        colStruct.dataOid = tableAUXColOid;
+        colStruct.tokenFlag = false;
+        colStruct.fCompressionType = colType.compressionType;
+        colStruct.colWidth = colType.colWidth;
+        colStruct.colDataType = colType.colDataType;
+        dctnryStruct.dctnryOid = 0;
+        dctnryStruct.columnOid = colStruct.dataOid;
+        dctnryStruct.fCompressionType = colType.compressionType;
+        dctnryStruct.colWidth = colType.colWidth;
+        colStructs.push_back(colStruct);
```
drrtuy, line 1176: I would also replace all these with 4 lines, 3 of which would be vector::emplace_back() calls + CalpontSystemCatalog::ColType initialization list ctor call.

primitives/linux-port/column.cpp
tntnatbry: This branching is outside the loop, so `in->hasAuxCol` is only checked per call, not for every value in the loop.

primitives/linux-port/column.cpp
tntnatbry: I can, but what is the point of creating a separate template function here?

writeengine/server/we_dmlcommandproc.cpp
tntnatbry: The existing code is more readable. `WriteEngine::DctnryStruct dctnryStruct;` calls the default ctor which initialises the struct members to some default values, followed by explicitly setting a subset of those members to other values here. Vs with the `emplace_back()` approach, we will have to interleave user-supplied values with defaults such as `dctnryList.emplace_back({colType.ddn.dictOID, 0, 0, dbRootExtent.fPartition ... etc.});` which makes it harder to read.

writeengine/server/we_dmlcommandproc.h
drrtuy: Let's left it as it might be needed by the bulk insertion API.

writeengine/server/we_readthread.cpp
drrtuy: Same thing. This is used by the bulk insertion API.

primitives/linux-port/column.cpp
drrtuy: You are right.

writeengine/server/we_dmlcommandproc.cpp
drrtuy: You can add a custom ctor for the purpose though.

tntnatbry: I would leave this as is, it goes outside the scope of this patch.

versioning/BRM/dbrm.cpp
```diff
@@ -4453,6 +4454,97 @@ void DBRM::deleteAISequence(uint32_t OID)
...
+          if (err)
+          {
+            ostringstream os;
+            os << "BRM lookup error. Could not get extents for OID " << objNum;
+            throw runtime_error(os.str());
+          }
+        }
+      }
+
+      uint32_t extentNumAux = fbo / (execplan::AUX_COL_WIDTH * 1024);
```
denis0x0D: @tntnatbry can you please clarify a little bit on the code.
Looking at the comment, this function checks if the given lbids are "aux columnd" lbids, and then for table where particular aux column belongs it collects lbids for all other columns in that table.

1. You use "execplan::AUX_COL_WIDTH" the size of "normalized number of blocks", it's correct until 
getExtentRows() or BLOCK_SIZE are not changed, but I would like to suggest to make it more general, because if decide to change the "BLOCK_SIZE" or "Number of rows" in extent you code will give the wrong result.

```

2826   // Convert extent size in rows to extent size in 8192-byte blocks.
2827   // extentRows should be multiple of blocksize (8192).
2828   const unsigned EXTENT_SIZE = (getExtentRows() * colWidth) / BLOCK_SIZE;

//  size is the number of multiples of 1024 blocks.
2843   //  ex: size=1 --> 1024 blocks
2844   //      size=2 --> 2048 blocks
2845   //      size=3 --> 3072 blocks, etc.
2846   uint32_t size = EXTENT_SIZE / 1024;
```

2. As far as I know FBO specifies an offset in the segment file for the particular extent, in case we have a multiple extents per one segment file. That means each extent has a unique lbid, in this case I do not understand a check at the line 4527. Can you please clarify on it? 
Thanks.

tntnatbry: @denis0x0D 

For Point 1, this is a very good catch, thanks for letting me know. I have changed this line:
```c++
uint32_t extentNumAux = fbo / (execplan::AUX_COL_WIDTH * 1024);
```
to:
```c++
uint32_t extentNumAux = fbo / ((getExtentRows() * execplan::AUX_COL_WIDTH) / BLOCK_SIZE);
```

Regarding Point 2:

Here is the check:
```c++
          if (dbRoot == extent.dbRoot && partitionNum == extent.partitionNum &&                                                             
              segmentNum == extent.segmentNum && extentNumAux == extentNum)                                                                 
          { 
            lbidSet.insert(extent.range.start);                                                                                             
            break;                                                                                                                          
          }                                                                                                                                 
```

Are you saying we should not have the check `extentNumAux == extentNum`? If we remove this check, then in case the segment file contains multiple extents, we will end up inserting multiple starting lbids into `lbidSet` for a single aux column lbid, which will be incorrect. We only want the lbids for all other columns in the table which correspond to a given extent number in the segment file, not for all extents in that segment file. Does this clarify?

denis0x0D: Can you please add a comment how does actually mapping between aux column and user columns does work.
Or if the description already exists, please point me on this, thanks.

tntnatbry: @denis0x0D  I will mention this when I create a high-level design doc for this feature. I will let you know once I have it.

writeengine/server/we_dmlcommandproc.cpp
tntnatbry: Done.

writeengine/server/we_dmlcommandproc.h
tntnatbry: Done. I have uncommented back this change.

writeengine/server/we_readthread.cpp
tntnatbry: Done. I have uncommented back this change.

primitives/linux-port/column.cpp
tntnatbry: I have created a separate function for this if block.

drrtuy (approved):
Nicely done @tntnatbry !
I will wait till @denis0x0D finishes and then merge it into develop.

denis0x0D:
@drrtuy I'm not holding this PR to merge, I just had some comments for some parts, I do not know the full design, @tntnatbry said, he will provide the design doc in the future. So it's better to merge if you approve it, because I do not see the full picture to approve or not approve this PR.

tntnatbry:
@drrtuy I will ask Daniel to run his DELETE and SELECT performance benchmark on this PR build. Once we have the numbers, we can merge this in.

PR #2414, opened 2022-06-10
MCOL-5089 EM::_createDictStoreExtent() and EM::_createColumnExtentExa…

drrtuy:
…ctFile() both lacks partition querying the EMIndex.

This forces EM to create another extent with a wrong partition number. See the issue for the reproduction scenario.


mariadb-DenisKhalikov (approved):
Thanks for the fix.

PR #2419, opened 2022-06-17
[MCOL-4699] Error message update.

denis0x0D:
This patch updates error message for circular joins -
we do support circular inner joins and do not support circular outer jois.


denis0x0D:
This should be implemented.

PR #2420, opened 2022-06-18
[MCOL-4778] Return if we have an error in push_down_init.

denis0x0D:


tntnatbry (changes requested):
LGTM. However, the crashing SQL test case should be added to MTR.

denis0x0D:
@tntnatbry in which part of mtr we adding tests cases for this kind of bugs `basics` or something else?

tntnatbry:
@denis0x0D Since this is a bug fix, we can put it under `mysql-test/columnstore/bugfixes/`.

denis0x0D:
@tntnatbry, I see, thanks!

davidhall586 (approved):

tntnatbry (approved):

PR #2422, opened 2022-06-21
MCOL-5044 FairThreadPool implementation + EXTRA threads functionality

drrtuy:


primitives/primproc/primitiveserver.h
```diff
@@ -36,7 +36,8 @@
 #include <boost/thread.hpp>
 
 #include "threadpool.h"
-#include "../../utils/threadpool/prioritythreadpool.h"
+#include "../../utils/threadpool/fair_threadpool.h"
+#include "fair_threadpool.h"
```
denis0x0D, line 39: Not related to this patch, but I also used a relative paths to the headers, in my previous patch, instead of adding includes into cmake, because for some reasons they do not work correctly, which I failed to understand, do you had the same problems?

utils/threadpool/fair_threadpool.cpp
```diff
@@ -0,0 +1,316 @@
...
+    ThreadPoolJobsList* txnJobsList = txnJobsMapPair.second;
+    auto job = txnJobsList->begin();
+    while (job != txnJobsList->end())
+    {
+      if (job->id_ == id)
+      {
+        job = txnJobsList->erase(job);  // update the job iter
+        if (txnJobsList->empty())
+        {
+          txn2JobsListMap_.erase(txnJobsMapPair.first);
```
denis0x0D: Do I understand correctly, when we remove specific jobs, we still keep associated transaction id for this joblist in priority queue and actual release of the priority queue happens in the `threadFcn` in this loop
`   while (txnAndJobListPair == txn2JobsListMap_.end() || txnAndJobListPair->second->empty())
      {`

denis0x0D (approved):
LGTM

primitives/primproc/primitiveserver.h
drrtuy: No, this was my mistake that I cleaned-up with a slightly changed version of the patch.

utils/threadpool/fair_threadpool.cpp
drrtuy: Yes, this is the way how PQ purges empty txn-s.

mariadb-LeonidFedorov (approved):

PR #2425, opened 2022-06-23
MCOL-785 Implement DISTRIBUTED JSON functions

Ziy1-Tan:
- [x] The Jira issue number for this PR is: [MCOL-785](https://jira.mariadb.org/browse/MCOL-785)

## Description
The following json function will be created:
- [x] JSON_ARRAY
- [x] JSON_ARRAY_APPEND
- [x] JSON_ARRAY_INSERT
- [x] JSON_COMPACT
- [x] JSON_CONTAINS
- [x] JSON_DEPTH
- [x] JSON_DETAILED
- [x] JSON_EQUALS
- [x] JSON_EXISTS
- [x] JSON_INSERT
- [x] JSON_KEYS
- [x] JSON_LENGTH
- [x] JSON_LOOSE
- [x] JSON_MERGE
- [x] JSON_MERGE_PATCH
- [x] JSON_MERGE_PRESERVE
- [x] JSON_NORMALIZE
- [x] JSON_OBJECT
- [x] JSON_QUERY
- [x] JSON_QUOTE
- [x] JSON_REMOVE
- [x] JSON_REPLACE
- [x] JSON_SET
- [x] JSON_TYPE
- [x] JSON_UNQUOTE
- [x] JSON_VALID
- [x] JSON_VALUE
- [x] JSON_CONTAINS_PATH
- [x] JSON_EXTRACT
- [x] JSON_OVERLAPS (Since MariaDB 10.9)
- [x] JSON_SEARCH

## How can this PR be tested?
The test files locate at test suite **columnstore/future** `func_json_xxx.test` and `func_json_xxx.result` 

## Basing the PR against the correct Columnstore version
- [x] This is a new feature and the PR is based against the latest MariaDB development branch
- [ ] This is a bug fix and the PR is based against the earliest branch in which the bug can be reproduce

### DIFF
--- a/dbcon/execplan/functioncolumn.cpp
+++ b/dbcon/execplan/functioncolumn.cpp
@@ -29,6 +29,7 @@ using namespace std;
 #include <boost/algorithm/string.hpp>
 using namespace boost;
 
+#include "functor_json.h"
 #include "bytestream.h"
 #include "functioncolumn.h"
 #include "constantcolumn.h"
@@ -323,6 +324,46 @@ void FunctionColumn::unserialize(messageqcpp::ByteStream& b)
   Func_decode* decode = dynamic_cast<Func_decode*>(fFunctor);
   if (decode)
     fFunctor = fDynamicFunctor = new Func_decode();
+
+  // Special treatment for json function contains the variable path. reset the variable path
+  if (dynamic_cast<Func_json_length*>(fFunctor))
+    fFunctor = fDynamicFunctor = new Func_json_length();
+
+  if (dynamic_cast<Func_json_keys*>(fFunctor))
+    fFunctor = fDynamicFunctor = new Func_json_keys();
+
+  if (dynamic_cast<Func_json_exists*>(fFunctor))
+    fFunctor = fDynamicFunctor = new Func_json_exists();
+
+  if (dynamic_cast<Func_json_value*>(fFunctor))
+    fFunctor = fDynamicFunctor = new Func_json_value();
+
+  if (dynamic_cast<Func_json_query*>(fFunctor))
+    fFunctor = fDynamicFunctor = new Func_json_query();
+
+  if (dynamic_cast<Func_json_contains*>(fFunctor))
+    fFunctor = fDynamicFunctor = new Func_json_contains();
+
+  if (dynamic_cast<Func_json_array_append*>(fFunctor))
+    fFunctor = fDynamicFunctor = new Func_json_array_append();
+
+  if (dynamic_cast<Func_json_array_insert*>(fFunctor))
+    fFunctor = fDynamicFunctor = new Func_json_array_insert();
+
+  if (auto f = dynamic_cast<Func_json_insert*>(fFunctor))
+    fFunctor = fDynamicFunctor = new Func_json_insert(f->getMode());
+
+  if (dynamic_cast<Func_json_remove*>(fFunctor))
+    fFunctor = fDynamicFunctor = new Func_json_remove();
+
+  if (dynamic_cast<Func_json_contains_path*>(fFunctor))
+    fFunctor = fDynamicFunctor = new Func_json_contains_path();
+
+  if (dynamic_cast<Func_json_search*>(fFunctor))
+    fFunctor = fDynamicFunctor = new Func_json_search();
+
+  if (dynamic_cast<Func_json_extract*>(fFunctor))
+    fFunctor = fDynamicFunctor = new Func_json_extract();
 }
 
 bool FunctionColumn::operator==(const FunctionColumn& t) const
--- a/dbcon/mysql/ha_mcs_execplan.cpp
+++ b/dbcon/mysql/ha_mcs_execplan.cpp
@@ -3550,6 +3550,35 @@ ReturnedColumn* buildReturnedColumn(Item* item, gp_walk_info& gwi, bool& nonSupp
   return rc;
 }
 
+// parse the boolean fields to string "true" or "false"
+ReturnedColumn* buildBooleanConstantColumn(Item* item, gp_walk_info& gwi, bool& nonSupport)
+{
+  ConstantColumn* cc = NULL;
+
+  if (gwi.thd)
+  {
+    {
+      if (!item->fixed())
+      {
+        item->fix_fields(gwi.thd, (Item**)&item);
+      }
+    }
+  }
+  int64_t val = static_cast<int64_t>(item->val_int());
+  cc = new ConstantColumnSInt(colType_MysqlToIDB(item), val ? "true" : "false", val);
+
+  if (cc)
+    cc->timeZone(gwi.timeZone);
+
+  if (cc && item->name.length)
+    cc->alias(item->name.str);
+
+  if (cc)
+    cc->charsetNumber(item->collation.collation->number);
+
+  return cc;
+}
+
 ArithmeticColumn* buildArithmeticColumn(Item_func* item, gp_walk_info& gwi, bool& nonSupport)
 {
   if (get_fe_conn_info_ptr() == NULL)
@@ -4013,7 +4042,24 @@ ReturnedColumn* buildFunctionColumn(Item_func* ifp, gp_walk_info& gwi, bool& non
           return NULL;
         }
 
-        ReturnedColumn* rc = buildReturnedColumn(ifp->arguments()[i], gwi, nonSupport);
+        ReturnedColumn* rc = NULL;
+
+
+        // Special treatment for json functions
+        // All boolean arguments will be parsed as boolean string true(false)
+        // E.g. the result of `SELECT JSON_ARRAY(true, false)` should be [true, false] instead of [1, 0]
+        bool mayHasBoolArg = ((funcName == "json_insert" || funcName == "json_replace" ||
+                              funcName == "json_set" || funcName == "json_array_append" ||
+                              funcName == "json_array_insert") && i != 0 && i % 2 == 0) ||
+                             (funcName == "json_array") ||
+                             (funcName == "json_object" && i % 2 == 1);
+        bool isBoolType =
+            (ifp->arguments()[i]->const_item() && ifp->arguments()[i]->type_handler()->is_bool_type());
+
+        if (mayHasBoolArg && isBoolType)
+          rc = buildBooleanConstantColumn(ifp->arguments()[i], gwi, nonSupport);
+        else
+          rc = buildReturnedColumn(ifp->arguments()[i], gwi, nonSupport);
 
         // MCOL-1510 It must be a temp table field, so find the corresponding column.
         if (!rc && ifp->arguments()[i]->type() == Item::REF_ITEM)
--- a/utils/funcexp/func_json_array_append.cpp
+++ b/utils/funcexp/func_json_array_append.cpp
@@ -1,0 +1,117 @@
+#include "functor_json.h"
+#include "functioncolumn.h"
+#include "constantcolumn.h"
+using namespace execplan;
+
+#include "rowgroup.h"
+using namespace rowgroup;
+
+#include "joblisttypes.h"
+using namespace joblist;
+
+#include "jsonhelpers.h"
+using namespace funcexp::helpers;
+
+namespace funcexp
+{
+CalpontSystemCatalog::ColType Func_json_array_append::operationType(FunctionParm& fp,
+                                                                    CalpontSystemCatalog::ColType& resultType)
+{
+  return fp[0]->data()->resultType();
+}
+
+string Func_json_array_append::getStrVal(rowgroup::Row& row, FunctionParm& fp, bool& isNull,
+                                         execplan::CalpontSystemCatalog::ColType& type)
+{
+  const string_view js = fp[0]->data()->getStrVal(row, isNull);
+  if (isNull)
+    return "";
+
+  const CHARSET_INFO* cs = getCharset(fp[0]);
+
+  json_engine_t jsEg;
+  const uchar* arrEnd;
+  size_t strRestLen;
+  string retJS;
+  retJS.reserve(js.size() + padding);
+
+  initJSPaths(paths, fp, 1, 2);
+
+  string tmpJS{js};
+  for (size_t i = 1, j = 0; i < fp.size(); i += 2, j++)
+  {
+    const char* rawJS = tmpJS.data();
+    const size_t jsLen = tmpJS.size();
+    JSONPath& path = paths[j];
+    if (!path.parsed && parseJSPath(path, row, fp[i], false))
+      goto error;
+
+    initJSEngine(jsEg, cs, tmpJS);
+
+    if (locateJSPath(jsEg, path))
+      goto error;
+
+    if (json_read_value(&jsEg))
+      goto error;
+
+    if (jsEg.value_type == JSON_VALUE_ARRAY)
+    {
+      int itemSize;
+      if (json_skip_level_and_count(&jsEg, &itemSize))
+        goto error;
+
+      arrEnd = jsEg.s.c_str - jsEg.sav_c_len;
+      strRestLen = jsLen - (arrEnd - (const uchar*)rawJS);
+      retJS.append(rawJS, arrEnd - (const uchar*)rawJS);
+      if (itemSize)
+        retJS.append(", ");
+      if (appendJSValue(retJS, cs, row, fp[i + 1]))
+        goto error;
+
+      retJS.append((const char*)arrEnd, strRestLen);
+    }
+    else
+    {
+      const uchar *start, *end;
+
+      /* Wrap as an array. */
+      retJS.append(rawJS, (const char*)jsEg.value_begin - rawJS);
+      start = jsEg.value_begin;
+
+      if (jsEg.value_type == JSON_VALUE_OBJECT)
+      {
+        if (json_skip_level(&jsEg))
+          goto error;
+        end = jsEg.s.c_str;
+      }
+      else
+        end = jsEg.value_end;
+
+      retJS.append("[");
+      retJS.append((const char*)start, end - start);
+      retJS.append(", ");
+      if (appendJSValue(retJS, cs, row, fp[i + 1]))
+        goto error;
+      retJS.append("]");
+      retJS.append((const char*)jsEg.s.c_str, rawJS + jsLen - (const char*)jsEg.s.c_str);
+    }
+
+    // tmpJS save the json string for next loop
+    tmpJS.swap(retJS);
+    retJS.clear();
+  }
+
+  initJSEngine(jsEg, cs, tmpJS);
+  retJS.clear();
+  if (doFormat(&jsEg, retJS, Func_json_format::LOOSE))
+    goto error;
+
+  isNull = false;
+  return retJS;
+
+error:
+  isNull = true;
+  return "";
+}
+
+}  // namespace funcexp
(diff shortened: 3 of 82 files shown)


utils/funcexp/func_json_length.cpp
```diff
@@ -0,0 +1,110 @@
...
+{
+  if (fp.size() > 1)
+    path.set_constant_flag(fp[2]->data()->getIntVal());
+  return resultType;
+}
+
+int64_t Func_json_length::getIntVal(rowgroup::Row& row, FunctionParm& fp, bool& isNull,
+                                    execplan::CalpontSystemCatalog::ColType& op_ct)
+{
+  const std::string tmp_js = fp[0]->data()->getStrVal(row, isNull);
```
drrtuy: This getStrVal returns a ref so I would use ref to avoid std::strin copy-ctor.

utils/funcexp/func_json_length.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+  while (!(err = json_scan_next(&je)) && je.state != JST_OBJ_END && je.state != JST_ARRAY_END)
+  {
+    switch (je.state)
+    {
+      case JST_VALUE:
+      case JST_KEY: length++; break;
+      case JST_OBJ_START:
+      case JST_ARRAY_START:
+        if (json_skip_level(&je))
```
drrtuy: Could you ellaborate on the semantics of this expression? If the next token is an array start - '[' and json_skip_level returns true we set isNull and set the length to 0. Why so?

utils/funcexp/func_json_valid.cpp
```diff
@@ -0,0 +1,40 @@
...
+{
+  const string& tmp_js = fp[0]->data()->getStrVal(row, isNull);
+  CHARSET_INFO* cs = fp[0]->data()->resultType().getCharset();
+
+  if (isNull)
+    return false;
+
+  const char* js = tmp_js.c_str();
+
+  return json_valid(js, strlen(js), cs);
```
drrtuy: I don't think we want to calculate the length again given that tmp_js has the length attribute.

drrtuy (changes requested):
Could I ask y

utils/funcexp/func_json_length.cpp
Ziy1-Tan: Well, First `json_skip_level` return 1 when error occurs. And when `JSON_LENGTH` is called, length of nested arrays or objects are not counted, they will be count as 1. `json_skip_level` is for skipping nested array or object. The outermost '[' or '{' have been skipped, So it they will never hit  `case JST_OBJ_START:
      case JST_ARRAY_START:
        if (json_skip_level(&je))...`.
In conclusion, this call is designed to skip nested structures. Refer to [here](https://github.com/qggcs/server/blob/ba5b2e7b291a9b4bfb97dcdf3c53ca49fc91a4e7/sql/item_jsonfunc.cc#L2664)

utils/funcexp/func_json_length.cpp
Ziy1-Tan: I have changed string to std::string_view

utils/funcexp/func_json_length.cpp
Ziy1-Tan: Take `JSON_LENGTH('[1,[2,3]]')`. When parsing the first '[',`je.state` is `JST_KEY`, when parsing the second '[', `je.state` is `JST_KEY` and then change to `JST_ARRAY_START` in the next loop.

Ziy1-Tan:
I must point out the function parameters `json_find_path` I use in columnstore is different before 10.8 and after 10.8. Before the 10.8 `int json_find_path(json_engine_t *je, json_path_t *p, json_path_step_t **p_cur_step, uint *array_counters)`.After 10.8 `int json_find_path(json_engine_t *je, json_path_t *p, json_path_step_t **p_cur_step, int *array_counters)`.This is causing my code to fail CI. How should I modify it?

drrtuy:
In short you should use the most recent API for jaon lib.
Full answer. This is a tricky question b/c the CI pipelines uses both API versions. Plz give me couple days so I can discuss with the server team about merging the most recent API into enterprise 10.6 ofthe server.

dbcon/execplan/functioncolumn.cpp
```diff
@@ -323,6 +324,50 @@ void FunctionColumn::unserialize(messageqcpp::ByteStream& b)
   Func_decode* decode = dynamic_cast<Func_decode*>(fFunctor);
   if (decode)
     fFunctor = fDynamicFunctor = new Func_decode();
+
+  Func_json_length* json_length = dynamic_cast<Func_json_length*>(fFunctor);
```
drrtuy: I am curios about this design decision?
Could you ellaborate on why do you need this RTTI-based checks to allocate fDynamicFunctor and replace fFunctor?

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -3550,6 +3550,35 @@ ReturnedColumn* buildReturnedColumn(Item* item, gp_walk_info& gwi, bool& nonSupp
...
+                              (int64_t)item->val_int());
+
+  if (cc)
+    cc->timeZone(gwi.timeZone);
+
+  if (cc && item->name.length)
+    cc->alias(item->name.str);
+
+  if (cc)
+    cc->charsetNumber(item->collation.collation->number);
```
drrtuy, line 3577: Are you sure that collation will never be null in the context? The same is actual about item->name.

utils/funcexp/func_json_array_append.cpp
```diff
@@ -0,0 +1,154 @@
...
+string Func_json_array_append::getStrVal(rowgroup::Row& row, FunctionParm& fp, bool& isNull,
+                                         execplan::CalpontSystemCatalog::ColType& type)
+{
+  string tmpJs = fp[0]->data()->getStrVal(row, isNull);
+  if (isNull)
+    return "";
+
+  const CHARSET_INFO* cs = fp[0]->data()->resultType().getCharset();
+
+  json_engine_t jsEg;
```
drrtuy, line 32: How heave is this json_engine_t ctor call is?

utils/funcexp/func_json_array_append.cpp
```diff
@@ -0,0 +1,154 @@
...
+CalpontSystemCatalog::ColType Func_json_array_append::operationType(FunctionParm& fp,
+                                                                    CalpontSystemCatalog::ColType& resultType)
+{
+  return fp[0]->data()->resultType();
+}
+
+string Func_json_array_append::getStrVal(rowgroup::Row& row, FunctionParm& fp, bool& isNull,
+                                         execplan::CalpontSystemCatalog::ColType& type)
+{
+  string tmpJs = fp[0]->data()->getStrVal(row, isNull);
```
drrtuy: Can we use string_view here also?

utils/funcexp/func_json_array_insert.cpp
```diff
@@ -0,0 +1,169 @@
...
+CalpontSystemCatalog::ColType Func_json_array_insert::operationType(FunctionParm& fp,
+                                                                    CalpontSystemCatalog::ColType& resultType)
+{
+  return fp[0]->data()->resultType();
+}
+
+string Func_json_array_insert::getStrVal(rowgroup::Row& row, FunctionParm& fp, bool& isNull,
+                                         execplan::CalpontSystemCatalog::ColType& type)
+{
+  string tmpJs = fp[0]->data()->getStrVal(row, isNull);
```
drrtuy: Can we use string_view here also?

utils/funcexp/func_json_insert.cpp
```diff
@@ -0,0 +1,286 @@
...
+CalpontSystemCatalog::ColType Func_json_insert::operationType(FunctionParm& fp,
+                                                              CalpontSystemCatalog::ColType& resultType)
+{
+  return fp[0]->data()->resultType();
+}
+
+string Func_json_insert::getStrVal(rowgroup::Row& row, FunctionParm& fp, bool& isNull,
+                                   execplan::CalpontSystemCatalog::ColType& type)
+{
+  string tmpJs = fp[0]->data()->getStrVal(row, isNull);
```
drrtuy: Can we use string_view here also?

utils/funcexp/func_json_remove.cpp
```diff
@@ -0,0 +1,215 @@
...
+CalpontSystemCatalog::ColType Func_json_remove::operationType(FunctionParm& fp,
+                                                              CalpontSystemCatalog::ColType& resultType)
+{
+  return fp[0]->data()->resultType();
+}
+
+string Func_json_remove::getStrVal(rowgroup::Row& row, FunctionParm& fp, bool& isNull,
+                                   execplan::CalpontSystemCatalog::ColType& type)
+{
+  string tmpJs = fp[0]->data()->getStrVal(row, isNull);
```
drrtuy: Can we use string_view here also?

utils/funcexp/func_json_length.cpp
drrtuy: Thx for the explanation.

drrtuy (commented):

dbcon/execplan/functioncolumn.cpp
Ziy1-Tan: Purpose: reinitialize the path variable.
Well, As we know, the variable `fDynamicFunctor` is used to reinitialize member variables(e.g `fSeed1、fSeed2、fSeedSet` in `Func_rand`). These functions use these variable to record the states in a SQL. Since the function state may be different in each SQL.
So these function must be assigned to `fDynamicFunctor` to ensure that is destructed after the call, does not retain the state of last SQL.
`FunctionColumn::~FunctionColumn()
{
  if (fDynamicFunctor)
    delete fDynamicFunctor;
}`

Ziy1-Tan: Same as json function. Some json function will accept path parameters. These functions use variable path to save path state. Taking `json_length` as example, When a SQL is called, json_length will be check if path argument is constant, If it is constant, it will be cached until next SQL call.Refer to [here](https://github.com/qggcs/mariadb-columnstore-engine/blob/dd95fcdde4ac5b66db8409aa0ead03e2f5a02f0f/utils/funcexp/func_json_length.cpp#L45-L64).

The cache must be reinitialized before the next call, otherwise, it will affect the next SQL call. So these function are assigned to the `fDynamicFunctor`.

[json_length](https://github.com/qggcs/server/blob/a5dc76a051ff908a5f506ed12e8d05311752e5fb/sql/item_jsonfunc.cc#L2721) in mariadb server.

Ziy1-Tan: You can run the func_json_length.test before and after delete RTTI code.

Ziy1-Tan: The small case:
`CREATE TABLE t1(l longtext) engine=columnstore;`
`INSERT INTO t1 VALUES('{"key1":[1,2,3],"key2":[1,2,3,4,5]}')`;
`SELECT JSON_LENGTH(l, '$.key1') FROM t1;`
`SELECT JSON_LENGTH(l, '$.key2') FROM t1;`
The 2nd argument is const. Without RTTI check, the path state will be cached and will not reinitialize, the result is always 3.

dbcon/mysql/ha_mcs_execplan.cpp
Ziy1-Tan: Before defining the function `buildBooleanConstColumn`,All the arguments will enter the function `rc = buildReturnedColumn(ifp->arguments()[i], gwi, nonSupport)` [here](https://github.com/qggcs/mariadb-columnstore-engine/blob/f4aba866ecf4d6e4c2aa3fc6e52808741e79b790/dbcon/mysql/ha_mcs_execplan.cpp#L4049).
After debugging, the `item.collation` and `item.name` have default value, So it will not be null. these code in `buildBooleanConstColumn` is same as [buildReturnedColumn](https://github.com/qggcs/mariadb-columnstore-engine/blob/f4aba866ecf4d6e4c2aa3fc6e52808741e79b790/dbcon/mysql/ha_mcs_execplan.cpp#L3544-L3548).

utils/funcexp/func_json_array_append.cpp
Ziy1-Tan: Why ask that? All json functions require `json_engine_t`.I don't understand why only this function is mentioned.

utils/funcexp/func_json_array_append.cpp
Ziy1-Tan: Fix.

utils/funcexp/func_json_array_insert.cpp
Ziy1-Tan: Fix.

utils/funcexp/func_json_insert.cpp
Ziy1-Tan: Fix.

utils/funcexp/func_json_remove.cpp
Ziy1-Tan: Fix.

Ziy1-Tan:
> In short you should use the most recent API for jaon lib. Full answer. This is a tricky question b/c the CI pipelines uses both API versions. Plz give me couple days so I can discuss with the server team about merging the most recent API into enterprise 10.6 ofthe server.

Now the lastest API have been used in the development. These code have passed all the CI on MariaDB 10.9. But CI on 10.6 still can't pass because the latest API is not forward compatible.

utils/funcexp/func_json_array_append.cpp
drrtuy: I just asked about this function taken it as an example b/c I don't want to repeat the same question many times.

drrtuy: If the ctor is heavy it might worth to consider call this ctor only once sharing the json_engine_t instance b/w calls somewhere.

dbcon/execplan/functioncolumn.cpp
```diff
@@ -323,6 +324,59 @@ void FunctionColumn::unserialize(messageqcpp::ByteStream& b)
   Func_decode* decode = dynamic_cast<Func_decode*>(fFunctor);
   if (decode)
     fFunctor = fDynamicFunctor = new Func_decode();
+
+  // Special treatment for json function contains the variable path. reset the variable path
+  Func_json_length* json_length = dynamic_cast<Func_json_length*>(fFunctor);
```
denis0x0D: I think it should rewritten to
`if (dynamic_cast<Type *>(pointer))`

instead of 
```
auto *typePointer = dynamic_cast<Type *>(pointer);
if (typePointer)
```
Here and below as well.

utils/funcexp/func_json_array_append.cpp
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    json_scan_start(&jsEg, cs, (const uchar*)js, (const uchar*)js + jsLen);
+
+    currPath.currStep = currPath.p.steps;
+
+    if (jsonFindPath(&jsEg, &currPath.p, &currPath.currStep))
+    {
+      if (jsEg.s.error)
+      {
```
denis0x0D: What this empty branch is about?

utils/funcexp/func_json_array_append.cpp
```diff
@@ -0,0 +1,140 @@
...
+  if (isNull)
+    return "";
+
+  const CHARSET_INFO* cs = fp[0]->data()->resultType().getCharset();
+
+  json_engine_t jsEg;
+  const uchar* arrEnd;
+  size_t strRestLen;
+  string ret;
+  ret.reserve(jsExp.size() + 8);
```
denis0x0D: Could you please use a naming constant instead of just 8?

utils/funcexp/func_json_array_insert.cpp
```diff
@@ -0,0 +1,156 @@
...
+}
+
+string Func_json_array_insert::getStrVal(rowgroup::Row& row, FunctionParm& fp, bool& isNull,
+                                         execplan::CalpontSystemCatalog::ColType& type)
+{
+  const string_view jsExp = fp[0]->data()->getStrVal(row, isNull);
+  if (isNull)
+    return "";
+
+  const CHARSET_INFO* cs = fp[0]->data()->resultType().getCharset();
```
denis0x0D: Some part of this function is the same, as function above, I would like to suggest to not copy past it, but just implement it once and use everywhere you need.

utils/funcexp/func_json_contains.cpp
```diff
@@ -0,0 +1,233 @@
...
+      return valEg->value_len == jsEg->value_len && memcmp(valEg->value, jsEg->value, valEg->value_len) == 0;
+    case JSON_VALUE_NUMBER:
+      if (valEg->value_type == JSON_VALUE_NUMBER)
+      {
+        double jsEgVal, valEgVal;
+        char* end;
+        int err;
+
+        jsEgVal = jsEg->s.cs->strntod((char*)jsEg->value, jsEg->value_len, &end, &err);
+        ;
```
denis0x0D, line 125: What this empty statement is about?

utils/funcexp/func_json_contains.cpp
```diff
@@ -0,0 +1,233 @@
...
+        goto error;
+
+      path.parsed = path.constant;
+    }
+
+    path.currStep = path.p.steps;
+    if (jsonFindPath(&jsEg, &path.p, &path.currStep))
+    {
+      if (jsEg.s.error)
+      {
```
denis0x0D: Empty branch is not needed.

utils/funcexp/func_json_contains_path.cpp
```diff
@@ -0,0 +1,152 @@
...
+    hasFound.assign(argSize, false);
+  }
+
+  for (size_t i = 2; i < fp.size(); i++)
+  {
+    JsonPath& currPath = paths[i - 2];
+
+    if (!currPath.parsed)
+    {
+      const string_view pathExp = fp[i]->data()->getStrVal(row, isNull);
```
denis0x0D: This function also has the same code with other functions, I would like to suggest to take the same code from all function and share it.

denis0x0D (commented):
Hey @Ziy1-Tan thanks for the patch, I've just started to look at your code. Some initial thoughts:
1. Some functions using the same code, could you please implement it once and use everywhere you need.
2. Have you tested your code on invalid json input? You implemented a lot of parsing logic which is usually predisposed for simple errors and security errors as well, I think we need to apply some fuzzing on them to make sure your code will work correctly on any input.

Ziy1-Tan:
thanks for your comments, @denis0x0D! I will take a look at these.

Ziy1-Tan:
> 2\. invalid json input

The invalid json input error will be handled by json_engin_t, all these errors are indicated by `jsEg.s.error`, thanks to this, I don't have to deal with this kind of errors extra,

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -3550,6 +3550,35 @@ ReturnedColumn* buildReturnedColumn(Item* item, gp_walk_info& gwi, bool& nonSupp
...
+  if (gwi.thd)
+  {
+    {
+      if (!item->fixed())
+      {
+        item->fix_fields(gwi.thd, (Item**)&item);
+      }
+    }
+  }
+  cc = new ConstantColumnSInt(colType_MysqlToIDB(item), (int64_t)item->val_int() ? "true" : "false",
```
drrtuy: I might better use static_cast and put item->val_int() into a variable to reduce number of calls.

Ziy1-Tan: it looks good! I have fixed it.

drrtuy (approved):

PR #2428, opened 2022-06-26
MCOL-4995 Research/implement basic vectorized filtering for ARM platforms 

NTH19:
### For the convenience of review
- [x]  The Jira issue number for this PR is: MCOL-4995
### Description
the filter part of it

### How can this PR be tested?
The test files locate at test suite columnstore/tests/simd_processors.cpp

### Basing the PR against the correct Columnstore version
- [x]  This is a new feature and the PR is based against the latest MariaDB development branch
- [ ]  This is a bug fix and the PR is based against the earliest branch in which the bug can be reproduce

### DIFF
--- a/utils/common/simd_arm.h
+++ b/utils/common/simd_arm.h
@@ -47,6 +47,28 @@ using vi128f_t = float32x4_t;
 using vi128d_t = float64x2_t;
 using int128_t = __int128;
 using MT = uint16_t;
+using MaskSimdType=vi1u_t;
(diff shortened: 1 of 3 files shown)


utils/common/simd_arm.h
```diff
@@ -17,6 +17,7 @@
 
 #pragma once
 
+enum ENUM_KIND;
```
drrtuy: Do we need this forward declaration?

PR #2432, opened 2022-06-29
MCOL-5013: Load Data from S3 into Columnstore

mariadb-LeonidFedorov:
Introduced UDF and stored prodecure.
usage:
```
set columnstore_s3_key='<s3_key>';
set columnstore_s3_secret='<s3_secret>';
set columnstore_s3_region='region';
```
and then use UDF
```
select columnstore_dataload("<tablename>", "<filename>", "<bucket>", "<db_name>");
```
for UDF ```db_name``` can be ommited, then current connection db will be used

or stored function
```
call calpontsys.columnstore_load_from_s3("<tablename>", "<filename>", "<bucket>", "<db_name>");```

### CODE REVIEW

File: dbcon/mysql/columnstore_dataload.cpp [hunk 1]
```diff
@@ -0,0 +1,190 @@
+/* Copyright (C) 2021 MariaDB Corporation
```
drrtuy: It is 2022 :)

dbcon/mysql/columnstore_dataload.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+    if (res != CURLE_OK)
+    {
+      std::string msg = std::string("Remote request failed: ") + curl_easy_strerror(res);
+      memcpy(result, msg.c_str(), msg.length());
+      *length = msg.length();
+      return result;
+    }
+
+    result = new char[readBuffer.length() + 1];
```
drrtuy, line 137: This is how it works. No SP just plain new and delete it in deinit?

dbcon/mysql/columnstore_dataload.cpp
```diff
@@ -0,0 +1,190 @@
...
+    const char* database = args->arg_count != 4 ? thd->get_db() : args->args[3];
+
+    return columnstore_dataload_impl(initData->curl, initData->result, length, bucket, table, filename,
+                                     database, secret, key, region, cmapi_host, cmapi_port, cmapi_version,
+                                     ::strlen(cmapi_key) == 0 ? parseCMAPIkey().c_str() : cmapi_key);
+  }
+
+  my_bool columnstore_dataload_init(UDF_INIT* initid, UDF_ARGS* args, char* message)
+  {
+    initid->max_length = 1000 * 1000;
```
drrtuy: Mb it is worth to put it behind the check that takes place at 182?

drrtuy:
Conceptually LGTM but plz allow us to fix MTR before we proceed with the merge.

dbcon/mysql/columnstore_dataload.cpp
mariadb-LeonidFedorov: That's it

dbcon/mysql/columnstore_dataload.cpp
mariadb-LeonidFedorov: ++

dbcon/mysql/columnstore_dataload.cpp
mariadb-LeonidFedorov: Fixed. I'm very old fashioned

drrtuy (approved):

PR #2433, opened 2022-06-30
Vectorized update min max

mariadb-AndreyPiskunov:
Solving MCOL-5140.


primitives/linux-port/column.cpp
```diff
@@ -1284,6 +1276,41 @@ inline SIMD_WRAPPER_TYPE simdSwapedOrderDataLoad(const ColRequestHeaderDataType
...
+    }
+  }
+}
+
+template<typename T, typename VT, typename SimdType>
+void extractMinMax(VT& simdProcessor, SimdType& simdMin, SimdType& simdMax, T& Min, T& Max)
+{
+  T* simdMinVec = reinterpret_cast<T*>(&simdMin);
+  T* simdMaxVec = reinterpret_cast<T*>(&simdMax);
+  Max = *std::max_element(simdMaxVec, simdMaxVec + VT::vecByteSize);
```
mariadb-LeonidFedorov: follow codestyle please, params should be lowercase

primitives/linux-port/column.cpp
```diff
@@ -1472,6 +1501,18 @@ void vectorizedFiltering(NewColRequestHeader* in, ColResultHeader* out, const T*
     valuesWritten = vectWriteRIDValues<T, VT, OUTPUT_TYPE, KIND, HAS_INPUT_RIDS>(
         simdProcessor, valuesWritten, validMinMax, ridOffset, dataVecTPtr, ridDstArray, writeMask, Min, Max,
         in, out, nonNullOrEmptyMask, ridArray);
+    if constexpr (KIND != KIND_FLOAT)
```
mariadb-LeonidFedorov: What will happen when ```KIND==KIND_FLOAT```

primitives/linux-port/column.cpp
mariadb-AndreyPiskunov: Renaming done, thank you.

primitives/linux-port/column.cpp
mariadb-AndreyPiskunov: I decided to just add min/max instructions for floats as well, so this branching is removed.

primitives/linux-port/column.cpp
```diff
@@ -1465,14 +1495,24 @@ void vectorizedFiltering(NewColRequestHeader* in, ColResultHeader* out, const T*
...
+        simdProcessor, writeMask, nonNullOrEmptyMask, validMinMax, ridOffset, dataVecTPtr, dstArray, min, max,
         in, out, ridDstArray, ridArray);
     // Some outputType modes saves RIDs also. vectWriteRIDValues is empty for
     // OT_DATAVALUE, OT_BOTH(vectWriteColValues takes care about RIDs).
     valuesWritten = vectWriteRIDValues<T, VT, OUTPUT_TYPE, KIND, HAS_INPUT_RIDS>(
-        simdProcessor, valuesWritten, validMinMax, ridOffset, dataVecTPtr, ridDstArray, writeMask, Min, Max,
+        simdProcessor, valuesWritten, validMinMax, ridOffset, dataVecTPtr, ridDstArray, writeMask, min, max,
         in, out, nonNullOrEmptyMask, ridArray);
 
+    if constexpr (HAS_INPUT_RIDS && KIND != KIND_TEXT)
```
drrtuy: What if we just run vectorizedUpdateMinMax if the KIND != KIND_TEXT?
JFYI We don't need scalarUpdateMinMax for KIND_TEXT b/c we don't support min/max for CHAR/VARCHAR extents.

drrtuy (changes requested):
I would like to discuss/change the way we apply scalar Min/Max calculation.

primitives/linux-port/column.cpp
```diff
@@ -118,6 +119,24 @@ inline int compareBlock(const void* a, const void* b)
   return ((*(T*)a) - (*(T*)b));
 }
 
+template <class To, class From>
```
drrtuy: I would rather put it into utils/common b/c this might be useful in many cases.

primitives/linux-port/column.cpp
```diff
@@ -1465,14 +1537,20 @@ void vectorizedFiltering(NewColRequestHeader* in, ColResultHeader* out, const T*
...
+        simdProcessor, writeMask, nonNullOrEmptyMask, validMinMax, ridOffset, dataVecTPtr, dstArray, min, max,
         in, out, ridDstArray, ridArray);
     // Some outputType modes saves RIDs also. vectWriteRIDValues is empty for
     // OT_DATAVALUE, OT_BOTH(vectWriteColValues takes care about RIDs).
     valuesWritten = vectWriteRIDValues<T, VT, OUTPUT_TYPE, KIND, HAS_INPUT_RIDS>(
-        simdProcessor, valuesWritten, validMinMax, ridOffset, dataVecTPtr, ridDstArray, writeMask, Min, Max,
+        simdProcessor, valuesWritten, validMinMax, ridOffset, dataVecTPtr, ridDstArray, writeMask, min, max,
         in, out, nonNullOrEmptyMask, ridArray);
 
+    if constexpr (KIND != KIND_TEXT)
```
drrtuy, line 1534: I will not force you to rewrite removing this if condition but as you can see you have two functions + additional if constexpr. If you just use SFINAE or concepts based template you would just have 2 specializations.

primitives/linux-port/column.cpp
```diff
@@ -1284,6 +1295,58 @@ inline SIMD_WRAPPER_TYPE simdSwapedOrderDataLoad(const ColRequestHeaderDataType
...
+{
+  constexpr const uint16_t size = VT::vecByteSize / sizeof(T);
+  T* simdMinVec = reinterpret_cast<T*>(&simdMin);
+  T* simdMaxVec = reinterpret_cast<T*>(&simdMax);
+  max = *std::max_element(simdMaxVec, simdMaxVec + size);
+  min = *std::min_element(simdMinVec, simdMinVec + size);
+}
+
+template <typename T, typename VT, typename SimdType>
+void getInitialSimdMinMax(VT& simdProcessor, SimdType& simdMin, SimdType& simdMax, T min, T max)
```
drrtuy: You are the functional guy not me an I would prefer to return a pair from this getInitialSimdMinMax(no side effects) and break it into to separate vars using the structured binding feature.

mariadb-AndreyPiskunov: Actually since all this function does is a simple assignment, maybe I'll just do it right in the vectorizedFiltering? This will also be more in line with other var declarations.

primitives/linux-port/column.cpp
mariadb-AndreyPiskunov: For me it feels like vectorizedTextUpdateMinMax is a bit different from vectorizedUpdateMinMax in how it does things, so I liked the idea of having functions with different logic (and different number of arguments) have different names.

primitives/linux-port/column.cpp
mariadb-AndreyPiskunov: Will do, thanks.

drrtuy (approved):

PR #2435, opened 2022-07-01
Mcol 5074 Case with In and aggregates asserts

davidhall586:
gwip->scsp wasn't set when buildPredicateItem() was called. buildPredicateItem() uses this object without checking for validity. This patch adds code to properly set the object in the case that it's NULL.


CMakeLists.txt
```diff
@@ -200,7 +200,7 @@ ELSE ()
     # Remove visibility flag for now as it breaks Ubuntu 18.05 and we need to
     # fix our libraries anyway
     STRING(REPLACE "-fvisibility=hidden" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
-    MY_CHECK_AND_SET_COMPILER_FLAG("-msse4.2" RELEASE RELWITHDEBINFO MINSIZEREL DEBUG)
+    MY_CHECK_AND_SET_COMPILER_FLAG("-msse4.2 -DBOOST_BIND_GLOBAL_PLACEHOLDERS" RELEASE RELWITHDEBINFO MINSIZEREL DEBUG)
```
mariadb-LeonidFedorov: Why do we need this flag in this PR?

mariadb-LeonidFedorov:
LGTM, but can we wait with merge till mtr tests fix?

mariadb-LeonidFedorov (approved):

CMakeLists.txt
davidhall586: I've removed this flag from the PR. However, it won't compile for me without it. THe new BOOST breaks itself.

PR #2444, opened 2022-07-09
[MCOL-4699] Add support for circular outer joins.

denis0x0D:


dbcon/joblist/jlf_tuplejoblist.cpp
```diff
@@ -1591,106 +1591,161 @@ bool addFunctionJoin(vector<uint32_t>& joinedTables, JobStepVector& joinSteps, s
   return added;
 }
 
-void collectCycles(JoinGraph& joinGraph, const JobInfo& jobInfo, TableInfoMap& tableInfoMap,
-                   uint32_t currentTable, uint32_t prevTable, JoinEdges& edgesToTransform, Cycles& cycles)
+// This class represents a circular inner join graph transformer.
+// It collects a cycles in the given join graph as disjoint set of the join edges, for each collected cycle
+// removes the join edge (if column statistics available it tries to remove join edge which could produce
+// highest intermidiate join result, otherwise just a random edge in a cycle) and adds that join edge to
```
drrtuy: typo "intermediate"

utils/loggingcpp/ErrorMessage.txt
```diff
@@ -22,7 +22,6 @@
 1000	ERR_MISS_JOIN	%1% not joined.
 1001	ERR_NON_SUPPORTED_FUNCTION	Function '%1%' isn't supported.
 1002	ERR_INCOMPATIBLE_JOIN	%1% incompatible column type specified for join condition.
-1003	ERR_CIRCULAR_JOIN	Circular joins are not supported.
```
drrtuy, line 25: I like this change the most :)

dbcon/joblist/jlf_tuplejoblist.cpp
```diff
@@ -3141,14 +3673,52 @@ void joinTablesInOrder(uint32_t largest, JobStepVector& joinSteps, TableInfoMap&
...
+        t1    t3
+
+       a. It's legal for CS to schedule this joins as one step join as follow:
+       join(small sides {t1, t3}, large side {t2})
+
+       b. Instead of generating two steps join as follow:
+       t1_t2 = join(small sides {t1}, large side {t2})
+       join(small sides {t1_t2}, large side {t3)}
+
+       In case of `a` we unable to implement a join with a multiple keys.
```
drrtuy, line 3691: multiple keys that differs for t1_t2 and t2_t3 joins, right?

dbcon/joblist/jlf_tuplejoblist.cpp
```diff
@@ -2374,10 +2903,12 @@ void matchEdgesInRowGroup(const JobInfo& jobInfo, const RowGroup& rg, JoinEdges&
   }
 
   std::vector<pair<uint32_t, uint32_t>> takenEdges;
-  for (const auto& edge : edgesToTransform)
+  for (const auto& joinEdgeWithWeight : edgesToRestore)
```
drrtuy: You might be interested in structured binding here to produce edge and weight as a separate variables available in the loop body.

dbcon/joblist/jlf_tuplejoblist.cpp
```diff
@@ -3356,6 +3932,10 @@ void joinTablesInOrder(uint32_t largest, JobStepVector& joinSteps, TableInfoMap&
         largeIndices.push_back(getKeyIndex(*k2, largeSideRG));
       }
 
+      // Try to restore `circular join edge` if possible.
+      tryToRestoreJoinEdges(jobInfo, info, largeSideRG, smallIndices, largeIndices, traces, joinIdIndexMap,
+                            smallSideIndex);
+      typeless.push_back(info->fJoinData.fTypeless);
```
drrtuy, line 3931: This is to allow the join to become typeless if we restored an edge that activates typeless join?

dbcon/joblist/jlf_tuplejoblist.cpp
denis0x0D: Thanks, fixed.

utils/loggingcpp/ErrorMessage.txt
denis0x0D: Yeah=)

dbcon/joblist/jlf_tuplejoblist.cpp
denis0x0D: I meant, that we could have a situation when we need a {t1, t2} to be on a small side as intemediate result of a previous join and t3 on a large side, in this case we can create a join for multiple keys, first keys represents an edge (t2, t3), and second keys represents an edge which we need to restore (t1, t3). But CS could apply an optimization and make a one large side {t2} and two small sides {t1}, {t3} for a single BPP, without intermediate joins, in this case we cannot restore edge (t1, t3) because those 2 rowgroups (tables) on a small side, 2 different hash maps created for them.

dbcon/joblist/jlf_tuplejoblist.cpp
denis0x0D: Right, to have a multiple keys join, the join type should be a `typeless`. This flag is set  in function `tryToRestoreJoinEdges` if we match keys in the smallside/largeside rowgroups.
```
// Mark as tupeless for multiple keys join.
 joinInfo->fJoinData.fTypeless = true;
```

dbcon/joblist/jlf_tuplejoblist.cpp
denis0x0D: Thanks, fixed!

drrtuy (approved):
Nicely done.

PR #2450, opened 2022-07-11
Mcol 5092 MODA uses wrong column width for some types

davidhall586:


utils/regr/modamysql.cpp
```diff
@@ -72,17 +75,16 @@ char* moda(CONTAINER& container, struct moda_data* data)
     }
   }
 
-  data->result = std::to_string(val);
-
-  return const_cast<char*>(data->result.c_str());
+  data->result = boost::lexical_cast<std::string>(val);
```
mariadb-LeonidFedorov: Did you consider using std::to_string instead? It's faster and do not use external deps

davidhall586: Yes, I had used to_string at first, but the output wasn't as nice as lexical cast for float types. We can switch it back. Float types are a bad choice for this function anyway.

datatypes/mcs_datatype.cpp
```diff
@@ -478,49 +478,49 @@ int TypeHandlerVarbinary::storeValueToField(rowgroup::Row& row, int pos, StoreFi
...
   int64_t val = row.getIntField<4>(pos);
-  return f->store_xlonglong(val);
+  return f->store_longlong(val);
 }
 
 int TypeHandlerInt::storeValueToFieldUInt32(rowgroup::Row& row, int pos, StoreField* f) const
 {
   uint64_t val = row.getUintField<4>(pos);
-  return f->store_xlonglong(static_cast<int64_t>(val));
+  return f->store_ulonglong(static_cast<int64_t>(val));
```
drrtuy: There is no difference b/w U64 and U32 - same Field::store_ulonglong() call ?

datatypes/mcs_datatype.h
```diff
@@ -953,7 +953,8 @@ class StoreField
   virtual int store_timestamp(int64_t val) = 0;
   virtual int store_string(const char* str, size_t length) = 0;
   virtual int store_varbinary(const char* str, size_t length) = 0;
-  virtual int store_xlonglong(int64_t val) = 0;
+  virtual int store_longlong(int64_t val) = 0;
+  virtual int store_ulonglong(int64_t val) = 0;
```
drrtuy: That is pretty strange interface when int64 is used to store an unsigned value.

utils/regr/moda.h
```diff
@@ -67,34 +70,77 @@ struct hasher
...
+  comparator(datatypes::Charset cs) {}
+
+  bool operator()(const T& lhs, const T& rhs) const
+  {
+    return lhs == rhs;
+  }
+};
+// A collation aware string comparator
+template <>
+struct comparator<std::string>
```
drrtuy, line 118: We are getting into the same trouble here producing another collation aware comparator. The only thing that stops me from suggesting ConstString here is the fact that ctor ConstString costs something. But I would rather use MariaDBHasher/MariaDBComparator then produce another Hasher/Comparator pair for a hashmap.

utils/regr/moda.h
```diff
@@ -105,30 +151,31 @@ struct ModaData : public UserData
...
 
   template <class T>
-  std::unordered_map<T, uint32_t, hasher<T> >* getMap()
+  std::unordered_map<T, uint32_t, hasher<T>, comparator<T> >* getMap()
   {
     if (!fMap)
     {
       // Just in time creation
-      fMap = new std::unordered_map<T, uint32_t, hasher<T> >;
+      fMap = new std::unordered_map<T, uint32_t, hasher<T>, comparator<T> >(
```
drrtuy, line 156: Could you ellaborate on what is this map for? If it will contain up to 100 entries flatmap might be more interesting in terms of memory consumption.

drrtuy (changes requested):

utils/regr/modamysql.cpp
davidhall586: Changed it back to to_string

datatypes/mcs_datatype.cpp
davidhall586: There is a difference between ulonglong and longlong. Even though we send field a signed, in the underlying field_store(), we need to pass the unsigned flag differently. The way the flag had been calculated in previous code was based on something that may or may not be true.

datatypes/mcs_datatype.h
davidhall586: The reason is the underlying field_>store(longlong nr, bool unsigned_val) takes a signed val and a flag indicating if it should be treated as signed or unsigned. We could have store_ulonglong() take an unsigned, but we'd then have to cast it to signed in the function. Or we could keep the store_xlonglong and pass the signed flag in there. All basically equivalent solutions.

utils/regr/moda.h
davidhall586: This map keys on the unique values of the result set with their respective counts. The result set can be 0 to infinity entries, limited by memory.

The algorithm is: for each value sent to the function from the query, look up that value in the map and increase it's count. If it isn't found, add with a count of 1. At the end, look for the entry with the largest count. If needed, apply tie-breakers.

utils/regr/moda.h
davidhall586: I'm not sure why this is a problem here. I need a template based hasher and comparator for the map. Simple specialization for string type using the charset.eq() operator and already existing CollationAwareHasher seems simple and extremely lightweight.

datatypes/mcs_datatype.cpp
davidhall586: We agreed that it is better to have the cast inside the store_ulonglong() function. 
Function is now store_ulonglong(uint64_t val);

mariadb-AndreyPiskunov (approved):

drrtuy (approved):

drrtuy:
@dhall-MariaDB You need to rebase the branch to be able to build at CI. Right now the branch can't be merged/build(take a look at CI).

PR #2472, opened 2022-07-22
MCOL-5162 This patch allows for an automatic upgrade of syscat tables via dbbuilder.

tntnatbry:
We have added a new function, SystemCatalog::upgrade(), which will
create a new column in the database and fill it with default values.
It uses a reference column to fill the column, which is passed as a
std::unordered_map<key, value> value to SystemCatalog::upgrade()
function, where key is the target column which needs to be created.
In addition, we pass additional std::unordered_map's to
SystemCatalog::upgrade() containing information such as the datatypes
and colwidths of the target and reference columns, as well as the
default value which will be used to fill the target column.

### DIFF
--- a/tools/dbbuilder/dbbuilder.cpp
+++ b/tools/dbbuilder/dbbuilder.cpp
@@ -189,26 +189,62 @@ int main(int argc, char* argv[])
     if (access(logFile.c_str(), W_OK) != 0)
       canWrite = false;
 
+    // MCOL-5162 Automatic syscat upgrade
+    // The below std::unordered_map's are used by the
+    // SystemCatalog::upgrade() call.
+    bool isUpgrade = false;
+
+    std::unordered_map<int, std::pair<int, bool>> upgradeOidMap;
+    upgradeOidMap[OID_SYSTABLE_AUXCOLUMNOID] =      // This is the candidate OID for the upgrade.
+      std::make_pair(OID_SYSTABLE_OBJECTID, false); // std::pair::first is the reference OID used
+                                                    // to fill the candidate OID with default vals
+                                                    // std::pair::second is set to false by default
+                                                    // which means the candidate OID will not be
+                                                    // upgraded. This is set to true in the code
+                                                    // below if a specific condition is met. Please
+                                                    // note that the candidate and reference OID
+                                                    // datatypes and colwidths are assumed to be the
+                                                    // same in SystemCatalog::upgrade().
+
+    std::unordered_map<int, OidTypeT> upgradeOidTypeMap;
+    upgradeOidTypeMap[OID_SYSTABLE_AUXCOLUMNOID] =
+      std::make_pair(CalpontSystemCatalog::INT, 4);
+
+    std::unordered_map<int, std::string> upgradeOidDefaultValStrMap;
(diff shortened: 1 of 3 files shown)


tntnatbry:
I have tagged this PR as WIP as it has a dependency on the merge of https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/2406. Lines containing the symbol `OID_SYSTABLE_AUXCOLUMNOID` need to be uncommented once https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/2406 is merged.

tools/dbbuilder/dbbuilder.cpp
```diff
@@ -189,26 +189,67 @@ int main(int argc, char* argv[])
...
+    std::unordered_map<int, std::pair<CalpontSystemCatalog::ColDataType, int>> upgradeOidTypeMap;
+//    upgradeOidTypeMap[OID_SYSTABLE_AUXCOLUMNOID] =
+//      std::make_pair(CalpontSystemCatalog::INT, 4);
+    upgradeOidTypeMap[OID_SYSCOLUMN_COMPRESSIONTYPE] =
+      std::make_pair(CalpontSystemCatalog::INT, 4);
+
+    std::unordered_map<int, std::string> upgradeOidDefaultValStrMap;
+//    upgradeOidDefaultValStrMap[OID_SYSTABLE_AUXCOLUMNOID] = "0";
+    upgradeOidDefaultValStrMap[OID_SYSCOLUMN_COMPRESSIONTYPE] = "0";
+
```
tntnatbry, line 215: @benthompson15 
To add support for the `collation` column, the change is as simple as adding an entry to the above three unordered_map's. Read the above comment as well as the commit description. I added the `OID_SYSCOLUMN_COMPRESSIONTYPE` entry just as a demonstration, this should be replaced with the entry for the `collation` columns in `systable` and `syscolumn`.

drrtuy:
@tntnatbry Is it still WIP ?

tntnatbry:
@drrtuy Yes. See my previous comment regarding lines containing `OID_SYSTABLE_AUXCOLUMNOID` that need to be uncommented once #2406 is merged. This PR has a dependency on the merge of #2406.

drrtuy (commented):
There are 2 things I would like to have answers for: 
- I would suggest to turn unreadable std::pair<CalpontSystemCatalog::ColDataType, int>> into something meaningful, e.g. using TypeOidT = std::pair<CalpontSystemCatalog::ColDataType, int>> to simplify the reading.
- Does the current dbbuilder code creates new columns when there is no syscat(new installation) or it doesn't and this is for a reason?

tntnatbry:
@drrtuy 
1. Sure. I will add a type alias for this.
2. Yes. The current dbbuilder code checks for the existence of the database file for OID 1001. If it doesn't exist, then it creates all columns for `calpontsys.systable` and `calpontsys.syscolumn`.

drrtuy:
Just to remind you @tntnatbry you were going to replace long type names with a simpler type aliases.

tntnatbry:
@drrtuy This is ready for merge. I have added a type-alias as you requested.

drrtuy (approved):

drrtuy (approved):

PR #2475, opened 2022-07-25
 [MCOL-5167] Add support for on clause filter for a table which is not involved in join.

denis0x0D:
This patch adds support for on clause filter for a table which is not involved in particular join
by disabling an `merge optimization` for those particular cases.
The `merge optimization` is optimization when CS
tries to create a one BPP join with one `large side` table and multiple `small sides` tables, in this
case we cannot apply a FE filter if this filter requires a columns from `small side` table which is not
involved in particular join.


mariadb-DenisKhalikov:
Associated fix for regression tests: https://github.com/mariadb-corporation/mariadb-columnstore-regression-test/pull/303

drrtuy (commented):
The patch LGTM so I will restart CI build to see if it passes.

PR #2477, opened 2022-07-26
MCOL-5166 This patch adds support for in-memory communication b/w EM …

drrtuy:
…to PP via a shared queue in DEC class

JobList low-level code relateod to primitive jobs now uses shared pointers instead of ByteStream refs talking to DEC
b/c same-node EM-PP communication now goes over a queue in DEC instead of a network hop.
PP now has a separate thread that processes the primitive job messages from that DEC queue.

### DIFF
--- a/dbcon/joblist/distributedenginecomm.cpp
+++ b/dbcon/joblist/distributedenginecomm.cpp
@@ -38,6 +38,7 @@
 #include <unistd.h>
 #include <chrono>
 #include <thread>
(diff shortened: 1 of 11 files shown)


denis0x0D (approved):
LGTM

denis0x0D:
@drrtuy you have to change the PR and commit tag to MCOL-5166

mariadb-LeonidFedorov (approved):

PR #2479, opened 2022-07-28
MCOL-5106 Port patches for develop from develop-6 related to rebuildEM tool.

denis0x0D:


mariadb-DenisKhalikov:
Some test failed and error in warning message, but I cannot reproduce it locally:
```
Logging: ./mtr  --force --suite=columnstore/basic mcs78_aliases.test
VS config: 
vardir: /usr/share/mysql/mysql-test/var
Checking leftover processes...
Removing old var directory...
Creating var directory '/usr/share/mysql/mysql-test/var'...
Checking supported features...
MariaDB Version 10.9.0-MariaDB
 - SSL connections supported
Collecting tests...
Installing system database...

==============================================================================

TEST                                      RESULT   TIME (ms) or COMMENT
--------------------------------------------------------------------------

worker[1] Using MTR_BUILD_THREAD 300, with reserved ports 16000..16019
worker[1] mysql-test-run: WARNING: running this script as _root_ will cause some tests to be skipped
columnstore/basic.mcs78_aliases          [ pass ]   4123
--------------------------------------------------------------------------
The servers were restarted 0 times
Spent 4.123 of 6 seconds executing testcases

Completed: All 1 tests were successful.
```

tools/rebuildEM/main.cpp
```diff
@@ -41,6 +40,20 @@ static void usage(const string& pname)
...
 }
 
+static bool isYes()
+{
+  std::string confirmation;
+  std::cin >> confirmation;
+  if (confirmation.size() == 0)
+    return false;
+
+  boost::algorithm::to_lower(confirmation);
```
drrtuy, line 50: JFYI to_lower uses locale internally. To access the locale it touches a mutex that is shared across all to_lower invocations. It is safe to use it here b/c it is a one shoot call but I would discourage from using it.

drrtuy (approved):

PR #2488, opened 2022-08-04
MCOL-5153 This patch replaces MDB collation aware hash function with the

drrtuy:
exact functionality that does not use MDB hash function.
This patch also takes a bit from Robin Hood hash map implementation forgotten
that reduces hash function collision rate.

### DIFF
--- a/utils/rowgroup/rowstorage.h
+++ b/utils/rowgroup/rowstorage.h
@@ -36,6 +36,9 @@ class RowGroupStorage;
 
 uint64_t hashRow(const rowgroup::Row& r, std::size_t lastCol);
 
+constexpr const size_t MaxConstStrSize = 2048ULL;
+constexpr const size_t MaxConstStrBufSize = MaxConstStrSize << 1;
+
 class RowAggStorage
 {
  public:
@@ -161,6 +164,9 @@ class RowAggStorage
    */
   inline void rowHashToIdx(uint64_t h, uint32_t& info, size_t& idx, const Data* curData) const
   {
+    // An addition from the original robin hood HM.
+    h *= fCurData->hashMultiplier_;
+    h ^= h >> 33U;
     info = curData->fInfoInc + static_cast<uint32_t>((h & INFO_MASK) >> curData->fInfoHashShift);
     idx = (h >> INIT_INFO_BITS) & curData->fMask;
   }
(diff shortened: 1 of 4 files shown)


utils/rowgroup/rowstorage.h
```diff
@@ -230,6 +236,13 @@ class RowAggStorage
     info = fCurData->fInfo[idx];
   }
 
+  void nextHashMultiplier()
+  {
+      // adding an *even* number, so that the multiplier will always stay odd. This is necessary
+      // so that the hash stays a mixing function (and thus doesn't have any information loss).
+      fCurData->hashMultiplier_ += 0xc4ceb9fe1a85ec54;
```
denis0x0D, line 243: Is this some theorem about using this constant, can you please clarify from did you take it?

utils/rowgroup/rowstorage.h
```diff
@@ -161,6 +164,9 @@ class RowAggStorage
    */
   inline void rowHashToIdx(uint64_t h, uint32_t& info, size_t& idx, const Data* curData) const
   {
+    // An addition from the original robin hood HM.
+    h *= fCurData->hashMultiplier_;
+    h ^= h >> 33U;
```
denis0x0D, line 169: Do they have some explanation on "h >> 33U" why do they do it?

utils/rowgroup/rowstorage.h
```diff
@@ -36,6 +36,9 @@ class RowGroupStorage;
 
 uint64_t hashRow(const rowgroup::Row& r, std::size_t lastCol);
 
+constexpr const size_t MaxConstStrSize = 2048ULL;
+constexpr const size_t MaxConstStrBufSize = MaxConstStrSize << 1;
```
denis0x0D, line 40: Can you please put a comment why is the buffer twiсe large than the string?

utils/rowgroup/rowstorage.h
drrtuy: The constants for hash map are usually picked up and not calculated. It is borrowed from robin hood hashmap impl [here](https://github.com/martinus/robin-hood-hashing/blob/master/src/include/robin_hood.h#L2494).

utils/rowgroup/rowstorage.h
drrtuy: Picked up from robin hood as well.

utils/rowgroup/rowstorage.h
drrtuy: Just in case that certain encoding technically can have 2 bytes of weights for every encoded string byte array.

denis0x0D (approved):
Since some magic part is a port from robin hood implementation, I guess they have match more experience in number theory, than me. Other parts of the code looks good to me, thanks.

mariadb-LeonidFedorov (approved):

PR #2491, opened 2022-08-05
WIP vectorize str  comparisions in p_Dicitonary and p_tokenByScan

NTH19:
- use weight array to replace str comparisions 
- Transform string into  weight array before hashing in eqFilter


mariadb-LeonidFedorov:
please apply clang-format on your code

PR #2496, opened 2022-08-09
Mcol 5092 MODA uses wrong column width for some types

mariadb-LeonidFedorov:


davidhall586:
This has been incorporated in the original PR

PR #2501, opened 2022-08-12
build boost during build phase, not during configure

vuvova:


vuvova:
unfortunate branch name, will recreate

PR #2506, opened 2022-08-15
MCOL-5092 addendum modify moda() mtr tests with order by

davidhall586:
also fix an issue in moda() evaluate() for string types


utils/regr/moda.cpp
```diff
@@ -589,7 +589,7 @@ mcsv1_UDAF::ReturnCode Moda_impl_T<string>::evaluate(mcsv1Context* context, stat
     else if (iter->second == maxCnt)
     {
       // Tie breaker: choose smallest according to collation
-      if (cs.strnncollsp(val, lastVal) < 0)
+      if (cs.strnncollsp(iter->first, lastVal) < 0)
```
drrtuy, line 592: Is it really the smallest according to collation?
I mean when one gets an iterator over a hashmap nobody guarantees an order. Is the comment is outdated or I am missing something?

davidhall586: Yes, it's smallest according to collation. That's what strnncollsp is for. We're walking the map and comparing counts. If the largest count to this point is the same as the count of the current item, then check collation to determine which should be kept as "largest". maxCnt is the largest count found so far and val is the value for the largest so far. The correction here is to fix a stupid mistake.

Test show that this works as expected.

drrtuy (approved):

PR #2507, opened 2022-08-15
MCOL-4715 Mixed inner and outer joins with "null" filter for the table which is not involved into the outer join produces wrong results.

denis0x0D:


denis0x0D:
Check mtr on CI.

drrtuy (approved):

PR #2522, opened 2022-08-22
Mcol 5196 replace triggers invalid capacity assertion

mariadb-SergeyZefirov:
When length of string to replace minus length of string to replace to is
bigger than input string and processing mode allows for binary (memcmp
or std::string::find()) comparison, REPLACE may trigger invalid capacity
assertion and query processing will stop.

The fix is to properly count the number of occurences of the string to replace, basically.


mariadb-LeonidFedorov:
This patch doesn't pass mtr replace test

mariadb-LeonidFedorov (approved):

PR #2534, opened 2022-08-25
Remove getBinaryField

mariadb-AndreyPiskunov:
Trying to remove getBinaryField and replace it with getTSInt128Field.


mariadb-LeonidFedorov:
This work was updated in other PR

PR #2551, opened 2022-08-31
Columnstore 22.08.1 merge back

davidhall586:


davidhall586:
There's a conflict in the way serg added arm support vs how it got done recently in develop. I'm not sure which is more correct. It won't drone build until the conflict is resolved.

PR #2554, opened 2022-09-01
MCOL-5188 Improved accuracy for regr_* functions

mariadb-AndreyPiskunov:
This is the online algorithm for `regr_sxy`. It does 2 things: improves precision of `regr_sxy` and seemingly resolves the issues with different test results for arm and x86 in regression.

PS. I got rid of the function dropValue, as it wasn't ever used anywhere and I wasn't sure how to add it to this algorithm.


davidhall586:
dropValue is needed in order for the function to work as a Window Function. It should reverse the effect of nextValue for the specified value. Please add this back in.

drrtuy (approved):
LGTM

mariadb-LeonidFedorov (changes requested):
Arm regression should be fixed

mariadb-LeonidFedorov (approved):

PR #2560, opened 2022-09-08
Unified AUX column scanning at ARM and x86 to re-enable vectorization at ARM.

drrtuy:


primitives/linux-port/column.cpp
```diff
@@ -144,36 +121,102 @@ constexpr getNonEmptyMaskPtrT getNonEmptyMaskPtrTemplate()
...
+  {
+    const char* ptr = reinterpret_cast<const char*>((uint64_t*)nonEmptyMaskAux + iter);
+    return proc.loadFrom(ptr);
+    // return MT{ptr[0], ptr[1], ptr[2], ptr[3], ptr[4], ptr[5], ptr[6], ptr[7]};
+    // return MT{0x0, 0x0};
+  }
+  else if constexpr (sizeof(T) == sizeof(uint32_t) && !std::is_same<T, float>::value)
+  {
+    const char* ptr = reinterpret_cast<const char*>((uint32_t*)nonEmptyMaskAux + iter);
+    return proc.loadFrom(ptr);
```
NTH19: Hey,I think loadFrom will not work right.For uint32_t,only first 32 bits from ptr are useful.We need to turn 0xFF into 0xFFFFFFFF in each element of mask because blend needs it.The loadFrom directly loads 128 bits from ptr.I think it maybe causes some problems.And arm has passed the  mtr testes(Before I fixed the error in aux column,it didn't pass it). So you don't need to consider  the correctness on aux column in arm .

drrtuy: Hey.
You are right indeed. I was going to add aux column filtering test first and then actually fix aux column processing.

NTH19 (commented):
LGTM

drrtuy:
Waiting for Kirill's patch to rebase and merge.

mariadb-LeonidFedorov (approved):

PR #2566, opened 2022-09-15
MCOL-5191 Add MCV statistics.

denis0x0D:
This patch adds:
1. Initial version of random sampling.
2. Initial version of MCV statistics.

### DIFF
--- a/primitives/primproc/sqlfrontsessionthread.cpp
+++ b/primitives/primproc/sqlfrontsessionthread.cpp
@@ -187,7 +187,8 @@ namespace exemgr
     fIos.write(emsgBs);
   }
 
-  void SQLFrontSessionThread::analyzeTableExecute(messageqcpp::ByteStream& bs, joblist::SJLP& jl, bool& stmtCounted)
+  void SQLFrontSessionThread::analyzeTableExecute(messageqcpp::ByteStream& bs, joblist::SJLP& jl,
+                                                  bool& stmtCounted)
   {
     auto* statementsRunningCount = globServiceExeMgr->getStatementsRunningCount();
     messageqcpp::ByteStream::quadbyte qb;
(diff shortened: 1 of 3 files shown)


primitives/primproc/sqlfrontsessionthread.cpp
```diff
@@ -238,19 +239,23 @@ namespace exemgr
...
-    msgHandler.stop();
-
-    auto outRG = (static_cast<joblist::TupleJobList*>(jl.get()))->getOutputRowGroup();
 
-    if (caep.traceOn())
-      std::cout << "Row count " << rowCount << std::endl;
-
-    // Process `RowGroup`, increase an epoch and save statistics to the file.
+    // Seemls like this a legacy parameter, not really needed.
+    const uint32_t dummyTableOid = 100;
```
drrtuy, line 245: It is indeed. Its semantics was important back in Columnstore prior to MCS 1.4.

utils/common/statistics.cpp
```diff
@@ -31,61 +32,135 @@ using namespace logging;
...
+    // MCV statistics.
+    std::vector<pair<uint64_t, uint32_t>> mcvList(columnMCV.begin(), columnMCV.end());
+    std::sort(mcvList.begin(), mcvList.end(),
+              [](const std::pair<uint64_t, uint32_t>& a, const std::pair<uint64_t, uint32_t>& b) {
+                return a.second > b.second;
+              });
+
+    // 200 buckets as Microsoft does.
+    const auto mcvSize = std::min(columnMCV.size(), static_cast<uint64_t>(200));
+    mcv[oid] = std::unordered_map<uint64_t, uint32_t>(mcvList.begin(), mcvList.begin() + mcvSize);
```
drrtuy, line 131: I would name a type for readability.

drrtuy:
I missed where how exactly do you treat/filter out char/varchar/text dict columns?

denis0x0D:
@drrtuy the char/varchar is not supported because it was not supported in the original pipeline, that means current analyze table command does not generate "job steps" to read the dictionary data. Probably that's what needs to be updated in analyze command.

drrtuy (approved):

PR #2568, opened 2022-09-16
[WIP] MCOL-5215 Fix bug from UNION on DECIMAL 

JigaoLuo:
Signed-off-by: Jigao Luo <luojigao@outlook.com>

The Jira issue number for this PR is: [MCOL-5215](https://jira.mariadb.org/browse/MCOL-5215)

# Task

- [x] Both sides DECIMAL with same signedness and precision.
- [ ] Both sides DECIMAL with same signedness and different precision.
  - [ ] Error when values saturated
- [ ] Both sides DECIMAL with different signedness and same precision.
- [ ] Both sides DECIMAL with different signedness and different precision.
- [ ] One side DECIMAL and other side integer type. Again, there should be variation in test cases in DECIMAL precision, integer types, and signedness.
- [ ] Tests on UNION and UNION ALL


drrtuy:
Hey @JigaoLuo.
Are you going to continue with this patch?

drrtuy:
TBD later.

PR #2572, opened 2022-09-21
MCOL-5225: Querytester build ad hoc

mariadb-LeonidFedorov:


drrtuy:
This is a valuable patch but we first have to resolve test001 issue for ARM.

drrtuy:
If we merge it right away this will break nightly-develop-6.

PR #2575, opened 2022-09-23
ci: temporary disable 10.9 community builds

mariadb-RomanNavrotskiy:


drrtuy:
We will follow another way (C)

PR #2586, opened 2022-10-13
MCOL-1170 Fix ANALYZE to not error

davidhall586:
Analyze needs to be completed differently than a normal query. In server, when an ANALYZE <statement> is seen, it calls init_scan() immediatly followed by end_scan(). This leaves the sqlfrontendsession (ExeMgr) in a state where it expects to return rows. This patch fixes end_scan to clean this up via reads and writes to get everything back in synch.

ANALYZE should display the number of rows to be displayed if the query were run normally. We have that information available, but no way to return it. A modification to server side to ask for that in the handler is required.

This patch also includes a beautification of sqlfrontsessionthread.cpp since it looked bad. The important change is at line 774
   if (!swallowRows)
which short circuits the actual return of data


primitives/primproc/sqlfrontsessionthread.cpp
```diff
@@ -21,381 +21,384 @@
...
+  int                     destructing                 = 0;
+  int                     gDebug                      = globServiceExeMgr->getDebugLevel();
+  logging::Logger&        msgLog                      = globServiceExeMgr->getLogger();
 
-  bool selfJoin = false;
-  bool tryTuples = false;
-  bool usingTuples = false;
-  bool stmtCounted = false;
-  auto* statementsRunningCount = globServiceExeMgr->getStatementsRunningCount();
+  bool                    selfJoin                    = false;
```
mariadb-DenisKhalikov: It seems like this format is different comparing to "clang-format: style we have.

mariadb-DenisKhalikov (approved):
Looks good to me.
Some question according code format changes - it seems like those changes are different comparing "clang-format" style we have.
Thanks.

primitives/primproc/sqlfrontsessionthread.cpp
davidhall586: I thought it funny as well. I will reload the formatter and see what happens

drrtuy:
Don't forget to squash before the merge.

PR #2588, opened 2022-10-14
MCOL-5000 Disable ALTER TABLE statement execution on replicas.

tntnatbry:
Exit early from the plugin execution of ALTER TABLE statements on the replica nodes. This is to prevent re-execution of syscat table population from the replica nodes which should only be executed once by the primary node in a CS cluster setup.

### DIFF
--- a/dbcon/mysql/ha_mcs_impl.cpp
+++ b/dbcon/mysql/ha_mcs_impl.cpp
@@ -2206,7 +2206,9 @@ int ha_mcs::impl_rnd_init(TABLE* table, const std::vector<COND*>& condStack)
   gp_walk_info gwi(timeZoneOffset);
   gwi.thd = thd;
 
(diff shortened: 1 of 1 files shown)


drrtuy:
impl_rnd_init for ALTER table, really?

drrtuy:
It looks like all ARM builds got stuck running MTR step.

tntnatbry:
Yes I was also a bit surprised when I noticed ALTER TABLE going through rnd_init(), rnd_next(), rnd_end(). I was thinking ALTER TABLE would go through ha_mcs_impl_pushdown_init(), but there is an early exit for ALTER TABLE in ha_mcs_impl_pushdown_init():
```c++
  if ((thd->lex)->sql_command == SQLCOM_ALTER_TABLE)
  {
    return 0;
  }
```
Not sure why this is the case.

drrtuy (approved):

PR #2592, opened 2022-10-20
MCOL-5227

mariadb-AndreyPiskunov:
Added tests to basic as well.


dbcon/joblist/jsonarrayagg.h
```diff
@@ -0,0 +1,174 @@
+/* Copyright (C) 2014 InfiniDB, Inc.
```
drrtuy: 2014, really?

utils/rowgroup/rowaggregation.cpp
```diff
@@ -3989,6 +4015,16 @@ void RowAggregationUM::setGroupConcatString()
...
       }
+
+      if (fFunctionCols[j]->fAggFunction == ROWAGG_JSON_ARRAY)
+      {
+        uint8_t* buff = data + fRow.getOffset(fFunctionCols[j]->fOutputColumnIndex);
+        uint8_t* gcString;
+        joblist::JsonArrayAggregatAgUM* gccAg = *((joblist::JsonArrayAggregatAgUM**)buff);
+        gcString = gccAg->getResult();
+        fRow.setStringField((char*)gcString, fFunctionCols[j]->fOutputColumnIndex);
+        // gccAg->getResult(buff);
```
drrtuy: Do we need this commented line?

dbcon/joblist/jsonarrayagg.cpp
```diff
@@ -0,0 +1,1079 @@
...
+
+        break;
+      }
+
+      case CalpontSystemCatalog::CHAR:
+      case CalpontSystemCatalog::VARCHAR:
+      case CalpontSystemCatalog::TEXT:
+      {
+        std::string maybeJson = row.getStringField(*i);
+        try
```
drrtuy: This try-catch doesn't look right and must be changed. It takes a lot to process every catch b/c it is a combination of internal interrupt + catch block processing.
There should be no try-catch that allows to go on processing in a low-level processing loop code. And outputRow is called for every record.

drrtuy (changes requested):
Plz consider inherit JsonArrayInfo from GroupConcatInfo.

utils/rowgroup/rowaggregation.cpp
mariadb-AndreyPiskunov: Will remove it and several other of my forgotten comments.

dbcon/joblist/jsonarrayagg.cpp
mariadb-AndreyPiskunov: Fortunately, I found a flag for this function to not throw exceptions. Thanks.

dbcon/joblist/jsonarrayagg.h
mariadb-AndreyPiskunov: Sorry, I always forget what year it is)

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -2904,6 +2905,14 @@ uint32_t setAggOp(AggregateColumn* ac, Item_sum* isp)
       return rc;
     }
 
+    case Item_sum::JSON_ARRAYAGG_FUNC:
+    {
+      Item_func_group_concat* gc = (Item_func_group_concat*)isp;
```
drrtuy: JSON_ARRAYAGG_FUNC uses Item_func_group_concat?

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -4909,7 +4918,8 @@ ReturnedColumn* buildAggregateColumn(Item* item, gp_walk_info& gwi)
   // Argument_count() is the # of formal parms to the agg fcn. Columnstore
   // only supports 1 argument except UDAnF, COUNT(DISTINC) and GROUP_CONCAT
```
drrtuy: It is worth to align the comment text also.

drrtuy (approved):

drrtuy:
There is one thing left, namely the MTR test that is introduced with the patch fails to pass b/c of a randomly sorted output. See the build for details.

drrtuy:
Look for sorted_results stanza [here](https://dev.mysql.com/doc/dev/mysql-server/latest/PAGE_MYSQL_TEST_COMMANDS.html).

drrtuy (changes requested):
Take a look at the last comment.

dbcon/mysql/ha_mcs_execplan.cpp
mariadb-AndreyPiskunov: Done

dbcon/mysql/ha_mcs_execplan.cpp
mariadb-AndreyPiskunov: My bad, missed it.

drrtuy:
LGTM

drrtuy (approved):

PR #2593, opened 2022-10-21
Adding NULL flag into ConstString class

drrtuy:


mariadb-SergeyZefirov (approved):
I generally agree, except that mStr can hold nullptr and that may indicate NULL in SQL sense. This means that isNull_ member is not strictly necessary.

It is necessary to assure that mLength is zero when mStr is equal to nullptr, this way all operations can be done blindly.

mariadb-SergeyZefirov:
Please disregard my comment on my approval, it was placed by accident and is of concern to the older version of code.

I completely approve changes.

PR #2596, opened 2022-10-24
ci: try community server 10.11

mariadb-RomanNavrotskiy:


drrtuy:
We will implement CS 10.11 pipelines in a separate build.

PR #2597, opened 2022-10-24
MCOL-603 Add error message for sum(a=1)

davidhall586:
This isn't currently supported, but rather than emitting an error, it asserted and crashed.

### DIFF
--- a/dbcon/mysql/ha_mcs_execplan.cpp
+++ b/dbcon/mysql/ha_mcs_execplan.cpp
@@ -5116,13 +5116,6 @@ ReturnedColumn* buildAggregateColumn(Item* item, gp_walk_info& gwi)
             {
               rc = buildFunctionColumn(ifp, gwi, gwi.fatalParseError);
             }
-#if 0
-            if (!rc)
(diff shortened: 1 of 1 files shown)


dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -5113,17 +5113,24 @@ ReturnedColumn* buildAggregateColumn(Item* item, gp_walk_info& gwi)
 
             // @bug 3603. for cases like max(rand()). try to build function first.
             if (!rc)
+            {
               rc = buildFunctionColumn(ifp, gwi, gwi.fatalParseError);
-
+            }
+#if 0
```
drrtuy: Sorry but I missed the reason why we need this commented out block here.

mariadb-AndreyPiskunov (approved):

dbcon/mysql/ha_mcs_execplan.cpp
davidhall586: removed the commented code

PR #2598, opened 2022-10-28
MCOL-5248 Change func_truncate() to use double for string

davidhall586:
rather than attempt to translate to decimal. Currently, the treenode.h conversion functions don't support string to decimal conversion, so there's no loss of functionality.

This new functionality brings us into alignment with MDB 10.6

When we add the ability to convert from string into decimal in TreeNode::getDecimalVal(), this should be re-evaluated.

I added an error rather than saturation to Func_truncate::getDecimalVal(). I believe this is preferred. However, since getDoubleVal() now doesn't call getDecimalVal() for string types (It used to), I don't know if this will ever come into play.


mariadb-LeonidFedorov:
Hi you should look at test
Compare failed - working_tpch1_compareLogOnly/misc/bug3483.sql

mariadb-LeonidFedorov (approved):

PR #2602, opened 2022-11-02
MDEV-25080 Add dummy support for UNION pushdown in the MCS 6.x plugin.

tntnatbry:
MCS 6.x does not support full/partial pushdown of UNIONs in outer select as implemented in MDEV-25080. This patch makes the bare minimum changes to the MCS 6.x plugin interface to make mariadb community server 10.11 compile with MCS 6.x.


drrtuy:
The server implemented the patch in ES 10.6 so this patch doesn't make sense unfortunately.

PR #2605, opened 2022-11-03
MCOL-5279 This commit fixes the corner case of the issue  that can ra…

drrtuy:
…rely happen when flow control is used to backpressure PPs that overflows EM with Primitive messages. It also adds a relevant test that makes sense in a multi-node scenario only.


davidhall586 (approved):

davidhall586 (commented):
Since this mtr case uses RAND, the result will be different every time and will not compare to .result

PR #2607, opened 2022-11-03
ci: add server branch, upd platforms

mariadb-RomanNavrotskiy:


drrtuy:
Unfortunately this won't fly b/c CS 10.11 was sent out with develop-6.

PR #2613, opened 2022-11-07
Mtrfix kp 221104 d6 ar

KirillPerov:


mysql-test/columnstore/basic/t/mcs118_charset_negative.test
```diff
@@ -12,18 +12,23 @@ CREATE DATABASE mcs118_db DEFAULT CHARACTER SET latin5;
 USE mcs118_db;
 
 CREATE TABLE t1 (a VARCHAR(10) CHARACTER SET latin1, b VARCHAR(10) CHARACTER SET utf8)ENGINE=Columnstore;
+--replace_regex /COLLATE latin1_swedish_ci // /COLLATE utf8mb3_general_ci // / COLLATE=latin5_turkish_ci//
 SHOW CREATE TABLE t1;
 
---error ER_CONFLICTING_DECLARATIONS
+--disable_warnings
+--disable_abort_on_error
+--replace_regex /42000/HY000/ /The storage engine for the table doesn't support The syntax or the data type\(s\) is not supported by Columnstore. Please check the Columnstore syntax guide for supported syntax or data types./Conflicting declarations: 'CHARACTER SET utf8mb3' and 'CHARACTER SET latin1'/
```
mariadb-SergeyZefirov, line 20: Is it right? This completely changes meaning of message.

mariadb-SergeyZefirov:
I have comments on what is changed here.

You changed messages to be in line with expected results from the tests. I am not quite OK with that, because some of the different messages can be genuine bugs or official behavior.

I would suggest to change expected results so that we do not have old incorrect behavior lingering.

mysql-test/columnstore/basic/t/mcs118_charset_negative.test
KirillPerov: Well, the purpose of these changes is to manage 2 exact different results (from different versions of mdb).
I have no other idea how to manage it.

mariadb-SergeyZefirov: After having some more thoughts, I am okay with that. Other approaches are more work-heavy.

drrtuy (approved):

PR #2617, opened 2022-11-10
Add ability to compile against MariaDB with new cost model

spetrunia:
If MariaDB defines MARIADB_NEW_COST_MODEL, then ha_mcs::scan_time() has a different signature.


drrtuy (approved):

drrtuy:
Let's wait for CI to converge.

PR #2627, opened 2022-11-14
Related to UNION pushdown Mtr fix kp 221111 columnstore 22.08.4 1 v2

KirillPerov:


drrtuy:
These changes will be used later when Gagan pushes UNION pushdown patch into develop.

PR #2637, opened 2022-11-24
MCOL-5264 This patch replaces boost mutex locks with std analogs

drrtuy:
boost::uniqie_lock dtor calls a fancy unlock logic that throws twice. First if the mutex is 0 and second lock doesn't own the mutex. The first condition failure causes unhandled exception for one of the clients in DEC::writeToClient(). I was unable to find out why Linux can have a 0 mutex and replaced boost::mutex with std::mutex b/c stdlibc++ should be more stable comparing with boost.


denis0x0D:
LGTM

mariadb-DenisKhalikov (approved):

PR #2638, opened 2022-11-24
MCOL-5302: Fix brm-save to prevent overwrites Extent Map files multiple times with shared(non-S3) storage setup.

mariadb-AlanMologorsky:
- [fix] oam/install_scripts/mcs-savebrm.py.in , now everytime using method to detect primary node
- [fix] condition style in oam/install_scripts/mcs-loadbrm.py.in


mariadb-AlanMologorsky:
@drrtuy PLS DNM while not tested with upcoming PR build.

drrtuy (approved):

PR #2642, opened 2022-12-02
MCOL-5306 Re-read the config (Columnstore.xml) file if it was updated.

tntnatbry:
The existing implementation of Config::makeConfig() factory method was returning a possibly stale config to the caller, without checking if the config file was updated since the last read. This bug triggered a scenario as described in MCOL-5306 where after a failover in an MCS cluster, the controllernode coordinates changed in the config file after failover and the existing mariadbd process was still using the old controllernode coordinates. This lead to failed network connection between mariadbd and the new controllernode.

The change in this fix, however, is more generic and not just limited to this above scenario.

### DIFF
--- a/utils/configcpp/configcpp.cpp
+++ b/utils/configcpp/configcpp.cpp
@@ -83,6 +83,21 @@ boost::mutex Config::fXmlLock;
 boost::mutex Config::fWriteXmlLock;
 std::atomic_bool globHasConfig;
 
+void Config::checkAndReloadConfig()
+{
+  struct stat statbuf;
+
+  if (stat(fConfigFile.c_str(), &statbuf) == 0)
+  {
+    if (statbuf.st_mtime != fMtime)
+    {
+      closeConfig();
+      fMtime = statbuf.st_mtime;
+      parseDoc();
+    }
+  }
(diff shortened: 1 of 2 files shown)


drrtuy (approved):
Nicely found Gagan.
I can accept this as a bug fix but in the long-term someone from the team must stop re-reading config all the way b/c makeConfig is called is many places and sometimes with a high frequency, e.g. once per block of values. This contributes to CPU time spent in system a lot IMHO.

tntnatbry:
Thanks @drrtuy for the feedback. I agree that re-reading the whole config is inefficient. However, if you look at the `checkAndReloadConfig()` function, the actual read is protected by the if condition:

```c++
if (stat(fConfigFile.c_str(), &statbuf) == 0)
{
}
```
So we only read if the config has actually be updated. In most cases this code block should not execute. We can however add the `UNLIKELY` branch prediction optimization here.

drrtuy:
The thing is that re-reading is not important here. FIrst the contents is buffered and comes from memory so it is relatively fast. However this stat() issues a syscall + it is behind the mutex that also uses syscall under the hood.

PR #2649, opened 2022-12-07
MCOL-5311 Add timezone to jobList in subquerytransformer

davidhall586:
TimeZone was uninitialized in this scenario and led to undefined behavior.


tntnatbry (approved):
LGTM. I know this commit is fixing an undefined behaviour which can only be reproduced after several runs of the query. So does it still make sense to add a test case for this?

PR #2655, opened 2022-12-10
MCOL-5263 Add support to ROLLBACK when PP were restarted.

denis0x0D:
DMLProc starts ROLLBACK when SELECT part of UPDATE fails b/c EM facility in PP were restarted. Unfortunately this ROLLBACK stuck if EM/PP are not yet available. DMLProc must have a t/o with re-try doing ROLLBACK.

### DIFF
--- a/dbcon/joblist/distributedenginecomm.cpp
+++ b/dbcon/joblist/distributedenginecomm.cpp
@@ -221,7 +221,7 @@ DistributedEngineComm::~DistributedEngineComm()
   fInstance = 0;
 }
 
-void DistributedEngineComm::Setup()
+int32_t DistributedEngineComm::Setup()
 {
   // This is here to ensure that this function does not get invoked multiple times simultaneously.
   boost::mutex::scoped_lock setupLock(fSetupMutex);
@@ -309,10 +309,9 @@ void DistributedEngineComm::Setup()
                  "Could not connect to PMS" + std::to_string(connectionId) + ": " + ex.what(),
                  LOG_TYPE_ERROR);
       if (newPmCount == 0)
-      {
         writeToLog(__FILE__, __LINE__, "No more PMs to try to connect to", LOG_TYPE_ERROR);
-        break;
-      }
+
+      return 1;
     }
     catch (...)
     {
@@ -322,10 +321,8 @@ void DistributedEngineComm::Setup()
       writeToLog(__FILE__, __LINE__, "Could not connect to PMS" + std::to_string(connectionId),
                  LOG_TYPE_ERROR);
       if (newPmCount == 0)
-      {
         writeToLog(__FILE__, __LINE__, "No more PMs to try to connect to", LOG_TYPE_ERROR);
-        break;
-      }
+      return 1;
     }
   }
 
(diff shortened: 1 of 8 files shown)


dbcon/joblist/distributedenginecomm.cpp
```diff
@@ -971,6 +973,7 @@ void DistributedEngineComm::write(messageqcpp::ByteStream& msg, uint32_t connect
   lk.unlock();
 
   newClients[connection]->write(msg, NULL, senderStats);
```
drrtuy: Is there a way to take newClients[connection]->write() call result into account before we return 0 in the end of DistributedEngineComm::write()?

dbcon/joblist/distributedenginecomm.cpp
```diff
@@ -1163,7 +1164,30 @@ int DistributedEngineComm::writeToClient(size_t aPMIndex, const SBS& bs, uint32_
       map_tok->second->queue.push(sbs);
     }
 
+    int tries = 0;
+    // Try to setup connection with PS, it could be a situation that PS is starting.
+    // MCOL-5263.
+    while (tries < 10 && Setup())
```
drrtuy, line 1171: I am ok with fixing the attempts number however the sleep timeout value should be exponential b/c it is painful to wait for 5 seconds if it was a network glitch.

writeengine/shared/we_brm.cpp
```diff
@@ -1134,234 +1135,243 @@ int BRMWrapper::rollBackBlocks(const VER_t transID, int sessionId)
 
   std::vector<BRM::FileInfo> files;
 
-  for (i = 0; i < lbidList.size(); i++)
+  try
```
drrtuy, line 1138: I missed the part where the patch looks for rollBackBlocks() result code. Could you point me to this line/-s?

drrtuy (changes requested):

writeengine/shared/we_brm.cpp
denis0x0D: It throws exception and we handle it https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/2655/files#diff-3da43348b5fbf465a6af2ecfd724226a5450af15fab2204485a9f1ad15358ad1R1371

denis0x0D: The return code is handling here https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/2655/files#diff-ff20b84c0c434c434ccb22533ef5e7810bb9cc4603e6b002447c931479bf5c3dR274 the problem there it's a 8 bit value, I didnt wanted to change it.

dbcon/joblist/distributedenginecomm.cpp
denis0x0D: Sure, will fix it.

dbcon/joblist/distributedenginecomm.cpp
denis0x0D: I will change the API back, it throws an exception "void MessageQueueClient::write(const", I believe some part of the code written in the way to handle exception for this particular function, if we change it here, it would require to search all parts of the code where this function is used and change it from exception to return code.

drrtuy (approved):

PR #2669, opened 2022-12-19
MCOL-5361 This patch is done to sync with MDB on MDEV-29988

drrtuy:


dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -396,8 +415,7 @@ bool sortItemIsInGroupRec(Item* sort_item, Item* group_item)
   found = group_item->eq(sort_item, false);
   if (!found && sort_item->type() == Item::REF_ITEM)
   {
-    Item_ref* ifp_sort_ref = reinterpret_cast<Item_ref*>(sort_item);
-    found = sortItemIsInGroupRec(*ifp_sort_ref->ref, group_item);
+    found = sortItemIsInGroupRec(directRefOrRefLoop(sort_item), group_item);
   }
```
tntnatbry, line 419: Why are we looping here? The original code did not have loop through the `Item`s here. We should follow the same code semantics as in the Spider example where the `real_item()` Item is pulled before any further processing. So the above code should look something like below:
```c++
  if (sort_item->type() == Item::REF_ITEM &&
      ((Item_ref*)sort_item)->ref_type() == Item_ref::DIRECT_REF)
  {
    sort_item= sort_item->real_item();
  }

  if (!found && sort_item->type() == Item::REF_ITEM)
  {
    Item_ref* ifp_sort_ref = reinterpret_cast<Item_ref*>(sort_item);
    found = sortItemIsInGroupRec(*ifp_sort_ref->ref, group_item);
  }
```

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -415,9 +433,7 @@ bool sortItemIsInGroupRec(Item* sort_item, Item* group_item)
     }
     else if (ifp_sort_arg->type() == Item::REF_ITEM)
     {
-      // dereference the Item
-      Item_ref* ifp_sort_ref = reinterpret_cast<Item_ref*>(ifp_sort_arg);
-      found = sortItemIsInGroupRec(*ifp_sort_ref->ref, group_item);
+      found = sortItemIsInGroupRec(directRefOrRefLoop(ifp_sort_arg), group_item);
```
tntnatbry, line 436: See my previous comment.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -443,8 +459,7 @@ void check_sum_func_item(const Item* item, void* arg)
 
   if (item->type() == Item::REF_ITEM)
   {
-    const Item_ref* ref_item = reinterpret_cast<const Item_ref*>(item);
-    Item* ref_item_item = (Item*)*ref_item->ref;
+    Item* ref_item_item = directRefOrRefLoop(const_cast<Item*>(item));
```
tntnatbry, line 462: See my previous comment.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -3452,6 +3467,10 @@ ReturnedColumn* buildReturnedColumn(Item* item, gp_walk_info& gwi, bool& nonSupp
     case Item::REF_ITEM:
     {
       Item_ref* ref = (Item_ref*)item;
+      if (ref->ref_type() == Item_ref::DIRECT_REF)
+      {
+        return buildReturnedColumn(item->real_item(), gwi, nonSupport);
+      }
```
tntnatbry, line 3473: `real_item()` call should happen before the switch case.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -3942,15 +3961,14 @@ ReturnedColumn* buildFunctionColumn(Item_func* ifp, gp_walk_info& gwi, bool& non
...
         args.add(info);
         gwi.parseErrorText = IDBErrorInfo::instance()->errorMsg(ERR_NON_SUPPORTED_FUNCTION, args);
         return NULL;
       }
-
-      if (!selectBetweenIn && (ifp->arguments()[0]->type() == Item::FIELD_ITEM ||
-                               (ifp->arguments()[0]->type() == Item::REF_ITEM &&
-                                (*(((Item_ref*)ifp->arguments()[0])->ref))->type() == Item::FIELD_ITEM)))
+      Item* ifpArg0 = directRefOrRefLoop(ifp->arguments()[0]);
+      if (!selectBetweenIn && ifpArg0->type() == Item::FIELD_ITEM)
```
tntnatbry, line 3971: Again, I am not clear as to why we have a call to `directRefOrRefLoop()` here.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -5914,7 +5930,10 @@ void gp_walk(const Item* item, void* arg)
 
         // bug 3137. If filter constant like 1=0, put it to ptWorkStack
         // MariaDB bug 750. Breaks if compare is an argument to a function.
-        //				if ((int32_t)gwip->rcWorkStack.size() <=  (gwip->rcBookMarkStack.empty() ? 0
+        //				if ((int32_t)gwip->rcWorkStack.size() <=
+        //(gwip->rcBookMarkStack.empty()
+        //?
+        // 0
         //: gwip->rcBookMarkStack.top())
         //				&& isPredicateFunction(ifp, gwip))
```
tntnatbry, line 5938: The above comment does not look readable, can you please fix this?

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -6175,23 +6195,19 @@ void gp_walk(const Item* item, void* arg)
...
 
-          if (ifp->arguments()[i]->type() == Item::REF_ITEM)
-          {
-            Item* op = *(((Item_ref*)ifp->arguments()[i])->ref);
-            operand = buildReturnedColumn(op, *gwip, gwip->fatalParseError);
-          }
-          else
-            operand = buildReturnedColumn(ifp->arguments()[i], *gwip, gwip->fatalParseError);
+          operand =
+              buildReturnedColumn(directRefOrRefLoop(ifp->arguments()[i]), *gwip, gwip->fatalParseError);
```
tntnatbry, line 6199: Same here, I am not clear on the usage of `directRefOrRefLoop()` in the above two places.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -6408,9 +6424,15 @@ void parse_item(Item* item, vector<Item_field*>& field_vec, bool& hasNonSupportI
 
     case Item::REF_ITEM:
     {
+      Item* thisBlockItem = directRefItemOrNull(item);
+      if (thisBlockItem)
+      {
+        parse_item(thisBlockItem, field_vec, hasNonSupportItem, parseInfo, gwi);
+        break;
+      }
```
tntnatbry, line 6432: I am not sure why a call to `parse_item()` is added here. `real_item()` should be pulled before the first switch statement.

tntnatbry (changes requested):

dbcon/mysql/ha_mcs_execplan.cpp
drrtuy: I disagree. Semantically we need a real element here. directRefOrRefLoop returns either the first real element or the item itself if itsn't a ref.

dbcon/mysql/ha_mcs_execplan.cpp
drrtuy: See my previous comment.

dbcon/mysql/ha_mcs_execplan.cpp
drrtuy: See my previous comment.

dbcon/mysql/ha_mcs_execplan.cpp
drrtuy: I am open to reconsider this if you elaborate with a counter example.

dbcon/mysql/ha_mcs_execplan.cpp
drrtuy: The original if condition checks if the ifp::arguments()[0] is Item_field or it is a Item_field behind a ref and directRefOrRefLoop loops till the first non-REF Item is found. directRefOrRefLoop can even return an item itself.

dbcon/mysql/ha_mcs_execplan.cpp
drrtuy: JFYI I am following the convention we all accepted previously, namely to use auto-formater. I am using clang-format in this case so this is an automated change. IMHO if the accepted tool formats the code this way it is OK with me.

dbcon/mysql/ha_mcs_execplan.cpp
drrtuy: directRefOrRefLoop either returns the first real Item or original Item if it isn't a ref.

drrtuy:
I need to confess that I will delay this patch b/c it doesn't bring any benefits except tiny code reduction.

PR #2670, opened 2022-12-20
MCOL-5195 Correlated subquery with equi/non-equi scalar filter and join condition

denis0x0D:


drrtuy (approved):

drrtuy:
Plz let me create a new develop before I proceed and merge the patch.

PR #2673, opened 2022-12-22
MCOL-4530: common conjuction top rewrite

mariadb-AndreyPiskunov:
Added logical transformation of the execplan::ParseTrees  with the taking out the common factor in expression of the form "(A and B) or (A and C)" for the purposes of passing a TPCH 19 query.


mariadb-AndreyPiskunov:
As the result, we are transforming queries of the general form:

```
(A_1 and A_2 and ... and A_n and B_1)
or
(A_1 and A_2 and ... and A_n and B_2)
or
(A_1 and A_2 and ... and A_n and B_3)
or ...
...
or 
(A_1 and A_2 and ... A_n and B_m)

```

factoring out `A_1 and ... A_n` common factors (where each factor is a single condition) resulting in a query of a general form:

```
(A_1 and A_2 and ... and A_n)
and
(B_1 or B_2 or ... or B_m)
```

The examples of queries can be found in added unit tests.

dbcon/execplan/rewrites.cpp
```diff
@@ -0,0 +1,417 @@
+#include "rewrites.h"
```
denis0x0D, line 13: Could you please add header with licence.

dbcon/execplan/rewrites.cpp
```diff
@@ -0,0 +1,417 @@
...
+    ++i;
+    if (i != container.end())
+      os << delimiter;
+  }
+  os << std::endl;
+}
+
+#define debug_rewrites false
+
+using CommonContainer =
```
denis0x0D, line 46: Nit: how about to put it into the header file?

dbcon/execplan/rewrites.cpp
```diff
@@ -0,0 +1,417 @@
...
+}
+
+bool simpleFiltersCmp(const SimpleFilter* left, const SimpleFilter* right)
+{
+  return normalizeNode(left->lhs()->data(), left->rhs()->data(), left->op()->op()) <
+         normalizeNode(right->lhs()->data(), right->rhs()->data(), right->op()->op());
+}
+
+// Walk the tree and find out common conjuctions
+void collectCommonConjuctions(execplan::ParseTree* root, CommonContainer& accumulator, int level = 0,
```
denis0x0D, line 113: Nit: I would suggest to add some comments with explanation why does this traversal depends on logical "or" operation, this code tries to match specific pattern [(... and .. and ..) or (... and .. and..) or ...] and collect intersection, may be we can explain more for the future developers, set links to mcol? what do you think?

dbcon/execplan/rewrites.cpp
```diff
@@ -0,0 +1,417 @@
...
+  return;
+}
+
+execplan::ParseTree* newAndNode()
+{
+  execplan::Operator* op = new execplan::Operator();
+  op->data("and");
+  return new execplan::ParseTree(op);
+}
+
```
denis0x0D, line 160: Nit: some comments to this function please.

dbcon/execplan/rewrites.cpp
```diff
@@ -0,0 +1,417 @@
...
+  GoTo direction;
+  ChildType containsLeft;
+  ChildType containsRight;
+  StackFrame(execplan::ParseTree** node_, GoTo direction_)
+   : node(node_), direction(direction_), containsLeft(ChildType::Leave), containsRight(ChildType::Leave)
+  {
+  }
+};
+
+using DFSStack = std::vector<StackFrame>;
```
denis0x0D, line 214: Nit: put to the header file? what do you think?

dbcon/execplan/rewrites.cpp
```diff
@@ -0,0 +1,417 @@
...
+}
+
+void replaceContainsTypeFlag(StackFrame& stackframe, ChildType containsflag)
+{
+  if (stackframe.direction == GoTo::Right)
+    stackframe.containsLeft = containsflag;
+  else
+    stackframe.containsRight = containsflag;
+}
+
```
denis0x0D, line 267: Nit: could you please add some comments with explanation for the better understanding.

dbcon/execplan/rewrites.cpp
```diff
@@ -0,0 +1,417 @@
...
+        deleteOneNode((*node)->rightRef());
+
+      if (parentframe)
+        replaceContainsTypeFlag(*parentframe, ChildType::Delete);
+      else
+        deleteOneNode(node);
+    }
+  }
+}
+
```
denis0x0D, line 307: Nit: please some comments here.

dbcon/execplan/rewrites.cpp
```diff
@@ -0,0 +1,417 @@
...
+  std::string dotname = dumpfolder + "filters." + name + ".dot";
+  filters->drawTree(dotname);
+  std::string dotInvoke = "dot -Tpng ";
+  std::string convert = dotInvoke + dotname + " -o " + dotname + ".png";
+  [[maybe_unused]] auto _ = std::system(convert.c_str());
+#endif
+}
+
+template <bool stableSort>
+execplan::ParseTree* extractCommonLeafConjunctionsToRoot(execplan::ParseTree* tree)
```
denis0x0D, line 371: Some thoughts for the future - what do you think about unification for pattern match/rewrite, for example:
1. We have unified tree traversal.
2. We have a list of patterns - pattern 1, pattern 2, pattern 3, they are applied iteratively and after each rewrites happens.
If we have some unified infrastructure we do not need to write again tree traversal, we can just specify what to do on some callbacks.
BTW, my point is not about to rewrite this current solution, I'm just interesting what do you think in general, may be it's not applicable here.

dbcon/execplan/simplecolumn.h
```diff
@@ -57,8 +57,14 @@ class SimpleColumn : public ReturnedColumn
   /**
    * Constructors
    */
+  class ForTestPurposeWithoutOID{};
+
+
+
```
denis0x0D: Nit: It seems like CS keeps 2 blank lines between next function definition. Could you remove unnecessary blank lines.

dbcon/execplan/simplefilter.cpp
```diff
@@ -55,6 +55,15 @@ SimpleFilter::SimpleFilter(const string& sql) : Filter(sql)
   parse(sql);
 }
 
+// TODO: only handled simplecolumn operands for now
+SimpleFilter::SimpleFilter(const string& sql, ForTestPurposesWithoutColumnsOIDS) : Filter(sql)
+{
+  parse(sql, ForTestPurposesWithoutColumnsOIDS{});
+}
+
+
```
denis0x0D: Nit: It seems like CS keeps 2 blank lines between next function definition. Could you remove unnecessary blank lines.

denis0x0D (commented):
This code looks great to me, this is best quality code I've seen in CS so far. I would like to take half of the day tomorrow to finish review. Thanks for the patch.

dbcon/execplan/rewrites.cpp
```diff
@@ -0,0 +1,417 @@
...
+{
+  Unchain,
+  Delete,
+  Leave
+};
+
+void printTreeLevel(execplan::ParseTree* root, int level)
+{
+#if debug_rewrites
+  auto sep = std::string(level * 4, '-');
```
denis0x0D, line 89: BTW, nice debug output
```
: or execplan::LogicOperator 0x7f94ec0e8f30
 ----: or execplan::LogicOperator 0x7f94ec0e8ee0
 --------: and execplan::LogicOperator 0x7f94ec0143b0
 ------------: `temp`.`t3`.`c` > `temp`.`t4`.`c` execplan::SimpleFilter 0x7f94ec058cc0
 ----------------: `temp`.`t3`.`a` = `temp`.`t3`.`a` execplan::SimpleFilter 0x7f94ec058c70
 ------------: and execplan::LogicOperator 0x7f94ec00fa00
 ----------------: `temp`.`t3`.`c` < `temp`.`t4`.`c` execplan::SimpleFilter 0x7f94ec011d50
 --------------------: `temp`.`t3`.`a` = `temp`.`t4`.`a` execplan::SimpleFilter 0x7f94ec015f90
 --------: and execplan::LogicOperator 0x7f94ec016870
 ------------: `temp`.`t3`.`b` < `temp`.`t4`.`b` execplan::SimpleFilter 0x7f94ec0d3b60
 ----------------: `temp`.`t3`.`a` = `temp`.`t4`.`a` execplan::SimpleFilter 0x7f94ec0e2d70
 Common Leaf Conjunctions:
 `temp`.`t3`.`a` = `temp`.`t4`.`a`
    Deleting: 0x7f94ec0eb6a0 execplan::ParseTree ptr: 0x7f94ec00fa00
    Deleting: 0x7f94ec0e6970 execplan::ParseTree ptr: 0x7f94ec0e2d70
    Deleting: 0x7f94ec0cad60 execplan::ParseTree ptr: 0x7f94ec016870
```

dbcon/execplan/rewrites.cpp
mariadb-AndreyPiskunov: Done

dbcon/execplan/rewrites.cpp
mariadb-AndreyPiskunov: Done

dbcon/execplan/rewrites.cpp
mariadb-AndreyPiskunov: Done, hopefully those are enough.

dbcon/execplan/rewrites.cpp
mariadb-AndreyPiskunov: I did that, but I'm unsure what exactly to comment here -- this function is pretty simple, just creates a new and node

dbcon/execplan/rewrites.cpp
mariadb-AndreyPiskunov: After some discussion we decided we don't think it's worth putting this local definition outside -- it is never used anywhere else.

dbcon/execplan/rewrites.cpp
mariadb-AndreyPiskunov: Done

dbcon/execplan/rewrites.cpp
mariadb-AndreyPiskunov: Done

dbcon/execplan/simplecolumn.h
mariadb-LeonidFedorov: Done

dbcon/execplan/simplefilter.cpp
mariadb-LeonidFedorov: Done

dbcon/execplan/rewrites.cpp
mariadb-LeonidFedorov: Thanks! I'm impressed, you did try it

dbcon/execplan/rewrites.cpp
drrtuy: A tiny note 2023 is going on already.

drrtuy: Feel free to leave this header as is but take into account doing changes in future.

dbcon/execplan/rewrites.cpp
mariadb-AndreyPiskunov: I like the idea and in fact I was thinking about something like that. However I have to say that the code here is somewhat specific for one particular case -- for example the collectCommonConjunctions function wouldn't work for the purpose of this transformation `a or (a and b) => a`. And it's not obvious to me how to change it right now so it would work.

I had a completely different approach in mind initially, which might've been more generalizeable. Maybe I'll come back to it at some point.

dbcon/execplan/rewrites.cpp
```diff
@@ -0,0 +1,437 @@
...
+      auto right = node->rightRef();
+      stack.emplace_back(right, GoTo::Left);
+    }
+  }
+}
+
+// this utility function reaplces the flag for in the stack frame,
+// indicating whether to delete, unchain or leave child node. It depends on the direction
+// specified in the stack frame
+void replaceContainsTypeFlag(StackFrame& stackframe, ChildType containsflag)
```
denis0x0D, line 260: @mariadb-LeonidFedorov @mariadb-AndreyPiskunov what do you think about to use google code style for the comments in MDB? https://google.github.io/styleguide/cppguide.html#Comment_Style
Currently everyone writes comments in the way it sees, but IMHO comments for the code is very important, as you can check top level google project, they almost have clear comments for every line of the code, I know it could be too much for us, we do not have so much time, but at least we can follow basic grammar rules - like start sentence with a capital letter, end sentence with a dot and so on.

denis0x0D (approved):
This a great work, thanks for the patch, I do not have more comments.

dbcon/execplan/rewrites.cpp
mariadb-LeonidFedorov: Denis, thank you for this notice! Going further I think we owe each other codestyle conversation on CS. Some hot discussion and document to refer. In my perspective, it's effective to use existing codestyle like Google, or LLVM, and refine it with enabling exceptions or specific places for curly braces. I will be glad to make this codestyle rules together. Some of this codestyle conventions are already implicitly implemented. But we still have no hot link to copy-paste on codereview

mariadb-LeonidFedorov:
Denis, thank you for your review

mariadb-LeonidFedorov (approved):

PR #2684, opened 2023-01-10
MCOL-5385 This patch reduces RAM consumption and adds GROUP_CONCAT RA…

drrtuy:
…M accounting feature

### DIFF
--- a/dbcon/joblist/groupconcat.cpp
+++ b/dbcon/joblist/groupconcat.cpp
@@ -1,5 +1,5 @@
 /* Copyright (C) 2014 InfiniDB, Inc.
-   Copyright (C) 2019 MariaDB Corporation
+   Copyright (C) 2019-2023 MariaDB Corporation
 
    This program is free software; you can redistribute it and/or
    modify it under the terms of the GNU General Public License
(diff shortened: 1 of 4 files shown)


dbcon/joblist/groupconcat.cpp
```diff
@@ -1038,17 +1064,18 @@ void GroupConcatNoOrder::getResult(uint8_t* buff, const string& sep)
       outputRow(oss, fRow);
       fRow.nextRow();
     }
-
+    size_t sizeDiff = oss.str().size() - prevResultSize;
```
mariadb-DenisKhalikov, line 1067: Could it be a situation when prevResultSize > oss.str().size(), if so we could face possible signed integer overflow.

drrtuy: I don't see how it is possible given that oss size is a monotonic non-decreasing function? Could you give an example?

mariadb-DenisKhalikov: Good to know, never heard about it.

mariadb-DenisKhalikov (approved):

dbcon/joblist/groupconcat.cpp
drrtuy: I mean every iteration of a while loop either adds or doesn't change oss so sizeDiff can't be negative.

PR #2703, opened 2023-01-20

mariadb-SergeyZefirov:


mariadb-LeonidFedorov:
Let's fix unittests first, then look at mtr and regression

mariadb-SergeyZefirov:
> Let's fix unittests first, then look at mtr and regression

You would not believe it, but I have unit-tests fixed in my branch: https://github.com/mariadb-SergeyZefirov/mariadb-columnstore-engine/blob/MCOL-271-empty-strings-should-not-be-null/tests/rowgroup-tests.cpp#L93

Most crucial difference is this:

``columnstore/columnstore/tests$ git diff develop rowgroup-tests.cpp
diff --git a/tests/rowgroup-tests.cpp b/tests/rowgroup-tests.cpp
index 0c2a2e789..5d8cbbd51 100644
--- a/tests/rowgroup-tests.cpp
+++ b/tests/rowgroup-tests.cpp
@@ -90,6 +90,9 @@ class RowDecimalTest : public ::testing::Test

     rg.initRow(&r);
     rg.initRow(&rOutMappingCheck);
+    rg.initRow(&rOut);
+    rg.initRow(&sameRow);
+    rg.initRow(&nonequiRow);
     rowSize = r.getSize();
     rg.getRow(0, &r);
``

But I cannot see it in the changeset.

So, believe me, I have unittests and MTR green. Can you please look at this monstrous thing while I dance around infrastructure beating my trusty tambourine?

mariadb-SergeyZefirov:
As a matter of fact, regression requires specific branch: https://github.com/mariadb-corporation/mariadb-columnstore-regression-test/tree/MCOL-271-empty-strings-should-not-be-nulls

There are a lot of "empty string is expected to be NULL" results fixed.

mariadb-SergeyZefirov:
It won't merge anyway.

PR #2707, opened 2023-01-23

mariadb-SergeyZefirov:

For regression test suite, it requires appropriate changes: https://github.com/mariadb-corporation/mariadb-columnstore-regression-test/tree/MCOL-271-empty-strings-should-not-be-nulls

Previous attempt was not successful, so I re-opened it.


mariadb-SergeyZefirov:
This is still wrong. I have 400 files to commit and 866 commits, github shows 399 files and 857 commits.

So I close it unless I figure out what is going on.

PR #2708, opened 2023-01-23

mariadb-SergeyZefirov:

For regression test suite, it requires appropriate changes: https://github.com/mariadb-corporation/mariadb-columnstore-regression-test/tree/MCOL-271-empty-strings-should-not-be-nulls


mariadb-SergeyZefirov:
Relevant PR for regression tests suite: https://github.com/mariadb-corporation/mariadb-columnstore-regression-test/pull/321

build/bootstrap_mcs.sh
```diff
@@ -16,7 +16,7 @@ fi
 
 message "Building Mariadb Server from $color_yellow$MDB_SOURCE_PATH$color_normal"
 
-BUILD_TYPE_OPTIONS=("Debug" "RelWithDebInfo")
+BUILD_TYPE_OPTIONS=("Debug" "RelWithDebInfo" "Release")
```
drrtuy: JFYI I am aware that Release build type is error-prone in case of MDB/MCS and it is never compiled or used thus this will be an unused option. Could you remove this?

mariadb-SergeyZefirov: Will do.

mariadb-SergeyZefirov:
Relevant PR for regression tests suite: https://github.com/mariadb-corporation/mariadb-columnstore-regression-test/pull/321

build/bootstrap_mcs.sh
mariadb-SergeyZefirov: Removed it is.

utils/common/conststring.h
```diff
@@ -38,12 +38,31 @@ class ConstString
   explicit ConstString(const std::string& str) : mStr(str.data()), mLength(str.length())
   {
   }
+  template <typename T>
```
drrtuy, line 41: Is it possible to calculate colWidth as sizeof(T) ?

utils/common/nullstring.h
```diff
@@ -0,0 +1,200 @@
+// nullstring.h
```
drrtuy, line 18: License plz.

tests/CMakeLists.txt
```diff
@@ -72,11 +72,11 @@ if (WITH_UNITTESTS)
     target_link_libraries(comparators_tests ${ENGINE_LDFLAGS} ${ENGINE_WRITE_LIBS} ${CPPUNIT_LIBRARIES} cppunit)
     add_test(NAME columnstore:comparators_tests COMMAND comparators_tests)
 
-    add_executable(patterns_test patterns_test.cpp)
-    add_dependencies(patterns_test googletest mpark_patterns)
-    target_include_directories(patterns_test PUBLIC ${ENGINE_COMMON_INCLUDES} ${ENGINE_BLOCKCACHE_INCLUDE} ${ENGINE_PRIMPROC_INCLUDE} )
-    target_link_libraries(patterns_test ${GTEST_LIBRARIES})
-    gtest_add_tests(TARGET patterns_test TEST_PREFIX mpark_patterns:)
+    #add_executable(patterns_test patterns_test.cpp)
```
drrtuy: This should be active.

primitives/primproc/primitiveserver.cpp
```diff
@@ -1598,7 +1598,7 @@ struct BPPHandler
   {
     uint32_t uniqueID, sessionID, stepID;
     BPPMap::iterator it;
-
+#if 0
```
drrtuy: You surely don't need this preprocessor if

utils/common/collation.h
```diff
@@ -166,7 +166,11 @@ class Charset
   }
   int strnncollsp(const utils::ConstString& str1, const utils::ConstString& str2) const
   {
-    return mCharset->strnncollsp(str1.str(), str1.length(), str2.str(), str2.length());
+    // nullptr handling below should return values as if nulls are substituted with empty string.
+    // please note that ConstString has an assertion so that nullptr data has zero length.
+    const char* s1 = str1.str();
+    const char* s2 = str2.str();
+    return mCharset->strnncollsp(s1 ? s1 : "", str1.length(), s2 ? s2 : "" , str2.length());
```
drrtuy, line 173: What is an empty string semantics here? Is it still NULL or what?

drrtuy: If it is NULL then two NULLs should be equal according with the code that doesn't sound right to me.

utils/common/string-with-null.h
```diff
@@ -0,0 +1,73 @@
...
+
+#pragma once
+
+#include <memory>
+#include <string>
+#include <string.h>
+
+namespace utils
+{
+class StringWithNull
```
drrtuy: Do we need this class if all of its methods are commented out?

utils/common/collation.h
drrtuy: What if you apply NullString::eq() logic here?

utils/rowgroup/rowgroup.h
```diff
@@ -143,8 +147,12 @@ class StringStore
   StringStore();
   virtual ~StringStore();
 
-  inline std::string getString(uint64_t offset) const;
-  uint64_t storeString(const uint8_t* data, uint32_t length);  // returns the offset
+  inline utils::NullString getString(uint64_t offset) const;
+  // returns the offset.
+  // it may receive nullptr as data and it is proper way to store NULL values.
+  uint64_t storeString(const uint8_t* data, uint32_t length);
+  //uint64_t storeString(const utils::NullString& str);  // returns the offset
```
drrtuy: We don't need these comments.

utils/rowgroup/rowgroup.h
```diff
@@ -61,11 +61,15 @@
...
 
+#include "stdlib.h"
+#include "execinfo.h"
+
 // Workaround for my_global.h #define of isnan(X) causing a std::std namespace
 
 namespace rowgroup
 {
 const int16_t rgCommonSize = 8192;
+const uint32_t rgNullOffset = std::numeric_limits<uint32_t>::max();
```
drrtuy: This const is not used in the patch plz remove it.

utils/rowgroup/rowgroup.h
```diff
@@ -592,9 +603,20 @@ class Row
...
+
+  void setNullMark(uint32_t col, bool isNull) const
+  {
+    data[getInternalSize() + col] = isNull;
+  }
+
   const CHARSET_INFO* getCharset(uint32_t col) const;
 
  private:
+  static const uint16_t null_string_length = 0xffffU;
```
drrtuy: This is also unused.

utils/rowgroup/rowgroup.h
```diff
@@ -984,17 +1021,30 @@ inline void Row::colUpdateHasherTypeless(datatypes::MariaDBHasher& h, uint32_t k
...
   // TODO: add multi-byte safe truncation here
   uint32_t length = str.length();
-  if (length > getColumnWidth(colIndex))
-    length = getColumnWidth(colIndex);
+
+  setNullMark(colIndex, !str.str());
 
   if (inStringTable(colIndex))
   {
+    if (length > getColumnWidth(colIndex))
```
drrtuy: There is a common if here and at line 1056. Plz move it back and put outside if that starts at 1043.

utils/rowgroup/rowgroup.h
```diff
@@ -1003,24 +1053,31 @@ inline void Row::setStringField(const utils::ConstString& str, uint32_t colIndex
...
-  return std::string((char*)&data[offsets[colIndex] + 2], *((uint16_t*)&data[offsets[colIndex]]));
+    uint8_t* buf = &data[offsets[colIndex]];
+    if (str.str())
+    {
+      memcpy(buf, str.str(), length);
+      memset(buf + length, 0, offsets[colIndex + 1] - (offsets[colIndex] + length));
+    }
+    else
+    {
+      setToNull(colIndex);
```
drrtuy, line 1046: I am curious if you tried to remove this setToNull() call b/c it was not here originally.

utils/rowgroup/rowgroup.h
```diff
@@ -1950,6 +2018,10 @@ inline void copyRow(const Row& in, Row* out, uint32_t colCount)
     {
       out->setLongDoubleField(in.getLongDoubleField(i), i);
     }
+    else if (UNLIKELY(in.getColTypes()[i] == execplan::CalpontSystemCatalog::DOUBLE))
+    {
+      out->setDoubleField(in.getDoubleField(i), i);
```
drrtuy, line 2004: Ehhm. Imagine in Row has NULL double value that is an invalid double from the perspective of IEEE-754. IMHO the integer NULL magic will be saturated when it is implicitly casted to double.

drrtuy: If this is not the case then where is if for FLOAT?

utils/rowgroup/rowgroup.h
```diff
@@ -1923,11 +1989,13 @@ inline void copyRow(const Row& in, Row* out, uint32_t colCount)
 
   out->setRid(in.getRelRid());
 
-  if (!in.usesStringTable() && !out->usesStringTable())
-  {
-    memcpy(out->getData(), in.getData(), std::min(in.getOffset(colCount), out->getOffset(colCount)));
-    return;
-  }
+  // XXX: this code will copy data incorrectly if sizes of columns differ. Not only for strings, but for eveything.
```
drrtuy: I don't mind an additional check but I am strongly against removing memcpy.

drrtuy: I suggest to resolve it this way. Move the commented code into a new method Row::copyRowAllColsInline(). What do you think?

datatypes/mcs_decimal.cpp
```diff
@@ -606,27 +606,46 @@ std::string Decimal::toStringTSInt64() const
 
 // Dispatcher method for toString() implementations
 std::string Decimal::toString(bool hasTSInt128) const
```
drrtuy, line 608: OK. Here is the thing.
You add utils::NullString Decimal::toNullString(bool hasTSInt128) thus std::string Decimal::toString(bool hasTSInt128) must become obsolete and deleted, right?
I had seen that code has expressions like this:
NullString value;
...
value.assign(dec.toString(true));
See we_dmlcommandproc.cpp:3114 and there are other places.
This returns std::string to assign to NullString? NullString looks like a more appropriate type to return here.

utils/common/conststring.h
mariadb-SergeyZefirov: This is called from filtering part of column.cpp and that code keeps colWidth as a part of template argument. ConstStrings constructed there previously used COL_WIDTH instead of sizeof, I just factored that out.

utils/common/nullstring.h
mariadb-SergeyZefirov: Will do.

tests/CMakeLists.txt
mariadb-SergeyZefirov: Will fix.

mariadb-SergeyZefirov: It produces "internal compiler error" on my Rocky Linux 8 builds with gcc version 8.5.0 20210514. The library is not used anywhere else.

I will uncomment, of course, after I am done with your other notes.

primitives/primproc/primitiveserver.cpp
mariadb-SergeyZefirov: You are right, I do not.

utils/common/collation.h
mariadb-SergeyZefirov: We should not.

strnncollsp should have the same semantics as strnncollsp from charsets in server, which does not know what SQL's NULLs are. The code we are talking about is a defensive programming.

utils/common/string-with-null.h
mariadb-SergeyZefirov: We do not.

utils/rowgroup/rowgroup.h
mariadb-SergeyZefirov: I think you are talking about commented out storeString. If this is the case, yes, we do not need this comment.

Three other comment lines are valuable in my opinion.

utils/rowgroup/rowgroup.h
mariadb-SergeyZefirov: Done.

utils/rowgroup/rowgroup.h
mariadb-SergeyZefirov: Done.

utils/rowgroup/rowgroup.h
mariadb-SergeyZefirov: It cannot be removed.

While we usually use rowgroup to fill it once and then only read, nothing in the interface precludes some creative person to set fields several times. Thus, it is possible that the aforementioned creative persong can set a field in a row to a non-NULL value and then to NULL. setToNull lets this to be the a valid use case.

utils/rowgroup/rowgroup.h
mariadb-SergeyZefirov: I think you may be right here. I'll test your suggestion.

Still, no tests are failing with that! What does that tell us?

utils/rowgroup/rowgroup.h
mariadb-SergeyZefirov: Done.

utils/common/nullstring.h
mariadb-SergeyZefirov: Added.

utils/common/conststring.h
mariadb-SergeyZefirov: No.

This method gets called from column.cpp, with COL_WIDTH template parameter as argument. If COL_WIDTH is not removed from column.cpp's templates, I cannot remove it here.

This is "when in Rome do as Romans do" approach. I do not fully understand why COL_WIDTH is not computed as sizeof in column.cpp so I assume it is not possible and sizeof(T) is not always equal to COL_WIDTH.

utils/rowgroup/rowgroup.h
mariadb-SergeyZefirov: I'll do that.

But.

Do you really think this can be more than a blip in the program profile?

utils/rowgroup/rowgroup.h
mariadb-SergeyZefirov: Upon inspection the setDoubleField(getDoubleField(...),...) can be safely handled by last leg of the if...else if...else conditional.

utils/rowgroup/rowgroup.h
drrtuy: 80% percent sure on that.

tests/CMakeLists.txt
drrtuy: Much appreciated.

utils/common/collation.h
drrtuy: Hm, but this approach will return true for NULL == NULL, right?

utils/rowgroup/rowgroup.h
drrtuy: This tells there are no relevant tests.

datatypes/mcs_datatype.h
```diff
@@ -248,6 +249,7 @@ class SystemCatalog
         @brief Convenience method to get int128 from a std::string.
     */
     int128_t decimal128FromString(const std::string& value, bool* saturate = 0) const;
+    int128_t decimal128FromString(const utils::NullString& value, bool* saturate = 0) const;
```
drrtuy, line 245: Do you think it is viable to remove the first overload where std::string is used?

dbcon/dmlpackage/dmlcolumn.cpp
```diff
@@ -104,17 +91,11 @@ int DMLColumn::write(messageqcpp::ByteStream& bytestream)
...
 
-  if (vectorSize > 0)
+  for (uint32_t i = 0; i < vectorSize; i++)
   {
-    for (uint32_t i = 0; i < vectorSize; i++)
-    {
-      bytestream << fColValuesList[i];
-    }
+    bytestream << fColValuesList[i];
   }
```
drrtuy, line 94: I am curious if you remember why bytestream << fData; becomes useless now or it was never used in the first place?

dbcon/dmlpackage/vendordmlstatement.cpp
```diff
@@ -25,6 +25,8 @@
 #include "vendordmlstatement.h"
 #undef VENDORDMLSTATEMENT_DLLEXPORT
 
+#include "exceptclasses.h"
```
drrtuy: It looks like this compilation unit doesn't need this include. What do you think?

dbcon/execplan/calpontsystemcatalog.cpp
```diff
@@ -6313,7 +6313,7 @@ void CalpontSystemCatalog::checkSysCatVer()
 
 CalpontSystemCatalog::ColType::ColType()
  : constraintType(NO_CONSTRAINT)
- , defaultValue("")
+ //, defaultValue("") // should be NULL, not an empty string.
```
drrtuy: Plz remove this if it commented.

dbcon/execplan/simplefilter.cpp
```diff
@@ -540,54 +540,54 @@ void SimpleFilter::convertConstant()
 
     if (fRhs->resultType().colDataType == CalpontSystemCatalog::DATE)
     {
-      if (lcc->constval().empty())
+      if (lcc->isNull())
       {
-        lcc->constval("0000-00-00");
+//        lcc->constval("0000-00-00");
```
drrtuy: Plz remove commented lines here and everywhere in this unit.

dbcon/execplan/treenode.h
```diff
@@ -458,28 +463,30 @@ inline bool TreeNode::getBoolVal()
...
-      if (fResultType.colWidth <= 8)
-        fResult.strVal = (char*)(&fResult.origIntVal);
-
-      break;
-
     case CalpontSystemCatalog::VARCHAR:
       if (fResultType.colWidth <= 7)
-        fResult.strVal = (char*)(&fResult.origIntVal);
+      {
+        //fResult.strVal.assign((uint8_t*)(&fResult.origIntVal), sizeof(fResult.origIntVal));
```
drrtuy: Plz remove this line.

dbcon/execplan/treenode.h
```diff
@@ -458,28 +463,30 @@ inline bool TreeNode::getBoolVal()
...
-    case CalpontSystemCatalog::VARBINARY:
+    case CalpontSystemCatalog::CHAR:
+    case CalpontSystemCatalog::VARBINARY: // XXX: TODO: we don't have varbinary support now, but it may be handled just like varchar.
     case CalpontSystemCatalog::BLOB:
     case CalpontSystemCatalog::TEXT:
-      if (fResultType.colWidth <= 7)
-        fResult.strVal = (char*)(&fResult.origIntVal);
+      if (fResultType.colWidth <= 8)
+      {
+        //fResult.strVal.assign((uint8_t*)(&fResult.origIntVal), sizeof(fResult.origIntVal));
```
drrtuy: Plz remove this line.

dbcon/execplan/treenode.h
```diff
@@ -619,38 +629,43 @@ inline const std::string& TreeNode::getStrVal(const long timeZone)
     case CalpontSystemCatalog::UDECIMAL:
     {
       if (fResultType.colWidth == datatypes::MAXDECIMALWIDTH)
+      {
         // Explicit path for TSInt128 decimals with low precision
-        fResult.strVal = fResult.decimalVal.toString(true);
+        //fResult.strVal.assign(fResult.decimalVal.toString(true));
```
drrtuy: Plz remove the line.

dbcon/joblist/groupconcat.cpp
```diff
@@ -267,6 +267,7 @@ void GroupConcatInfo::mapColumns(const RowGroup& projRG)
       i2++;
     }
 
+//idbassert(false);
```
drrtuy: Plz remove the line.

dbcon/joblist/groupconcat.cpp
```diff
@@ -878,6 +880,7 @@ uint8_t* GroupConcatOrderBy::getResultImpl(const string& sep)
 
   size_t prevResultSize = 0;
   size_t rowsProcessed = 0;
+  bool isNull = true;;
```
drrtuy: Plz remove the second semicolon.

dbcon/joblist/groupconcat.cpp
```diff
@@ -400,29 +396,32 @@ void GroupConcator::initialize(const rowgroup::SP_GroupConcat& gcc)
   // MCOL-901 This value comes from the Server and it is
   // too high(3MB) to allocate it for every instance.
   fGroupConcatLen = gcc->fSize;
-  fCurrentLength -= strlen(gcc->fSeparator.c_str());
+  char t[100];
```
drrtuy: These two lines looks like a debug printout.

dbcon/joblist/jlf_execplantojoblist.cpp
```diff
@@ -1802,12 +1803,14 @@ const JobStepVector doSimpleFilter(SimpleFilter* sf, JobInfo& jobInfo)
       }
 
 #else
-      bool isNull = ConstantColumn::NULLDATA == cc->type();
+      bool isNull = cc->isNull();
+//idblog("outside of try block: is null " << ((int)isNull) <<", data type " << ((int)ct.colDataType) << ", CHAR is " << ((int)CalpontSystemCatalog::CHAR) << ", width " << ct.colWidth);
```
drrtuy: Plz remove the commented line.

dbcon/joblist/jlf_execplantojoblist.cpp
```diff
@@ -1802,12 +1803,14 @@ const JobStepVector doSimpleFilter(SimpleFilter* sf, JobInfo& jobInfo)
...
 
       if (ct.isWideDecimalType())
-        convertValueNum(constval, ct, isNull, rf, jobInfo.timeZone, value128);
+        convertValueNum(constval.safeString(""), ct, isNull, rf, jobInfo.timeZone, value128);
       else
-        convertValueNum(constval, ct, isNull, rf, jobInfo.timeZone, value);
+        convertValueNum(constval.safeString(""), ct, isNull, rf, jobInfo.timeZone, value);
+//idblog("value " << std::hex << ((uint64_t)value));
```
drrtuy: Plz remove the commented line.

utils/rowgroup/rowgroup.h
mariadb-SergeyZefirov: Notable exception is group by processing and likes - they read and write the same rowgroup.

This process usually does not set NULL after setting a non-NULL value, but it is safe to assume someone will.

utils/common/collation.h
mariadb-SergeyZefirov: Right.

NULL's usualy are dealt with elsewhere. Typical use cases for strnncollsp are "compute ranges" and sorting. These use cases insulate strnncollsp from NULLs.

This thing is just wrapper over what server provides, coded in "defensive style".

utils/rowgroup/rowgroup.h
mariadb-SergeyZefirov: Okay. ;)

dbcon/mysql/ha_mcs_impl.cpp
```diff
@@ -419,7 +419,9 @@ int fetchNextRow(uchar* buf, cal_table_info& ti, cal_connection_info* ci, long t
             colType.colDataType == CalpontSystemCatalog::VARCHAR ||
             colType.colDataType == CalpontSystemCatalog::VARBINARY)
         {
-          (*f)->store("", 0, (*f)->charset());
+          //(*f)->store("", 0, (*f)->charset());
```
drrtuy, line 422: Plz remove.

drrtuy: Plz remove the comment.

dbcon/mysql/ha_mcs_impl_if.h
```diff
@@ -261,9 +265,9 @@ struct cal_group_info
 };
 
 typedef std::tr1::unordered_map<TABLE*, cal_table_info> CalTableMap;
-typedef std::vector<std::string> ColValuesList;
+//typedef std::vector<std::string> ColValuesList;
```
drrtuy: Plz remove commented lines.

primitives/linux-port/column.cpp
```diff
@@ -1123,6 +1139,7 @@ void scalarFiltering_(
   T curValue = 0;
   primitives::RIDType rid = 0;
   bool isEmpty = false;
+bool ourCase = filterCount == 2 && ((uint64_t)filterValues[0]) == 0xfeffffffffffffffULL && ((uint64_t)filterValues[1]) == 0x31235247464dULL;
```
drrtuy: You need to clean ourCase from the whole hierarchy now :)

primitives/linux-port/column.cpp
```diff
@@ -322,22 +322,41 @@ inline bool colCompareDispatcherT(T1 columnValue, T2 filterValue, uint8_t cop, u
...
       return colCompare_(order_swap(columnValue), order_swap(filterValue), cop);
-    utils::ConstString s1{reinterpret_cast<const char*>(&columnValue), COL_WIDTH};
-    utils::ConstString s2{reinterpret_cast<const char*>(&filterValue), COL_WIDTH};
-    return colCompareStr(typeHolder, cop, s1.rtrimZero(), s2.rtrimZero());
+    }
+    utils::ConstString s1((&columnValue), nullValue, COL_WIDTH);
+    utils::ConstString s2((&filterValue), (T2)nullValue, COL_WIDTH);
+    s1.rtrimZero();
+    s2.rtrimZero();
+//    if (ourCase) { idblog("comparing as strings, column value '" << s1.toString() << "', filter value '" << s2.toString() << "', cop " << ((int)cop) << ", compare NULLEQ is " << ((int)COMPARE_NULLEQ) ); }
```
drrtuy: Plz clean the comments up in this unit.

drrtuy: The same is valid for debug printouts.

primitives/linux-port/dictionary.cpp
```diff
@@ -285,6 +283,7 @@ void PrimitiveProcessor::nextSig(int NVALS, const PrimToken* tokens, p_DataValue
   const uint8_t* niceBlock = reinterpret_cast<const uint8_t*>(block);
   const uint16_t* offsets = reinterpret_cast<const uint16_t*>(&niceBlock[10]);
 
+  //idbassert(oldGetSigBehavior); // XXX: new behavior is completely broken.
```
drrtuy: Plz remove the line.

primitives/linux-port/dictionary.cpp
```diff
@@ -615,8 +619,10 @@ void PrimitiveProcessor::p_Dictionary(const DictInput* in, vector<uint8_t>* out,
     DataValue* tmpDV = reinterpret_cast<DataValue*>(&(*out)[header.NBYTES + sizeof(uint16_t)]);
 
     *tmp16 = aggCount;
+    tmpDV->isnull = 0;
     tmpDV->len = min.len;
     memcpy(tmpDV->data, min.data, min.len);
+    idbassert(0); /// this is just plain wrong.
```
drrtuy, line 624: Plz remove this.

primitives/primproc/dictstep.cpp
```diff
@@ -183,13 +183,19 @@ void DictStep::copyResultToTmpSpace(OrderedToken* ot)
...
+      uint8_t isnull = *pos;
+      pos += 1;
       len = *((uint16_t*)pos);
       pos += 2;
-      ot[rid16].str = string((char*)pos, len);
+      if (!isnull) {
+        ns.assign(pos, len);
+      }
       pos += len;
+      ot[rid16].str = ns; //string((char*)pos, len);
```
drrtuy: Plz remove the comments.

primitives/primproc/dictstep.cpp
```diff
@@ -237,22 +243,29 @@ void DictStep::processResult()
...
     /* XXXPAT: disclaimer: this is how we do it in DictionaryStep; don't know
             if it's necessary or not yet */
-    if ((bpp->absRids[tmpResultCounter] & 0x8000000000000000LL) != 0)
-    {
-      if (primMsg->OutputType & OT_DATAVALUE)
-        (*strValues)[tmpResultCounter] = joblist::CPNULLSTRMARK.c_str();
-
-      bpp->absRids[tmpResultCounter] &= 0x7FFFFFFFFFFFFFFFLL;
-    }
+    //if ((bpp->absRids[tmpResultCounter] & 0x8000000000000000LL) != 0)
```
drrtuy: Plz remove the lines.

primitives/primproc/primitiveserver.cpp
```diff
@@ -1598,21 +1598,18 @@ struct BPPHandler
...
-
-    try
-    {
-      bs.advance(sizeof(ISMPacketHeader));
-      bs >> sessionID;
-      bs >> stepID;
-      bs >> uniqueID;
-    }
-    catch (...)
+    if (bs.length() < sizeof(ISMPacketHeader) + sizeof(sessionID) + sizeof(stepID) + sizeof(uniqueID))
```
drrtuy, line 1570: This changes the behavior. 
You can sure have this if but at the same time you better get the try block back.

utils/dataconvert/dataconvert.h
```diff
@@ -1271,10 +1273,13 @@ class DataConvert
 
   // convert string to date
   EXPORT static int64_t stringToDate(const std::string& data);
```
drrtuy, line 1243: I suppose it is time to remove all overloads with std::string.

utils/rowgroup/rowgroup.cpp
```diff
@@ -993,8 +979,8 @@ bool Row::equals(const Row& r2, uint32_t lastCol) const
   // hasCollation is true if there is any column of type CHAR, VARCHAR or TEXT
   // useStringTable is true if any field declared > max inline field size, including BLOB
   // For memcmp to be correct, both must be false.
-  if (!hasCollation && !useStringTable && !r2.hasCollation && !r2.useStringTable)
-    return !(memcmp(&data[offsets[0]], &r2.data[offsets[0]], offsets[lastCol + 1] - offsets[0]));
+  //if (!hasCollation && !useStringTable && !r2.hasCollation && !r2.useStringTable)
```
drrtuy: I strongly disagree with this comment. This should work.

utils/rowgroup/rowgroup.cpp
```diff
@@ -1123,6 +1109,7 @@ RowGroup::RowGroup(uint32_t colCount, const vector<uint32_t>& positions, const v
 
   // Set all the charsets to NULL for jit initialization.
   charsets.insert(charsets.begin(), charsetNumbers.size(), NULL);
+  //idbassert(columnCount);
```
drrtuy: Plz remove the comment

utils/rowgroup/rowgroup.cpp
```diff
@@ -1185,6 +1173,7 @@ RowGroup& RowGroup::operator=(const RowGroup& r)
   else if (!useStringTable && !oldOffsets.empty())
     offsets = &oldOffsets[0];
 
+  //idbassert(columnCount);
```
drrtuy: Plz remove the comment.

utils/rowgroup/rowgroup.h
```diff
@@ -887,8 +912,17 @@ inline void Row::setBinaryField_offset<int128_t>(const int128_t* value, uint32_t
 
 inline utils::ConstString Row::getShortConstString(uint32_t colIndex) const
 {
-  const char* src = (const char*)&data[offsets[colIndex]];
-  return utils::ConstString(src, strnlen(src, getColumnWidth(colIndex)));
+  uint32_t offset = offsets[colIndex];
+  //idbassert(getColumnWidth(colIndex) < 8); // we have to be sure these are SHORT strings, not VARCHAR(8191).
```
drrtuy: Plz remove the comment.

utils/rowgroup/rowgroup.h
drrtuy: Looking from this angle the whole codebase is unsafe b/c it doesn't restrict ownership property. setToNull is a huge switch-case spagetti that is called every time here. This is just doesn't look right.

utils/rowgroup/rowgroup.h
```diff
@@ -2030,49 +2095,54 @@ inline const uint8_t* StringStore::getPointer(uint64_t off) const
...
-  uint32_t length;
-
   if (off == std::numeric_limits<uint64_t>::max())
     return true;
+  return false;
 
-  // Long strings won't be NULL
-  if (off & 0x8000000000000000)
-    return false;
+#if 0
```
drrtuy: Plz remove the comments.

utils/windowfunction/wf_percentile.cpp
```diff
@@ -334,11 +334,13 @@ void WF_percentile<T>::operator()(int64_t b, int64_t e, int64_t c)
         implicit2T(idx, cv, 0);
         vd = (crn - rn) * fv + (rn - frn) * cv;
       }
-
+      // XXX an attempt to quickly fix compilation error regarding out-of-bounds reference.
```
drrtuy: Nothing changed so plz remove the comments.

utils/windowfunction/windowfunctiontype.cpp
```diff
@@ -692,6 +691,8 @@ void* WindowFunctionType::getNullValueByType(int ct, int pos)
 
     case CalpontSystemCatalog::CHAR:
     case CalpontSystemCatalog::VARCHAR:
+    case CalpontSystemCatalog::TEXT:
+    case CalpontSystemCatalog::VARBINARY: // XXX: I guess it is right to do that. TODO: we can add TEXT here too.
```
drrtuy, line 695: VARBINARY is still unsupported.

writeengine/dictionary/we_dctnry.cpp
```diff
@@ -1061,12 +1061,14 @@ int Dctnry::insertDctnry(const int& sgnature_size, const unsigned char* sgnature
     return ERR_DICT_SIZE_GT_2G;
   }
 
+#if 0
```
drrtuy: Plz remove the comments.

writeengine/server/we_dmlcommandproc.cpp
```diff
@@ -1798,32 +1789,34 @@ uint8_t WE_DMLCommandProc::processBatchInsertBinary(messageqcpp::ByteStream& bs,
         {
           for (uint32_t i = 0; i < valuesPerColumn; i++)
           {
-            bs >> tmp8;
-            isNULL = tmp8;
+            // XXX XXX XXX XXX now isNull is in the NullString, this is not necessary!!!
+            // XXX XXX XXX XXX but it may crash the application!!! XXX XXX XXX XXX
+            // XXX XXX XXX XXX I need to figure out where the opposite part is!!! XXX XXX XXX XXX
+            //bs >> tmp8;
```
drrtuy: Plz remove the comments.

dbcon/dmlpackage/dmlcolumn.cpp
mariadb-SergeyZefirov: I do not have actual commit handy, but fData is (and most probably was) not used anywhere.

datatypes/mcs_datatype.h
mariadb-SergeyZefirov: Let me try.

utils/rowgroup/rowgroup.h
mariadb-SergeyZefirov: It is not called every time here. It is called only when ConstString is NULL (nullptr). Which is not all that often.

Also, there were not setToNull there because empty string was a NULL indicator.

Thus, I very much like setToNull() to be there until it is proven to be a detriment performance-wise.

utils/rowgroup/rowgroup.h
mariadb-SergeyZefirov: Okay, I made a copyRowInline with a better memcpy guard check and several lines of comments describing pitfalls to avoid. copyRowInline tries to memcpy and falls back into copyRow when it can't.

primitives/linux-port/dictionary.cpp
mariadb-SergeyZefirov: I cannot. The assertion should be there.

primitives/primproc/primitiveserver.cpp
mariadb-SergeyZefirov: "Exceptions are for exceptional cases". This exception happens so often it renders futile any and all attempts to catch actual error in primitiveserver.

Also, you didn't explain the change in behavior. What is different for code that calls this routine?

utils/dataconvert/dataconvert.h
mariadb-SergeyZefirov: I will spent two weeks fixxing build and ironing bugs then. Why do you hate me? ;)

utils/rowgroup/rowgroup.cpp
mariadb-SergeyZefirov: It will not for inline stored strings, with byte sizes 9..20. The NULL marks for them are stored at the end and value is not changed when string field is set to a NULL value.

That means that if you set string field to "a" and then to NULL, the data at appropriate colIndex will still be "a".

This, in turn, means that if you construct rows A and B by setting inline string field to "a" and then to NULL for A and setting string field to "b" and then to NULL for B, the memcmp will consider them different despite the fact they are the same.

So i strongly disagree with your disagreement. The memcmp will not work.

utils/windowfunction/windowfunctiontype.cpp
mariadb-SergeyZefirov: So, let it be. ;)

utils/rowgroup/rowgroup.h
mariadb-SergeyZefirov: But!

I can make a specialized fast version for setToNull, which will just set the flag. I'll do that.

mariadb-SergeyZefirov: Actually, it was superfluous because of setNullMark few lines above. Deleted that.

datatypes/mcs_decimal.cpp
mariadb-SergeyZefirov: This requires a lot of code changes. I propose to postpone this change.

datatypes/mcs_datatype.h
mariadb-SergeyZefirov: It is unviable. Too many changes required.

utils/rowgroup/rowgroup.cpp
mariadb-SergeyZefirov: I think I've found the way for this to work. It is in testing now.

utils/rowgroup/rowgroup.h
mariadb-SergeyZefirov: The setToNull() call is essential for handling short strings.

primitives/linux-port/column.cpp
mariadb-SergeyZefirov: Done.

utils/rowgroup/rowgroup.h
mariadb-SergeyZefirov: setToNull is needed - it sets magic values for short character columns. I can speed up code by partially inlining it here but it will produce more copy-paste.

Right now I am on the "make it right" side of development on what is going on here.

utils/rowgroup/rowgroup.cpp
mariadb-SergeyZefirov: It works.

tests/CMakeLists.txt
mariadb-SergeyZefirov: Did that.

utils/dataconvert/dataconvert.h
mariadb-SergeyZefirov: Can't be done now.

drrtuy: We can delay this for now.

utils/funcexp/func_concat_ws.cpp
```diff
@@ -104,18 +105,25 @@ string Func_concat_ws::getStrVal(Row& row, FunctionParm& parm, bool& isNull,
...
-  if (str.empty())
+  if (firstTime) {
+    // all arguments are NULL.
     isNull = true;
-  else
-    isNull = false;
+    return str;
+  }
+
+  //if (str.empty()) // XXX: this is not right, again. empty strings are not nulls anymore.
```
drrtuy: Remove the comments plz.

utils/funcexp/func_hex.cpp
```diff
@@ -78,10 +78,12 @@ string Func_hex::getStrVal(rowgroup::Row& row, FunctionParm& parm, bool& isNull,
...
-      const string& arg = parm[0]->data()->getStrVal(row, isNull);
-      scoped_array<char> hexPtr(new char[strlen(arg.c_str()) * 2 + 1]);
-      octet2hex(hexPtr.get(), arg.c_str(), strlen(arg.c_str()));
-      return string(hexPtr.get(), strlen(arg.c_str()) * 2);
+      const auto& arg = parm[0]->data()->getStrVal(row, isNull);
+      //scoped_array<char> hexPtr(new char[strlen(arg.str()) * 2 + 1]);
+      scoped_array<char> hexPtr(new char[arg.length() * 2 + 1]); // XXX: code now the same as for BLOB.
+      octet2hex(hexPtr.get(), arg.str(), arg.length());
+      return string(hexPtr.get(), arg.length() * 2);
+//      return string(hexPtr.get(), strlen(arg.str()) * 2);
```
drrtuy: Comments plz.

utils/funcexp/func_hex.cpp
```diff
@@ -78,10 +78,12 @@ string Func_hex::getStrVal(rowgroup::Row& row, FunctionParm& parm, bool& isNull,
     case CalpontSystemCatalog::DATE:
     case CalpontSystemCatalog::TIME:
     {
-      const string& arg = parm[0]->data()->getStrVal(row, isNull);
-      scoped_array<char> hexPtr(new char[strlen(arg.c_str()) * 2 + 1]);
-      octet2hex(hexPtr.get(), arg.c_str(), strlen(arg.c_str()));
-      return string(hexPtr.get(), strlen(arg.c_str()) * 2);
+      const auto& arg = parm[0]->data()->getStrVal(row, isNull);
+      //scoped_array<char> hexPtr(new char[strlen(arg.str()) * 2 + 1]);
```
drrtuy: Comments plz.

utils/funcexp/func_in.cpp
```diff
@@ -272,8 +272,10 @@ inline bool getBoolForIn(rowgroup::Row& row, funcexp::FunctionParm& pm, bool& is
       for (uint32_t i = 1; i < pm.size(); i++)
       {
         isNull = false;
-        const string& str1 = pm[i]->data()->getStrVal(row, isNull);
-        if (cs->strnncoll(val.c_str(), val.length(), str1.c_str(), str1.length()) == 0 && !isNull)
+        const auto& str1 = pm[i]->data()->getStrVal(row, isNull);
+        // XXX: str1 may result in is NULL argument assing to strnncoll. The length is 0 in that case.
+        // We'll see how it will go.
```
drrtuy: Could you either remove the last line or alter the statement b/c this one doesn't make sense to me or make me comfortable regarding the code.

utils/funcexp/func_ltrim_oracle.cpp
```diff
@@ -91,7 +89,7 @@ std::string Func_ltrim_oracle::getStrVal(rowgroup::Row& row, FunctionParm& fp, b
   std::string ret(pos, binLen);
   if (binLen == 0)
   {
-    isNull = true;
+    isNull = true; // XXX: is it goog?
```
drrtuy: Good question. Did you try this function?

utils/funcexp/func_nullif.cpp
```diff
@@ -348,16 +348,16 @@ uint64_t Func_nullif::getUintVal(rowgroup::Row& row, FunctionParm& parm, bool& i
...
                               CalpontSystemCatalog::ColType& op_ct)
 {
-  string exp1 = parm[0]->data()->getStrVal(row, isNull);
+  string exp1 = parm[0]->data()->getStrVal(row, isNull).safeString("");
   CHARSET_INFO* cs = parm[0]->data()->resultType().getCharset();
 
   if (isNull)
   {
-    isNull = false;
+    //isNull = false; // NULLIF(NULL, ...) is NULL, according to server's results.
```
drrtuy: Plz remove the //isNull = false; part and left the comment itself.

utils/funcexp/func_quote.cpp
```diff
@@ -45,15 +45,14 @@ std::string Func_quote::getStrVal(rowgroup::Row& row, FunctionParm& fp, bool& is
     return "NULL";
   }
 
-  if (str.empty())
-    return "NULL";
+  size_t strSize = strlen(str.c_str());
```
drrtuy, line 48: You have a shared_ptr to string why do you need to run strlen here?

utils/funcexp/func_space.cpp
```diff
@@ -29,6 +29,7 @@ namespace funcexp
 CalpontSystemCatalog::ColType Func_space::operationType(FunctionParm& fp,
                                                         CalpontSystemCatalog::ColType& resultType)
 {
+  resultType.colWidth = 32;
```
drrtuy: Could you elaborate on the value of the constant?

utils/funcexp/func_time_format.cpp
```diff
@@ -152,19 +152,22 @@ string Func_time_format::getStrVal(rowgroup::Row& row, FunctionParm& parm, bool&
...
-      *ptr++ = format[i];
+    char fi = format.unsafeStringRef()[i];
+    if (fi != '%')
+      *ptr++ = fi;
     else
     {
       i++;
 
-      switch (format[i])
+      fi = format.unsafeStringRef()[i];
```
drrtuy, line 168: There is a fi assignment duplication here and at line 161.

utils/funcexp/func_quote.cpp
mariadb-SergeyZefirov: Because std::string can contain a data with NUL ('\0') inside.

mariadb-SergeyZefirov: By the way, it is common in our code to have std::string and take strlen() on it's c_str().

utils/funcexp/func_space.cpp
mariadb-SergeyZefirov: I cannot. Having smaller values may break some tests, from I remember.

It can be that now it is not necessary.

utils/funcexp/func_time_format.cpp
mariadb-SergeyZefirov: Note the `i++` between them.

utils/funcexp/func_space.cpp
mariadb-SergeyZefirov: I'll look.

utils/funcexp/func_ltrim_oracle.cpp
mariadb-SergeyZefirov: We have no tests for that function in our MTR.

https://mariadb.com/docs/skysql/ref/mdb/functions/LTRIM_ORACLE/

"This accesses a version of the LTRIM function that returns NULL instead of an empty string when returning an empty string."

Wowsers. ;)

mariadb-SergeyZefirov: I'll look into that.

mariadb-SergeyZefirov: ltrim_oracle does not work as intended.

mariadb-SergeyZefirov: Fixed, added to the appropriate test.

utils/funcexp/func_space.cpp
mariadb-SergeyZefirov: It is not needed.

tests/CMakeLists.txt
drrtuy: Don't forget to uncomment this.

PR #2716, opened 2023-01-26
adding initial backup and restore scripts

mariadb-AllenHerrera:
Not sure what folder to include the initial backup/restore scripts in.
Seeking feedback/ open to changes

### DIFF
--- a/support/simpleBackup.bash
+++ b/support/simpleBackup.bash
@@ -1,0 +1,639 @@
+#!/bin/bash
+########################################################################
+#
+# Purpose: Simple Backup of columnstore
+# Example: sudo ./simpleBackup.bash 
+#
+# If doing a Local Storage backup:
+#    - backupLocation
+#    - backupDestination
+#    - scp (only required when saving backup to remote location)
+#   
+#   Examples:
+#        ./simpleBackup.bash -bl /tmp/backups/ -bd Local -s LocalStorage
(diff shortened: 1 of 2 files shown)


mariadb-AlexeyVorovich:
Allen, 2 questions
1. How is  SSH involved ? can we avoid that if we assume that the invoker (human or operator code) is already in the shell of the node ?
2. Todd believes that S3 action is not needed. For now can we make it optional

mariadb-AllenHerrera:
1) ssh is optional already, you'd have to add the flag -bd Remote and provide the ssh user/ip -scp root@172.31.6.163 so that the backup folder and files are saved remote relative to where the backup script is run from. 
and Yes the default is to save the backup locally to /tmp/backups unless otherwise specified via -bl /tmp/backups/

2) the default invocation is  based on what the script see in storagemanager.cnf service =
storage=$(grep -m 1 "service = " /etc/columnstore/storagemanager.cnf | awk '{print $3}').
is that the question?
else if the question is "for S3 deployments, can we skip copying the S3 bucket to another bucket i,e make this part optional to the backup" then at the moment its not optional but would be an easy enhancement/flag to add

toddstoffel:
Guys this is not going into the engine code. Why are we discussing on GitHub?

toddstoffel:
Please move this conversation to https://github.com/mariadb-corporation/mariadb-backup-manager

PR #2724, opened 2023-02-08
Enable ASAN build 

mariadb-LeonidFedorov:
Fixed config leaks that stopped services to start with asan
Systemd files hacked for ASAN
Create database test on finish of build


mariadb-LeonidFedorov:
There should be fix for non ASAN builds to get rid of emty ENVIRONMENT set in systemd files

PR #2731, opened 2023-02-16
MCOL-5242: fix Columnstore tmp_files dir owner

mariadb-LeonidFedorov:
When dbrmroot created, it has columnstore_tmp_files as parent, and the parent was still with root owner


mariadb-AlanMologorsky:
LGTM

mariadb-AlanMologorsky (approved):
LGTM

PR #2760, opened 2023-03-03
MCOL-1205 Choose a last join edge in cycle to transform without statistics

denis0x0D:
When we do not have statistics we can choose any join edge to transform, choosing the last one helps to solve TCP-H Q5 without proper optimizer/statistics.


drrtuy:
Close this for now b/c it will change a default behavior.

PR #2774, opened 2023-03-14
Fix to thrift from serg

mariadb-LeonidFedorov:


tntnatbry:
@mariadb-LeonidFedorov What about the other changes related to mpark?

mariadb-LeonidFedorov:
I want to skip mpark changes from serg, mpark is not just test library we will either use it or delete

tntnatbry (approved):

PR #2775, opened 2023-03-14
ci: use local files for core_dumps scripts

mariadb-RomanNavrotskiy:
- use local scripts for coredumps
- compress core-dumps

@mariadb-AndreyPiskunov please review and also check core dumps are processed and uploaded correctly (may be we have to make some false positive cores?)

### DIFF
--- a/core_dumps/core_dump_format.sh
+++ b/core_dumps/core_dump_format.sh
@@ -29,3 +29,5 @@ echo "<b> Step: ${STEP_NAME}<br> Binary name: ${BINARY}<br> </b>" >> "${FILENAME
 save_ansi_to_html "Backtrace"
 save_ansi_to_html "Arguments"
 save_ansi_to_html "Locals"
+
+bzip2 --compress --best "${COREDUMP}"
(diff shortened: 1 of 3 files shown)


core_dumps/core_dump_format.sh
```diff
@@ -29,3 +29,5 @@ echo "<b> Step: ${STEP_NAME}<br> Binary name: ${BINARY}<br> </b>" >> "${FILENAME
 save_ansi_to_html "Backtrace"
 save_ansi_to_html "Arguments"
 save_ansi_to_html "Locals"
+
+bzip2 --compress --best "${FILENAME}"
```
mariadb-LeonidFedorov: Here we should compress `${COREDUMP},` not `${FILENAME}`
COREDUMP is big blob, FILENAME here is lightweight HTML report

mariadb-RomanNavrotskiy: I am not sure actually!

https://github.com/mariadb-corporation/mariadb-columnstore-engine/blob/core-dumps-merge/core_dumps/core_dump_format.sh#L15

`cat "$DUMPNAME" | bash "${SCRIPT_LOCATION}"/ansi2html.sh --palette=solarized >> "${FILENAME}"`

`${FILENAME}` is just a report file? Ok. I will update to `${COREDUMP}`!

mariadb-LeonidFedorov (approved):

PR #2786, opened 2023-03-23
[MCOL-4991] Solving TRUNCATE/ROUND/CEILING functions on TIME/DATETIME/TIMESTAMP

MuHe03:
Solved the metioned issues on TIME, DATETIME and TIMESTAMP datatypes executions in TRUNCATE/ROUND/CEILING functions.


utils/funcexp/func_round.cpp
```diff
@@ -734,13 +664,109 @@ string Func_round::getStrVal(Row& row, FunctionParm& parm, bool& isNull, Calpont
...
                                       CalpontSystemCatalog::ColType& op_ct)
 {
-  return parm[0]->data()->getIntVal(row, isNull);
+  int32_t s = parm.size() > 1 ? parm[1]->data()->getIntVal(row, isNull) : 0;
+  if (isNull)
+    return 0;
+  s = (s > 6) ? 6 : s;
+  if (s < 0) 
+  {
+    // int64_t x = parm[0]->data()->getDatetimeIntVal(row, isNull);
```
drrtuy: It is good to clean the commented text up.

utils/funcexp/func_round.cpp
```diff
@@ -734,13 +664,109 @@ string Func_round::getStrVal(Row& row, FunctionParm& parm, bool& isNull, Calpont
 int64_t Func_round::getDatetimeIntVal(Row& row, FunctionParm& parm, bool& isNull,
                                       CalpontSystemCatalog::ColType& op_ct)
 {
-  return parm[0]->data()->getIntVal(row, isNull);
+  int32_t s = parm.size() > 1 ? parm[1]->data()->getIntVal(row, isNull) : 0;
+  if (isNull)
+    return 0;
+  s = (s > 6) ? 6 : s;
```
drrtuy: Could you make this 6 a constexpr const int32_t constant with a meaningful name?

drrtuy: It would be easier to read/maintain.

utils/funcexp/func_round.cpp
```diff
@@ -734,13 +664,109 @@ string Func_round::getStrVal(Row& row, FunctionParm& parm, bool& isNull, Calpont
...
+    //   if (x % 10 >= 5)
+    //     x += 10;
+    //   x *= helpers::powerOf10_c[s];
+    // }
+    // return dataconvert::DataConvert::intToDatetime(x);
+    s = 0;
+  }
+  // else 
+  // {
+    int64_t x = parm[0]->data()->getDatetimeIntVal(row, isNull) + (s <= 5 ? 5 * helpers::powerOf10_c[5 - s] : 0) + (s == 0 ? 48576 : 0); // 48576 = 0xfffff - 1000000;
```
drrtuy: Technically 5 - s can be an arbitrary number and there will be a SEGV called accessing, let's say powerOf10_c[2**31-1] so it is worth to add add a value range check here and everywhere else.

drrtuy (changes requested):
Thank you very much for the bug fix @MuHe03. We appreciate you contribution. I like the patch in general. There are some bits that will make the patch even better. 
There is a lot of boilerplate code so I suggest you to produce a inline pure function that calculates the value. It should be reused in the actual compilation units, e.g. func_ceil.cpp etc.
It is up to you to have a single pure func or func per compilation unit.

utils/funcexp/func_round.cpp
MuHe03: Sure, but I'm still not sure, whether it's necessary to do truncate and round before the decimal point, which is not supported in the InnoDB. If it's not, then I will remove the commented part, which I wrote to support this.

utils/funcexp/func_round.cpp
MuHe03: It's not necessary, because I already make sure that s is between 0 and 6 which won't cause out of index.

drrtuy (approved):

mariadb-AndreyPiskunov (approved):

drrtuy (changes requested):
Could you take a look at the MTR test failed.

drrtuy (approved):

PR #2792, opened 2023-03-27
MCOL-5464: Fixes of bugs from ASAN warnings, part one

mariadb-LeonidFedorov:


writeengine/bulk/we_bulkloadbuffer.cpp
```diff
@@ -2415,7 +2415,7 @@ void BulkLoadBuffer::tokenize(const boost::ptr_vector<ColumnInfo>& columnsInfo,
       //------------------------------------------------------------------
       case FLD_PARSE_ENCLOSED_STATE:
       {
-        char next = *(p + 1);
+        char next = p + 1 < pEndOfData ? *(p + 1) : 0;
```
mariadb-LeonidFedorov: better to pass the error here

drrtuy: Plz note here you introduce a magic value semantics, that hides a potential row parsing error. Such errors should be noted.

drrtuy (approved):

PR #2795, opened 2023-03-28
fix dependency error in bootstrap_mcs.sh

HanpyBin:
There exists an error in build/bootstrap_mcs.sh which causes unsuccessful building.The error exists in yum install list which contains an extra `install` string here. After removing it, shell can be handled correctly.


drrtuy:
Hi @HanpyBin,
Thanks a lot for the patch.

mariadb-AndreyPiskunov (approved):

PR #2805, opened 2023-04-11
MCOL-4414 Fix TIME_TO_SEC() gives unexpected results for seconds having decimals

noorall:
Fix TIME_TO_SEC() gives unexpected results for seconds having decimals.
see: https://jira.mariadb.org/browse/MCOL-4414


drrtuy:
Hey @noorall. Thank you for your contribution! Looking forward for the changes I mentioned in zulip.

noorall:
> Hey @noorall. Thank you for your contribution! Looking forward for the changes I mentioned in zulip.

I'm sorry for replying to you so late.

I noticed Func_ time_ to_ Sec:: getIntVal will always return a value of type int_64. Is this check still necessary?

stay https://github.com/mariadb-corporation/mariadb-columnstore-engine/blob/develop/dbcon/execplan/functioncolumn.h#L266 I saw similar code in, and it will perform a conversion on the s128Value of decimal.

utils/funcexp/func_time_to_sec.cpp
```diff
@@ -194,4 +210,44 @@ int64_t Func_time_to_sec::getIntVal(rowgroup::Row& row, FunctionParm& parm, bool
...
+      value = value * IDB_pow[-scaleDiff];
+    }
+    decimal.value = value;
+  }
+  else
+  {
+    int128_t s128Value = value;
+    if (scaleDiff > 0)
+    {
+      s128Value = (int128_t)(s128Value > 0 ? (double)s128Value / IDB_pow[scaleDiff] + 0.5
```
mariadb-SergeyZefirov: Here can be a loss of precision. Better cast to long double. It is as good as double in worst case but certainly can be beneficial when it is present.

Also, there is a nice function round: https://cplusplus.com/reference/cmath/round/

It does more stable rounding than what is in code (a/b + 0.5).

noorall: This issue has been fixed.

PR #2806, opened 2023-04-11
Add missing stdlib includes

FooBarrior:
Looks that these includes were missing. Either compiler was silently skipping this, or the team is using precompiled header, which I am unaware of. These includes were required to build columnstore with clang from MariaDB root.

### DIFF
--- a/storage-manager/src/MetadataFile.cpp
+++ b/storage-manager/src/MetadataFile.cpp
@@ -20,6 +20,7 @@
  */
 #include "MetadataFile.h"
 #include <set>
+#include <vector>
 #include <boost/filesystem.hpp>
 #define BOOST_SPIRIT_THREADSAFE
 #ifndef __clang__


drrtuy:
Hi @FooBarrior.
I am curious what is the clang version that complains about headers missing b/c some of our pipelines compiles MCS + ES-10.6 with clang 16.

drrtuy:
Any updates on this one @FooBarrior ?

FooBarrior:
Hi @drrtuy, my current CMakeCache is configured to gcc, and I didn't compile it anymore after April 11, so I could have mistaken on the compiler been used.

FooBarrior:
I see that most include are already at develop, only one <vector> include is missing...

mariadb-LeonidFedorov:
@FooBarrior vector in included in corresponding header file. Feel free and welcomed for any new PR's

PR #2808, opened 2023-04-12
MCOL-5385: Delete RowGroup::setData and make Pointer ctor explicit

mariadb-LeonidFedorov:
We have a proxy class 
```
class Row
{
 public:
  struct Pointer
  {
    inline Pointer() = default;

    // Pointer(uint8_t*) implicitly makes old code compatible with the string table impl;
    inline Pointer(uint8_t* d) : data(d)
    {
    }
    inline Pointer(uint8_t* d, StringStore* s) : data(d), strings(s)
    {
    }
    inline Pointer(uint8_t* d, StringStore* s, UserDataStore* u) : data(d), strings(s), userDataStore(u)
    {
    }
    uint8_t* data = nullptr;
    StringStore* strings = nullptr;
    UserDataStore* userDataStore = nullptr;
  };
```
Note, first ctor is not explicit, and doesn't respect anything about ```StringStore```
when using ```Row::setData``` on pure byte memory, all the ```StringStore``` does not work

This PR removes ugly ```RowGroup::setData``` and makes Pointer ctor explicit, to highlight all the places we need to inspect

### DIFF
--- a/dbcon/joblist/groupconcat.cpp
+++ b/dbcon/joblist/groupconcat.cpp
@@ -322,7 +322,8 @@ void GroupConcatAgUM::initialize()
 
   fGroupConcat->fRowGroup.initRow(&fRow, true);
   fData.reset(new uint8_t[fRow.getSize()]);
-  fRow.setData(fData.get());
+
+  fRow.setData(rowgroup::Row::Pointer(fData.get()));
 }
 
 void GroupConcatAgUM::processRow(const rowgroup::Row& inRow)
--- a/dbcon/joblist/jsonarrayagg.cpp
+++ b/dbcon/joblist/jsonarrayagg.cpp
@@ -315,7 +315,7 @@ void JsonArrayAggregatAgUM::initialize()
 
   fGroupConcat->fRowGroup.initRow(&fRow, true);
   fData.reset(new uint8_t[fRow.getSize()]);
-  fRow.setData(fData.get());
+  fRow.setData(rowgroup::Row::Pointer(fData.get()));
 }
 
 void JsonArrayAggregatAgUM::processRow(const rowgroup::Row& inRow)
--- a/utils/joiner/tuplejoiner.cpp
+++ b/utils/joiner/tuplejoiner.cpp
@@ -574,7 +574,7 @@ void TupleJoiner::match(rowgroup::Row& largeSideRow, uint32_t largeRowIndex, uin
           return;
 
         for (; range.first != range.second; ++range.first)
-          matches->push_back(range.first->second);
+          matches->push_back(rowgroup::Row::Pointer(range.first->second));
       }
     }
     else
@@ -615,7 +615,7 @@ void TupleJoiner::match(rowgroup::Row& largeSideRow, uint32_t largeRowIndex, uin
       pair<iterator, iterator> range = h[bucket]->equal_range(nullVal);
 
       for (; range.first != range.second; ++range.first)
-        matches->push_back(range.first->second);
+        matches->push_back(rowgroup::Row::Pointer(range.first->second));
     }
     else
     {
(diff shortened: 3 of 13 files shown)


dbcon/joblist/jsonarrayagg.cpp
```diff
@@ -315,7 +315,7 @@ void JsonArrayAggregatAgUM::initialize()
 
   fGroupConcat->fRowGroup.initRow(&fRow, true);
   fData.reset(new uint8_t[fRow.getSize()]);
-  fRow.setData(fData.get());
+  fRow.setData(rowgroup::Row::Pointer(fData.get()));
```
mariadb-LeonidFedorov, line 318: That's the place, where GroupConcat do not use any StrngStore

dbcon/joblist/groupconcat.cpp
```diff
@@ -322,7 +322,8 @@ void GroupConcatAgUM::initialize()
 
   fGroupConcat->fRowGroup.initRow(&fRow, true);
   fData.reset(new uint8_t[fRow.getSize()]);
-  fRow.setData(fData.get());
+
+  fRow.setData(rowgroup::Row::Pointer(fData.get()));
```
mariadb-LeonidFedorov, line 326: That's the place, where ```GroupConcat``` does't  use any ```StringStore```

mariadb-LeonidFedorov:
So we can search using of ```Row::Pointer::Pointer``` with one argument, and find all the old code places

mariadb-SergeyZefirov (approved):
I think this is very important change.

tntnatbry:
Please don't merge, there are potential performance implications. I need to review this.

mariadb-LeonidFedorov:
There are no potential perfomance issues here. implicit ctor was replaced by explicit and unused function was deleted. Generated code will be exactly the same

drrtuy (commented):
I would replace push_back calls with emplace_back where it is appropriate.

mariadb-LeonidFedorov:
> I would replace push_back calls with emplace_back where it is appropriate.

Anytime

utils/joiner/tuplejoiner.cpp
```diff
@@ -1033,10 +1033,10 @@ void TupleJoiner::getUnmarkedRows(vector<Row::Pointer>* out)
       for (uint i = 0; i < bucketCount; i++)
         for (it = h[i]->begin(); it != h[i]->end(); ++it)
         {
-          smallR.setPointer(it->second);
+          smallR.setPointer(rowgroup::Row::Pointer(it->second));
 
           if (!smallR.isMarked())
-            out->push_back(it->second);
+            out->push_back(rowgroup::Row::Pointer(it->second));
```
tntnatbry: Can you replace this also with emplace_back() call?

utils/rowgroup/rowgroup.h
```diff
@@ -339,7 +339,7 @@ class Row
     inline Pointer() = default;
 
     // Pointer(uint8_t*) implicitly makes old code compatible with the string table impl;
-    inline Pointer(uint8_t* d) : data(d)
+    explicit inline Pointer(uint8_t* d) : data(d)
```
tntnatbry, line 341: Row::setData() function declaration at line 364 below has the comment: "convenience fcn, can go away". Can you please remove this comment since this function is clearly used at other places in the code.

tntnatbry (changes requested):
Changes look good to me, I just added a couple of comments to be addressed. There are no performance changes in the patch, so please ignore my previous comment.

tntnatbry (approved):

utils/rowgroup/rowgroup.h
mariadb-LeonidFedorov: ++

utils/joiner/tuplejoiner.cpp
mariadb-LeonidFedorov: ++

tntnatbry:
@mariadb-LeonidFedorov It looks like the build failed with test001 and MTR failures.

PR #2817, opened 2023-04-18
Rewrite collectCommonConjunctions without recursion

mariadb-AndreyPiskunov:


dbcon/execplan/rewrites.cpp
```diff
@@ -83,6 +80,13 @@ enum class ChildType
   Leave
 };
 
+enum class GoTo
+{
```
drrtuy: Is it possible to re-use GoTo enum defined in parstree.h?

dbcon/execplan/rewrites.cpp
```diff
@@ -110,44 +114,104 @@ bool simpleFiltersCmp(const SimpleFilter* left, const SimpleFilter* right)
...
+void advanceSetUp(std::vector<StackFrameWithSet>& stack, CommonContainer& accumulator)
+{
+  if (stack.size() == 1)
+    accumulator = stack.back().localset;
+  else
   {
-    // we want to collect it if it is a child of and node and or node was met before
-    if (castToFilter(root))
+    auto sz = stack.size();
+    if (operatorType(stack.at(sz - 2).node) == OP_OR)
```
drrtuy, line 132: Do we check the invariant that there is something at(sz-2)? If not std::optional looks like a natural fit here.

drrtuy (commented):
LGTM. There are couple questions in the patch.

dbcon/execplan/rewrites.cpp
mariadb-AndreyPiskunov: Yeah, this was created after, but maybe now it makes sense to reuse.

dbcon/execplan/rewrites.cpp
mariadb-AndreyPiskunov: stack size is always > 0  in the while loop outside. and the = 1 case is checked on line 132.

dbcon/execplan/rewrites.cpp
mariadb-AndreyPiskunov: Done.

mariadb-LeonidFedorov (approved):

drrtuy (approved):

mariadb-LeonidFedorov (changes requested):
You should check the new test for MTR proposed, it's not passing on CI

dbcon/execplan/objectreader.cpp
```diff
@@ -195,33 +195,102 @@ void ObjectReader::writeParseTree(const ParseTree* tree, messageqcpp::ByteStream
...
 
   ret = new ParseTree();
-  ret->left(createParseTree(b));
-  ret->right(createParseTree(b));
-  ret->data(createTreeNode(b));
+  stack.emplace_back(ret);
+  while (!stack.empty())
+  {
+    auto [node, dir] = stack.back();
+    if (dir == ParseTree::GoTo::Left)
```
drrtuy, line 248: There is a pattern matching lib in the repo. Use it next time if you decide to create a enum to follow FP patterns.

dbcon/execplan/parsetree.h
```diff
@@ -378,65 +394,174 @@ inline ParseTree::ParseTree(const ParseTree& rhs)
...
+        if (node->fRight != nullptr)
+        {
+          stack.emplace_back(node->fRight);
+        }
+      }
+      else
+      {
+        if (stack.size() == 1)
+        {
+          node->fLeft = nullptr;
```
drrtuy, line 433: else scope shares these two lines so you can put them before the if. Both if and else shares `stack,pop_back()` so you can put them outside if IMHO.

drrtuy (approved):
LGTM except minor boilerplate part that could be fixed before the merge.

mariadb-LeonidFedorov (approved):

PR #2823, opened 2023-04-19
MCOL-5429 Fix high memory consumption in GROUP_CONCAT() processing.

tntnatbry:
1. Input and output RowGroup's used in GROUP_CONCAT classes are currently allocating a raw memory buffer of size equal to the actual width of the string datatype. As an example, for the following query:
  SELECT col1, GROUP_CONCAT(col2) FROM t GROUP BY col1;
If col2 is a TEXT field with default width, the input RowGroup containing the target rows to be concatenated will assign 64kb of memory for every input row in the RowGroup. This is wasteful as actual field values in real workloads would be much smaller. We fix this by enabling the RowGroup to use the StringStore when the RowGroup contains long strings.

2. RowAggregation::initialize() allocates a memory buffer for a NULL row. The size of this buffer is equal to the row size for the output RowGroup. For the above scenario, using the default group_concat_max_len (which is a server variable that sets the maximum length of the GROUP_CONCAT string) value of 1mb, the buffer size would be
(1mb + 64kb + some additional metadata). If the user sets group_concat_max_len to a higher value, say 3gb, this buffer size would be ~3gb. Now if the runtime initiates several instances of RowAggregation, total memory consumption by PrimProc could exceed the hardware memory limits causing the OS OOM to kill the process. We fix this problem by again enabling the StringStore for the NULL row allocation.

3. In the plugin code in buildAggregateColumn(), there is an integer overflow when the server group_concat_max_len variable (which is an uint32_t) is set to a value > INT32_MAX (such as 3gb) and is assigned to
CalpontSystemCatalog::ColType::colWidth (which is an int32_t). As a short term fix, we saturate the assigned value to colWidth to INT32_MAX. Proper fix would be to upgrade
CalpontSystemCatalog::ColType::colWidth to an uint32_t.

### DIFF
--- a/dbcon/mysql/ha_mcs_execplan.cpp
+++ b/dbcon/mysql/ha_mcs_execplan.cpp
@@ -5349,7 +5349,24 @@ ReturnedColumn* buildAggregateColumn(Item* item, gp_walk_info& gwi)
         // Item_func_group_concat* gc = (Item_func_group_concat*)isp;
         CalpontSystemCatalog::ColType ct;
         ct.colDataType = CalpontSystemCatalog::VARCHAR;
-        ct.colWidth = isp->max_length;
+
+        // MCOL-5429 CalpontSystemCatalog::ColType::colWidth is currently
+        // stored as an int32_t (see calpontsystemcatalog.h). However,
+        // Item_sum::max_length is an uint32_t. This means there will be an
+        // integer overflow when Item_sum::max_length > colWidth. This ultimately
+        // causes an array index out of bound in GroupConcator::swapStreamWithStringAndReturnBuf()
+        // in groupconcat.cpp when ExeMgr processes groupconcat. As a temporary
+        // fix, we cap off the max groupconcat length to INT32_MAX. The proper fix
+        // would be to change colWidth type to uint32_t.
+        if (isp->max_length <= INT32_MAX)
+        {
+          ct.colWidth = isp->max_length;
+        }
+        else
+        {
+          ct.colWidth = INT32_MAX;
+        }
+
         ct.precision = 0;
         ac->resultType(ct);
       }
--- a/utils/rowgroup/rowaggregation.cpp
+++ b/utils/rowgroup/rowaggregation.cpp
@@ -649,7 +649,7 @@ void RowAggregation::resetUDAF(RowUDAFFunctionCol* rowUDAF, uint64_t funcColsIdx
 // Initilalize the data members to meaningful values, setup the hashmap.
 // The fRowGroupOut must have a valid data pointer before this.
 //------------------------------------------------------------------------------
-void RowAggregation::initialize()
+void RowAggregation::initialize(bool hasGroupConcat)
 {
   // Calculate the length of the hashmap key.
   fAggMapKeyCount = fGroupByCols.size();
@@ -687,9 +687,25 @@ void RowAggregation::initialize()
   makeAggFieldsNull(fRow);
 
   // Keep a copy of the null row to initialize new map entries.
-  fRowGroupOut->initRow(&fNullRow, true);
-  fNullRowData.reset(new uint8_t[fNullRow.getSize()]);
-  fNullRow.setData(rowgroup::Row::Pointer(fNullRowData.get()));
+  // MCOL-5429 Use stringstore if the datatype of the groupconcat
+  // field is a long string.
+  if (hasGroupConcat && fRowGroupOut->hasLongString())
+  {
+    fNullRowGroup = *fRowGroupOut;
+    fNullRowGroup.setUseStringTable(true);
+    fNullRowRGData.reinit(fNullRowGroup, 1);
+    fNullRowGroup.setData(&fNullRowRGData);
+    fNullRowGroup.resetRowGroup(0);
+    fNullRowGroup.initRow(&fNullRow);
+    fNullRowGroup.getRow(0, &fNullRow);
+  }
+  else
+  {
+    fRowGroupOut->initRow(&fNullRow, true);
+    fNullRowData.reset(new uint8_t[fNullRow.getSize()]);
+    fNullRow.setData(rowgroup::Row::Pointer(fNullRowData.get()));
+  }
+
   copyRow(fRow, &fNullRow);
 
   // Lazy approach w/o a mapping b/w fFunctionCols idx and fRGContextColl idx
@@ -2407,7 +2423,7 @@ void RowAggregationUM::endOfInput()
 //------------------------------------------------------------------------------
 // Initilalize the Group Concat data
 //------------------------------------------------------------------------------
-void RowAggregationUM::initialize()
+void RowAggregationUM::initialize(bool hasGroupConcat)
 {
   if (fGroupConcat.size() > 0)
     fFunctionColGc = fFunctionCols;
@@ -2417,7 +2433,7 @@ void RowAggregationUM::initialize()
     fKeyRG = fRowGroupIn.truncate(fGroupByCols.size());
   }
 
-  RowAggregation::initialize();
+  RowAggregation::initialize(fGroupConcat.size() > 0);
 }
 
 //------------------------------------------------------------------------------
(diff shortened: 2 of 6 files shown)


tntnatbry:
Please don't merge until QA tests the crashing cases on this PR build.

mariadb-SergeyZefirov (approved):
I agree with the changes.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -5349,7 +5349,24 @@ ReturnedColumn* buildAggregateColumn(Item* item, gp_walk_info& gwi)
...
+
+        // MCOL-5429 CalpontSystemCatalog::ColType::colWidth is currently
+        // stored as an int32_t (see calpontsystemcatalog.h). However,
+        // Item_sum::max_length is an uint32_t. This means there will be an
+        // integer overflow when Item_sum::max_length > colWidth. This ultimately
+        // causes an array index out of bound in GroupConcator::swapStreamWithStringAndReturnBuf()
+        // in groupconcat.cpp when ExeMgr processes groupconcat. As a temporary
+        // fix, we cap off the max groupconcat length to INT32_MAX. The proper fix
+        // would be to change colWidth type to uint32_t.
+        if (isp->max_length <= INT32_MAX)
```
drrtuy, line 5361: Does INT32_MAX const come from MDB namespaces?

utils/rowgroup/rowaggregation.cpp
```diff
@@ -687,9 +687,25 @@ void RowAggregation::initialize()
...
 
   // Keep a copy of the null row to initialize new map entries.
-  fRowGroupOut->initRow(&fNullRow, true);
-  fNullRowData.reset(new uint8_t[fNullRow.getSize()]);
-  fNullRow.setData(rowgroup::Row::Pointer(fNullRowData.get()));
+  // MCOL-5429 Use stringstore if the datatype of the groupconcat
+  // field is a long string.
+  if (hasGroupConcat && fRowGroupOut->hasLongString())
+  {
+    fNullRowGroup = *fRowGroupOut;
```
drrtuy, line 694: I didn't get it. Do you add fNullRowGroup to just use it for init here? Why then do you need to have the fNullRowGroup as RowAggregation attribute?

drrtuy:
The patch LGTM. There are some questions to be clarified before we merge the patch though.
A separate but relevant observation. There is clone of GroupConcat to deliver JSON_ARRAYAGG functionality. Could you @tntnatbry  plz look if it is also affected but the inefficient memory allocations so we can file another Jira ?

dbcon/mysql/ha_mcs_execplan.cpp
tntnatbry: No, this is the c const literal defined in the `<cstdint>` header.

utils/rowgroup/rowaggregation.cpp
tntnatbry: Yes, fNullRowGroup is used here to init a single row RGData with StringStore enabled, and then make Row point to this raw data via fNullRowGroup.

tntnatbry:
@drrtuy Ok, let me look into the JSON_ARRAYAGG scenario.

dbcon/mysql/ha_mcs_execplan.cpp
drrtuy: GTK.
JFYI Modern cpp favors std::numeric_limits<int32_t>::max() instead of those constants.

utils/rowgroup/rowaggregation.cpp
drrtuy: IMHO It is worth to limit the span of this RG and remove it from the list of RowAggregation attributes.

dbcon/mysql/ha_mcs_execplan.cpp
tntnatbry: That's correct. Thanks for highlighting the use of `std::numeric_limits` here.

utils/rowgroup/rowaggregation.cpp
tntnatbry: `fNullRowData` which is also just used in the else block is also an attribute of RowAggregation. We can keep this as is to maintain the code semantics here.

drrtuy: Feel free to remove fNullRowData also b/c these two are wastes RAM in runtime.

drrtuy: Or you can just save some some bytes not following the wrecked pattern and left fNullRowData intact if you don't want to increase the scope.

tntnatbry: No. This is not wrecked pattern. The reason fNullRowData is an attribute of the class is because it owns a memory buffer to which fNullRow points. If we remove fNullRowData as a class member and make it local variable with the scope of the else block, then the RGData dtor would get called and destroy the underlying memory buffer, causing fNullRow to point to an invalid memory address.

drrtuy: Makes sense. You don't need RG to stay around as RowStorage attribute though, right?

tntnatbry: `fRowGroup` is tightly coupled with `fRowRGData`, so I am keeping the declarations of both at the same place.

drrtuy: But this tight coupling spends 312 bytes per each RowAggregation instance just for nothing and you use RG only once!

drrtuy: I didn't get your logic behind this. IMHO this approach values the form more then the semantics but I won't push on this anymore.

drrtuy (approved):

utils/rowgroup/rowaggregation.cpp
tntnatbry: If I make `fNullRowGroup` local to the scope of the if block, now `fNullRowGroup.initRow(&fNullRow);` would be invalid. This is because `initRow()` is making some `Row` pointer attributes point to the `RowGroup` objects (`Row::oldOffsets`, `Row::stOffsets` are pointers to their corresponding objects defined in the `RowGroup` class). As soon as `fNullRowGroup` goes out of scope, these `offset` objects are destroyed and the corresponding `Row` pointers to these offsets would now be invalid.

Below is the `initRow` definition. If you look at the if-else blocks, you will see the pointer assignments to the offset attributes:
```c++
void RowGroup::initRow(Row* r, bool forceInlineData) const                                                                                                                                 
{                                                                                                                                                                  
  r->columnCount = columnCount;                                                                                                                                    
                                                                                                                                                                                   
  if (LIKELY(!types.empty()))                                                                                                                                                                                                                       
  {                                                                                                                                                                                                                                      
    r->colWidths = (uint32_t*)&colWidths[0];                                                                                                                                                                                       
    r->types = (execplan::CalpontSystemCatalog::ColDataType*)&(types[0]);                                                                                                
    r->charsetNumbers = (uint32_t*)&(charsetNumbers[0]);                                                                                                                                                                                            
    r->charsets = (CHARSET_INFO**)&(charsets[0]);                                                                                                                         
    r->scale = (uint32_t*)&(scale[0]);                                                                                                                                                                                                              
    r->precision = (uint32_t*)&(precision[0]);                                                                                                                      
  }                                                                                                                                                                
                                                                                                                                                                                                     
  if (forceInlineData)                                                                                                                                              
  {                                                                                                                                                                                                                                                 
    r->useStringTable = false;                                                                                                                                                                                       
    r->oldOffsets = (uint32_t*)&(oldOffsets[0]);                                                                                                                   
    r->stOffsets = (uint32_t*)&(stOffsets[0]);                                                                                                                                                
    r->offsets = (uint32_t*)&(oldOffsets[0]);                                                                                                                                                           
  }                                                                                                                                                                
  else                                                                                                                                                                                     
  {                                                                                                                                                                
    r->useStringTable = useStringTable;                                                                                                                                            
    r->oldOffsets = (uint32_t*)&(oldOffsets[0]);                                                                                                                                                                                                    
    r->stOffsets = (uint32_t*)&(stOffsets[0]);                                                                                                                                                                                           
    r->offsets = offsets;                                                                                                                                                                                                          
  }                                                                                                                                                                      
                                                                                                                                                                                                                   
  r->hasLongStringField = hasLongStringField;                                                                                                                             
  r->sTableThreshold = sTableThreshold;                                                                                                                             
  r->forceInline = forceInline;                                                                                                                                    
  r->hasCollation = hasCollation;                                                                                                                                                                                            
}
```

drrtuy: This RowGroup placement makes sense to me now. Thank you for the patience and the detailed explanation.

PR #2831, opened 2023-04-26
Managed shmem segment remap race

drrtuy:


versioning/BRM/extentmap.cpp
```diff
@@ -219,15 +219,20 @@ boost::mutex ExtentMapRBTreeImpl::fInstanceMutex;
 ExtentMapRBTreeImpl* ExtentMapRBTreeImpl::fInstance = nullptr;
 
 /*static*/
-ExtentMapRBTreeImpl* ExtentMapRBTreeImpl::makeExtentMapRBTreeImpl(unsigned key, off_t size, bool readOnly)
+ExtentMapRBTreeImpl* ExtentMapRBTreeImpl::makeExtentMapRBTreeImpl(unsigned key, off_t size, bool readOnly,
+                                                                  const MasterSegmentTable* emSegTable)
```
tntnatbry, line 223: `emSegTable` should not be const here.

versioning/BRM/extentmap.cpp
```diff
@@ -219,15 +219,20 @@ boost::mutex ExtentMapRBTreeImpl::fInstanceMutex;
...
 
   if (fInstance)
   {
     if (key != fInstance->fManagedShm.key())
     {
+      if (emSegTable)
+        emSegTable->getTable_upgrade(MasterSegmentTable::EMTable);
       fInstance->fManagedShm.reMapSegment();
+      if (emSegTable)
+        emSegTable->getTable_downgrade(MasterSegmentTable::EMTable);
```
tntnatbry, line 235: This should be changed to:
```c++
      if (emSegTable)
      {
        emSegTable->getTable_upgrade(MasterSegmentTable::EMTable);
        emLocked = true;
      }
      fInstance->fManagedShm.reMapSegment();
      if (emSegTable)
      {
        emLocked = false;
        emSegTable->getTable_downgrade(MasterSegmentTable::EMTable);
      }
```
Obviously, the function signature would need to be changed appropriately to pass `emLocked` from the caller.

versioning/BRM/extentmap.h
```diff
@@ -1187,6 +1187,8 @@ class ExtentMap : public Undoable
   ExtentMapRBTreeImpl* fPExtMapRBTreeImpl;
   FreeListImpl* fPFreeListImpl;
   ExtentMapIndexImpl* fPExtMapIndexImpl_;
+
+  friend class UTTester;
```
tntnatbry, line 1191: I am not sure what this change is.

tntnatbry (changes requested):

versioning/BRM/extentmap.cpp
drrtuy: Why?

drrtuy:
Thank you for the review. JFYI this version was not to be reviewed. I pushed it to check the build.

versioning/BRM/extentmap.cpp
drrtuy: Agree

versioning/BRM/extentmap.cpp
drrtuy: It is a constant pointer to a MST and I don't change it so I don't see any reason to remove const here.

tntnatbry: I just noticed `getTable_upgrade` and `getTable_downgrade` are const functions, so this looks good to me.

versioning/BRM/extentmap.h
drrtuy: Doesn't matter b/c it is gone.

PR #2835, opened 2023-04-27
MCOL-5487 Managed shmem segment access/remap race

drrtuy:
There is a potential edge case when a thread remaps EM
    managed shmem segment in grabEMEntryTable() whilst another
    thread access the current shmem segment mapping in virtual memory

### DIFF
--- a/versioning/BRM/extentmap.cpp
+++ b/versioning/BRM/extentmap.cpp
@@ -219,7 +219,9 @@ boost::mutex ExtentMapRBTreeImpl::fInstanceMutex;
 ExtentMapRBTreeImpl* ExtentMapRBTreeImpl::fInstance = nullptr;
 
 /*static*/
-ExtentMapRBTreeImpl* ExtentMapRBTreeImpl::makeExtentMapRBTreeImpl(unsigned key, off_t size, bool readOnly)
+ExtentMapRBTreeImpl* ExtentMapRBTreeImpl::makeExtentMapRBTreeImpl(unsigned key, off_t size, bool readOnly,
+                                                                  bool& emLocked,
+                                                                  const MasterSegmentTable* emSegTable)
 {
   boost::mutex::scoped_lock lk(fInstanceMutex);
 
@@ -227,7 +229,17 @@ ExtentMapRBTreeImpl* ExtentMapRBTreeImpl::makeExtentMapRBTreeImpl(unsigned key, 
   {
     if (key != fInstance->fManagedShm.key())
     {
+      if (emSegTable)
+      {
+        emSegTable->getTable_upgrade(MasterSegmentTable::EMTable);
+        emLocked = true;
+      }
       fInstance->fManagedShm.reMapSegment();
+      if (emSegTable)
+      {
+        emLocked = false;
+        emSegTable->getTable_downgrade(MasterSegmentTable::EMTable);
+      }
     }
 
     return fInstance;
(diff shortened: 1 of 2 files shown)


versioning/BRM/extentmap.cpp
```diff
@@ -1941,7 +1953,11 @@ void ExtentMap::grabEMEntryTable(OPS op)
     }
     else
     {
-      fPExtMapRBTreeImpl = ExtentMapRBTreeImpl::makeExtentMapRBTreeImpl(fEMRBTreeShminfo->tableShmkey, 0);
+      const bool isReadOnly = false;
+      // The ternary sends EM segment table only if this grab is for a read to allow to upgrade the RWLock if
+      // case of a shmem segment remap.
+      fPExtMapRBTreeImpl = ExtentMapRBTreeImpl::makeExtentMapRBTreeImpl(
+          fEMRBTreeShminfo->tableShmkey, 0, isReadOnly, emLocked, op == READ ? &fMST : nullptr);
```
tntnatbry, line 1960: Can you please remove the var `isReadOnly` and directly pass `false` to the function.

Also, a minor grammar error in the comment: "if case of a shmem" should be replaced by "in case of a shmem".

tntnatbry (changes requested):
Patch LGTM. I have added one comment to be addressed.

versioning/BRM/extentmap.cpp
```diff
@@ -219,15 +219,27 @@ boost::mutex ExtentMapRBTreeImpl::fInstanceMutex;
...
+                                                                  bool& emLocked,
+                                                                  const MasterSegmentTable* emSegTable)
 {
   boost::mutex::scoped_lock lk(fInstanceMutex);
 
   if (fInstance)
   {
     if (key != fInstance->fManagedShm.key())
     {
+      if (emSegTable)
```
mariadb-DenisKhalikov, line 232: @drrtuy I have more conservative idea, how about to grab "write lock" inside grabEmEntyTable
right after this condition 
 `if (!fPExtMapRBTreeImpl || fPExtMapRBTreeImpl->key() != (uint32_t)fEMRBTreeShminfo->tableShmkey)`
 
in this case have a "write lock" either EM is nullptr for current process or the keys are different between current process and global segment table. What do you think?

versioning/BRM/extentmap.cpp
drrtuy: This variable is here for readability so I am not inclined to remove it.

versioning/BRM/extentmap.cpp
drrtuy: Reasonable suggestion however with this approach we:
- increase crit section scope thus we increase a potential shmem RWLock contention
- crit section now covers risk-free cases, e.g. the condition key == fInstance->fManagedShm.key()  evaluates to true in makeExtentMapRBTreeImpl.

versioning/BRM/extentmap.cpp
drrtuy: Changed the comment wording though.

versioning/BRM/extentmap.cpp
mariadb-DenisKhalikov: Does not look like a possible solution we either have a makeExtentMapRBTreeImpl == nullptr or keys are not equal, in other cases we are not going to this part.

versioning/BRM/extentmap.cpp
```diff
@@ -219,15 +219,27 @@ boost::mutex ExtentMapRBTreeImpl::fInstanceMutex;
...
-ExtentMapRBTreeImpl* ExtentMapRBTreeImpl::makeExtentMapRBTreeImpl(unsigned key, off_t size, bool readOnly)
+ExtentMapRBTreeImpl* ExtentMapRBTreeImpl::makeExtentMapRBTreeImpl(unsigned key, off_t size, bool readOnly,
+                                                                  bool& emLocked,
+                                                                  const MasterSegmentTable* emSegTable)
 {
   boost::mutex::scoped_lock lk(fInstanceMutex);
 
   if (fInstance)
   {
     if (key != fInstance->fManagedShm.key())
```
tntnatbry, line 230: Can you please replace this if condition with:
```c++
if (key != fInstance->key())
```
Similarly, in `FreeListImpl::makeFreeListImpl()`, please replace the if condition with:
```c++
if (key != fInstance->key())
```

tntnatbry (dismissed):

versioning/BRM/extentmap.cpp
drrtuy: Internally the keys have the same value so I am not going to make any additional changes for 23.02. I will probably change it in develop after discussion with @denis0x0D .

tntnatbry: Sure. Here, `fInstance->fManagedShm.key() == fInstance->key()`. Using `fInstance->key()` here is more readable.

The calling function `grabEMEntryTable()` also uses the member function `ExtentMapRBTreeImpl::key()`:
`if (!fPExtMapRBTreeImpl || fPExtMapRBTreeImpl->key() != (uint32_t)fEMRBTreeShminfo->tableShmkey)`

mariadb-DenisKhalikov (approved):

PR #2836, opened 2023-05-02
MCOL-5357 Fix TPC-DS query error "MCS-3009: Unknown column '.<colname>'".

tntnatbry:
For the following query:

select item from (
select item from (select a as item from t1) tt
union all
select item from (select a as item from t1) tt
) ttt;

There is an if predicate in buildSimpleColFromDerivedTable() that compares the outermost query field name (ttt.item) to the returned column list of the inner query (tt.item) when building the returned column list of the outer most query. In the above query example, the inner query field name is an alias set in the inner most query and is set to "`tt`.`item`", while the outermost query field name is set to "item". The use of backticks "`" in the inner query alias is causing the execution to not enter the if block which creates the SimpleColumn for the outermost query field name. As a fix, we strip off the backticks from the inner query alias.

### DIFF
--- a/dbcon/mysql/ha_mcs_execplan.cpp
+++ b/dbcon/mysql/ha_mcs_execplan.cpp
@@ -2454,6 +2454,20 @@ SimpleColumn* buildSimpleColFromDerivedTable(gp_walk_info& gwi, Item_field* ifp)
         SimpleColumn* col = dynamic_cast<SimpleColumn*>(cols[j].get());
         string alias = cols[j]->alias();
 
+        // MCOL-5357 For the following query:
+
+        // select item from (
+        // select item from (select a as item from t1) tt
+        // union all
+        // select item from (select a as item from t1) tt
+        // ) ttt;
+
+        // When the execution reaches the outermost item (ttt.item),
+        // alias = "`tt`.`item`" and ifp->field_name.str = "item".
+        // To make the execution enter the if block below, we strip off
+        // the backticks from alias.
+        boost::erase_all(alias, "`");
+
         if (strcasecmp(ifp->field_name.str, alias.c_str()) == 0 ||
             (col && alias.find(".") != string::npos &&
              (strcasecmp(ifp->field_name.str, col->columnName().c_str()) == 0 ||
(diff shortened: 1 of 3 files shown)


dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -2454,6 +2454,20 @@ SimpleColumn* buildSimpleColFromDerivedTable(gp_walk_info& gwi, Item_field* ifp)
...
+        // select item from (select a as item from t1) tt
+        // union all
+        // select item from (select a as item from t1) tt
+        // ) ttt;
+
+        // When the execution reaches the outermost item (ttt.item),
+        // alias = "`tt`.`item`" and ifp->field_name.str = "item".
+        // To make the execution enter the if block below, we strip off
+        // the backticks from alias.
+        boost::erase_all(alias, "`");
```
drrtuy, line 2469: I am not sure if this function doesn't trigger locale creation under the hood. Some of text-related functions in boos do this. If it does this hits OS level global mutex that I would avoid. Feel free to replace erase_all with a simple loop over cols[j]->alias() chars.

drrtuy (changes requested):
Take a look at the comment inline.

dbcon/mysql/ha_mcs_execplan.cpp
tntnatbry: It does not look like this function is triggering locale creation. There is a case-insensitive version of this function which does create a locale `boost::ierase_all()`. Here is the boost doc for both the functions:
https://www.boost.org/doc/libs/1_82_0/doc/html/boost/algorithm/erase_all.html
https://www.boost.org/doc/libs/1_82_0/doc/html/boost/algorithm/ierase_all.html

But there is an equivalent way to achieve the same result using STL:
```c++
alias.erase(std::remove(alias.begin(), alias.end(), '`'), alias.end());
```
Let me know if you want me to use this instead. The boost version does look more readable to me however.

drrtuy: No, boost expression should be fine. Thx for the detailed explanation.

drrtuy (approved):

PR #2842, opened 2023-05-18
MCOL-5493: First portion of UBSan fixes

mariadb-LeonidFedorov:


utils/rowgroup/rowgroup.h
```diff
@@ -1329,7 +1329,8 @@ inline void Row::setVarBinaryField(const uint8_t* val, uint32_t len, uint32_t co
       len = getColumnWidth(colIndex);
 
     *((uint16_t*)&data[offsets[colIndex]]) = len;
-    memcpy(&data[offsets[colIndex] + 2], val, len);
+    if (val != nullptr)
```
drrtuy, line 1334: IMHO, if we compare val with null ptr the else should log this.
We could replace this if with idbassert but it doesn't look good to assert so logging suits better IMHO.

primitives/linux-port/dictionary.cpp
```diff
@@ -597,7 +597,8 @@ void PrimitiveProcessor::p_Dictionary(const DictInput* in, vector<uint8_t>* out,
         outValue = reinterpret_cast<DataValue*>(&(*out)[header.NBYTES]);
         outValue->isnull = sigptr.data == nullptr;
         outValue->len = sigptr.len;
-        memcpy(outValue->data, sigptr.data, sigptr.len);
+        if (sigptr.data != nullptr)
```
drrtuy, line 600: IMHO, if we compare sigptr.data with null ptr the else should log this.
We could replace this if with idbassert but it doesn't look good to assert so logging suits better IMHO.

drrtuy (changes requested):

utils/rowgroup/rowgroup.h
mariadb-LeonidFedorov: As you can see in line 1316, nullptr is expected here. I've added assert here ` idbassert(val != nullptr || !len);`

dbcon/execplan/calpontselectexecutionplan.cpp
```diff
@@ -106,55 +105,15 @@ CalpontSelectExecutionPlan::CalpontSelectExecutionPlan(
...
- , fDistinct(false)
- , fOverrideLargeSideEstimate(false)
- , fDistinctUnionNum(0)
- , fSubType(MAIN_SELECT)
- , fLimitStart(0)
- , fLimitNum(-1)
- , fHasOrderBy(false)
- , fStringScanThreshold(ULONG_MAX)
- , fQueryType(SELECT)
  , fPriority(querystats::DEFAULT_USER_PRIORITY_LEVEL)
```
drrtuy, line 108: Why didn't you move this assignment into a class declaration also?

drrtuy (approved):

dbcon/execplan/calpontselectexecutionplan.cpp
mariadb-LeonidFedorov: I was going to write my hand got a case of the jitters, but there was the reason.
That's because `querystats::DEFAULT_USER_PRIORITY_LEVEL` is defined in `querystats.h` header which includes 
`resourcemanager.h` header which includes `calpontselectexecutionplan.h` and I will get circular dependency if simply add required header, so I decided it's appropriate tradeoff here not to refactor more code and leave this field

PR #2851, opened 2023-05-26
MCOL-5477 Disk join step improvement.

denis0x0D:
This patch:
1. Handles corner case when the bucket exceeded the memory limit, but we cannot redistribute the data in this bucket into new buckets based on a hash algorithm, because the rows have the same values.
2. Adds force option for disk join step.

### DIFF
--- a/dbcon/joblist/tuplehashjoin.cpp
+++ b/dbcon/joblist/tuplehashjoin.cpp
@@ -106,6 +106,8 @@ TupleHashJoinStep::TupleHashJoinStep(const JobInfo& jobInfo)
   djsSmallLimit = jobInfo.smallSideLimit;
   djsLargeLimit = jobInfo.largeSideLimit;
   djsPartitionSize = jobInfo.partitionSize;
+  djsMaxPartitionTreeDepth = jobInfo.djsMaxPartitionTreeDepth;
+  djsForceRun = jobInfo.djsForceRun;
   isDML = jobInfo.isDML;
 
   config::Config* config = config::Config::makeConfig();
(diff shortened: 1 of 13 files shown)


dbcon/joblist/tuplehashjoin.cpp
```diff
@@ -1971,53 +1977,69 @@ void TupleHashJoinStep::segregateJoiners()
...
+    done by DJS. */
+    if (allInnerJoins)
     {
-      // if (joiners[i]->isFinished() && i < limit) {  // debugging
-      if (joiners[i]->isFinished())
+      for (i = 0; i < smallSideCount; i++)
       {
-        // cout << "2joiner " << i << "  " << hex << (uint64_t) joiners[i].get() << dec << " -> TBPS" << endl;
-        tbpsJoiners.push_back(joiners[i]);
+        // if (joiners[i]->isFinished() && (rand() % 2)) {    // for debugging
```
drrtuy: Plz rm the line commented out.

dbcon/joblist/tuplehashjoin.cpp
```diff
@@ -1971,53 +1977,69 @@ void TupleHashJoinStep::segregateJoiners()
...
+        {
+          // cout << "1joiner " << i << "  " << hex << (uint64_t) joiners[i].get() << dec << " -> TBPS" <<
+          // endl;
+          tbpsJoiners.push_back(joiners[i]);
+        }
+        else
+        {
+          joinIsTooBig = true;
+          joiners[i]->setConvertToDiskJoin();
+          // cout << "1joiner " << i << "  " << hex << (uint64_t) joiners[i].get() << dec << " -> DJS" <<
```
drrtuy: dItto

dbcon/joblist/tuplehashjoin.cpp
```diff
@@ -1971,53 +1977,69 @@ void TupleHashJoinStep::segregateJoiners()
...
-      // cout << "2joiner " << i << "  " << hex << (uint64_t) joiners[i].get() << dec << " -> DJS" << endl;
-      djsJoiners.push_back(joiners[i]);
-      djsJoinerMap.push_back(i);
+      // uint limit = rand() % smallSideCount;
+      for (i = 0; i < smallSideCount; i++)
+      {
+        // if (joiners[i]->isFinished() && i < limit) {  // debugging
+        if (joiners[i]->isFinished())
+        {
+          // cout << "2joiner " << i << "  " << hex << (uint64_t) joiners[i].get() << dec << " -> TBPS" <<
```
drrtuy: ditto

dbcon/joblist/tuplehashjoin.cpp
```diff
@@ -1971,53 +1977,69 @@ void TupleHashJoinStep::segregateJoiners()
...
-
-    for (; i < smallSideCount; i++)
+    else
     {
-      joinIsTooBig = true;
-      joiners[i]->setConvertToDiskJoin();
-      // cout << "2joiner " << i << "  " << hex << (uint64_t) joiners[i].get() << dec << " -> DJS" << endl;
-      djsJoiners.push_back(joiners[i]);
-      djsJoinerMap.push_back(i);
+      // uint limit = rand() % smallSideCount;
```
drrtuy: ditto

dbcon/joblist/tuplehashjoin.cpp
```diff
@@ -1971,53 +1977,69 @@ void TupleHashJoinStep::segregateJoiners()
...
+        }
+        else
+          break;
+      }
+
+      for (; i < smallSideCount; i++)
+      {
+        joinIsTooBig = true;
+        joiners[i]->setConvertToDiskJoin();
+        // cout << "2joiner " << i << "  " << hex << (uint64_t) joiners[i].get() << dec << " -> DJS" << endl;
```
drrtuy: ditto

oam/etc/Columnstore.xml
```diff
@@ -189,6 +189,8 @@
 		<TotalUmMemory>25%</TotalUmMemory>
 		<CPUniqueLimit>100</CPUniqueLimit>
 		<AllowDiskBasedJoin>N</AllowDiskBasedJoin>
+		<ForceDiskBasedJoin>N</ForceDiskBasedJoin>
```
drrtuy: I would suggest you to write a markdown file with these options and short explanations in doc directory. IMHO you shouldn't put these options into config. Nobody would use the first option except developers.

drrtuy (approved):
Plz remove lines commented out.

oam/etc/Columnstore.xml
mariadb-AlexeyVorovich: Agree.. these options should be hidden from users and support

denis0x0D: Without Force option there is no guarantee that DJS will be enabled, and therefore for the large join we wil have OOM.

denis0x0D: UPD: After discussion, I'll put it into session variables and remove from config file.

drrtuy: Much appreciated

PR #2863, opened 2023-06-08
CI from develop + cmapi [develop-23.02]

mariadb-LeonidFedorov:
This patch do not try no merge .drone.jsonnet from original PR, but replaces it with develop version.
scripts used by CI were also replaced with develop versions
Let's see the builds result


tntnatbry:
@mariadb-LeonidFedorov It looks like test001/test000 failed for Centos7 and Ubuntu20.

mariadb-AlanMologorsky (approved):
LGTM
Builds seems to be ok, except centos7 and ubuntu20.
The reason is in missed `./regression_logs.sh` file.

mariadb-RomanNavrotskiy (approved):

mariadb-LeonidFedorov:
@tntnatbry the only test that fails is multinode MTR on rocky 8

drrtuy (approved):
There are two points regarding the patch that must be addressed ASAP IMHO:
- license should be less restrictive
- engine_files dir is obsolete and misguiding so must be deleted

PR #2864, opened 2023-06-08
Columnstore 23.02 CS update for stock GCC[develop-23.02]

mariadb-LeonidFedorov:


dbcon/execplan/rewrites.cpp
```diff
@@ -59,12 +59,12 @@ SimpleFilter* castToSimpleFilter(execplan::TreeNode* node)
 bool commonContainsSemantic(const CommonContainer& common, execplan::ParseTree* node)
 {
   auto filter = castToFilter(node);
-  return filter && common.first.contains(node);
+  return filter && common.first.count(node) != 0;
```
tntnatbry, line 62: Instead of `count`, calling `find` might be more efficient here:
```c++
return filter && common.first.find(node) != common.first.end();
```

dbcon/joblist/batchprimitiveprocessor-jl.cpp
```diff
@@ -890,7 +890,8 @@ void BatchPrimitiveProcessorJL::getRowGroupData(ByteStream& in, vector<RGData>*
 
         if (joinResults.get() == NULL)
         {
-          joinResults.reset(new vector<uint32_t>[8192]);
+          auto v = new vector<uint32_t>[8192];
+          joinResults.reset(v);
```
tntnatbry, line 894: I don't see any issue with the original code. Is this change related to the `set::contains()` change?

tntnatbry (changes requested):

dbcon/execplan/rewrites.cpp
mariadb-LeonidFedorov: std::set doesn't allow duplicates, so count is pretty the same, but shorter version

dbcon/joblist/batchprimitiveprocessor-jl.cpp
mariadb-LeonidFedorov: This is changes serg made for getting rid of buggy compiler warnings on Ubuntu 22.10. I'm ok with it. 
You could get this from the commit message from serg within this PR

dbcon/execplan/rewrites.cpp
mariadb-LeonidFedorov: that's how count implemented in std::set
```
      size_type
      count(const key_type& __x) const
      { return _M_t.find(__x) == _M_t.end() ? 0 : 1; }
```

tntnatbry: Sounds good.

dbcon/joblist/batchprimitiveprocessor-jl.cpp
tntnatbry: Thanks for clarifying.

tntnatbry (approved):
@mariadb-LeonidFedorov Changes look good to me. JFYI Rocky8 build is failing due to the known MTR issue.

tntnatbry:
@mariadb-LeonidFedorov Just a reminder that we need to merge this PR into develop-23.02 before the release. There is a conflict in the VERSION file. Can you please rebase this branch on top of latest develop-23.02.

Also, please create the duplicate PR for develop.

PR #2867, opened 2023-06-14
MCOL-5509, MCOL-5510: Fixed stdout and stderr forking procs handling

mariadb-LeonidFedorov:


cmapi/cmapi_server/controllers/s3dataload.py
```diff
@@ -288,24 +288,32 @@ def prepare_google_storage(
...
+                if name == 'downloader_error' and line:
                     downloader_error += line + '\n'
-                if name == 'cpimport_error':
+                if name == 'cpimport_error' and line:
                     cpimport_error += line + '\n'
-                if name == 'cpimport_output':
+                if name == 'cpimport_output' and line:
                     cpimport_output += line + '\n'
 
+            if downloader_error:
```
mariadb-AlanMologorsky, line 303: If we have `downloader_error` we'll never know about `cpimport_error`.

cmapi/cmapi_server/controllers/s3dataload.py
```diff
@@ -288,24 +288,32 @@ def prepare_google_storage(
...
+                response_error(downloader_error)
+
+            if cpimport_error:
+                response_error(cpimport_error)
+
+            cpimport_status = cpimport_proc.poll()
+            download_status = download_proc.poll()
+
+            if cpimport_status is not None \
```
mariadb-AlanMologorsky, line 312: Prefer to use braces in multiline condition.

mariadb-AlanMologorsky (changes requested):

cmapi/cmapi_server/controllers/s3dataload.py
mariadb-LeonidFedorov: That's sounds good, downloader error is premature for cpimport

mariadb-AlanMologorsky (approved):

PR #2878, opened 2023-06-23
MCOL-5514 Parallel disk join step

denis0x0D:


utils/joiner/joinpartition.h
```diff
@@ -137,7 +149,7 @@ class JoinPartition
   uint64_t largeSizeOnDisk;
   utils::Hasher_r hasher;
   bool rootNode;
-
+  bool canSplit;
```
drrtuy: According with the style guide we add a suffix '\_' to an attribute,e.g canSplit\_.

utils/joiner/joinpartition.h
```diff
@@ -163,6 +174,9 @@ class JoinPartition
   /* file descriptor reduction */
   size_t nextSmallOffset;
   size_t nextLargeOffset;
-};
 
+  // Options to control partition tree depth.
+  uint32_t currentPartitionTreeDepth;
```
drrtuy: ditto

dbcon/joblist/jlf_tuplejoblist.cpp
```diff
@@ -1841,7 +1841,7 @@ void CircularJoinGraphTransformer::chooseEdgeToTransform(Cycle& cycle,
     std::cout << "FK FK key not found, removing the first one inner join edge" << std::endl;
 
   // Take just a first.
-  resultEdge = std::make_pair(cycle.front(), 0 /*Dummy weight*/);
+  resultEdge = std::make_pair(cycle.back(), 0 /*Dummy weight*/);
```
drrtuy, line 1844:
```suggestion
  resultEdge = {cycle.back(), 0 /*Dummy weight*/};
```

dbcon/joblist/diskjoinstep.h
```diff
@@ -80,28 +88,55 @@ class DiskJoinStep : public JobStep
 
   uint64_t mainThread;  // thread handle from thread pool
 
+  struct JoinPartitionsProcessor
```
drrtuy, line 93: I suppose you can eliminate this class using Lambda instead.

dbcon/joblist/diskjoinstep.h
```diff
@@ -80,28 +88,55 @@ class DiskJoinStep : public JobStep
...
   struct LoaderOutput
   {
     std::vector<rowgroup::RGData> smallData;
     uint64_t partitionID;
     joiner::JoinPartition* jp;
   };
-  boost::shared_ptr<joblist::FIFO<boost::shared_ptr<LoaderOutput> > > loadFIFO;
+  std::vector<boost::shared_ptr<joblist::FIFO<boost::shared_ptr<LoaderOutput>>>> loadFIFO;
 
   struct Loader
```
drrtuy, line 124: This Loader class is also a runner that can be replaced with lambda.

dbcon/joblist/diskjoinstep.h
```diff
@@ -112,47 +147,53 @@ class DiskJoinStep : public JobStep
     joiner::JoinPartition* jp;
   };
 
-  boost::shared_ptr<joblist::FIFO<boost::shared_ptr<BuilderOutput> > > buildFIFO;
+  std::vector<boost::shared_ptr<joblist::FIFO<boost::shared_ptr<BuilderOutput>>>> buildFIFO;
 
   struct Builder
```
drrtuy, line 157: ditto

dbcon/joblist/diskjoinstep.h
```diff
@@ -112,47 +147,53 @@ class DiskJoinStep : public JobStep
...
+    uint32_t threadID;
   };
-  void buildFcn();
+  void buildFcn(const uint32_t threadID);
 
   /* Joining structs */
   struct Joiner
   {
-    Joiner(DiskJoinStep* d) : djs(d)
+    Joiner(DiskJoinStep* d, const uint32_t threadID) : djs(d), threadID(threadID)
```
drrtuy, line 175: ditto

dbcon/joblist/diskjoinstep.h
```diff
@@ -112,47 +147,53 @@ class DiskJoinStep : public JobStep
...
   int64_t largeLimit;
   uint64_t partitionSize;
+  uint32_t maxPartitionTreeDepth;
 
   void reportStats();
 
   uint32_t joinerIndex;
   bool closedOutput;
+
+  std::mutex outputMutex;
```
drrtuy, line 200: According with the style guide we add a suffix '' to an attribute,e.g canSplit.

dbcon/joblist/tuplehashjoin.cpp
```diff
@@ -198,10 +200,10 @@ void TupleHashJoinStep::join()
   joinRan = true;
   jobstepThreadPool.join(mainRunner);
 
-  if (djs)
+  if (djs.size())
   {
-    for (int i = 0; i < (int)djsJoiners.size(); i++)
-      djs[i].join();
+    for (uint32_t i = 0, size = djs.size(); i < size; ++i)
```
drrtuy:
```suggestion
    for (auto& djsThread: djs)
    {
      djs.join();
    }
```

dbcon/joblist/tuplehashjoin.cpp
```diff
@@ -629,10 +631,10 @@ void TupleHashJoinStep::djsReaderFcn(int index)
   while (more)
     more = fifos[index]->next(it, &rgData);
 
-  for (int i = 0; i < (int)djsJoiners.size(); i++)
+  for (uint32_t i = 0, size = djs.size(); i < size; ++i)
```
drrtuy: ditto

dbcon/joblist/tuplehashjoin.cpp
```diff
@@ -881,7 +883,7 @@ void TupleHashJoinStep::hjRunner()
   }
 
   // todo: forwardCPData needs to grab data from djs
-  if (!djs)
+  if (!djs.size())
```
drrtuy:
```suggestion
  if (djs.empty())
```
This form looks easier to read IMHO. What do you think?

dbcon/joblist/tuplehashjoin.cpp
```diff
@@ -931,7 +933,7 @@ void TupleHashJoinStep::hjRunner()
   {
     largeBPS->useJoiners(tbpsJoiners);
 
-    if (djs)
+    if (djs.size())
```
drrtuy, line 936: ditto

dbcon/joblist/tuplehashjoin.cpp
```diff
@@ -1971,53 +1973,55 @@ void TupleHashJoinStep::segregateJoiners()
...
+      joiners[i]->setConvertToDiskJoin();
+      djsJoiners.push_back(joiners[i]);
+      djsJoinerMap.push_back(i);
     }
   }
   else
   {
-    // uint limit = rand() % smallSideCount;
-    for (i = 0; i < smallSideCount; i++)
+    /* If they are all inner joins they can be segregated w/o respect to
```
drrtuy: I would rephrase the first sentence like this: "they can be segregated not taking ordering into account". I would also explain ordering of what are we talking about.

dbcon/joblist/tuplehashjoin.cpp
```diff
@@ -1971,53 +1973,55 @@ void TupleHashJoinStep::segregateJoiners()
...
+    done by DJS. */
+    if (allInnerJoins)
     {
-      // if (joiners[i]->isFinished() && i < limit) {  // debugging
-      if (joiners[i]->isFinished())
+      for (i = 0; i < smallSideCount; i++)
       {
-        // cout << "2joiner " << i << "  " << hex << (uint64_t) joiners[i].get() << dec << " -> TBPS" << endl;
-        tbpsJoiners.push_back(joiners[i]);
+        if (joiners[i]->isFinished())
```
drrtuy: I would add parenthesis.

dbcon/joblist/tuplehashjoin.cpp
```diff
@@ -1971,53 +1973,55 @@ void TupleHashJoinStep::segregateJoiners()
...
     {
-      joinIsTooBig = true;
-      joiners[i]->setConvertToDiskJoin();
-      // cout << "2joiner " << i << "  " << hex << (uint64_t) joiners[i].get() << dec << " -> DJS" << endl;
-      djsJoiners.push_back(joiners[i]);
-      djsJoinerMap.push_back(i);
+      for (i = 0; i < smallSideCount; i++)
+      {
+        if (joiners[i]->isFinished())
+          tbpsJoiners.push_back(joiners[i]);
```
drrtuy: ditto

dbcon/joblist/tuplehashjoin.h
```diff
@@ -620,6 +620,8 @@ class TupleHashJoinStep : public JobStep, public TupleDeliveryStep
   int64_t djsSmallLimit;
   int64_t djsLargeLimit;
   uint64_t djsPartitionSize;
+  uint32_t djsMaxPartitionTreeDepth;
```
drrtuy: According with the style guide we add a suffix '\_' to an attribute,e.g djsMaxPartitionTreeDepth\_.

dbcon/joblist/diskjoinstep.cpp
```diff
@@ -527,49 +560,117 @@ void DiskJoinStep::joinFcn()
...
+}
+
+void DiskJoinStep::spawnJobs(const std::vector<std::vector<JoinPartition*>>& joinPartitionsJobs,
+                             const uint32_t smallSideSizeLimitPerThread)
+{
+  const uint32_t threadsCount = joinPartitionsJobs.size();
+  std::vector<uint64_t> processorThreadsId;
+  processorThreadsId.reserve(threadsCount);
+  for (uint32_t threadID = 0; threadID < threadsCount; ++threadID)
+    processorThreadsId.push_back(jobstepThreadPool.invoke(
```
drrtuy, line 640: It is worth to add parenthesis here to avoid potential issues if the scope will be extended in future.

dbcon/joblist/diskjoinstep.cpp
```diff
@@ -527,49 +560,117 @@ void DiskJoinStep::joinFcn()
...
 }
 
+void DiskJoinStep::initializeFIFO(const uint32_t threadCount)
+{
+  loadFIFO.clear();
+  buildFIFO.clear();
+
+  for (uint32_t i = 0; i < threadCount; ++i)
+  {
+    boost::shared_ptr<joblist::FIFO<boost::shared_ptr<LoaderOutput>>> lFIFO(
```
drrtuy: My eyes got broken on these types :)
Could you define and use an alias?

dbcon/joblist/jlf_tuplejoblist.cpp
```diff
@@ -1841,7 +1841,7 @@ void CircularJoinGraphTransformer::chooseEdgeToTransform(Cycle& cycle,
     std::cout << "FK FK key not found, removing the first one inner join edge" << std::endl;
```
drrtuy: The first or the last now?

dbcon/joblist/jlf_tuplejoblist.cpp
```diff
@@ -1841,7 +1841,7 @@ void CircularJoinGraphTransformer::chooseEdgeToTransform(Cycle& cycle,
...
 
   // Take just a first.
```
drrtuy: Last ?

dbcon/joblist/diskjoinstep.cpp
```diff
@@ -260,30 +249,85 @@ void DiskJoinStep::largeReader()
...
+        rowGroup.setData(&rgData);
+
+        // Check that current `RowGroup` has rows.
+        if (!rowGroup.getRowCount())
+        {
+          partitionDone = true;
+          break;
+        }
+
+        currentSize += rowGroup.getRowCount() * rowGroup.getColumnCount() * 16;
```
drrtuy: This heuristic isn't clear for me. Why do you decide to go with 16  for a column width? There are methods to retrieve the actual RGData size that should be used IMHO. Or did I overlook something?

dbcon/joblist/diskjoinstep.cpp
```diff
@@ -527,49 +560,117 @@ void DiskJoinStep::joinFcn()
...
+  pipelineThreads.push_back(jobstepThreadPool.invoke(Joiner(this, threadID)));
+  jobstepThreadPool.join(pipelineThreads);
+}
+
+void DiskJoinStep::prepareJobs(const std::vector<JoinPartition*>& joinPartitions,
+                               std::vector<std::vector<JoinPartition*>>& joinPartitionsJobs)
+{
+  const uint32_t issuedThreads = jobstepThreadPool.getIssuedThreads();
+  const uint32_t maxNumOfThreads = jobstepThreadPool.getMaxThreads();
+  const uint32_t numOfThreads =
```
drrtuy, line 606: Nice one!

dbcon/joblist/diskjoinstep.cpp
```diff
@@ -527,49 +560,117 @@ void DiskJoinStep::joinFcn()
...
-        break;
+      // Collect all join partitions.
+      std::vector<JoinPartition*> joinPartitions;
+      jp->collectJoinPartitions(joinPartitions);
 
-      loadFIFO.reset(
-          new FIFO<boost::shared_ptr<LoaderOutput> >(1, 1));  // double buffering should be good enough
-      buildFIFO.reset(new FIFO<boost::shared_ptr<BuilderOutput> >(1, 1));
+      // Split partitions for each threads.
+      std::vector<std::vector<JoinPartition*>> joinPartitionsJobs;
```
drrtuy: It is worth to make a type alias with a semantic name.

drrtuy (changes requested):
Good job!
I don't see significant issues except maybe RGData size estimate formula in loadFcn. Other stuff is merely syntactical to make the code easy to read/align it with the styling.

mariadb-LeonidFedorov (approved):
I like the patch. Good work indeed! Same concerns about magic 16 in heuristics, maybe some comment will resolve it. IMHO   functor classes are suitable and expressive, and change them to the lambdas is not required. About typedefs for complex type I am absolutely agree. Nevertheless, for me it's NIT's

drrtuy:
> I like the patch. Good work indeed! Same concerns about magic 16 in heuristics, maybe some comment will resolve it. IMHO functor classes are suitable and expressive, and change them to the lambdas is not required. About typedefs for complex type I am absolutely agree. Nevertheless, for me it's NIT's

When I read the code Functor classes force me to navigate twice over the code to understand what does this simple functor do. Lambda is a syntactical sugar so why shouldn't we use it to reduce a number of code navigation moves?

dbcon/joblist/tuplehashjoin.h
denis0x0D: I agree, but all other variables in this code has different style, so we will have zoo of variables with different styles, how about keep this style, until we change them everywhere.

dbcon/joblist/diskjoinstep.cpp
denis0x0D: Thanks, fixed this, looks like a bug.

denis0x0D:
Thanks for review! I fixed all issues except:
1. Variable style - I think changing the name adding suffix will make a zoo of variables style in this code file. How about to keep this at it is, until fix everywhere.
2. Also do not like a lamdas and functional programming as well, those really hard to debug, also this the current code style allows to keep functor class.

drrtuy (approved):

PR #2885, opened 2023-06-29
MCOL-5521: Use Columnstore VERSION file for CMAPI versioning.

mariadb-AlanMologorsky:
- [x] [fix] columnstore_version.cmake setting extra variables for CMAPI versioning
- [x] [fix] CMAPI CmakeLists.txt using columnstore_version.cmake file
- [x] [fix] CMAPI VERSION file to template file due to using it inside CMAPI application


drrtuy (approved):
I personally would just reuse CS_VERSION_* vars in the template file but this is just my opinion and not mandatory request for change.

PR #2891, opened 2023-07-04
Feat[WIP]: JIT compiled SQL expressions

jun0315:
## Summary

MCS uses interepreted execution to calculate SQL expressions results. Here is an example of a SQL expression 'table1.col1 + FLOOR(table2.col1)'. Given that table1.col1 is DECIMAL and table2.col1 is DOUBLE there is a number of conditions that drives the calculation of this relatively simple example in runtime. Given that SQL types and expression tree are known before the query begins it is possible to replace interpretation with JIT to produce specialized compiled bytecode that is:

small
has no or almost no branches
optimized for the specific platform it is run at
This is a research mostly project which goal is to produce a set of microbenchmarks that:

leverages any JIT compiler available, e.g. LLVM, MIR
demonstrates a negative and positive effects of using JIT

## Related

[gsoc-2023](https://summerofcode.withgoogle.com/programs/2023/projects/hLufFI1y)


jit/jit.cpp
```diff
@@ -0,0 +1,467 @@
...
+void JIT::deleteCompiledModule(JIT::CompiledModule& module)
+{
+  std::lock_guard<std::mutex> lock(jit_lock);
+  auto module_memory_manager_it = module_identifier_to_memory_manager.find(module.identifier);
+  if (module_memory_manager_it == module_identifier_to_memory_manager.end())
+  {
+    throw std::runtime_error("There is no compiled module with identifier");
+  }
+  module_identifier_to_memory_manager.erase(module_memory_manager_it);
+  compiled_code_size.fetch_sub(module.size, std::memory_order_relaxed);
```
dr-m: Which code or data structure owns the memory that `module` referes to here? Could the memory actually be freed in the `erase` call? If so, then these statements should be swapped with each other to make the code safe.

jit/jit.h
```diff
@@ -0,0 +1,46 @@
...
+  llvm::LLVMContext context;
+  std::unique_ptr<llvm::TargetMachine> target_machine;
+  llvm::DataLayout data_layout;
+  std::unique_ptr<JITCompiler> compiler;
+  std::unique_ptr<JITSymbolResolver> symbol_resolver;
+  std::unordered_map<uint64_t, std::unique_ptr<JITModuleMemoryManager>> module_identifier_to_memory_manager;
+  uint64_t current_module_key = 0;
+  std::atomic<size_t> compiled_code_size;
+  // TODO lock when compiling a module
+  mutable std::mutex jit_lock;
```
dr-m: For any mutex or any potentially concurrently accessed data structure, it would be useful to document the protection rules. Many of these data members are actually protected by `jit_lock` in `deleteCompiledModule()`.

jit/jit.h
```diff
@@ -0,0 +1,46 @@
...
+  std::string getMangledName(const std::string& name_to_mangle) const;
+  void runOptimizationPassesOnModule(llvm::Module& module) const;
+  llvm::LLVMContext context;
+  std::unique_ptr<llvm::TargetMachine> target_machine;
+  llvm::DataLayout data_layout;
+  std::unique_ptr<JITCompiler> compiler;
+  std::unique_ptr<JITSymbolResolver> symbol_resolver;
+  std::unordered_map<uint64_t, std::unique_ptr<JITModuleMemoryManager>> module_identifier_to_memory_manager;
+  uint64_t current_module_key = 0;
+  std::atomic<size_t> compiled_code_size;
```
dr-m: I’d like to see comment for  documenting the lifetime and purpose of each data member. For `compiled_code_size` it would be useful to know what (if any) kind of padding or alignment overhead it would include.

Due to the lack of comments, it took me some time to find a counterpart of `deleteCompiledModule()` that would increment `compiled_code_size` and add something to `module_identifier_to_memory_manager`. It is `compileModule(std::unique_ptr<llvm::Module>)`. It seems to lack proper `jit_lock` protection.

jit/jit.cpp
```diff
@@ -0,0 +1,467 @@
...
+#include <boost/noncopyable.hpp>
+namespace msc_jit
+{
+// Arena Memory pool for JIT
+class Arena : private boost::noncopyable
+{
+ public:
+  Arena() : page_size(8)
+  {
+  }
```
dr-m: We could live with an implicit default constructor if the initializer expression for `page_size` was revised to this value.

jit/jit.cpp
```diff
@@ -0,0 +1,467 @@
...
+    return allocated_size;
+  }
+
+  inline size_t getPageSize() const
+  {
+    return page_size;
+  }
+
+ private:
+  size_t page_size = 0;
```
dr-m: I do not see any modification of this `page_size` in the code. I think that it could be declared simply as
```c++
  const size_t page_size = 8;
```
or even `static const` or `static constexpr` to save some memory, and then we could avoid explicitly declaring `Arena()`. I think that it is rather pointless to declare `const` data members `private` and to define public accessors for them.

What is the unit here? Bytes? Kilobytes? Some comments would be helpful.

Typically, `sysconf(PAGESIZE)` should return 4096 or a larger value.

utils/funcexp/functor.h
```diff
@@ -82,6 +84,14 @@ class Func
     throw logging::IDBExcept(oss.str(), logging::ERR_DATATYPE_NOT_SUPPORT);
   }
 
+  bool isCompilable(const execplan::CalpontSystemCatalog::ColType& colType)
+  {
+    return false;
+  }
+  llvm::Value* compile()
+  {
+  }
```
dr-m: It looks like the actual interesting part is still missing. The body of `compile()` seems to be missing `return nullptr;`.

jit/compilefunction.cpp
```diff
@@ -0,0 +1,23 @@
...
+#include <llvm/IR/Function.h>
+#include <llvm/IR/IRBuilder.h>
+
+namespace funcexp
+{
+void compileFunction(llvm::Module& module, const Func& function)
+{
+  llvm::IRBuilder<> b(module.getContext());
+  function->
+}
```
dr-m: The member function body does not look like valid C++ to me.

utils/funcexp/functor.h
```diff
@@ -38,6 +38,8 @@
 
 #include "dataconvert.h"
 
+#include <llvm/IR/IRBuilder.h>
```
dr-m, line 42: This is a rather large header file that will `#include` many other header files. I think that including that kind of a "jungle" in a rather popular header file (`functor.h` is being included from many other header files, which in turn could be included from many compilation units) could significantly include the compilation time.

We only seem to need this for the type name `llvm::Value`, which `compile()` would return a pointer to. I think that it would be better to define an opaque type and avoid including the header here. Something like this might work:
```c++
namespace llvm { class Value; }
```

dr-m (commented):
I understand that this is work in progress. Hopefully these comments are useful. It looks like we so far only have some interfaces but not any actual compilation, which would have been the main focus of my review.

I tried to compile this in the submodule `storage/columnstore/columnstore` under MariaDB/server@313c5a1dfb744aaef10586526dda89d2b4a50651. I got the build to complete, but the directory `storage/columnstore/columnstore/jit/CMakeFiles/jit.dir` in my `cmake -G Ninja` based build directory is empty. Other ColumnStore compilation units were being built.

jit/compileoperator.cpp
```diff
@@ -0,0 +1,92 @@
...
+      return_type = b.getInt64Ty(); break;
+    case execplan::CalpontSystemCatalog::DOUBLE:
+    case execplan::CalpontSystemCatalog::UDOUBLE: return_type = b.getDoubleTy(); break;
+    case execplan::CalpontSystemCatalog::FLOAT:
+    case execplan::CalpontSystemCatalog::UFLOAT: return_type = b.getFloatTy(); break;
+    default: throw logic_error("compileOperator: unsupported type");
+  }
+  auto* func_type = llvm::FunctionType::get(return_type, {data_type, isNull_type}, false);
+  auto* func =
+      llvm::Function::Create(func_type, llvm::Function::ExternalLinkage, expression->alias(), module);
```
dr-m: I would suggest
```c++
func->setDoesNotThrow();
```
so that no `.eh_frame` or `.rela.eh_frame` section will be generated.

jit/compileoperator.h
```diff
@@ -0,0 +1,24 @@
...
+template <typename t>
+using JITCompiledOperator = t (*)(uint8_t* data, bool& isNull);
+
+struct CompiledOperator
+{
+  msc_jit::JIT::CompiledModule compiled_module;
+  JITCompiledOperator<int64_t> compiled_function_int64;
+  JITCompiledOperator<uint64_t> compiled_function_uint64;
+  JITCompiledOperator<double> compiled_function_double;
+  JITCompiledOperator<float> compiled_function_float;
```
dr-m: Can multiple `compiled_function_` be set at a time?

jit/expressionjit.h
```diff
@@ -0,0 +1,105 @@
...
+
+ public:
+  int64_t getIntVal(rowgroup::Row& row, bool& isNull) override
+  {
+    return compiledOperator.compiled_function_int64(row.getData(), isNull);
+  }
+  uint64_t getUintVal(rowgroup::Row& row, bool& isNull) override
+  {
+    return compiledOperator.compiled_function_int64(row.getData(), isNull);
+  }
```
dr-m: `compiled_function_uint64` seems to be unused. Do we really need this distinction? The ABI for signed and unsigned should be the same.

I would suggest to add some tests and to demonstrate that all this code is covered by the test cases.

jit/expressionjit.h
```diff
@@ -0,0 +1,105 @@
...
+  double getDoubleVal(rowgroup::Row& row, bool& isNull) override
+  {
+    return compiledOperator.compiled_function_double(row.getData(), isNull);
+  }
+  float getFloatVal(rowgroup::Row& row, bool& isNull) override
+  {
+    return compiledOperator.compiled_function_float(row.getData(), isNull);
+  }
```
dr-m: Could we simply have one `compiled_function` member that will be `reinterpret_cast` to the appropriate type? That could avoid some run-time code duplication.

CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ endif()
 INCLUDE(ExternalProject)
 INCLUDE(CheckCXXSourceCompiles)
 
-SET(CMAKE_CXX_STANDARD 11)
+SET(CMAKE_CXX_STANDARD 14)
```
drrtuy: We use c++20 if compiler handles it [via compiler flags](https://github.com/mariadb-corporation/mariadb-columnstore-engine/blob/develop/CMakeLists.txt#L53C1-L53C1). Is this set is necessary?

CMakeLists.txt
```diff
@@ -358,6 +372,7 @@ ADD_SUBDIRECTORY(writeengine/splitter)
 ADD_SUBDIRECTORY(storage-manager)
 ADD_SUBDIRECTORY(datatypes)
 ADD_SUBDIRECTORY(tests)
+ADD_SUBDIRECTORY(jit)
```
drrtuy: This directory must be moved into ./utils and the paths changed accordingly.

jit/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+        jit.cpp
+        jit.h
+        compileoperator.cpp
+        compileoperator.h
+        expressionjit.h)
+
+
+add_library(jit SHARED ${jit_LIB_SRCS})
+
+target_link_libraries(jit ${NETSNMP_LIBRARIES} ${LLVM_LIBS} ${LLVM_SYSTEM_LIBS} ${LLVM_LDFLAGS})
```
drrtuy: You can drop ${NETSNMP_LIBRARIES} as this is an obsolete prerequisite.

jit/compileoperator.cpp
```diff
@@ -0,0 +1,105 @@
...
+      llvm::FunctionType::get(b.getInt64Ty(), {b.getInt64Ty(), b.getInt64Ty()}, false);
+  llvm::Function::Create(func_timestampValueToInt_type, llvm::Function::ExternalLinkage,
+                         "dataconvert::DataConvert::timestampValueToInt", module);
+}
+
+void compileOperator(llvm::Module& module, const execplan::SRCP& expression, rowgroup::Row& row)
+{
+  auto columns = expression.get()->simpleColumnList();
+  llvm::IRBuilder<> b(module.getContext());
+  compileExternalFunction(module, b);
```
drrtuy: How many times compileExternalFunction will be actually called?

jit/compileoperator.cpp
```diff
@@ -0,0 +1,105 @@
...
+  llvm::Value* data_ptr = args++;
+  llvm::Value* isNull_ptr = args++;
+  auto* entry = llvm::BasicBlock::Create(b.getContext(), "entry", func);
+  b.SetInsertPoint(entry);
+  auto* ret = expression->compile(b, data_ptr, isNull_ptr, row, expression->resultType().colDataType);
+
+  b.CreateRet(ret);
+
+  // Print the IR
+  llvm::verifyFunction(*func);
```
drrtuy: I suppose these two lines will go away.

jit/compileoperator.h
noorall: No, only one pointer will be set at a time.

dbcon/execplan/arithmeticoperator.h
```diff
@@ -210,6 +210,14 @@ class ArithmeticOperator : public Operator
   inline void execute(IDB_Decimal& result, IDB_Decimal op1, IDB_Decimal op2, bool& isNull);
   long fTimeZone;
   bool fDecimalOverflowCheck;
+
+ public:
+  inline llvm::Value* compile(llvm::IRBuilder<>& b, llvm::Value* data, llvm::Value* isNull,
+                              rowgroup::Row& row, ParseTree* lop, ParseTree* rop,
+                              CalpontSystemCatalog::ColDataType dataType) override;
+  inline llvm::Value* compileI(llvm::IRBuilder<>& b, llvm::Value* l, llvm::Value* r);
```
drrtuy:
```suggestion
  inline llvm::Value* compileInt_(llvm::IRBuilder<>& b, llvm::Value* l, llvm::Value* r);
```

dbcon/execplan/arithmeticoperator.h
```diff
@@ -210,6 +210,14 @@ class ArithmeticOperator : public Operator
...
+  inline llvm::Value* compileF(llvm::IRBuilder<>& b, llvm::Value* l, llvm::Value* r);
```
drrtuy:
```suggestion
  inline llvm::Value* compileFloat_(llvm::IRBuilder<>& b, llvm::Value* l, llvm::Value* r);
```

utils/rowgroup/rowgroup.cpp
```diff
@@ -1043,6 +1043,103 @@ const CHARSET_INFO* Row::getCharset(uint32_t col) const
   return charsets[col];
 }
 
+void Row::compileIsNull(llvm::IRBuilder<>& b, llvm::Value* dataValue, llvm::Value* isNull, uint32_t colIndex)
```
drrtuy: I am curious why do you think this functionality belongs here and not to jit directory?

noorall: The initial idea was to maintain consistency in function call behavior. It seems better to add a tool class in the jit directory to make the JIT code more centralized.

CMakeLists.txt
noorall: It is necessary. If it is not specified as 14 during compilation, an error will be reported.

CMakeLists.txt
noorall: Done.

jit/CMakeLists.txt
noorall: Done.

jit/compileoperator.cpp
noorall: Done.

jit/compileoperator.cpp
noorall: Done.

utils/rowgroup/rowgroup.cpp
noorall: > I am curious why do you think this functionality belongs here and not to jit directory?

I have moved it to a compilehelper class.

mariadb-LeonidFedorov:
work moved to https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/3165

PR #2893, opened 2023-07-06
Update 'configureEngine.cmake' to not write to source dir

tgross35:
It is currently not possible to do a full MariaDB build without touching the source tree, i.e. such that the the source tree could exist on a read only filesystem. This columnstore build file currently creates temporary files in the source tree to evaluate how the operating system handles them, so it is one blocker to these isolated builds.

This patch adjusts these tests to create the temporary files in the output directory, rather than in the source tree.

(I have tested this on linux but have not been able to test on Windows)

### DIFF
--- a/cmake/configureEngine.cmake
+++ b/cmake/configureEngine.cmake
@@ -314,9 +314,9 @@ int
 main ()
 {
 struct stat s, t;
-  return ! (stat (\"conftest.data\", &s) == 0
-        && utime (\"conftest.data\", 0) == 0
-        && stat (\"conftest.data\", &t) == 0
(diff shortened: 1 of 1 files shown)


drrtuy:
Thx for the suggestion. I don't think that Windows builds are affected b/c one can't build MCS on Windows. Activating this engine build would fail anyway.

mariadb-LeonidFedorov (approved):

mariadb-LeonidFedorov:
@tgross35 thank you for the PR!

tgross35:
Thanks for the review!

PR #2900, opened 2023-07-13
[WIP] MCOL-4740 UPDATE involving multi-tables returns wrong "Rows matched".

tntnatbry:
For UPDATEs involving a single table, the server call to handler::direct_update_rows() is used to correctly set the count for the number of updated rows in the UPDATE statement.

However, for UPDATEs involving multi-tables, the server uses the select_handler execution which does not support setting the count for the number of updates rows in ColumnStore. This patch adds support for correctly reporting the number of updated rows to the client by setting multi_update::updated and multi_update::found in ha_columnstore_select_handler::end_scan().

### DIFF
--- a/dbcon/mysql/ha_mcs_pushdown.cpp
+++ b/dbcon/mysql/ha_mcs_pushdown.cpp
@@ -1169,6 +1169,25 @@ int ha_columnstore_select_handler::end_scan()
 
   scan_ended = true;
 
+  // MCOL-4740 select_handler::send_eof() is called after
+  // select_handler::end_scan() in select_handler::execute().
+  // So we set multi_update::updated and multi_update::found here
+  // which are used in send_eof() to output the affected number
+  // of rows to the client.
+  if (current_thd->lex->sql_command == SQLCOM_UPDATE_MULTI)
+  {
+    multi_update* multi = (multi_update*) result;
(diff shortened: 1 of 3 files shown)


tntnatbry:
Please note that the compile against the server is expected to fail. This patch expects the following members of `multi_update` to have public visibility:
`multi_update::update_tables`
`multi_update::table_to_update`
`multi_update::updated`
`multi_update::found`

Here is the draft server side patch:
```c++
diff --git a/sql/sql_class.h b/sql/sql_class.h
index c472009a480..be79ec7c130 100644
--- a/sql/sql_class.h
+++ b/sql/sql_class.h
@@ -7095,10 +7095,15 @@ class multi_update :public select_result_interceptor
   TABLE_LIST *all_tables; /* query/update command tables */
   List<TABLE_LIST> *leaves;     /* list of leaves of join table tree */
   List<TABLE_LIST> updated_leaves;  /* list of of updated leaves */
+ public:
   TABLE_LIST *update_tables;
-  TABLE **tmp_tables, *main_table, *table_to_update;
+  TABLE *table_to_update;
+ private:
+  TABLE **tmp_tables, *main_table;
   TMP_TABLE_PARAM *tmp_table_param;
+ public:
   ha_rows updated, found;
+ private:
   List <Item> *fields, *values;
   List <Item> **fields_for_table, **values_for_table;
   uint table_count;
```

PR #2905, opened 2023-07-17
[WIP] MCOL-4740 UPDATE involving multi-tables returns wrong "Rows matched".

tntnatbry:
For UPDATEs involving a single table, the server call to handler::direct_update_rows() is used to correctly set the count for the number of updated rows in the UPDATE statement.

However, for UPDATEs involving multi-tables, the server does not call handler::direct_update_rows(). This patch adds support to correctly report the number of updated rows to the client by setting multi_update::updated and multi_update::found in handler::rnd_end().

### DIFF
--- a/dbcon/mysql/ha_mcs.cpp
+++ b/dbcon/mysql/ha_mcs.cpp
@@ -618,6 +618,25 @@ int ha_mcs::rnd_end()
 {
   DBUG_ENTER("ha_mcs::rnd_end");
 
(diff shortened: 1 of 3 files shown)


tntnatbry:
Please do not merge this PR until the changes required in MDEV-31725 are merged in the server tree. Here is the MDEV-31725 PR: https://github.com/MariaDB/server/pull/2697

PR #2909, opened 2023-07-21
MCOL-5522 Properly process pm join result count.

denis0x0D:
This patch:
1. Properly processes situation when pm join result count is exceeded.
2. Adds session variable 'columnstore_max_pm_join_result_count` to control the limit.


drrtuy (approved):

drrtuy:
Could you add an explanation about the behavior that needs to track the last small side row number?

mariadb-LeonidFedorov (approved):

PR #2912, opened 2023-07-28
MCOL-5005 Add charset number to system catalog.

tntnatbry:


tntnatbry:
Please don't merge yet.

tntnatbry:
The PR is now ready for review. When merging, please do not squash, but keep both the commits.

mariadb-LeonidFedorov (approved):

tntnatbry:
@drrtuy Do you have any review feedback for this patch? If not, then I will proceed with the merge.

dbcon/ddlpackage/ddl.y
```diff
@@ -56,24 +56,42 @@ int ddllex(YYSTYPE* ddllval, void* yyscanner);
...
+void fix_column_length(SchemaObject* elem, const CHARSET_INFO* def_cs)
+{
     auto* column = dynamic_cast<ColumnDef*>(elem);
+
     if (column == NULL || column->fType == NULL)
     {
         return;
     }
 
+    if (column->fType->fType == DDL_BLOB ||
```
drrtuy, line 68: The name of the function is misguiding. fix_column_length() starts to assign charset related stuff out of a sudden.

drrtuy: I suggest to rename the function or put the functionality into a separate function.

dbcon/execplan/calpontsystemcatalog.cpp
```diff
@@ -6223,6 +6223,14 @@ void CalpontSystemCatalog::buildSysColinfomap()
   aCol.colPosition++;
   aCol.columnOID = OID_SYSCOLUMN_NEXTVALUE;
   fColinfomap[aCol.columnOID] = aCol;
+
+  aCol.colWidth = 4;
```
drrtuy: IMHO this is an example how the previously set attributes of a shared class instance can be reused accidentally. Despite this is a pattern I suggest to create another ColType instance inside a scope.

dbcon/execplan/calpontsystemcatalog.cpp
```diff
@@ -6223,6 +6223,14 @@ void CalpontSystemCatalog::buildSysColinfomap()
...
+  aCol.constraintType = NOTNULL_CONSTRAINT;
```
drrtuy:
```suggestion
  
     fColinfomap[OID_SYSCOLUMN_CHARSETNUM] = ColType(4, INT, ddn, colPosition,OID_SYSCOLUMN_CHARSETNUM);

drrtuy: Doesn't re-use ColType in an error-prone way and saves space.

drrtuy (commented):
There are some aspects that IMHO can be done in a different way.

dbcon/ddlpackage/ddl.y
tntnatbry: Good point. I have renamed the function.

dbcon/execplan/calpontsystemcatalog.cpp
tntnatbry: I am not fully convinced about this. I don't see why we should create a new ColType ctor just to address a special pattern that only exists in this function.

drrtuy: Consider the case when someone forgets to set colWidth. This would cause an incorrect colWidth to be set. When you use a ctor one is forced to ship all values needed.

tntnatbry: I have added a new ctor for ColType to address this use case.

tntnatbry:
Changes requested have been incorporated, please review.

drrtuy (approved):
untested ack

PR #2920, opened 2023-08-04
Mcol 678 support WITH ROLLUP

mariadb-SergeyZefirov:
The logic is to add as many RowStorage's as there are GROUP BY columns. The row gets added to RowStorages[0] as it is and then it gets added to other RowStorages with NULLs progressively added.

Let's say we have GROUP BY with two columns. The first row with key (1,2) will be added (grouped) to RowStorages]0], the row with (1,NULL) will be added (grouped) to RowStorages[1] and the row (NULL, NULL) will be added to RowStorage[2].

The row groups from these storages are output in order of the RowStorages indices. Thus, first we have our regular GROUP BY results, then subtotal results for one column coerced to NULL, then two columns, etc.


mariadb-SergeyZefirov:
As it is a draft stage, "get thing to work," all GROUP BY clauses work as they have WITH ROLLUP added.

After I'll get the thing to work correctly, I'll pass necessary information to enable this functionality selectively. I expect a flag argument in some construuctors.

PR #2928, opened 2023-08-11
Shorter returns

mariadb-AndreyPiskunov:
This request contains 2 types of changes:
1) changing the pattern of 
```
if (condition)
  return true;
return false;
```
and similar to it. I went all out on that, which means that there might be places where it became less readable. However, I think that this way it is consistent with both codestyle in the repo and general guidelines. Also, there are definitely places in the codebase where the returned boolean expressions are enormous. Maybe those should be reduced, but that is a different task.

2) Changing the `while (1)` conditions to `while (true)`
Not the best pattern anyway, but it seems that in most places changing it to a meaningful condition would take serious refactoring effort. The way I put it it is at least a better codestyle + it is more consistent now across codebase.

### DIFF
--- a/dbcon/ddlpackageproc/altertableprocessor.cpp
+++ b/dbcon/ddlpackageproc/altertableprocessor.cpp
@@ -728,7 +728,7 @@ void AlterTableProcessor::addColumn(uint32_t sessionID, execplan::CalpontSystemC
       {
         fWEClient->write(bs, (uint32_t)pmNum);
 
-        while (1)
+        while (true)
         {
           bsIn.reset(new ByteStream());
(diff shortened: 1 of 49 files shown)


phoeinx:
Hi,
More of a question here than a comment. Nice work on code style changes! (I feel you're right that regarding pattern 1 some expressions become a bit less readable.) Just wanted to ask which general guidelines you are referring to. Is there a CodeStyle guideline for our repo or are you working with the coding standards from MariaDB server?

mariadb-LeonidFedorov:
@mariadb-AndreyPiskunov please rebase the PR and mark it ready, if you want it to be merged

PR #2930, opened 2023-08-15
Logging of memory 

mariadb-AlexeyVorovich:
this is incomplete.  I am doing this to have drone create the package so that I can test docker

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -181,4 +181,9 @@ build/.cmake
 *.vtg
 *.vtg-back
 '*.vtg-Stashed changes'
+versioning/BRM/brmshmimpl.h
+versioning/BRM/shmkeys.cpp
+obj/
+build/build
+
 
--- a/.vscode/gdb.config
+++ b/.vscode/gdb.config
@@ -1,3 +1,0 @@
-define target hookpost-remote
-set scheduler-locking step
-show scheduler-locking
--- a/.vscode/gdb.config
+++ b/.vscode/gdb.config
@@ -1,0 +1,3 @@
+define target hookpost-remote
+set scheduler-locking step
+show scheduler-locking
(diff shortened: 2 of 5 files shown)


mariadb-AlexeyVorovich:
@mariadb-LeonidFedorov  @drrtuy  I am trying to improve logging for total and available memory for both cgroup and non-cgroup. Question :
getFreeMemory is called on timer . The current version does inconsistent logging 
-for cgroup (I see that in container) each call  is logged
-for non-cgroup (i see in VM) no calls are logged
I added logging to both places in this PR . However this creates unreadable huge logs. What do you want to do ? remove logging from both cases ? make it depend on setting ? other suggestions ?

mariadb-AlexeyVorovich:
printing msg on every invocation clearly makes the primproc log unusable.  So it cannot stay.

I did notice that in pure develop (without my PR)  i get 1k messages
Aug 23 15:41:43 av2 env[169323]: [140B blob data]
Aug 23 15:41:43 av2 env[169323]: [140B blob data]

mariadb-AlexeyVorovich:
log each 1000 call per LF. effectively 15min

drrtuy:
The size of a log was a reason why I removed logging from getFreeMemory in the first place.
IMHO one-shot logging that is now presented is useful. One can see the amount of RAM on startup. You can prettify it printing outGBs instead of byte units.

mariadb-AlexeyVorovich:
- The GB printout at start is in PR 
- getfreememory was inconsistent  (with logging for cgroup and w/o logging for no cgroup). per LF , i made it  every 1000 calls or effectively 15 min at current polling rate.  Please decide if ongoing mem diagnostic is useful or should I comment it out 

Let me know 

@mariadb-LeonidFedorov  make the decision on above , pls

mariadb-LeonidFedorov (approved):

PR #2938, opened 2023-08-21
MCOL-4931 Make cpimport charset-aware.

tntnatbry:
1. Extend the following CalpontSystemCatalog member functions to set CalpontSystemCatalog::ColType::charsetNumber, after the system catalog update to add charset number to calpontsys.syscolumn in MCOL-5005: CalpontSystemCatalog::lookupOID CalpontSystemCatalog::colType CalpontSystemCatalog::columnRIDs CalpontSystemCatalog::getSchemaInfo

2. Update cpimport to use the CHARSET_INFO object associated with the charset number retrieved from the system catalog, for a dictionary/non-dictionary CHAR/VARCHAR/TEXT column, to truncate long strings that exceed the target column character length.

3. Add MTR test cases.

### DIFF
--- a/dbcon/execplan/calpontsystemcatalog.cpp
+++ b/dbcon/execplan/calpontsystemcatalog.cpp
@@ -553,8 +553,9 @@ CalpontSystemCatalog::OID CalpontSystemCatalog::lookupOID(const TableColName& ta
   string autoincrement = CALPONT_SCHEMA + "." + SYSCOLUMN_TABLE + "." + AUTOINC_COL;
   string nextVal = CALPONT_SCHEMA + "." + SYSCOLUMN_TABLE + "." + NEXTVALUE_COL;
   string nullable = CALPONT_SCHEMA + "." + SYSCOLUMN_TABLE + "." + NULLABLE_COL;
+  string charsetnum = CALPONT_SCHEMA + "." + SYSCOLUMN_TABLE + "." + CHARSETNUM_COL;
 
-  SimpleColumn* col[17];
+  SimpleColumn* col[18];
   col[0] = new SimpleColumn(columnlength, fSessionID);
   col[1] = new SimpleColumn(objectid, fSessionID);
   col[2] = new SimpleColumn(datatype, fSessionID);
@@ -572,6 +573,7 @@ CalpontSystemCatalog::OID CalpontSystemCatalog::lookupOID(const TableColName& ta
   col[14] = new SimpleColumn(autoincrement, fSessionID);
   col[15] = new SimpleColumn(nextVal, fSessionID);
   col[16] = new SimpleColumn(nullable, fSessionID);
+  col[17] = new SimpleColumn(charsetnum, fSessionID);
 
   SRCP srcp;
   srcp.reset(col[0]);
@@ -609,13 +611,15 @@ CalpontSystemCatalog::OID CalpontSystemCatalog::lookupOID(const TableColName& ta
   colMap.insert(CMVT_(nextVal, srcp));
   srcp.reset(col[16]);
   colMap.insert(CMVT_(nullable, srcp));
+  srcp.reset(col[17]);
+  colMap.insert(CMVT_(charsetnum, srcp));
   csep.columnMapNonStatic(colMap);
 
   // ignore returnedcolumn, because it's not read by Joblist for now
   csep.returnedCols(returnedColumnList);
-  OID oid[17];
+  OID oid[18];
 
-  for (int i = 0; i < 17; i++)
+  for (int i = 0; i < 18; i++)
     oid[i] = col[i]->oid();
 
   // Filters
@@ -709,6 +713,8 @@ CalpontSystemCatalog::OID CalpontSystemCatalog::lookupOID(const TableColName& ta
         ct.constraintType = NOTNULL_CONSTRAINT;
       }
     }
+    else if ((*it)->ColumnOID() == oid[17])
+      ct.charsetNumber = ((*it)->GetData(0));
     else if ((*it)->ColumnOID() == DICTOID_SYSCOLUMN_DEFAULTVAL)
     {
       ct.defaultValue = ((*it)->GetStringData(0));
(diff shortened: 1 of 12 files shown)


writeengine/dictionary/we_dctnry.cpp
```diff
@@ -837,12 +836,28 @@ int Dctnry::insertDctnry(const char* buf, ColPosPair** pos, const int totalRow,
...
-    // @Bug 2565: Truncate any strings longer than schema's column width
-    if (curSig.size > m_colWidth)
+    if (cs->mbmaxlen > 1)
     {
-      uint8_t truncate_point = utf8::utf8_truncate_point((const char*)curSig.signature, m_colWidth);
-      curSig.size = m_colWidth - truncate_point;
-      ++truncCount;
+      const char* start = (const char*) curSig.signature;
+      const char* end = (const char*)(curSig.signature + curSig.size);
+      size_t numChars = cs->numchars(start, end);
```
drrtuy, line 843: I need to double check this but the complexity of this func is `O(N)` where `N` is a length of a string being processed.

dbcon/execplan/calpontsystemcatalog.cpp
```diff
@@ -702,6 +706,8 @@ CalpontSystemCatalog::OID CalpontSystemCatalog::lookupOID(const TableColName& ta
     }
     else if ((*it)->ColumnOID() == oid[15])
       ct.nextvalue = ((*it)->GetData(0));
+    else if ((*it)->ColumnOID() == oid[17])
```
drrtuy: I am curious why does the if condition for the 17th element goes before the check for the 16th?

writeengine/bulk/we_bulkloadbuffer.cpp
```diff
@@ -515,14 +513,32 @@ void BulkLoadBuffer::convert(char* field, int fieldLength, bool nullFlag, unsign
...
-        if (fieldLength > column.definedWidth)
+        if (column.cs->mbmaxlen > 1)
         {
-          uint8_t truncate_point = utf8::utf8_truncate_point(field, column.definedWidth);
-          memcpy(charTmpBuf, field, column.definedWidth - truncate_point);
-          bufStats.satCount++;
+          const CHARSET_INFO* cs = column.cs;
+          const char* start = (const char*) field;
+          const char* end = (const char*)(field + fieldLength);
+          size_t numChars = cs->numchars(start, end);
```
drrtuy, line 521: This op might have `O(N)` complexity where `N` is a length of utf8 string. AFAIK numchars() will run full blown decoding that is complex 3 tables(in the worst-case) lookup. What I suggest is to set a length using maxCharLength and skip numChars calcultion. Surely if this is possible.

writeengine/bulk/we_bulkloadbuffer.cpp
```diff
@@ -515,14 +513,32 @@ void BulkLoadBuffer::convert(char* field, int fieldLength, bool nullFlag, unsign
...
+          size_t maxCharLength = column.definedWidth / cs->mbmaxlen;
+
+          if (numChars > maxCharLength)
+          {
+            MY_STRCOPY_STATUS status;
+            cs->well_formed_char_length(start, end, maxCharLength, &status);
```
drrtuy, line 527: Is there anything to read from status, e.g. errors to react on?

writeengine/dictionary/we_dctnrystore.h
```diff
@@ -135,22 +135,6 @@ class DctnryStore : public DbFileOp
...
-   * @brief Update dictionary store with tokenized strings (for Bulk use)
-   *
-   * @param buf       - bulk buffer containing strings to be parsed
-   * @param pos       - list of offsets into buf
-   * @param totalRow  - total number of rows in buf
-   * @param col       - the column to be parsed from buf
-   * @param colWidth  - width of the dictionary column being parsed
-   * @param tokenBuf  - (output) list of tokens for the parsed strings
-   */
-  const int updateDctnryStore(const char* buf, ColPosPair** pos, const int totalRow, const int col,
```
drrtuy, line 148: Did I get it right this method was not used by the code?

drrtuy (commented):
I have couple questions on this patch:
- why do you add charsetnum into multiple syscat query functions: lookupOID, colType, columnRID, getSchemaInfo? I suppose that cpimport uses only some of them to render syscat info?
- did you check how does pure cpimport now behaves if he sees a partial utf8 glyph, let's say there are two bytes out of 3 necessary?
- not directly related to this patch but important. I am missing what is the default for charsetnum and what is the upgrade plan for the syscat change?

writeengine/bulk/we_bulkloadbuffer.cpp
tntnatbry: Yes, `cs->numchars()` will be `O(N)` where N is the number of bytes in the input string.

For `utf8mb3` charset (as an example), here is the flow of execution for `cs->numchars()`:
```c++
my_numchars_mb() (defined in string/ctype-mb.c) -> my_ismbchar() (defined in include/m_ctype.h) -> my_valid_mbcharlen_utf8mb3() (defined in strings/ctype-utf8.c)
```
In this execution flow, I don't see a 3-table complex lookup anywhere.

I don't see how it is possible to determine if a string needs to be truncated or not without first finding it's character length.

writeengine/bulk/we_bulkloadbuffer.cpp
tntnatbry: `status` is used to set the `fieldLength` on line 528.

writeengine/dictionary/we_dctnry.cpp
tntnatbry: I addressed this comment earlier.

writeengine/dictionary/we_dctnrystore.h
tntnatbry: Yes, this was dead code.

dbcon/execplan/calpontsystemcatalog.cpp
tntnatbry: Fixed.

tntnatbry:
> I have couple questions on this patch:
> 
> * why do you add charsetnum into multiple syscat query functions: lookupOID, colType, columnRID, getSchemaInfo? I suppose that cpimport uses only some of them to render syscat info?
> * did you check how does pure cpimport now behaves if he sees a partial utf8 glyph, let's say there are two bytes out of 3 necessary?
> * not directly related to this patch but important. I am missing what is the default for charsetnum and what is the upgrade plan for the syscat change?

1. The addition of charsetnum to `lookupOID()`, `colType()`, `columnRIDs()`, `getSchemaInfo()` is a generic implementation. It is not just limited to the cpimport use case. These functions are also setting other `execplan::CalpontSystemCatalog::ColType` class attributes such as compressionType, autoincrement etc. This way, charsetnum will be correctly set and will be readily available as an attribute of the `ColType` class to other parts of the engine code which use these syscat functions.
2. cpimport will now correctly handle any broken input multi-byte strings by calling `CHARSET_INFO::well_formed_char_length()`.
3. The default for charset will be 0 in syscat for all old columns that are upgraded to the syscat version with charsetnum field in `calpontsys.syscolumn` table. Upgrades will be handled by dbbuilder. This is done in the previous PR: https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/2912

mariadb-LeonidFedorov (approved):

PR #2939, opened 2023-08-22
MCOL-5559 This patch extends the scope of the crit section to cover remap operation to avoid access races running EMIndex operations

drrtuy:
Similar to MCOL-5487. Remap op wasn't covered by an exclusive shmem lock.

### DIFF
--- a/versioning/BRM/extentmap.cpp
+++ b/versioning/BRM/extentmap.cpp
@@ -2080,20 +2080,54 @@ void ExtentMap::grabEMIndex(OPS op)
     }
     else
     {
-      // Sending down current Managed Shmem size. If EMIndexImpl instance size doesn't match
-      // fEMIndexShminfo->allocdSize makeExtentMapIndexImpl will remap managed shmem segment.
-      fPExtMapIndexImpl_ =
-          ExtentMapIndexImpl::makeExtentMapIndexImpl(getInitialEMIndexShmkey(), fEMIndexShminfo->allocdSize);
+      if (op == READ)
+      {
+        fMST.getTable_upgrade(MasterSegmentTable::EMIndex);
+        emIndexLocked = true;
 
-      if (r_only)
-        fPExtMapIndexImpl_->makeReadOnly();
+        // Sending down current Managed Shmem size. If EMIndexImpl instance size doesn't match
+        // fEMIndexShminfo->allocdSize makeExtentMapIndexImpl will remap managed shmem segment.
+        fPExtMapIndexImpl_ = ExtentMapIndexImpl::makeExtentMapIndexImpl(getInitialEMIndexShmkey(),
+                                                                        fEMIndexShminfo->allocdSize);
+
+        if (r_only)
+          fPExtMapIndexImpl_->makeReadOnly();
+
+        emIndexLocked = false;
+        fMST.getTable_downgrade(MasterSegmentTable::EMIndex);
+      }
+      else
+      {
+        fPExtMapIndexImpl_ = ExtentMapIndexImpl::makeExtentMapIndexImpl(getInitialEMIndexShmkey(),
+                                                                        fEMIndexShminfo->allocdSize);
+
+        if (r_only)
+          fPExtMapIndexImpl_->makeReadOnly();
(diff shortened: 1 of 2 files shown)


versioning/BRM/extentmap.cpp
```diff
@@ -2080,10 +2092,12 @@ void ExtentMap::grabEMIndex(OPS op)
...
+      const bool isReadOnly = false;
       // Sending down current Managed Shmem size. If EMIndexImpl instance size doesn't match
       // fEMIndexShminfo->allocdSize makeExtentMapIndexImpl will remap managed shmem segment.
       fPExtMapIndexImpl_ =
-          ExtentMapIndexImpl::makeExtentMapIndexImpl(getInitialEMIndexShmkey(), fEMIndexShminfo->allocdSize);
+          ExtentMapIndexImpl::makeExtentMapIndexImpl(getInitialEMIndexShmkey(), fEMIndexShminfo->allocdSize,
+                                                     isReadOnly, emIndexLocked, op == READ ? &fMST : nullptr);
 
       if (r_only)
         fPExtMapIndexImpl_->makeReadOnly();
```
tntnatbry: `makeReadOnly()` should be under WRITE lock.

versioning/BRM/extentmap.cpp
```diff
@@ -2092,8 +2106,10 @@ void ExtentMap::grabEMIndex(OPS op)
   else if (fPExtMapIndexImpl_->getShmemImplSize() != (unsigned)fEMIndexShminfo->allocdSize)
   {
     fPExtMapIndexImpl_->refreshShm();
```
tntnatbry: `refreshShm()` should be under WRITE lock.

tntnatbry (changes requested):

versioning/BRM/extentmap.cpp
drrtuy: JFYI you are mixing locks up. mutex lock that is taken for the scope of the function controls multithread races whilst shmem RWLock controls interprocess races.  fPExtMapIndexImpl_ is process-specific entity.

versioning/BRM/extentmap.cpp
drrtuy: Same as previously.

tntnatbry: The write op in `refreshShm()` is not synchronised. Here is the race scenario:
  Reader1: Execute `ExtentMap::getExtents()` -> `fPExtMapIndexImpl_->find()`
  Reader2: Execute `ExtentMap::getExtents()` -> `grabEMEntryTable(READ)` -> `fPExtMapIndexImpl_->refreshShm()`

versioning/BRM/extentmap.cpp
tntnatbry: `makeReadOnly()` is calling `BRMManagedShmImpl::remap()` under the hood, which is a write op to the shmem segment.

tntnatbry (changes requested):
It looks like similar race also exists in `ExtentMap::grabFreeList()`.

Also, I noticed there is no call to `ExtentMapIndexImpl::refreshShm()` in `DBRM::refreshShm()`.

drrtuy:
> It looks like similar race also exists in `ExtentMap::grabFreeList()`.
> 
> Also, I noticed there is no call to `ExtentMapIndexImpl::refreshShm()` in `DBRM::refreshShm()`.

The story is different for FreeList. It is based on a different kind of shmem wrapper, namely unmanaged shmem. JFYI there were no races before we switched EM and introduced EMIndex on top of managed shmem implementation. Thus I won't change it in 23.02 but I have the fact in mind doing the patch for 23.08.

versioning/BRM/extentmap.cpp
drrtuy: After some consideration I agree with the point.

versioning/BRM/extentmap.cpp
drrtuy: After some consideration I agree with the point.

tntnatbry (approved):

PR #2942, opened 2023-08-24
Mcol 5559 emindex shmem race dev 2

drrtuy:
This patch incorporates fixes for race that happens when a process remaps managed shared memory segments whilst another process reads from it. It is possible b/c some control path takes a a shared RWLock remaping shmem segment. Now the first process takes an exclusive lock remaping.


drrtuy:
Don't mind the comments and two commits at this point.

mariadb-DenisKhalikov (approved):

tntnatbry (approved):
@drrtuy Changes LGTM. Can you please fix the commit messages and the PR title.

PR #2944, opened 2023-08-26
Create Arrow Buffer

MuHe03:
Create arrow buffers for scanned and filtered data for a BPP.


mariadb-LeonidFedorov:
@MuHe03 are any works to continue here?

PR #2945, opened 2023-08-28
MCOL-5539 Put table on small side if it was involved in prev.join.

denis0x0D:


mariadb-SergeyZefirov (approved):
It does not disagree with me.

PR #2946, opened 2023-08-30
Rename a limit var + change error message

mariadb-AndreyPiskunov:


mariadb-LeonidFedorov:
You should change corresponding MTR test according new error message

mariadb-LeonidFedorov (approved):

PR #2951, opened 2023-09-07
MCOL-678 support GROUP BY WITH ROLLUP

mariadb-SergeyZefirov:
A code to support WITH ROLLUP modifier of the GROUP BY clause.


mariadb-SergeyZefirov:
I need to add a test for a rare corner case I've found.

utils/libmarias3/libmarias3
drrtuy, line 1: This change doesn't look appropriate.

utils/rowgroup/rowaggregation.cpp
```diff
@@ -812,49 +815,61 @@ void RowAggregationUM::aggReset()
...
-  updateEntry(row, rgContextColl);
+    updateEntry(row, rgContextColl);
+    // these quantities are unsigned and comparing z and cnt - 1 can be incorrect
+    // because cnt can be zero.
+    if ((z + 1 < cnt)) {
+      // if we are rolling up, we mark appropriate field as NULL and also increment
+      // value in the "mark" column, so that we can differentiate between data and
+      // various rollups.
+      row.setIntField(row.getIntField(cnt - 1) + 1, cnt - 1);
+      row.setToNull(cnt - 2 - z);
```
drrtuy, line 876: How do we know that cnt -2 - z doesn't force uint32 wraparound?

utils/rowgroup/rowaggregation.h
```diff
@@ -630,6 +632,7 @@ class RowAggregation : public messageqcpp::Serializeable
   joblist::ResourceManager* fRm = nullptr;
   boost::shared_ptr<int64_t> fSessionMemLimit;
   std::unique_ptr<RGData> fCurRGData;
+  bool fRollupFlag = true;
```
drrtuy: Could you elaborate why do you set this to true by default?

mysql-test/columnstore/basic/r/mcs84_rollup.result
```diff
@@ -16,9 +16,35 @@ year	SUM(sales)
 SELECT year, SUM(sales) FROM booksales GROUP BY year WITH ROLLUP ORDER BY year;
 ERROR HY000: Incorrect usage of CUBE/ROLLUP and ORDER BY
 SELECT year, SUM(sales) FROM booksales GROUP BY year WITH ROLLUP;
```
drrtuy, line 45: I suggest to extend the test case with at least wide decimal and dictionary-based string types.

mysql-test/columnstore/basic/r/mcs84_rollup.result
```diff
@@ -16,9 +16,35 @@ year	SUM(sales)
...
+SELECT key1, key2, SUM(value), AVG(value), MIN(value), MAX(value), COUNT(value), COUNT(*) FROM three_cols GROUP BY key1, key2 WITH ROLLUP;
+key1	key2	SUM(value)	AVG(value)	MIN(value)	MAX(value)	COUNT(value)	COUNT(*)
+1	1	3	3.0000	3	3	1	1
+1	2	9	4.5000	4	5	2	2
+1	NULL	12	4.0000	3	5	3	3
+2	3	13	6.5000	6	7	2	2
+2	NULL	13	6.5000	6	7	2	2
+NULL	1	2	2.0000	2	2	1	1
+NULL	NULL	1	1.0000	1	1	1	2
+NULL	NULL	28	4.0000	1	7	7	8
```
drrtuy, line 78: I am curious if this row has a proper avg given there are 8 values(one is NULL though) and their sum is 28. I expect this avg to be a decimal number.

utils/rowgroup/rowaggregation.cpp
mariadb-SergeyZefirov: I've added a comment.

We have a condition z + 1 < cnt, the biggest value of z that satisfy it is cnt-2. Thus, the least value of cnt - 2 - z is zero.

utils/rowgroup/rowaggregation.h
mariadb-SergeyZefirov: I should not, of course. An artifact of the development process.

drrtuy (approved):

PR #2956, opened 2023-09-12
MCOL-5572: set syscat back to latin1

mariadb-LeonidFedorov:


tntnatbry:
Please don't merge yet.

mariadb-AlexeyVorovich:
Gagan's concern is that this may break existing customers.
He will try another approach today

my interpretation  is that the change should not affect existing  catalog due to ` not exists` clause 
`create table if not exists syscolumn`

PR #2958, opened 2023-09-19
MCOL-5555 Reduce number of direct writes to the file

denis0x0D:


versioning/BRM/brmtypes.cpp
```diff
@@ -447,6 +447,46 @@ void TableLockInfo::deserialize(istream& i)
     i.read((char*)&dbrootList[j], 4);
 }
 
+uint32_t TableLockInfo::getInternalSize() const
+{
+  const uint32_t nameLen = ownerName.length();
+  const uint32_t dbrootListSize = dbrootList.size();
+  return (32 + sizeof(time_t) + nameLen + (dbrootListSize * 4));
```
drrtuy: Could you turn 32 into a constant dn replace 4 with sizeof(uint32_t)?

versioning/BRM/brmtypes.cpp
```diff
@@ -447,6 +447,46 @@ void TableLockInfo::deserialize(istream& i)
...
+}
+
+void TableLockInfo::serialize(std::unique_ptr<char[]>& buffer, uint32_t& offset) const
+{
+  const uint32_t nameLen = ownerName.length();
+  const uint32_t dbrootListSize = dbrootList.size();
+
+  std::memcpy(&buffer[offset], (char*)&id, 8);
```
drrtuy: I would add a variadic template to send down all the args at once but I am fine if you define an inline function `char* serializeElement(char* dst,  const char* src, size_t size)` that returns a new destination and apply it instead of calling memcpy.

versioning/BRM/brmtypes.cpp
```diff
@@ -447,6 +447,46 @@ void TableLockInfo::deserialize(istream& i)
...
+uint32_t TableLockInfo::getInternalSize() const
+{
+  const uint32_t nameLen = ownerName.length();
+  const uint32_t dbrootListSize = dbrootList.size();
+  return (32 + sizeof(time_t) + nameLen + (dbrootListSize * 4));
+}
+
+void TableLockInfo::serialize(std::unique_ptr<char[]>& buffer, uint32_t& offset) const
+{
+  const uint32_t nameLen = ownerName.length();
```
drrtuy, line 465: It is worth to add an assertion the buffer size is big enough to store a lock info.

versioning/BRM/brmtypes.cpp
```diff
@@ -447,6 +447,46 @@ void TableLockInfo::deserialize(istream& i)
...
+  const uint32_t dbrootListSize = dbrootList.size();
+
+  std::memcpy(&buffer[offset], (char*)&id, 8);
+  offset += 8;
+  std::memcpy(&buffer[offset], (char*)&tableOID, 4);
```
drrtuy:
```suggestion
  std::memcpy(&buffer[offset], (char*)&tableOID, sizeof(decltype(tableOID));
```

drrtuy (approved):

drrtuy (approved):

PR #2961, opened 2023-09-19
Cgroup log fix

mariadb-AlexeyVorovich:


build/bootstrap_mcs.sh
```diff
@@ -529,6 +533,8 @@ generate_svgs()
     fi
 }
 
+disable_git_restore_frozen_revision
```
drrtuy: This might be unwanted behavior to be honest. @mariadb-LeonidFedorov  what do you think?

utils/common/cgroupconfigurator.cpp
```diff
@@ -210,7 +201,19 @@ uint64_t CGroupConfigurator::getTotalMemory()
...
+  string cGroupVersion_str="";
+  switch(cGroupVersion_){
+    case v1:
+      cGroupVersion_str="v1";
+      break;
+    case v2:
+      cGroupVersion_str="v2";
+      break;
+  }
+  cout << __func__ << " cGroupVersion_str " <<  cGroupVersion_str  << endl;
```
drrtuy:
```suggestion
  cout << __func__ << " cGroupVersion_str " <<  ((cGroupVersion_ == v2) ? std::string("v2") : std::string("v1")  << endl;
```

tests/rewritetest.cpp
```diff
@@ -593,7 +593,7 @@ TEST_P(ParseTreeComparatorTest, CompareContains)
   auto filter = std::make_unique<execplan::ParseTree>(new execplan::SimpleFilter(
       GetParam().filter, execplan::SimpleFilter::ForTestPurposesWithoutColumnsOIDS{}));
 
-  ASSERT_EQ(GetParam().contains, container.contains(filter));
```
drrtuy: What is this and how it is related to logging?

mariadb-AlexeyVorovich: oh , boy.. This branch was based on the source tree   that was fixed with @mariadb-LeonidFedorov help as the the compilation of develop branch in bootstrapMCS failed.   My bad for including [tests/rewritetest.cpp]

mariadb-LeonidFedorov (approved):

drrtuy (changes requested):
I am concerned by the fact getTotalMemory is called in ResourceManager ctor that is used universally. I suggest to run MTR or test001 and see if this patch prints out extra lines in both journalctl and debug.log.
If it doesn't I will ack this patch.

mariadb-AlexeyVorovich:
wherever the old code printed one line , we will now print 2 (or 3 lines if the cgroup is present)

drrtuy (approved):

PR #2962, opened 2023-09-19
MCOL-5573 Fix cpimport truncation of TEXT columns.

tntnatbry:
1. Restore the utf8_truncate_point() function in utils/common/utils_utf8.h that I removed as part of the patch for MCOL-4931.

2. As per the definition of TEXT columns, the default column width represents the maximum number of bytes that can be stored in the TEXT column. So the effective maximum length is less if the value contains multi-byte characters. However, if the user explicitly specifies the length of the TEXT column in a table DDL, such as TEXT(65535), then the DDL logic ensures that enough number of bytes are allocated (upto a system maximum) to allow upto that many number of characters (multi-byte characters if the charset for the column is multi-byte, such as utf8mb3).

### DIFF
--- a/utils/common/utils_utf8.h
+++ b/utils/common/utils_utf8.h
@@ -91,6 +91,33 @@ inline std::string wstring_to_utf8(const std::wstring& str)
   return ret;
 }
 
+inline uint8_t utf8_truncate_point(const char* input, size_t length)
+{
+  // Find the beginning of a multibyte char to truncate at and return the
+  // number of bytes to truncate1`
+  if (length < 3)
+  {
+    return 0;
+  }
+
+  const unsigned char* b = (const unsigned char*)(input) + length - 3;
+
+  if (b[2] & 0x80)
+  {
+    // First byte in a new multi-byte sequence
+    if (b[2] & 0x40)
+      return 1;
+    // 3 byte sequence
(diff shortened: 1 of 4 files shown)


utils/common/utils_utf8.h
```diff
@@ -91,6 +91,33 @@ inline std::string wstring_to_utf8(const std::wstring& str)
   return ret;
 }
 
+inline uint8_t utf8_truncate_point(const char* input, size_t length)
+{
+  // Find the beginning of a multibyte char to truncate at and return the
+  // number of bytes to truncate1`
+  if (length < 3)
```
drrtuy, line 98: I understand this code had existed for some time but I am curious how does this handles 2 byte glyphs, e.g. russian alphabet is encoded in 2 bytes in utf-8.

drrtuy (approved):

utils/common/utils_utf8.h
tntnatbry: This is a good observation. I believe in this case, the function will return 1 and reduce the byte-length of the input string by one so that the last byte in the input after truncation belongs to a well-formed character.

But I admit that the robustness of this function handling various edge cases for truncation around multi-byte character boundaries needs to be tested.

PR #2964, opened 2023-09-20
MCOL-5572 - fixing bad handling of UTF-8 CHAR(1) column in calpont system catalog table processing

mariadb-SergeyZefirov:
UTF-8 requires at least 3-byte wide columns and internal processing of ``autoincrement`` columns of ``syscolumn`` table used 1. ASCII characters have least byte zero in the strnxfrm encoding and, because of this, 'y' and 'n' got converted into zeroes (as would 'z', 'Z', '?' and even '~') and, thusly, that resulted in 'y' being equal to 'n'.

We have here a fix and a test.

### DIFF
--- a/mysql-test/columnstore/basic/t/MCOL-5572-autoincrement-filtering.test
+++ b/mysql-test/columnstore/basic/t/MCOL-5572-autoincrement-filtering.test
@@ -1,0 +1,22 @@
+--source ../include/have_columnstore.inc
+
+SET default_storage_engine=columnstore;
+
+--disable_warnings
+DROP DATABASE IF EXISTS mcol5572;
+--enable_warnings
+
+
+CREATE DATABASE mcol5572;
+USE mcol5572;
+
+create table foo (col1 int) engine=columnstore;
+insert into foo values ( 1 ), ( 2 ), ( 3 ), ( 4 ), ( 5 );
+Alter table foo add column newcol bigint comment 'autoincrement';
+select callastinsertid('foo');
+select tablename, autoincrement, nextvalue from calpontsys.syscolumn where autoincrement = 'y' and `schema`='mcol5572' and tablename='foo';
+select tablename, autoincrement, nextvalue from calpontsys.syscolumn where autoincrement != 'n' and `schema`='mcol5572' and tablename='foo';
+select tablename, autoincrement, nextvalue from calpontsys.syscolumn where autoincrement = 'n' and `schema`='mcol5572' and tablename='foo';
+
+DROP DATABASE mcol5572;
+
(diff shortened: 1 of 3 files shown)


mariadb-SergeyZefirov:
This is  not squashed. Also, libmarias3 link is incorrect. I ill fix it shortly.

mysql-test/columnstore/basic/t/MCOL-5572-autoincrement-filtering.test
```diff
@@ -0,0 +1,22 @@
...
+CREATE DATABASE mcol5572;
+USE mcol5572;
+
+create table foo (col1 int) engine=columnstore;
+insert into foo values ( 1 ), ( 2 ), ( 3 ), ( 4 ), ( 5 );
+Alter table foo add column newcol bigint comment 'autoincrement';
+select callastinsertid('foo');
+select tablename, autoincrement, nextvalue from calpontsys.syscolumn  where autoincrement = 'y';
+select tablename, autoincrement, nextvalue from calpontsys.syscolumn  where autoincrement != 'n';
+select tablename, autoincrement, nextvalue from calpontsys.syscolumn  where autoincrement = 'n';
```
tntnatbry: Can you please include the `schema` and `tablename` in the WHERE clause of the above 3 SELECT queries on `calpontsys.syscolumn`. This is because if this test is run on an existing installation with other databases and tables, it will return more rows than expected and the test will fail. Here is the updated statements (please note the usage of back tick ` around the schema keyword which is required):
```sql
select tablename, autoincrement, nextvalue from calpontsys.syscolumn  where autoincrement = 'y' and `schema`='mcol5572' and tablename='foo';
select tablename, autoincrement, nextvalue from calpontsys.syscolumn  where autoincrement != 'n' and `schema`='mcol5572' and tablename='foo';
select tablename, autoincrement, nextvalue from calpontsys.syscolumn  where autoincrement = 'n' and `schema`='mcol5572' and tablename='foo';
```

tntnatbry (changes requested):
@mariadb-SergeyZefirov Code change looks good to me. However, I have added a comment to update the MTR test.

mysql-test/columnstore/basic/t/MCOL-5572-autoincrement-filtering.test
mariadb-SergeyZefirov: There are an "i_s.test" and "mcol641-create.test" tests in the basic test suite that most probably have the same problem.

Notwithsdanding that, I've changed these SELECT statements per your request.

mariadb-LeonidFedorov (approved):

PR #2969, opened 2023-09-25
init columnstore_review.sh & cs_package_manager.sh into extras

mariadb-AllenHerrera:


extra/columnstore_review.sh
```diff
@@ -0,0 +1,2330 @@
+#!/bin/bash
```
drrtuy, line 1: It might worth to add GPLv2 header here.

mariadb-LeonidFedorov (approved):

PR #2977, opened 2023-09-29
MCOL-5583: bug(engine) 

mariadb-LeonidFedorov:
suggested invariant is broken by some reasons, so commenting assert till further investigation


drrtuy (approved):
untested ACK

PR #2985, opened 2023-10-05
Fix disk-based distinct aggregation. (E.g. queries like `AGG(DISTINCT ...)` and `AGG(DISTINCT ...) ... GROUP BY ...`).

phoeinx:
Description to be added.

### DIFF
--- a/utils/rowgroup/rowaggregation.cpp
+++ b/utils/rowgroup/rowaggregation.cpp
@@ -4576,7 +4576,8 @@ void RowAggregationDistinct::addRowGroup(const RowGroup* pRows,
 //------------------------------------------------------------------------------
 void RowAggregationDistinct::doDistinctAggregation()
 {
-  while (dynamic_cast<RowAggregationUM*>(fAggregator.get())->nextRowGroup())
+  fAggregator->finalAggregation();  // TODO: Perhaps move somewhere else?
+  while (dynamic_cast<RowAggregationUM*>(fAggregator.get())->nextOutputRowGroup())
   {
     fRowGroupIn.setData(fAggregator->getOutputRowGroup()->getRGData());
 
--- a/utils/rowgroup/rowaggregation.cpp
+++ b/utils/rowgroup/rowaggregation.cpp
@@ -4105,6 +4105,18 @@ bool RowAggregationUM::nextRowGroup()
   return more;
 }
 
+bool RowAggregationUM::nextOutputRowGroup()
+{
+  bool more = fRowAggStorage->getNextOutputRGData(fCurRGData);
+
+  if (more)
+  {
+    fRowGroupOut->setData(fCurRGData.get());
+  }
+
+  return more;
+}
+
 //------------------------------------------------------------------------------
 // Row Aggregation constructor used on UM
 // For 2nd phase of two-phase case, from partial RG to final aggregated RG
--- a/utils/rowgroup/rowstorage.cpp
+++ b/utils/rowgroup/rowstorage.cpp
@@ -666,6 +666,156 @@ class RowGroupStorage
     }
   }
 
+  /** @brief Used to output aggregation results from memory and disk in the current generation in the form of
+   * RGData. Returns next RGData, loads from disk if necessary. Skips finalized rows as they would contain
+   * duplicate results, compacts actual rows into start of RGData and adapts number of rows transmitted in
+   * RGData.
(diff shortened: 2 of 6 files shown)


utils/rowgroup/rowstorage.cpp
```diff
@@ -666,6 +666,156 @@ class RowGroupStorage
...
+      fRowGroupOut->setData(rgdata.get());
+      fRowGroupOut->setRowCount(pos);
+      int64_t memSz = fRowGroupOut->getSizeWithStrings(fMaxRows);
+
+      // TODO: why do we release the memory if we just loaded something?, or are we realesing for file
+      // managment?
+      fMM->release(memSz);
+      unlink(makeRGFilename(rgid).c_str());
+
+      fLRU->remove(rgid);  // TODO: why? if we are compeletly dissolving this storage, what does it help?
```
drrtuy, line 812: This call cleans the memory marked as freed. Otherwise control flow frees all the memory at once when it finishes the processing that is not a concurrent-query-friendly behavior.

utils/rowgroup/rowaggregation.cpp
```diff
@@ -4560,7 +4572,8 @@ void RowAggregationDistinct::addRowGroup(const RowGroup* pRows,
 //------------------------------------------------------------------------------
 void RowAggregationDistinct::doDistinctAggregation()
 {
-  while (dynamic_cast<RowAggregationUM*>(fAggregator.get())->nextRowGroup())
+  fAggregator->finalAggregation();  // TODO: Perhaps move somewhere else?
```
drrtuy, line 4579: You are right. I suggest to add a wrapper method in RowAggregationDistinct to call it in TAS control flow.

utils/rowgroup/rowaggregation.cpp
```diff
@@ -4560,7 +4572,8 @@ void RowAggregationDistinct::addRowGroup(const RowGroup* pRows,
...
+  while (dynamic_cast<RowAggregationUM*>(fAggregator.get())->nextOutputRowGroup())
```
drrtuy, line 4580: There at least two points in the existing code that don't look right. 
Dynamic cast traverses classes hierarchy every time it is called so I suggest you to run it once.
It is correct to check if dynamic_cast is successful before one uses it. I mean, I see there is an implicit invariant telling fAggregator is always RowAggregationUM but in future this might become a false statement.

utils/rowgroup/rowstorage.cpp
```diff
@@ -666,6 +666,156 @@ class RowGroupStorage
...
   }
 
+  /** @brief Used to output aggregation results from memory and disk in the current generation in the form of
+   * RGData. Returns next RGData, loads from disk if necessary. Skips finalized rows as they would contain
+   * duplicate results, compacts actual rows into start of RGData and adapts number of rows transmitted in
+   * RGData.
+   * @returns A pointer to the next RGData or an empty pointer if there are no more RGDatas in this
+   * generation.
+   */
+  bool getNextOutputRGData(std::unique_ptr<RGData>& rgdata)
```
drrtuy, line 676: This method is a close replica of 
```
bool getNextRGData(std::unique_ptr<RGData>& rgdata, std::string& fname)
```
I suggest to separate the common code into methods so future changes affects both this method and  getNextRGData mentioned.

utils/rowgroup/rowaggregation.cpp
```diff
@@ -4104,6 +4104,18 @@ bool RowAggregationUM::nextRowGroup()
   return more;
 }
 
+bool RowAggregationUM::nextOutputRowGroup()
+{
+  bool more = fRowAggStorage->getNextOutputRGData(fCurRGData);
```
drrtuy, line 4110: ~An optional suggestion is to leverage std::option<unique_ptr<RGData>> here. c++23 enables a monadic flow based on option. 
I suggest to apply this despite the fact that all other methods use bool to signal the state of a stream.~

utils/rowgroup/rowaggregation.cpp
drrtuy, line 4580: Every dynamic_cast costs a lot, so I suggest to cast it only once. I also suggest to check dynamic_cast result before the cast is used.

drrtuy (changes requested):
Nice patch @phoeinx . I put some comments so plz address them.

utils/rowgroup/rowstorage.cpp
phoeinx: will do

PR #2990, opened 2023-10-09
fix(build): Fixing build for S3 with "LocalStorage" type.

denis0x0D:


build/bootstrap_mcs.sh
```diff
@@ -170,15 +171,24 @@ start_service()
     check_service mcs-writeengineserver
 }
 
+start_storage_manager_if_needed()
+{
+  if [ ! -z "$MCS_USE_S3_STORAGE" ] && [ $MCS_USE_S3_STORAGE -eq 1 ]; then
```
mariadb-LeonidFedorov: Don't you think It's more convenient to use flag for build script instead of environment variable? There is a very simple and useful option parser at the bottom

denis0x0D: This variable is already needed to be set for `post install script` at line 313

mariadb-LeonidFedorov: OK, maybe this variable should be set with the command line option then?
I like the idea of having --help with full options set

denis0x0D: I can add a new option, but this not only one variable to be set, to enable sm, you have to set this variables as well
export MCS_USE_S3_STORAGE=1;
export MCS_S3_BUCKET=;
export MCS_S3_ACCESS_KEY_ID=
export MCS_S3_SECRET_ACCESS_KEY=

denis0x0D: @mariadb-LeonidFedorov since you maintainer of this script, any suggestion about naming for this kind of variable?

mariadb-LeonidFedorov: I believe your choice

mariadb-AlexeyVorovich: please update help regardless of the choice of env vs options.
If help contains a set of sample invocations for common scenarios, that would help as well

mariadb-LeonidFedorov (approved):

PR #3008, opened 2023-10-24
Fix/mcol 5568 stable 23.10

drrtuy:
This is a backport of the path from the develop.


mariadb-SergeyZefirov (approved):
I agree with these changes.

PR #3011, opened 2023-10-26
MCOL-4740: This fixes update rows counter for multi-table update

drrtuy:


dbcon/mysql/ha_mcs.cpp
```diff
@@ -16,6 +16,14 @@
    Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
    MA 02110-1301, USA. */
 
+// This makes specific MDB classes' attributes public to implement
+// MCOL-4740 temporary solution. Search for MCOL-4740
+// to get the actual place where it is used.
+#define updated_leaves \
```
mariadb-LeonidFedorov, line 22: I have so many words to say, I'd better no

drrtuy: I know, I know but it is better to save the precious breath )

dbcon/mysql/ha_mcs_pushdown.cpp
```diff
@@ -715,22 +715,29 @@ int ha_mcs_group_by_handler::end_scan()
...
- * PARAMETERS:
- *    thd - THD pointer.
- *    sel_lex - SELECT_LEX* that describes the query.
- *    sel_unit - SELECT_LEX_UNIT* that describes the query.
- * RETURN:
- *    select_handler if possible
- *    NULL in other case
- ***********************************************************/
+* DESCRIPTION:
+* Creates a select handler if there is no non-equi JOIN, e.g
```
denis0x0D, line 719: This is interesting comment, CS does not support non-equi join.

dbcon/mysql/ha_mcs.cpp
```diff
@@ -618,6 +626,29 @@ int ha_mcs::rnd_end()
 {
   DBUG_ENTER("ha_mcs::rnd_end");
 
+  // MCOL-4740 multi_update::send_eof(), which outputs the affected
+  // number of rows to the client, is called after handler::rnd_end().
+  // So we set multi_update::updated and multi_update::found here.
+  if (current_thd->lex->sql_command == SQLCOM_UPDATE_MULTI)
```
denis0x0D: IMHO, it could be placed in separate function.

dbcon/mysql/ha_mcs_pushdown.cpp
drrtuy: I didn't get you TBH. MCS doesn't support non-equi JOIN with SH but MDB can do non-equi for MCS if SH is disabled bu this will be sloooow.

dbcon/mysql/ha_mcs.cpp
drrtuy: Good point.

mariadb-LeonidFedorov:
I still prefer friend class instead of disgusting hack with preprocessor

drrtuy:
> I still prefer friend class instead of disgusting hack with preprocessor

Though I understand you intention, I am not going to fight for the `friend` solution b/c it is also a hack. The hack also needs server-side changes that Serg is reluctant to introduce. What I would like to fight for is a long-term solution that is described in [this MDEV](https://jira.mariadb.org/browse/MDEV-32256).

drrtuy:
I am very sorry but I had to rewrite the patch. The major difference with the prev patch is that I moved bunch of common utility functions into a separate ha_mcs_common.h to reuse them both in the plugin facade and plugin internals.

mariadb-DenisKhalikov (approved):
LGTM

PR #3017, opened 2023-11-02
fix(rowstorage): MCOL-5472 SplitMix64 PRNG implementation to replace stdlib MT PRNG that uses /dev/urandom guarded by spinlock

drrtuy:
This adds a new PRNG to produce a pseudo random number for probability checks.

### DIFF
--- a/utils/rowgroup/rowstorage.cpp
+++ b/utils/rowgroup/rowstorage.cpp
@@ -1495,9 +1495,6 @@ RowAggStorage::RowAggStorage(const std::string& tmpDir, RowGroup* rowGroupOut, R
  , fTmpDir(tmpDir)
  , fRowGroupOut(rowGroupOut)
  , fKeysRowGroup(keysRowGroup)
- , fRD()
- , fRandGen(fRD())
- , fRandDistr(0, 100)
 {
   char suffix[PATH_MAX];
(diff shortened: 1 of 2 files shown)


mariadb-SergeyZefirov (approved):
I agree with these changes.

mariadb-SergeyZefirov:
SplitMix is also used as a standard RNG in Haskell's System.Random and I had some problems with the quality of numbes generated by it. It appears that higher bits of the output value are very interdependent..

It is good enough for most purposes, but when you try to simulate something with it, it can pull the rug underneath you.

It would be good to have a fast projection-based (COUNTER mode of encryption) RNG for someone as picky as me. ;)

E.g., something like Siphash from variable number of BIGINTs. SIPHASH(a,b,...,z).

drrtuy:
> SplitMix is also used as a standard RNG in Haskell's System.Random and I had some problems with the quality of numbes generated by it. It appears that higher bits of the output value are very interdependent..
> 
> It is good enough for most purposes, but when you try to simulate something with it, it can pull the rug underneath you.
> 
> It would be good to have a fast projection-based (COUNTER mode of encryption) RNG for someone as picky as me. ;)
> 
> E.g., something like Siphash from variable number of BIGINTs. SIPHASH(a,b,...,z).

I agree that SplitMix64 isn't good enough for simulation and modeling.

PR #3020, opened 2023-11-05
MCOL-5544 Disable saving/reading statistics to/from file.

denis0x0D:
Temporal workaround until MCOL-5544 is fixed.


drrtuy:
Thx for the patch but we shouldn't disable it.

denis0x0D:
Let's close this

PR #3021, opened 2023-11-07
MCOL-5603 Online alter can break data consistency[develop]

FooBarrior:
Propagate the TL lock to TL_READ_NO_INSERT, blocking any concurrent writes until the end of a copy stage.

After that, this lock will be unlocked, allowing a window to write into a table before MDL is upgraded to exclusive (see online_alter_read_from_binlog call and the "alter_table_online_before_lock" debug syncpoint). This window will be covered by the second replication iteration after the lock upgrade


FooBarrior:
The CI failure looks suspicious, I hope someone can rerun the failing nodes

mariadb-LeonidFedorov:
Work moved to https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/3559

PR #3027, opened 2023-11-17
MCOL-5597 Set length for `nullptr` string to 0.

denis0x0D:


mariadb-DenisKhalikov:
The point here is to follow the same semantics which we have for another ConstString constructor which set length = 0 if *value == nullptr

drrtuy (approved):
untested ACK

mariadb-SergeyZefirov (approved):
Completely agree.

PR #3029, opened 2023-11-17
MCOL-5597 Set length for `nullptr` string to 0.

denis0x0D:


drrtuy:
@denis0x0D kindly change the branch to stable-23.10

PR #3035, opened 2023-11-24
MCOL-5463: identical column names in table and table in view

mariadb-SergeyZefirov:


drrtuy (approved):

denis0x0D:
@mariadb-SergeyZefirov @mariadb-LeonidFedorov This is a completely different problem comparing MCOL-5476, what we have there is problem with matching duplicates in select statement and group by statement for builtin function, for example:
select sum(a), length(b), length(b) from t1 group by length(b), length(b) - does not work, because every `function column` is associated with unique expression id, and currently cannot be matched with columns in group by statement.

PR #3045, opened 2023-11-29
MCOL-5467 Add support for duplicate expressions in group by.

denis0x0D:
This patch adds support for duplicate expressions (builtin_functions) with one argument in select statement and group by statement.

### DIFF
--- a/dbcon/joblist/jlf_common.h
+++ b/dbcon/joblist/jlf_common.h
@@ -93,6 +93,20 @@ struct TupleInfo
   uint32_t csNum;  // For collations
 };
 
+// This struct holds information about `FunctionColumn`.
+struct FunctionColumnInfo
+{
+  // Function argument.
+  uint64_t associatedColumnOid;
+  // Function name.
+  std::string functionName;
+
+  FunctionColumnInfo(uint64_t colOid, std::string funcName)
+   : associatedColumnOid(colOid), functionName(funcName)
+  {
+  }
+};
+
 // for compound join
 struct JoinData
 {
@@ -383,6 +397,8 @@ struct JobInfo
   std::map<std::pair<uint32_t, uint32_t>, int64_t> joinEdgesToRestore;
   // Represents a pair of `table` to be on a large side and weight associated with that table.
   std::unordered_map<uint32_t, int64_t> tablesForLargeSide;
+  // Represents a pair of `tupleId` and `FunctionColumnInfo`.
+  std::unordered_map<uint32_t, FunctionColumnInfo> functionColumnMap;
 
  private:
   // defaults okay
--- a/dbcon/joblist/joblistfactory.cpp
+++ b/dbcon/joblist/joblistfactory.cpp
@@ -172,7 +172,7 @@ void projectSimpleColumn(const SimpleColumn* sc, JobStepVector& jsv, JobInfo& jo
       // This is a double-step step
       //			if (jobInfo.trace)
       //				cout << "doProject Emit pGetSignature for SimpleColumn " << dictOid <<
-      //endl;
+      // endl;
 
       pds = new pDictionaryStep(dictOid, tbl_oid, ct, jobInfo);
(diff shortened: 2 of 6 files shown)


dbcon/joblist/jlf_common.h
```diff
@@ -93,6 +93,15 @@ struct TupleInfo
   uint32_t csNum;  // For collations
 };
 
+// This struct holds information about `FunctionColumn`.
+struct FunctionColumnInfo
+{
+  // Function argument.
+  uint64_t associatedColumnOid;
+  // Function name.
+  std::string functionName;
```
drrtuy, line 102: Mb ref?

dbcon/joblist/joblistfactory.cpp
```diff
@@ -1246,6 +1248,7 @@ const JobStepVector doAggProject(const CalpontSelectExecutionPlan* csep, JobInfo
         const WindowFunctionColumn* wc = NULL;
         bool hasAggCols = false;
         bool hasWndCols = false;
+        bool hasFuncCols = false;
```
drrtuy: It is better to rename the flag so it brings the specific semantics, e.g. hasFuncColsWithSingleArg

dbcon/joblist/joblistfactory.cpp
```diff
@@ -1291,6 +1297,14 @@ const JobStepVector doAggProject(const CalpontSelectExecutionPlan* csep, JobInfo
         {
           jobInfo.expressionVec.push_back(tupleKey);
         }
+
+        if (hasFuncCols)
+        {
+          FunctionColumnInfo fcInfo;
```
drrtuy: You should better add a proper ctor.

mysql-test/columnstore/bugfixes/mcol-5476.result
```diff
@@ -0,0 +1,71 @@
...
+insert into t1 values (1, 1), (2, 1), (3, 1), (4, 2), (5, 2);
+select sum(a), abs(b), abs(b) from t1 group by abs(b), abs(b) order by abs(b);
+sum(a)	abs(b)	abs(b)
+6	1	1
+9	2	2
+select sum(a), abs(b), abs(b) from t1 group by abs(b) order by abs(b);
+sum(a)	abs(b)	abs(b)
+6	1	1
+9	2	2
+select sum(distinct a), abs(b), abs(b) from t1 group by abs(b), abs(b) order by abs(b);
```
drrtuy: You should use #sorted_result MTR interpreter clause. This way MTR sorts the given and the expected outputs on its own before comparison. You can drop ORDER BY using #sorted_result.

drrtuy: This is valid for all similar cases in this test.

dbcon/joblist/tupleaggregatestep.cpp
```diff
@@ -5953,4 +5984,42 @@ void TupleAggregateStep::formatMiniStats()
...
+uint32_t TupleAggregateStep::getTupleKeyFromTuple(uint32_t key)
+{
+  return key;
+}
+
+template <class GroupByMap>
+bool TupleAggregateStep::tryToFindEqualFunctionColumnByTupleKey(JobInfo& jobInfo, GroupByMap& groupByMap,
+                                                                const uint32_t tupleKey, uint32_t& foundKey)
+{
+  if (jobInfo.functionColumnMap.count(tupleKey))
```
drrtuy: Change to find/cmp with map::end() as per offline discussion.

dbcon/joblist/jlf_common.h
denis0x0D: I think it's ok to hold a copy of string, because I'm not sure when FunctionColumn destructor is about to call.

drrtuy (approved):

drrtuy:
untested ACK

PR #3049, opened 2023-11-30
fix(DEC): MCOL-5602 fixing potentially endless loop in DEC

drrtuy:


dbcon/joblist/distributedenginecomm.cpp
```diff
@@ -1206,9 +1181,15 @@ DistributedEngineComm::MQE::MQE(const uint32_t pCount, const uint32_t initialInt
                                 const uint64_t flowControlEnableBytesThresh)
  : ackSocketIndex(0), pmCount(pCount), hasBigMsgs(false), targetQueueSize(flowControlEnableBytesThresh)
 {
-  unackedWork.reset(new volatile uint32_t[pmCount]);
   interleaver.reset(new uint32_t[pmCount]);
-  memset((void*)unackedWork.get(), 0, pmCount * sizeof(uint32_t));
+  const uint32_t localPmCount = pmCount;
```
denis0x0D, line 1185: Didn't get why `localPmCount` is needed, if `pmCount` already exists

dbcon/joblist/distributedenginecomm.cpp
```diff
@@ -761,65 +766,66 @@ void DistributedEngineComm::sendAcks(uint32_t uniqueID, const vector<SBS>& msgs,
   }
 }
 
-void DistributedEngineComm::nextPMToACK(boost::shared_ptr<MQE> mqe, uint32_t maxAck, uint32_t* sockIndex,
-                                        uint16_t* numToAck)
+size_t DistributedEngineComm::ackAndRotatePM(boost::shared_ptr<MQE> mqe, const size_t maxAck,
```
denis0x0D: Can you please explain why did you change the name of this function, it seems to me the ACK is not sending to PM from this function.

dbcon/joblist/distributedenginecomm.cpp
drrtuy: pmCount is atomic and it would be wasteful to use it in for comparison expression.

mariadb-DenisKhalikov (approved):

PR #3052, opened 2023-12-05
MCOL-5467 Add support for duplicate expressions in group by.

denis0x0D:
This patch adds support for duplicate expressions (builtin_functions) with one argument in select statement and group by statement.


drrtuy (approved):
untested ACK

PR #3066, opened 2023-12-13
MCOL-4202: crossdb rename

mariadb-LeonidFedorov:
``` 
uint8_t updateSyscolumnTablename(messageqcpp::ByteStream& bs, std::string& err);
uint8_t updateSystableTablename(messageqcpp::ByteStream& bs, std::string& err);
```
this functions were written in assumption we are renaming table only on the same schema
the code is ugly mix of low level manipulations and logic, so I refactored a helper function for both and call it twice


mariadb-SergeyZefirov (approved):
I see no problems here.

mysql-test/columnstore/basic/t/mcol4202-crossdb-rename.test
```diff
@@ -0,0 +1,23 @@
+#
+# Test ALTER TABLE schemas in various possible ways.
+#
+-- source ../include/have_columnstore.inc
+-- source include/have_innodb.inc
+
+--disable_warnings
+DROP DATABASE IF EXISTS db1;
```
drrtuy: I suggest you to rename db1 to mcol4202_db1 so there will be no overlap if MTR test fails and doesn't remove the leftovers.

drrtuy (approved):
untested ACK.
See the comment on the schema names in the MTR test though.

PR #3067, opened 2023-12-15
MCOL-4202: crossdb rename [stable23.10]

mariadb-LeonidFedorov:
``` 
uint8_t updateSyscolumnTablename(messageqcpp::ByteStream& bs, std::string& err);
uint8_t updateSystableTablename(messageqcpp::ByteStream& bs, std::string& err);
```
this functions were written in assumption we are renaming table only on the same schema
the code is ugly mix of low level manipulations and logic, so I refactored a helper function for both and call it twice


drrtuy (approved):
Untested ACK

PR #3073, opened 2023-12-18
perf(primproc) MCOL-5601: Initilize two fields once in ctor instead of calling makeConfig

mariadb-LeonidFedorov:
```
std::string fTmpDir = config::Config::makeConfig()->getTempFileDir(config::Config::TempDirPurpose::Aggregates); 
std::string fCompStr = config::Config::makeConfig()->getConfig("RowAggregation", "Compression");
```

### DIFF
--- a/utils/rowgroup/rowaggregation.cpp
+++ b/utils/rowgroup/rowaggregation.cpp
@@ -675,19 +675,16 @@ void RowAggregation::initialize(bool hasGroupConcat)
     }
   }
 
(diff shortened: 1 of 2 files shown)


mariadb-LeonidFedorov:
Further action point inho is to replace `boost::mutex` for `xmlDoc` and `instanceMap` with `std::shared_mutex` to avoid contention on any readonly `getConfig`

drrtuy (approved):
untested ACK

PR #3076, opened 2023-12-19
MCOL-5623 This patch changes the logic applied to `save` function for EM

denis0x0D:


versioning/BRM/extentmap.cpp
```diff
@@ -1802,26 +1802,28 @@ void ExtentMap::save(const string& filename)
 
 #endif
 
-  grabEMEntryTable(READ);
-  grabEMIndex(READ);
+  // MCOL-5623 Grag the `WRITE` lock, we cannot let BRM sctructres to be modified (new Extent), while saving
+  // them to the file.
+  grabEMEntryTable(WRITE);
```
drrtuy: READ lock taken assumes there will be no writes. 
Taking a WRITE lock you disallows other threads to read EM or EMIndex.

denis0x0D:
When SM is enabled one `write` takes about ~2 milliseonds, so if em has about 150k elements, it takes about 300 seconds to `save` function instead ~4 milliseconds.

versioning/BRM/extentmap.cpp
```diff
@@ -1802,26 +1802,28 @@ void ExtentMap::save(const string& filename)
 
 #endif
 
-  grabEMEntryTable(READ);
-  grabEMIndex(READ);
+  // MCOL-5623 Grag the `WRITE` lock, we cannot let BRM sctructres to be modified (new Extent), while saving
```
drrtuy: Type in Grag I suppose.

drrtuy (changes requested):
Take a look at the comments plz.

versioning/BRM/extentmap.cpp
denis0x0D: I see, thanks for showing this, fixed

drrtuy (approved):
Untested ACK

PR #3085, opened 2023-12-25
MCOL-5631 Optimize BRM load operation.

denis0x0D:


denis0x0D:
We have a fail, when trying to start a cluster with cmapi, because of timing on "load_brm" (Extent map 2.5kk elements)
`0.327067 |0|0|0| C 29 CAL0000: ExtentMap::loadVersion4(): read :  Connection reset by peer        %%10%%`
But it works fine with just systemctl start mariadb-columnstore;
` mcs-loadbrm.service: Succeeded.`

drrtuy (approved):

mariadb-AlanMologorsky:
@denis0x0D Any action item needed from my side about cmapi?

denis0x0D:
@mariadb-AlanMologorsky it seems like the problem is fixed by this patch. @drrtuy just upgraded system with that patch and everything is ok.

mariadb-LeonidFedorov (approved):

PR #3087, opened 2023-12-26
MCOL-5624: wrong config option stable

mariadb-LeonidFedorov:
…o be required to start mariadb server

### DIFF
--- a/dbcon/mysql/columnstore.cnf
+++ b/dbcon/mysql/columnstore.cnf
@@ -4,4 +4,4 @@ plugin-load-add=ha_columnstore.so
 collation_server = utf8_general_ci
 character_set_server = utf8
 
-columnstore_use_import_for_batchinsert = ON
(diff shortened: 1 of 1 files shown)


mariadb-AlanMologorsky (approved):
LGTM

drrtuy (changes requested):
Double check if this must be `loose_` but not `loose-`.
See [this](https://www.percona.com/blog/using-the-loose_-option-prefix-in-my-cnf/) for details.

mariadb-LeonidFedorov:
> Double check if this must be `loose_` but not `loose-`. See [this](https://www.percona.com/blog/using-the-loose_-option-prefix-in-my-cnf/) for details.

I prefer our own documentation https://mariadb.com/kb/en/mariadbd-options/#-loose-
And I have tested modified name option against corresponding value for started server

drrtuy (approved):

PR #3088, opened 2023-12-26
stable-23.10 remove boost from deps, add procsps, due without ps installation halts

mariadb-LeonidFedorov:


mariadb-AlanMologorsky (approved):
LGTM

cmake/cpackEngineRPM.cmake
```diff
@@ -71,20 +71,14 @@ IF (EXISTS "/etc/redhat-release")
...
-    SETA(CPACK_RPM_columnstore-engine_PACKAGE_REQUIRES "MariaDB-columnstore-shared" "snappy" "net-tools" "MariaDB-server" "python3")
-    # Disable auto require as this will also try to pull Boost via RPM
-    SET(CPACK_RPM_PACKAGE_AUTOREQPROV " no")
-elseif (${SUSE_VERSION_NUMBER} EQUAL 12)
-    SETA(CPACK_RPM_columnstore-engine_PACKAGE_REQUIRES "boost-devel >= 1.54.0" "libsnappy1" "net-tools" "MariaDB-server" "python3" "jemalloc")
-else ()
-    SETA(CPACK_RPM_columnstore-engine_PACKAGE_REQUIRES "boost >= 1.53.0" "snappy" "net-tools" "MariaDB-server" "python3" "jemalloc")
-endif()
+
+SETA(CPACK_RPM_columnstore-engine_PACKAGE_REQUIRES "snappy" "net-tools" "MariaDB-server" "python3" "jemalloc" "procps-ng")
```
drrtuy, line 81: elfutils for eu-stack?

mariadb-LeonidFedorov: There is the logic difference, procs is required for Columnstore to install, we use ps in install scripts, eustack is used on ci only, and IMHO should be installed separately

drrtuy (approved):
Untested ACK. All ARM builds had failed to start so we don't know if it breaks anything there or not.

PR #3093, opened 2023-12-27
MCOL-4886 skip chown if top folder permissions good

mariadb-AllenHerrera:


mariadb-AllenHerrera:
The problem we're trying to solve is re-running chown mysql on millions of files if theres a shorter way to validate that we can skip this step.

mariadb-AlanMologorsky (approved):
LGTM 
Tested on Rocky8. No files with other then mysql owner exists, except dbroot1-lock file (root root)

PR #3099, opened 2024-01-02
Boost1.84[stable-23.10]

mariadb-LeonidFedorov:
Something happened with boost 1.81 we used
```
-- SHA256 hash of
    /mdb/verylongdirnameforverystrangecpackbehavior/storage/columnstore/columnstore/.boost/src/boost_1_81_0.tar.bz2
  does not match expected value
    expected: '71feeed900fbccca04a3b4f2f84a7c217186f28a940ed8b7ed4725986baf99fa'
      actual: '5e89103d9b70bba5c91a794126b169cb67654be2051f90cf7c22ba6893ede0ff'
-- Hash mismatch, removing...
CMake Error at external_boost-stamp/download-external_boost.cmake:159 (message):
  Each download failed!
```
IMHO it's time to update

### DIFF
--- a/cmake/boost.cmake
+++ b/cmake/boost.cmake
@@ -37,8 +37,8 @@ ENDFOREACH()
 
 ExternalProject_Add(external_boost
   PREFIX .boost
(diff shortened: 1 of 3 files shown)


drrtuy (approved):
untested ACK

PR #3101, opened 2024-01-05
MCOL-5630: fix multi node MTR. Add "cmapi is-ready" cli command [develop-23.02]

mariadb-LeonidFedorov:
The docker repo is isolated from engine, so we need duplicates to all branches we use

This task is a next logical part of MCOL-5470 that implements AppManager and API endpoint
to check if CMAPI ready or not.

- [x] [add] cmapi_app.py with is-ready command implementation
- [x] [fix] add cmapi is-ready command to main typer app


drrtuy (approved):
untested ACK

PR #3106, opened 2024-01-12
fix(messageqcpp): MCOL-5636 same node communication crashes transmiting PP errors to EM b/c error messaging leveraged socket that was a nullptr.

drrtuy:


mariadb-LeonidFedorov (approved):

utils/messageqcpp/samenodepseudosocket.cpp
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// This is the only working method of this class. It puts SBS directly into DEC queue.
+void SameNodePseudoSocket::write(SBS msg, Stats* stats)
+{
+  dec_->addDataToOutput(msg);
+}
+
+void SameNodePseudoSocket::write(const ByteStream& msg, Stats* stats)
+{
```
denis0x0D, line 68: Just wondering, could it be a situation that in a future we call this method, instead of one on line 61?

drrtuy: The write() here is private within the child class so it should be impossible.

denis0x0D (approved):

PR #3116, opened 2024-01-31
Fixes MCOL-5599 where LIKE operator never finishes [develop-23.02]

mariadb-LeonidFedorov:
This is a fix of logging subsystem, nothing else.

The old code expanded an argument into string and advanced too little and, if expansion contained argument's index, it expanded it again. And again.


drrtuy (approved):
untested ACK

PR #3117, opened 2024-02-01
MCOL-5597: large side read errors

mariadb-SergeyZefirov:
The large side read errors mentioned there can be due to failure to close file stream properly. Some of the data may still reside in the file stream buffers, closing must flush it. The flush is an I/O operation and can fail, leading to partial write and subsequent partial read.

This patch tries to provide better diagnostics.


mariadb-DenisKhalikov (commented):
Makes sense to me.

mariadb-DenisKhalikov (approved):
LGTM

drrtuy (approved):
untested ACK

PR #3120, opened 2024-02-08
MCOL-5352 Properly handle PP restart for DML/DDL operations.

denis0x0D:


mariadb-DenisKhalikov:
This patch solves the following problem - we could have a situation when PP is restarted, in this case we have to restart the connection from DML/DDL to PP. When we start the DML/DDL command without connection restarted, `process` command will return an error, in this case we have to 1) restart a connection 2) try to run DML/DDL command one more time. This is implemented by following approach - we have a function `processPackage` for each base DML/DDL class which implements a wrapper around `processPackageInternal`, the wrapper calls `processPackageInternal` which is virtual function and overrided for each DML/DDL class and implements the special logic. The `processPackage` handles the case when we need to restart connection and try to call `process` command one more time.

dmlproc/dmlprocessor.cpp
```diff
@@ -976,7 +987,17 @@ void PackageHandler::run()
             CalpontSystemCatalog::TableName tableName;
             tableName.schema = updatePkg->get_Table()->get_SchemaName();
             tableName.table = updatePkg->get_Table()->get_TableName();
-            CalpontSystemCatalog::ROPair roPair = fcsc->tableRID(tableName);
+            CalpontSystemCatalog::ROPair roPair;
```
drrtuy: This snippet is used 3 times in this unit. It is worth to make it a function.

dbcon/dmlpackageproc/dmlpackageprocessor.cpp
```diff
@@ -287,6 +287,31 @@ int32_t DMLPackageProcessor::tryToRollBackTransaction(uint64_t uniqueId, BRM::Tx
   return weRc;
 }
 
+DMLPackageProcessor::DMLResult DMLPackageProcessor::processPackage(dmlpackage::CalpontDMLPackage& cpackage)
+{
+  auto result = processPackageInternal(cpackage);
+  uint32_t tries = 0;
+  // Try to setup connection and process package one more time.
+  while ((result.result == PP_LOST_CONNECTION) && (tries < 5))
```
drrtuy, line 295: Is it possible to use do-while here? What do you think @denis0x0D ?

denis0x0D: I do not think the "do while" is good choice here, because we have to try new connection only if we got "PP_LOST_CONNECTION" error.

dmlproc/dmlprocessor.cpp
denis0x0D: Fixed

mariadb-LeonidFedorov (approved):

drrtuy (approved):
untested ACK

PR #3121, opened 2024-02-09
MCOL-5241: incorrect AVG(YEAR(col)) result

mariadb-SergeyZefirov:
The problem was due to Field_new_decimal that receiving the results from our code had too small a precision to hold the value. This resulted in overflow capping and in an incorrect value in the result.

Changing the precision of the Field fixed this incorrect behavior.


mariadb-LeonidFedorov (approved):

mariadb-SergeyZefirov:
There are problems with regression sute, I have to work on them.

mariadb-SergeyZefirov:
This is completely on the side of MDB.

PR #3122, opened 2024-02-09
MCOL-5650: CMAPI python3 binary crash.

mariadb-AlanMologorsky:
- [x] [fix] python prebuilded interpreter to interpreter without AVX512 instructions


drrtuy (approved):
untested ACK

PR #3137, opened 2024-03-06
MCOL-5328: use PCRE2 and JPCRE wrapper

mariadb-LeonidFedorov:
to implement functions
```
regexp
regexp_substr
regexp_instr
regexp_replace
```


utils/funcexp/func_regexp.cpp
```diff
@@ -212,39 +215,126 @@ inline bool getBool(rowgroup::Row& row, funcexp::FunctionParm& pm, bool& isNull,
...
+/*
+  returns the string subject with all occurrences of the regular expression pattern replaced by
+  the string replace. If no occurrences are found, then subject is returned as is.
+  https://mariadb.com/kb/en/regexp_replace/
+*/
+std::string Func_regexp_replace::getStrVal(rowgroup::Row& row, FunctionParm& fp, bool& isNull,
+                                           execplan::CalpontSystemCatalog::ColType& ct)
+
+{
+  if (isNull)
```
mariadb-SergeyZefirov, line 301: Is this right? We test that something is NULL and exactly what is this something?

mariadb-SergeyZefirov (approved):
The C++ code looks good. I have only one question there.

The test may include something like ``REGEXP_REPLACE(text_field, 'q', 'KittyCat')`` to test that string with replacements expanded has all replacements correct. The code uses what appears a thoroughly tested implementation of ``replace``, this is just for us to be on safe side in case we change regexp implementation somewhere down the road.

utils/funcexp/func_regexp.cpp
mariadb-LeonidFedorov: ++

drrtuy (approved):
untested ACK

PR #3141, opened 2024-03-14
Fixes MCOL-5056, an error of joining TEXT column from InnoDB

mariadb-SergeyZefirov:
We incorrectly identified TEXT columns from external tables as BLOB. Alexander Barkov suggested a way to discriminate them which I implemented here.


mysql-test/columnstore/basic/t/MCOL-5056-text-is-not-blob.test
```diff
@@ -0,0 +1,34 @@
...
+--disable_warnings
+CREATE USER IF NOT EXISTS'cejuser'@'localhost' IDENTIFIED BY 'Vagrant1|0000001';
+--enable_warnings
+GRANT ALL PRIVILEGES ON *.* TO 'cejuser'@'localhost';
+FLUSH PRIVILEGES;
+
+CREATE TABLE foo5 (a TEXT);
+CREATE TABLE bar5 (b TEXT) ENGINE=COLUMNSTORE;
+# no data, just the fact that we can run this SELECT without errors.
+SELECT * FROM foo5, bar5 WHERE foo5.a=bar5.b;
```
drrtuy, line 30: There should be a test with multiple text key columns and mixed int/text key columns.

mariadb-SergeyZefirov: Can you elaborate on why?

I did the test that way because this particular type of column gets mistreated. I saw no other types that will get such mistreatment.

drrtuy: B/c of the mechanism that produces key for joining is fragile and previously demonstrates different behavior for keys types that are on different positions.

mariadb-SergeyZefirov: Again, can you provide an example of what can go wrong and how to achieve that? You tell that there is demonstrable problem, can you point to a case about that?

Where to look for a mechanism that produces keys for joining? If it is in ha_mcs_execplan, this patch cover all server Large OBjects (look at sql/sql_yacc.yy, field_type_log rule) cases, other cases were covered by VARCHAR handling.

mariadb-LeonidFedorov (approved):

mysql-test/columnstore/basic/t/MCOL-5056-text-is-not-blob.test
drrtuy: The case that I  was referring to happened about 2 years ago and happened in the join code. Based on this I suggest to extend the test list. The more tests we have the better.

mariadb-SergeyZefirov:
I close this PR. I moved my work on this problem here: https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/3160

PR #3142, opened 2024-03-14
MCOL-4480: TEXT type added

mariadb-LeonidFedorov:
the field `columnDefPtr->fType->fType` in `AlterTableProcessor::addColumn` is written as `enum DDL_DATATYPES` and was compared with values from `enum ColDataType` 
`CalpontSystemCatalog::TEXT` was missed in the same function and in two more or less similar cases


drrtuy (approved):
untested ACK

PR #3144, opened 2024-03-18
added dbrm_restore for LocalStorage & S3

mariadb-AllenHerrera:


extra/cs_package_manager.sh
```diff
@@ -253,8 +255,42 @@ check_package_managers() {
     fi;
 }
 
+check_mac_dependancies() {
```
mariadb-AlanMologorsky: Let's add  some docstring for this function.
Short description is enough IMO.

mariadb-AlanMologorsky: I'm curious why `dependAncy` not `dependEncy` in multiple places?

extra/cs_package_manager.sh
```diff
@@ -253,8 +255,42 @@ check_package_managers() {
...
+        echo "which ggrep"
+        echo ""
+        exit 1;
+    fi
+
+}
+
 check_operating_system() {
 
+    if [[ -e "/System/Library/CoreServices/SystemVersion.plist" ]]; then
```
mariadb-AlanMologorsky: How it should work on Mac OS?

extra/cs_package_manager.sh
```diff
@@ -917,13 +953,80 @@ do_install() {
...
+    os_options=("centos" "rhel" "rocky" "ubuntu" "debian")
+    select opt in "${os_options[@]}"; do
+        case $opt in
+            "centos" |  "rhel" | "rocky" )
+                distro_info=$opt
+                echo "What major version of $distro_info:"
+                short_options=("7" "8" "9")
+                select short in "${short_options[@]}"; do
+                         case $short in
+                            "7" | "8" | "9")
```
mariadb-AlanMologorsky, line 975: Here we potentially have wrong combinations like `centos 8`/`centos 9`/`rocky 7`, etc.

Could it potentially affect something.

extra/mcs_backup_manager.sh
```diff
@@ -2763,25 +2773,98 @@ validation_prechecks_for_dbrm_backup() {
...
             printf "\n[!] Failed testS3Connection\n\n"
             exit 1;
+        else
+            echo " - S3 Connection works"
         fi
     fi;
 
+    # Download cs_package_manager.sh if not exists
+    if [ ! -f "cs_package_manager.sh" ]; then
+        wget https://raw.githubusercontent.com/mariadb-corporation/mariadb-columnstore-engine/develop/extra/cs_package_manager.sh; chmod +x cs_package_manager.sh;
```
mariadb-AlanMologorsky: Conceptually we should pack and deliver this script with cmapi or engine packages.
I prefer cmapi.
I could add it to latest PR related to mcs cli backup/restore.

Is there any case when we don't have engine and cmapi packages installed but could run `mcs_backup_manager.sh`?

extra/mcs_backup_manager.sh
```diff
@@ -2763,25 +2773,98 @@ validation_prechecks_for_dbrm_backup() {
...
+        else
+            printf "[!] No files found in the DBRM backup directory\n"
+            printf "Path: ${backup_location}/${backup_folder_to_restore_dbrms}\n\n"
+            exit 2;
+        fi
     fi
 
     # Confirm bucket connection
     if [ "$storage" == "S3" ]; then
         if ! testS3Connection 1>/dev/null 2>/dev/null; then
```
mariadb-AlanMologorsky: JFYI It's a good practice to build conditions starting from positive.
It feels more natural and readable.
EG
```
if something; then
    echo "It's true"
else
    echo "Not true"
    exit 1;
fi
```
Compare to
```
if ! something; then
    echo "Not true"
    exit 1;
else
    echo "It's true"
fi
```

extra/mcs_backup_manager.sh
```diff
@@ -2844,62 +2930,248 @@ process_dbrm_backup() {
     if ! $quiet; then printf "[+] Complete\n\n"; fi;
 }
 
-# for next release
+confirm_cmapi_online_and_configured() {
+    cmapi_current_status=$(mcs cmapi is-ready 2> /dev/null);
+    if [ $? -ne 0 ]; then
+
+        # if cmapi is not online - check systemd is running and start cmapi
+            if [ "$(ps -p 1 -o comm=)" = "systemd" ]; then
```
mariadb-AlanMologorsky: Wrong indentation.

extra/mcs_backup_manager.sh
```diff
@@ -2844,62 +2930,248 @@ process_dbrm_backup() {
...
+            eval $cmapi_start_command
+            sleep 2;
```
mariadb-AlanMologorsky: Probably could be not enough for cmapi becomes ready.
Take a look at https://github.com/mariadb-corporation/mariadb-columnstore-engine/blob/develop/cmapi/check_ready.sh

extra/mcs_backup_manager.sh
```diff
@@ -2844,62 +2930,248 @@ process_dbrm_backup() {
...
+            echo "mcs cmapi is-ready"
+            exit 1
+        fi
+    fi
+
+    confirm_cmapi_configured
+}
+
+# currently supports singlenode only
+confirm_cmapi_configured() {
```
mariadb-AlanMologorsky: [Just about naming] Way better to name like `confirm_nodes_configured`.

extra/cs_package_manager.sh
mariadb-AllenHerrera: worst case the user gets 0 results if they self pick it.  ill add a comment in the code
Was looking for a quick way when a mac user runs "cs_package_manager.sh check" since theres no /etc/os-release to auto detect what OS / version to search the mariadb repos. not perfect as this will require update in the future when new OS/versions are released.
Ideally theres an easy endpoint or list to get from our downloads site of the OS, version combinations that we support as opposed to parsing the download links

extra/cs_package_manager.sh
mariadb-AllenHerrera: was what i found at the time, i'll change it to 
```
if [[ $OSTYPE == 'darwin'* ]]; then
  echo 'macOS'
fi
```

reminder, on mac the only call supported is "cs_package_manager check" to find the version mappings between mariadb server and columnstore

extra/mcs_backup_manager.sh
mariadb-AllenHerrera: cs_package_manager.sh is in the engine repo, but not installed/packaged.
I guess since cmapi now bundles mcs_backup_manager.sh, it should also bundle cs_package_manager.sh.
Yes both scripts use each other for some functions

extra/mcs_backup_manager.sh
mariadb-AllenHerrera: for now im writing a function cmapi_check_ready thats a small augmentation of that script, let me know if you think theres a more maintainable/better way to import it because i could technically run it as it with the function piping to dev/null

mariadb-AlanMologorsky (approved):
LGTM. 
BUT
Still have some unresolved tasks related to this PR.
1. Pack and deliver cs_package_manager.sh https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/3144/files#r1538107358 
Should be implemented after merging Allen's PRs.
2. cmapi_check_ready() augmentation and import for check_ready.sh should be implemented
3. It's not clear for me, should script check if there installed cmapi or not on a system? Any cases when we don't have cmapi installed but have script on board (except manual delivering)?

Will post it in slcak too.

mariadb-AllenHerrera:
Posting link to my response
https://mariadb.slack.com/archives/G2CHFHD2L/p1717765724825959?thread_ts=1717756120.639929&cid=G2CHFHD2L

PR #3145, opened 2024-03-19
Fix/mcol 5691 distinct aggregate disk based

drrtuy:


dbcon/joblist/tupleaggregatestep.cpp
```diff
@@ -5728,116 +5742,147 @@ uint64_t TupleAggregateStep::doThreadedAggregate(ByteStream& bs, RowGroupDL* dlp
...
+          // easy way of multi-threading this and it's done in a single thread for now.
+          for (uint32_t bucketNum = 0; bucketNum < fNumOfBuckets; bucketNum++)
+          {
+            if (fEndOfResult == false)
+            {
+              fAggregator->append(fAggregators[bucketNum].get());
+            }
+          }
+        }
       }
```
mariadb-LeonidFedorov: looks the whole block could be simplified to
```
else if (hasGroupByColumns && !fEndOfResult && !fDoneAggregate) 
{
    for (uint32_t bucketNum = 0; bucketNum < fNumOfBuckets; bucketNum++)
    {
         fAggregator->append(fAggregators[bucketNum].get());
    }
}
```
please apply clang-format as well, curly braces style is inconsistent here

dbcon/joblist/tupleaggregatestep.cpp
```diff
@@ -5912,6 +5951,25 @@ uint64_t TupleAggregateStep::doThreadedAggregate(ByteStream& bs, RowGroupDL* dlp
...
+  }
+  else
+  {
+    bs.restart();
+    fRowGroupDelivered.serializeRGData(bs);
+    return false;
+  }
+  return true;
+}
+
```
mariadb-LeonidFedorov: NIT
I'd prefer early return style, _elses are evil_

```
  if (dlp)
  {
    RGData rgData = fRowGroupDelivered.duplicate();
    dlp->insert(rgData);
    return true;
  }
  bs.restart();
  fRowGroupDelivered.serializeRGData(bs);
  return false;
  }

File: utils/rowgroup/rowstorage.cpp [hunk 3]
```diff
@@ -666,6 +666,156 @@ class RowGroupStorage
...
+
+    while (!fRGDatas.empty())
+    {
+      uint64_t rgid = fRGDatas.size() - 1;
+      rgdata = std::move(fRGDatas[rgid]);
+      fRGDatas.pop_back();
+
+      // Calculate from first and last uint64_t entry in fFinalizedRows BitMap
+      // which contains information about rows in the RGData.
+      uint64_t fgid = rgid * fMaxRows / 64;
```
mariadb-LeonidFedorov: magic number, I'd replace with constant or sizeof expression

utils/rowgroup/rowstorage.cpp
```diff
@@ -666,6 +666,156 @@ class RowGroupStorage
...
+      if (fFinalizedRows.size() <= fgid)
+      {
+        // There are no finalized rows in this RGData. We can just return it.
+        // Load from disk if necessary and unlink DumpFile.
+        if (!rgdata)
+        {
+          loadRG(rgid, rgdata, true);
+        }
+        return true;
+      }
```
mariadb-LeonidFedorov, line 774: the block could be simplified for [Cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity)
```
      if (fFinalizedRows.size() <= fgid && !rgdata)
      {
        loadRG(rgid, rgdata, true);
        return true;
      }
```

utils/rowgroup/rowstorage.cpp
```diff
@@ -666,6 +666,156 @@ class RowGroupStorage
...
+          // Load RGData from disk, unlink dump file and continue processing
+          loadRG(rgid, rgdata, true);
+        }
+        else
+        {
+          // All rows are finalized. Unlink dump file and continue search for return RGData
+          unlink(makeRGFilename(rgid).c_str());
+          continue;
+        }
+      }
```
mariadb-LeonidFedorov, line 813: no need for else and for inner if here as well
```
  if (rgdata && !hasReturnRows)
  {
    continue;
  }
  // Load RGData from disk, unlink dump file and continue processing
  loadRG(rgid, rgdata, true);
       
  if (!hasReturnRows)
  {
   // All rows are finalized. Unlink dump file and continue search for return RGData
    unlink(makeRGFilename(rgid).c_str());
    continue;
  }
```

dbcon/joblist/tupleaggregatestep.cpp
drrtuy: Can't group all conditions together b/c there is a tail that happens even if (!fEndOfResult && !fDoneAggregate) is false. Can group two of these conds together though.

utils/rowgroup/rowstorage.cpp
```diff
@@ -666,6 +666,156 @@ class RowGroupStorage
...
+      // TODO: Finish comments (make statements out of assumptions)
+      // assumption: this shifts data within RGData such that it compacts the non finalized rows
+      // move next non-finalized rows on finalized row positions
+      //
+      uint64_t pos = 0;
+      uint64_t opos = 0;
+      fRowGroupOut->setData(rgdata.get());
+      for (auto i = fgid; i < tgid; ++i)
+      {
+        if ((i - fgid) * 64 >= fRowGroupOut->getRowCount())
```
mariadb-LeonidFedorov: magic number 64

utils/rowgroup/rowstorage.cpp
```diff
@@ -666,6 +666,156 @@ class RowGroupStorage
...
+          continue;
+        }
+
+        if (mask == 0)
+          continue;
+
+        while (mask != 0)
+        {
+          // assmp.: find position until block full of not finalized rows?
+          size_t b = __builtin_ffsll(mask);
```
mariadb-LeonidFedorov: I'd like to see intrinsics in isolated fuctions

utils/rowgroup/rowstorage.cpp
```diff
@@ -666,6 +666,156 @@ class RowGroupStorage
...
+
+      // set RGData with number of not finalized rows which have been compacted at front of RGData
+      fRowGroupOut->setData(rgdata.get());
+      fRowGroupOut->setRowCount(pos);
+      int64_t memSz = fRowGroupOut->getSizeWithStrings(fMaxRows);
+
+      // TODO: why do we release the memory if we just loaded something?, or are we realesing for file
+      // managment?
+      fMM->release(memSz);
+      unlink(makeRGFilename(rgid).c_str());
```
mariadb-LeonidFedorov: That's not production ready speculation

dbcon/joblist/tupleaggregatestep.cpp
drrtuy: Fixed

utils/rowgroup/rowstorage.cpp
drrtuy: It was like this before but I will make it a const.

utils/rowgroup/rowstorage.cpp
```diff
@@ -666,6 +666,156 @@ class RowGroupStorage
...
+      }
+
+      // assmp.:: nothing got shifted at all -> all rows must be finalized? (but shouldn't it be the
+      // other way around?) if all rows finalized remove RGData and file and don't give it out why would
+      // we check this here again? see: fRowGroupOut->setRowCount(pos); -> pos is the number of valid
+      // rows, interesting
+      if (pos == 0)
+      {
+        fLRU->remove(rgid);
+        unlink(makeRGFilename(rgid).c_str());
```
mariadb-LeonidFedorov, line 822: probably unlinks should be done in RAII manner

utils/rowgroup/rowstorage.cpp
```diff
@@ -1748,6 +1898,44 @@ std::unique_ptr<RGData> RowAggStorage::getNextRGData()
...
+  if (!fStorage)
+  {
+    return {};
+  }
+
+  cleanup();
+  freeData();
+
+  // fGeneration is an unsigned int, we need a signed int for a comparison >= 0
+  int32_t gen = fGeneration;
```
mariadb-LeonidFedorov, line 1913: const

utils/rowgroup/rowstorage.cpp
drrtuy: What will happen if `fFinalizedRows.size() <= fgid` is true but `!rgdata` is not? It seems this suggestions alters the control flow.

mariadb-LeonidFedorov:
Mostly all the comments are on the code style and the functions complexity. 
all are NIT's here, with except of intrinsics  and unlink usage.
clang format should be applied for sure

mariadb-LeonidFedorov (approved):

utils/rowgroup/rowstorage.cpp
drrtuy: Looks like `rgdata && !hasReturnRows` doesn't cover the case when `rgdata` is true but `hasReturnRows` is false. In this case the current code does nothing yours will execute the previous else block.

dbcon/joblist/tupleaggregatestep.cpp
```diff
@@ -5696,14 +5697,27 @@ void TupleAggregateStep::doAggregate()
   return;
 }
 
+/** @brief Aggregate input row groups in two-phase multi-threaded aggregation.
+ * In second phase handle three different aggregation cases differently:
```
phoeinx, line 5701:
```suggestion
 * In second phase handle three aggregation cases differently:
```

dbcon/joblist/tupleaggregatestep.cpp
```diff
@@ -5728,116 +5742,144 @@ uint64_t TupleAggregateStep::doThreadedAggregate(ByteStream& bs, RowGroupDL* dlp
...
+          }
+          done = true;
+        }
+        if (done)
+          fEndOfResult = true;
+      }
+    }
+    // CASE 3: Query contains no aggregation on a DISTINCT column, but at least one GROUP BY column
+    // e.g. SELECT SUM(col1) FROM test GROUP BY col2;
+    // As the aggregation algorithm of first phase calculates aggregation results online, e.g. for SUM()
```
phoeinx: I think this comment is a left-over from what I first thought to be true (being able to just deliver the row groups), before I realised the generational nature of fAggregator. I think that the comment from line 5872 is correct and should probably be moved up here?

utils/rowgroup/rowaggregation.cpp
```diff
@@ -4554,19 +4570,31 @@ void RowAggregationDistinct::addRowGroup(const RowGroup* pRows,
 //------------------------------------------------------------------------------
 void RowAggregationDistinct::doDistinctAggregation()
 {
-  while (dynamic_cast<RowAggregationUM*>(fAggregator.get())->nextRowGroup())
+  fAggregator->finalAggregation();  // TODO: Perhaps move somewhere else?
```
phoeinx: Not sure if that TODO should stay in. If I remember correctly I added it, because I first tried putting the `fAggregator->finalAggregation()` line into the `doThreadedAggregate()` function before `doDistinctAggregation()` gets called there, but it somehow didn't work because it wasn't able to access the correct fAggregator object in that place. 
I wasn't able to figure out why though and it puzzled me.

utils/rowgroup/rowstorage.cpp
phoeinx: Thank you for your comment! What do you mean by RAll manner? (Sorry, I am probably missing vocabulary here.)

drrtuy: RAII in this case means we call a hypotetical `AggTempFile` struct ctor that going out of scope call `unlink` in its dtor.

dbcon/joblist/tupleaggregatestep.cpp
drrtuy: Good point @phoeinx

utils/rowgroup/rowaggregation.cpp
drrtuy: Moved this method into `doThreadedAggregate()` to give this idea a try.

PR #3148, opened 2024-03-21
MCOL-5597 Rollback changes

denis0x0D:


mariadb-SergeyZefirov (approved):
I approve these changes in general, as a palliative measue. This is the best we can do in our current circumstances.

utils/joiner/joinpartition.cpp
```diff
@@ -481,8 +486,8 @@ int64_t JoinPartition::processSmallBuffer(RGData& rgData)
 
     ret = writeByteStream(0, bs);
 
-    if (rg.getRowCount())
-      htSizeEstimate += rg.getDataSize();
+    // FIXME(MCOL-5597): Properly calculate the size of the bucket.
```
drrtuy, line 489: Are you sure this is the properly calculated size of the RG ? 
I mean the wording doesn't express the fact that Stringstore size is ignored. I suggest to elaborate on the fact that this is a heuristict and we ignore the Stringstore b/c the tests tells us to do so.

drrtuy (changes requested):
Plz comment on the note I left in the patch.

utils/joiner/joinpartition.cpp
denis0x0D: Right, it's not a proper calculation and tag "FIXME" is added to specify this.
The reason of this patch  - @mariadb-LeonidFedorov asked to apply rollback based on @mariadb-SergeyZefirov research to pass 5597

drrtuy: Ah, I misinterpreted `FIXME` part. 
Could you describe this FIXME in more words?

drrtuy (approved):
untested ACK

PR #3149, opened 2024-03-21
chore(arrow) bump apache arrow version and fix test load generator for memory usage

mariadb-LeonidFedorov:
parquet test generator collected all the test data in memory and then dumped it to the file, rewritten to block stream style, that does not affect performance and limits memory usage
Apache arrow version dumped to the latest release and build flags added to limit SIMD usage by SSE4.2


drrtuy (approved):
untested ACK

PR #3152, opened 2024-03-25
chore(build) fix boost version for arrow library

mariadb-LeonidFedorov:
boost 1.81 was repacked and sha256 sum was changed, columnstore suffered from this inappropriate action before and this is the second time, arrow using it's own internal dependencies is broken the same way
```
       file='/mdb/verylongdirnameforverystrangecpackbehavior/storage/columnstore/columnstore/external/arrow/bin/boost_ep-prefix/src/boost_1_81_0.tar.gz'

-- SHA256 hash of
    /mdb/verylongdirnameforverystrangecpackbehavior/storage/columnstore/columnstore/external/arrow/bin/boost_ep-prefix/src/boost_1_81_0.tar.gz
  does not match expected value
    expected: '9e0ffae35528c35f90468997bc8d99500bf179cbae355415a89a600c38e13574'
    actual: '205666dea9f6a7cfed87c7a6dfbeb52a2c1b9de55712c9c1a87735d7181452b6'
-- Hash mismatch, removing...
```


drrtuy (approved):
untested ACK

PR #3153, opened 2024-03-26
Minor fix in timeout for /notfound "endpoint" call.

mariadb-AlanMologorsky:
- [x] [fix] timeout to urlopen /notfound "endpoint" in mcs-savebrm and mcs-loadbrm


drrtuy (changes requested):
Named constant instead of a literal @mariadb-AlanMologorsky ?

drrtuy (approved):
untested ACK

PR #3159, opened 2024-03-27
Fixes MCOL-5700, Oracle mode test results

mariadb-SergeyZefirov:
This changeset contains fixes in Oracle mode tests and for the implementation of the CONCAT_ORACLE. Also, we harmonise our translation process with the recent changes in the server.

Due to changed behavior of the server, some CREATE VIEW/EXPLAIN statements' results begun to output unexpected results and need to be fixed.

Also, concatenation operation's name also changed. This lead to disabled func_concat_oracle test to be enabled to test it and it turned out that our implementation of this function was broken and need to be fixed too.


utils/funcexp/func_concat_oracle.cpp
```diff
@@ -46,25 +46,21 @@ string Func_concat_oracle::getStrVal(Row& row, FunctionParm& parm, bool& isNull,
...
-    isNull = false;
-  }
+  // we default to true as any non-NULL operand will reset it to false
+  // and all NULL operands will leave it as true, which is intended.
+  isNull = true;
   // TODO: do a better job of cutting down the number re-allocations.
   // look at Item_func_concat::realloc_result for ideas and use
   // std::string:resize() appropriatly.
-  for (unsigned int id = 1; id < parm.size(); id++)
+  for (unsigned int id = 0; id < parm.size(); id++)
```
drrtuy, line 55: Really. The code and the author ignored the first param previously or starting id was set to 1 for a reason?

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -3883,15 +3883,23 @@ ReturnedColumn* buildFunctionColumn(Item_func* ifp, gp_walk_info& gwi, bool& non
   cal_connection_info* ci = static_cast<cal_connection_info*>(get_fe_conn_info_ptr());
 
   string funcName = ifp->func_name();
-  const Schema* funcSchema = ifp->schema();
-  if (funcSchema)
+  if ( nullptr != dynamic_cast<Item_func_concat_operator_oracle*>(ifp))
```
drrtuy, line 3886: If the idea is to check whether `ifp` is `Item_func_concat_operator_oracle` and not its child class I would suggest to use `typeid` here.

drrtuy (commented):
Plz comment back on the comments.

utils/funcexp/func_concat_oracle.cpp
mariadb-SergeyZefirov: Previously the handling of the first parameter was out of loop but with the same logic as applied to parameters handled inside the loop. There are deleted lines just above, take a look.

This is why I moved all handling inside the loop.

And I changed logics because ours was not conformant to what InnoDB does.

dbcon/mysql/ha_mcs_execplan.cpp
mariadb-SergeyZefirov: Server uses dynamic_cast and does not use typeid. The only place we use typeid is in isUDFSumItem function (BTW, it should use dynamic cast to Item_udf_sum, in my opinion). In all other places we use static casts of server's Item_XXX objects and dynamic casts for our objects.

I think I will stand by the use of dynamic_cast.

drrtuy: Server treats CPP as C in general and we are not, so IMHO following the server in this case is inefficient. Typeid is faster than dynamic_cast() b/c it bypasses the classes hierarchy. 
Plz find the list of files where `typeid` is used in the code.
Снимок экрана от 2024-04-02 11-06-43
This suggestion is not obligatory though.

drrtuy (approved):
untested ACK

PR #3160, opened 2024-03-28
Fixes MCOL-5056, an error of joining TEXT column from InnoDB

mariadb-SergeyZefirov:
We incorrectly identified TEXT columns from external tables as BLOB. Alexander Barkov suggested a way to discriminate them which I implemented here.


mariadb-LeonidFedorov (approved):

drrtuy (approved):
untested ACK

PR #3162, opened 2024-03-28
MCOL-5236 Take `Item` from `Ref_Item` for group by list.

denis0x0D:


dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -4387,7 +4396,8 @@ ReturnedColumn* buildFunctionColumn(Item_func* ifp, gp_walk_info& gwi, bool& non
 
     // A few functions use a different collation than that found in
     // the base ifp class
-    if (funcName == "locate" || funcName == "find_in_set" || funcName == "strcmp" || funcName == "regexp_instr")
+    if (funcName == "locate" || funcName == "find_in_set" || funcName == "strcmp" ||
```
mariadb-SergeyZefirov, line 4399: https://mariadb.com/kb/en/regexp_instr/ - REGEXP_INSTR(subject, pattern)

https://mariadb.com/kb/en/find_in_set/ - FIND_IN_SET(pattern, strlist)

These two have different position of "pattern" argument.

mariadb-SergeyZefirov (changes requested):
I have a question on the validity of string function substitution. It does not seem right to me and I would like to see comment in the code, explaining why it is correct if it is, in fact, correct.

dbcon/mysql/ha_mcs_execplan.cpp
denis0x0D: @mariadb-SergeyZefirov I'm not quite understand a question. The change introduced in this patch for code line you commented is just a code formatting, which was applied by clang-format tool automatically based on clang-format config we have. 
In details: the expression `funcName == "regexp_instr"` was moved to line below, because we have limit for number of characters in one line.

mariadb-SergeyZefirov: Sorry, I misread equality check as an assignment of the funcName.

mariadb-SergeyZefirov (approved):

mariadb-SergeyZefirov:
Thus I approved this changeset.

drrtuy (approved):
untested ACK

drrtuy:
I think @denis0x0D you should make a clone of this for stable-23.10.

PR #3171, opened 2024-04-25
MCOL-5715 Changes the second phase aggregation.

denis0x0D:
This patch changes the second phase aggregation pipeline - takes into account current memory consumption.


drrtuy (approved):
untested ACK

PR #3172, opened 2024-05-01
Docker multinode provision script expects container names in the cli

mariadb-SusilBehera:
For multi-node setup, docker provision [script](https://github.com/mariadb-corporation/mariadb-columnstore-docker/blob/86c9c28ebef027c18d56d0236bdbb43b6c5c353d/scripts/provision)  expects all container names to be passed in the cli, otherwise it only provisions the localhost (the present behavior). So, pass container names in the provision script cli.


drrtuy (approved):
untestedACK

PR #3174, opened 2024-05-09
Correct handling of utf8 charsets and collations in DDL

mariadb-SergeyZefirov:
This patch makes our DDL handling to properly handle contents of ``OLD_MODE`` environment variable, namely, the ``UTF8_IS_UTF8MB3`` setting.

The test is made after InnoDB's results and we are doing what server does.


mariadb-LeonidFedorov (approved):

drrtuy (approved):
Untested ACK

PR #3177, opened 2024-05-13
chore(build) add Ubuntu 24.04 to the CI

mariadb-LeonidFedorov:


.drone.jsonnet
```diff
@@ -143,7 +153,7 @@ local Pipeline(branch, platform, event, arch='amd64', server='10.6-enterprise')
   local etc_path = '/etc/columnstore',
   local socket_path = if (pkg_format == 'rpm') then '/var/lib/mysql/mysql.sock' else '/run/mysqld/mysqld.sock',
   local config_path_prefix = if (pkg_format == 'rpm') then '/etc/my.cnf.d/' else '/etc/mysql/mariadb.conf.d/50-',
-  local img = if (platform == 'centos:7' || platform == 'rockylinux:8') then platform else 'romcheck/' + std.strReplace(platform, '/', '-'),
+  local img = if (platform == 'centos:7' || platform == 'rockylinux:8') then platform else 'detravi/' + std.strReplace(platform, '/', '-'),
```
mariadb-AlanMologorsky: May be better to make some special user, eg "columnstore" or "columnstore_img" for such a thing.
Then move all images there one by one.

mariadb-LeonidFedorov: You are perfectly right. After the brief communication, it happens the only thing I can do, just substitute obsolete credentials with my own, actual one.

I'm fond of your perfect eye, but we have no resources to escape it, and we have to match new policies, that were never discussed with us. 

Nothing new, right?

mariadb-AlanMologorsky: Nothing new(
Let's make a task on it ;)
And anyway we need only man hours on it =)

timofeyturenko (approved):

PR #3178, opened 2024-05-14
MCOL-5747 gcc-14.1.1 compile error - calloc - transposed args

grooverdan:
The arguments of calloc are the number of members and the sizeof the member. Gcc-14.1.1 worked out how to tell the difference.

We correct this by transposing to gcc's will.

### DIFF
--- a/writeengine/wrapper/writeengine.cpp
+++ b/writeengine/wrapper/writeengine.cpp
@@ -1099,8 +1099,8 @@ inline void allocateValArray(void*& valArray, ColTupleList::size_type totalRow, 
     case WriteEngine::WR_VARBINARY:  // treat same as char for now
     case WriteEngine::WR_CHAR:
     case WriteEngine::WR_BLOB:
-    case WriteEngine::WR_TEXT: valArray = calloc(sizeof(char), totalRow * MAX_COLUMN_BOUNDARY); break;
-    case WriteEngine::WR_TOKEN: valArray = calloc(sizeof(Token), totalRow); break;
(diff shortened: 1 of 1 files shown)


drrtuy:
Hi @grooverdan .
Thx for the patch. Could you elaborate what is the minimal MDB version do you want to get this patch into?

drrtuy (approved):
untested ACK

writeengine/wrapper/writeengine.cpp
```diff
@@ -1099,8 +1099,8 @@ inline void allocateValArray(void*& valArray, ColTupleList::size_type totalRow,
     case WriteEngine::WR_VARBINARY:  // treat same as char for now
     case WriteEngine::WR_CHAR:
     case WriteEngine::WR_BLOB:
-    case WriteEngine::WR_TEXT: valArray = calloc(sizeof(char), totalRow * MAX_COLUMN_BOUNDARY); break;
-    case WriteEngine::WR_TOKEN: valArray = calloc(sizeof(Token), totalRow); break;
+    case WriteEngine::WR_TEXT: valArray = calloc(totalRow * MAX_COLUMN_BOUNDARY, sizeof(char)); break;
+    case WriteEngine::WR_TOKEN: valArray = calloc(totalRows, sizeof(Token)); break;
```
mariadb-LeonidFedorov: there is a typo here `totalRows` should be replaced with `totalRow`

grooverdan: yep, corrected.

grooverdan:
> Hi @grooverdan . Thx for the patch. Could you elaborate what is the minimal MDB version do you want to get this patch into?

At the moment 11.1 as its currently only Fedora 40 that has such a new gcc version . I'm not sure why 10.11 is building CS ok.

PR #3194, opened 2024-05-30
Mcol 4234 order by group by mishap

mariadb-SergeyZefirov:
This is a fix for a long (before 2016) standing Columnstore's deficiency in handling GROUP BY and ORDER BY and SELECT expressions that are not aggregates and not a references to GROUP BY columns.

As other engines from MariaDB's family (and even SLite) can handle such cases, this fix brings MCS closer to server's behavior.

The fix, itself, adds a wrapping of expressions into a dummy aggregate and looses checks of what can go into ORDER BY expression list.

Also, it changes some of the tests and appropriate test results to reflect new functionality.

### DIFF
--- a/dbcon/mysql/ha_mcs_execplan.cpp
+++ b/dbcon/mysql/ha_mcs_execplan.cpp
@@ -474,6 +474,22 @@ bool sortItemIsInGrouping(Item* sort_item, ORDER* groupcol)
   if (sort_item->type() == Item::SUM_FUNC_ITEM)
   {
     found = true;
+    return found;
+  }
+
+  {
+    // as we now can warp ORDER BY or SELECT expression into
+    // an aggregate, we can pass FIELD_ITEM as "found" as well.
+    Item* item = sort_item;
+    while (item->type() == Item::REF_ITEM)
+    {
+      const Item_ref* ref_item = static_cast<const Item_ref*>(item);
+      item = (Item*)*ref_item->ref;
+    }
+    if (item->type() == Item::FIELD_ITEM || item->type() == Item::CONST_ITEM || item->type() == Item::NULL_ITEM)
+    {
+      return true;
+    }
   }
 
   // A function that contains an aggregate function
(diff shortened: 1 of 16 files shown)


mariadb-SergeyZefirov:
https://github.com/mariadb-corporation/mariadb-columnstore-regression-test/pull/363 - regression suite changes for this PR.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -474,6 +474,22 @@ bool sortItemIsInGrouping(Item* sort_item, ORDER* groupcol)
...
+  {
+    // as we now can warp ORDER BY or SELECT experssion into
+    // an aggregate, we can pass FIELD_ITEM as "found" as well.
+    Item* item = sort_item;
+    while (item->type() == Item::REF_ITEM)
+    {
+      const Item_ref* ref_item = static_cast<const Item_ref*>(item);
+      item = (Item*)*ref_item->ref;
+    }
+    if (item->type() == Item::FIELD_ITEM || item->type() == Item::CONST_ITEM || item->type() == Item::NULL_ITEM)
```
drrtuy, line 489: Can we got here other types, e.g. Item::FUNC_ITEM ?

utils/rowgroup/rowaggregation.cpp
```diff
@@ -1254,6 +1254,111 @@ void RowAggregation::doMinMax(const Row& rowIn, int64_t colIn, int64_t colOut, i
     }
   }
 }
+//------------------------------------------------------------------------------
+// Update the min/max fields if input is not null.
```
drrtuy: I might be mistaken but `min/max` looks like a typo here.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -474,6 +474,22 @@ bool sortItemIsInGrouping(Item* sort_item, ORDER* groupcol)
   if (sort_item->type() == Item::SUM_FUNC_ITEM)
   {
     found = true;
+    return found;
+  }
+
+  {
+    // as we now can warp ORDER BY or SELECT experssion into
```
drrtuy: `experssion` looks like a typo here.

drrtuy (approved):
There are number of typos and a question regarding an invariant. 
untested ACK

dbcon/mysql/ha_mcs_execplan.cpp
mariadb-SergeyZefirov: We can, but analyzing them would complicate the solution.

One of such cases is a function of an aggregate - if we wrap it into an aggregate, it can trigger completely unrelated and puzzling error ("invalid use of group by" or something like that).

So, to not jump into a rabbit hole, I decided to play safe.

mariadb-LeonidFedorov (approved):

dbcon/mysql/ha_mcs_execplan.cpp
mariadb-SergeyZefirov: And, regarding the FUNC_ITEM, if it cannot be processed, it will trigger internal error elsewhere in joblist part. But proper fix for that is lengthy.

PR #3202, opened 2024-06-11
MCOL-5708 Calculate precision and scale for constant decimal.

denis0x0D:
This patch calculates precision and scale for constant decimal value for SUM aggregation function.

### DIFF
--- a/datatypes/mcs_datatype.cpp
+++ b/datatypes/mcs_datatype.cpp
@@ -66,6 +66,41 @@ int128_t TypeAttributesStd::decimal128FromString(const utils::NullString& value,
   return decimal128FromString(value.unsafeStringRef(), saturate);
 }
 
+// SQL parser checks that given `value` is in a valid format.
+// The first symbol can be `-`. The `value` can contain `.` symbol.
+void decimalPrecisionAndScale(const utils::NullString& value, int& precision, int& scale)
+{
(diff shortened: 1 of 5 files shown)


mariadb-LeonidFedorov (approved):

datatypes/mcs_datatype.cpp
```diff
@@ -66,6 +66,41 @@ int128_t TypeAttributesStd::decimal128FromString(const utils::NullString& value,
   return decimal128FromString(value.unsafeStringRef(), saturate);
 }
 
+// SQL parser checks that given `value` is in a valid format.
+// The first symbol can be `-`. The `value` can contain `.` symbol.
+void decimalPrecisionAndScale(const utils::NullString& value, int& precision, int& scale)
```
drrtuy, line 71: IMHO it is better to return std::pair<precision, scale> from the f() b/c one doesn't need to watch out side effects that happens with refs inside `decimalPrecisionAndScale`

drrtuy (commented):
I don't mind the changes but I would better change the iface exposed by the introduced `decimalPrecisionAndScale`.

denis0x0D:
@drrtuy I see your points, let's merge to develop and test the patch first, since we have a plan to deliver it in upcoming release, if it works fine, I can update function interface on your request.

PR #3209, opened 2024-06-20
Containers memory limits for CI [stable-23.10]

mariadb-LeonidFedorov:
Limit test containers by memory, fix cgroup path inside the containers by introducing new ugly setting name 

---------


drrtuy (approved):
untested ACK

PR #3211, opened 2024-06-24
MCOL-5771 fixed -spoll for newer versions and config-file defaults

mariadb-AllenHerrera:
fix for skip_mdb when running localstorage restore

fixes for s3 backups 
* remove redundant eval and piping

fixes for cmapi config file detection
* would throw errors if cmapi not installed

Adjusted some white space around function names and output formatting


mariadb-AlanMologorsky (approved):
LGTM

PR #3213, opened 2024-06-25
MCOL-5454: Self-signed certificate autorenew.

mariadb-AlanMologorsky:
- [x] [add] managers/certificate.py with CertificateManger class
- [x] [mv] creating self-signed certificate logic into CertificateManger class
- [x] [add] renew and days_before_expire methods to CertificateManger class
- [x] [mv] several certificate dependent constants to managers/certificate.py
- [x] [add] CherryPy BackgroundTask to invoke certificate check hourly (3600 secs)
- [x] [fix] tests
- [x] [fix] bug with txn timer clean (clean_txn_by_timeout, worker and invoking of it)

### DIFF
--- a/cmapi/cmapi_server/__main__.py
+++ b/cmapi/cmapi_server/__main__.py
@@ -8,15 +8,10 @@ import logging
 import os
 import threading
 import time
-from datetime import datetime, timedelta
+from datetime import datetime
 
 import cherrypy
 from cherrypy.process import plugins
-from cryptography import x509
-from cryptography.hazmat.backends import default_backend
-from cryptography.hazmat.primitives import serialization, hashes
-from cryptography.hazmat.primitives.asymmetric import rsa
-from cryptography.x509.oid import NameOID
 
 # TODO: fix dispatcher choose logic because code executing in endpoints.py
 #       while import process, this cause module logger misconfiguration
@@ -27,28 +22,26 @@ from cmapi_server import helpers
 from cmapi_server.constants import DEFAULT_MCS_CONF_PATH, CMAPI_CONF_PATH
 from cmapi_server.controllers.dispatcher import dispatcher, jsonify_error
 from cmapi_server.failover_agent import FailoverAgent
-from cmapi_server.managers.process import MCSProcessManager
 from cmapi_server.managers.application import AppManager
+from cmapi_server.managers.process import MCSProcessManager
+from cmapi_server.managers.certificate import CertificateManager
 from failover.node_monitor import NodeMonitor
 from mcs_node_control.models.dbrm_socket import SOCK_TIMEOUT, DBRMSocketHandler
 from mcs_node_control.models.node_config import NodeConfig
 
 
-cert_filename = './cmapi_server/self-signed.crt'
-
-
-def worker():
+def worker(app):
     """Background Timer that runs clean_txn_by_timeout() every 5 seconds
     TODO: this needs to be fixed/optimized. I don't like creating the thread
     repeatedly.
     """
     while True:
-        t = threading.Timer(5.0, clean_txn_by_timeout)
+        t = threading.Timer(5.0, clean_txn_by_timeout, app)
         t.start()
         t.join()
 
 
-def clean_txn_by_timeout():
+def clean_txn_by_timeout(app):
     txn_section = app.config.get('txn', None)
     timeout_timestamp = txn_section.get('timeout') if txn_section is not None else None
     current_timestamp = int(datetime.now().timestamp())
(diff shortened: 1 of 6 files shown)


mariadb-AlanMologorsky:
@drrtuy Pls pay attention that I found 4 yo bug that means that clean txn by timeout never properly works. I have no idea how it works somehow, may be I missed something.

cmapi/cmapi_server/managers/certificate.py
```diff
@@ -0,0 +1,106 @@
...
+        return days_before_expire
+
+    @staticmethod
+    def renew_certificate() -> None:
+        """Creates new self signed certificate.
+        
+        Creates self-signed cetificate if certificate doesn't exist or
+        expires in a 1 day or less.
+        """
+        CertificateManager.create_self_signed_certificate_if_not_exist()
```
drrtuy: I might be too strict but "create_self_signed_certificate_if_not_exist" shouldn't be a part of "renew" and you should explicitly call it before "renew". Otherwise there is an implicit call that a reader must find on his own.

drrtuy: Now I see that you decided to use this method as a guard. I don't like this approach personally but I am OK with it, so plz disregard my prev comment on this line.

cmapi/cmapi_server/__main__.py
```diff
@@ -224,6 +167,10 @@ def create_self_signed_certificate():
     # subscribe FailoverBackgroundThread plugin code to bus channels
     # code below not starting "real" failover background thread
     FailoverBackgroundThread(cherrypy.engine, turn_on_failover).subscribe()
+    cherrypy.engine.certificate_monitor = plugins.BackgroundTask(
+        3600, CertificateManager.renew_certificate
```
drrtuy, line 172: Raise the number up to 86400. Once a day should be enough.

cmapi/cmapi_server/managers/certificate.py
```diff
@@ -0,0 +1,106 @@
...
+                encoding=serialization.Encoding.PEM,
+                format=serialization.PrivateFormat.TraditionalOpenSSL,
+                encryption_algorithm=serialization.NoEncryption()),
+            )
+
+        subject = issuer = x509.Name([
+            x509.NameAttribute(NameOID.COUNTRY_NAME, 'US'),
+            x509.NameAttribute(NameOID.STATE_OR_PROVINCE_NAME, 'California'),
+            x509.NameAttribute(NameOID.LOCALITY_NAME, 'Redwood City'),
+            x509.NameAttribute(NameOID.ORGANIZATION_NAME, 'MariaDB'),
```
drrtuy: Change this to `MariaDB Corporation`.

cmapi/cmapi_server/managers/certificate.py
```diff
@@ -0,0 +1,106 @@
...
+CERT_FILENAME = './cmapi_server/self-signed.crt'
+KEY_FILENAME = './cmapi_server/self-signed.key'
+CERT_DAYS = 365
+
+
+class CertificateManager():
+    """Class with methods to manage self-signed certificate."""
+
+    @staticmethod
+    def create_self_signed_certificate() -> None:
```
drrtuy, line 22: Don't we need to sync with mutex running this method when renew a cert?

drrtuy (commented):
Plz address the comments.

cmapi/cmapi_server/managers/certificate.py
mariadb-AlanMologorsky: Ok, got it, but why you don't like such an approach?
Pros and cons?

drrtuy: An implicit action that is a side effect mostly unexpected. One has to read the code of the functions to realize this.

cmapi/cmapi_server/__main__.py
mariadb-AlanMologorsky: Checking if certificate will expire in 1 day, so setting 86400 could be potentially harmful.

cmapi/cmapi_server/managers/certificate.py
mariadb-AlanMologorsky: Have no idea which mutex to sync you're talking about.

cmapi/cmapi_server/managers/certificate.py
mariadb-AlanMologorsky: Done.

cmapi/cmapi_server/managers/certificate.py
drrtuy: You write into a file. You can do this from multiple threads thus this is an resource access race.

cmapi/cmapi_server/__main__.py
drrtuy: I agree to disagree on this.

drrtuy (approved):
untested ACK

PR #3214, opened 2024-06-25
MCOL-5772: incorrect ORDER BY ordering for a columns not in GROUP BY

mariadb-SergeyZefirov:
When ORDER BY column is not in GROUP BY, is not an aggregate and there is a SELECT column that is also not an aggregate, there was a problem: ordering happened on the SELECTed column, not ORDERed one.

This patch fixes that particular problem and also performs some tidying around newly added aggregate.


drrtuy (approved):
untested ACK.
CI should finish before one merges.

mariadb-LeonidFedorov (approved):

PR #3216, opened 2024-06-26
MCOL-5747: gcc-14.1.1 compile error - calloc - transposed args [stable-23.10]

mariadb-LeonidFedorov:
The arguments of calloc are the number of members and the sizeof the member. Gcc-14.1.1 worked out how to tell the difference.
We correct this by transposing to gcc's will.

cherry-pick from stable-23.10


drrtuy (approved):
untested ACK

PR #3219, opened 2024-06-26
fix(plugin) MCOL-5699: [stable-23.10]

mariadb-LeonidFedorov:
throw error for unimplemented INTERSECT and EXCEPT


drrtuy (approved):
untested ACK

PR #3220, opened 2024-06-26
MCOL-5035: a difference in INSERT and UPDATE behavior [stable-23.10]

mariadb-LeonidFedorov:
The UPDATE statement wrote NULL when the column set is DATETIME and value is '0000-00-00 00:00:00'. The problem was inside WriteEngine's handling of UPDATE statements and this is where heart of change is.

Other changes are related to some obsolete data structures in DML/DDL handling that just hanging around there, doing nothing.


drrtuy (approved):
untested ACK

mariadb-SergeyZefirov (approved):

PR #3228, opened 2024-06-28
MCOL-5670, MCOL-5669: bug(priproc) fix of the last_day function

mariadb-SergeyZefirov:
This fixes discrepance with the server, which assigns DATE type to last_day()'s result.

Now we also assigns DATE result type and, also, use proper dataconvert::Day data structure to return date.

Tests agree with InnoDB.

Also, this patch includes test for MCOL-5669, to show we fixed it.


LaGrunge (approved):
Good work! Lets add some union related tests here

mariadb-LeonidFedorov (approved):

PR #3234, opened 2024-07-03
fix(funcexp): MCOL-4671: Fix behaviour of LEFT/RIGHT functions when negative trim value is passed

mariadb-AlexanderPresniakov:
Both LEFT and RIGHT functions can receive a negative trim value, but such values don't make any sense.
There was a bug: trim value was decoded as unsigned, negative values were overflowing the int, resulting in giant values, so input strings were not trimmed at all


mariadb-LeonidFedorov (approved):
untested ACK

mariadb-SergeyZefirov (approved):
I consulted server's code, that's what they do.

PR #3243, opened 2024-07-09
fix(client): MCOL-5587: enable quick mode for predictable performance

mariadb-SergeyZefirov:
This changeset enables quick (mariadb -q) mode when columnstore is installed. Quick mode precludes client CLI program from storing too much data in memory, preventing out of memory conditions.


drrtuy (approved):
untested ACK

PR #3244, opened 2024-07-10
fix: MCOL-5755: incorrect handling of BLOB (and TEXT) in GROUP BY

mariadb-SergeyZefirov:
BLOB fields did not work as grouping keys at all, they were assigned value NULL for any value, be it NULL or not. The fix is in the rowaggregation.cpp in the initMapping(), a switch/case branch was added to handle BLOB field copying there.

Also, TEXT columns did not distinguish between NULL and empty string in the grouping algorithm, now they do. The fix is in the equals() function, now we specifically check for isNull() equality between values.


mariadb-LeonidFedorov (approved):

mariadb-SergeyZefirov:
https://ci.columnstore.mariadb.net/mariadb-corporation/mariadb-columnstore-engine/10698/2 - builds and tests are green when platforms' pipelines have no timeouts. So I merge it.

PR #3248, opened 2024-07-19
Add scripts to run Multinode MTR

timofeyturenko:


tests/scripts/mdbci/multinode_mtr/columnstore_operations.py
```diff
@@ -0,0 +1,65 @@
+import requests
+import json
+import re
+
+from mdb_buildbot_scripts.operations import host
+
+REPL_PASS = 'ct570c3521fCCR#'
+REPL_USER = 'repl'
+MAX_PASS = 'ct570c3521fCCR#ef'
```
mariadb-LeonidFedorov, line 9: this is open source repo, I believe we should not commit the secrets

PR #3249, opened 2024-07-19
MCOL-4499: Filters [NOT] LIKE NULL stop filtering

mariadb-AlexanderPresniakov:
Varchars with width > 7 are dictionary objects that keep the value in the blob, and the position of value inside the blob is stored in the token structure. 
Here we introduce columnFilterMode == ALWAYS_FALSE that means "don't even start filtering, the query won't find anything" (because LIKE NULL will never match anything). We enable it if operation is LIKE/NOT LIKE and the right hand value is NULL.
The hardest part here is to determine the NULL value. There are many implementation of getting the null value (I found six of them). Now getNullValue<T> in primitiveprocessor.h also takes width argument that we use to check if VARCHAR is wide enough to use dictionary format

### DIFF
--- a/mysql-test/columnstore/bugfixes/mcol-4499-float-null.test
+++ b/mysql-test/columnstore/bugfixes/mcol-4499-float-null.test
@@ -1,0 +1,19 @@
+-- source ../include/have_columnstore.inc
+
+--disable_warnings
+DROP DATABASE IF EXISTS `doubles_and_nulls`;
+CREATE DATABASE `doubles_and_nulls`;
+USE `doubles_and_nulls`;
+
+DROP TABLE IF EXISTS qatablefloat;
+--enable_warnings
+
+CREATE TABLE qatablefloat (col float) engine=columnstore;
+INSERT INTO qatablefloat VALUES (null);
+INSERT INTO qatablefloat VALUES (null);
+INSERT INTO qatablefloat VALUES (null);
+DELETE FROM qatablefloat WHERE col IS NULL;
+SELECT * FROM qatablefloat;
+
+DROP TABLE qatablefloat;
+DROP DATABASE `doubles_and_nulls`;
--- a/mysql-test/columnstore/bugfixes/mcol-4499-float-null.test
+++ b/mysql-test/columnstore/bugfixes/mcol-4499-float-null.test
@@ -12,6 +12,9 @@ CREATE TABLE qatablefloat (col float) engine=columnstore;
 INSERT INTO qatablefloat VALUES (null);
 INSERT INTO qatablefloat VALUES (null);
 INSERT INTO qatablefloat VALUES (null);
+INSERT INTO qatablefloat VALUES (3.14);
+INSERT INTO qatablefloat VALUES (0);
+INSERT INTO qatablefloat VALUES (1.44);
 DELETE FROM qatablefloat WHERE col IS NULL;
 SELECT * FROM qatablefloat;
 
--- a/mysql-test/columnstore/bugfixes/mcol-4499.test
+++ b/mysql-test/columnstore/bugfixes/mcol-4499.test
@@ -8,20 +8,120 @@
 DROP DATABASE IF EXISTS `mcol_4499`;
 CREATE DATABASE `mcol_4499`;
 USE `mcol_4499`;
-DROP TABLE IF EXISTS twidevarchar;
-DROP TABLE IF EXISTS twidechar;
+
+DROP TABLE IF EXISTS t_char_2;
+DROP TABLE IF EXISTS t_char_3;
+DROP TABLE IF EXISTS t_char_4;
+DROP TABLE IF EXISTS t_char_5;
+DROP TABLE IF EXISTS t_char_6;
+DROP TABLE IF EXISTS t_char_7;
+DROP TABLE IF EXISTS t_char_8;
+DROP TABLE IF EXISTS t_char_9;
+
+DROP TABLE IF EXISTS t_varchar_2;
+DROP TABLE IF EXISTS t_varchar_3;
+DROP TABLE IF EXISTS t_varchar_4;
+DROP TABLE IF EXISTS t_varchar_5;
+DROP TABLE IF EXISTS t_varchar_6;
+DROP TABLE IF EXISTS t_varchar_7;
+DROP TABLE IF EXISTS t_varchar_8;
+DROP TABLE IF EXISTS t_varchar_9;
 --enable_warnings
 
-CREATE TABLE twidevarchar (col VARCHAR(4)) ENGINE=ColumnStore;
-INSERT INTO twidevarchar VALUES (NULL),('a');
-SELECT col FROM twidevarchar WHERE col LIKE NULL;
-SELECT col FROM twidevarchar WHERE col NOT LIKE NULL;
-DROP TABLE twidevarchar;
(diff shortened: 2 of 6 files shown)


primitives/linux-port/column.cpp
```diff
@@ -1763,7 +1772,8 @@ void filterColumnData(NewColRequestHeader* in, ColResultHeader* out, uint16_t* r
 
   // Bit patterns in srcArray[i] representing EMPTY and NULL values
   T emptyValue = getEmptyValue<T>(dataType);
-  T nullValue = getNullValue<T>(dataType);
+  // TODO why don't we get null value here like in _parseColumnFilter? It can be unsigned
```
mariadb-SergeyZefirov: Please, provide examples where nullValue can be unsugned and type is signed, if you meant that.

As far as I can remember, colDataType is properly dispatched to specific C++ signed/unsigned type.

primitives/linux-port/primitiveprocessor.h
```diff
@@ -492,14 +493,17 @@ T getNullValue(uint8_t type)
     case execplan::CalpontSystemCatalog::BLOB:
     case execplan::CalpontSystemCatalog::TEXT: return joblist::CHAR8NULL;
 
+    // VARCHARs with width >= 8 are stored as dictionaries
```
mariadb-SergeyZefirov: Same is for CHAR type. Please add it here and also add appropriate column into a test case.

mariadb-SergeyZefirov (changes requested):
i generally agree with what was done here.

The only thing of importance is that CHAR type processed differently than VARCHAR/TEXT etc. CHAR columns have to be processed in same manner. Also, all affected types should be in test.

Thus, I  finish my review with "changes requested."

.gitignore
```diff
@@ -187,3 +187,4 @@ obj/
 build/build
 
 
+
```
drrtuy: I assume this is an automatic change but I don't see reason why it is here.

drrtuy: The main concern is this change distracts the attention of the reviewer.

primitives/linux-port/column.cpp
```diff
@@ -1747,12 +1747,21 @@ void filterColumnData(NewColRequestHeader* in, ColResultHeader* out, uint16_t* r
 
   // Cache some structure fields in local vars
   auto dataType = (CalpontSystemCatalog::ColDataType)in->colType.DataType;  // Column datatype
+  uint16_t colWidth = in->colType.DataSize;                                    // Column width
```
drrtuy: Why not const auto ?

primitives/linux-port/primitiveprocessor.h
```diff
@@ -469,21 +471,20 @@ class PrimitiveProcessor
...
 template <typename T, typename std::enable_if<sizeof(T) == sizeof(int128_t), T>::type* = nullptr>
-T getNullValue(uint8_t type)
+T getNullValue(uint8_t type, uint32 width)
 {
   return datatypes::Decimal128Null;
 }
 
 template <typename T, typename std::enable_if<sizeof(T) == sizeof(int64_t), T>::type* = nullptr>
-T getNullValue(uint8_t type)
+T getNullValue(uint8_t type, uint32 width)
```
drrtuy: JFYI @mariadb-AlexanderPresniakov here the template uses ugly SFINAE(enable_if) to narrow the number of candidates found by the compiler. There were no concepts available when I refactored this part. Put it plain this template is triggered for types that are 8 bytes long. I might be missing some details but isn't this info enough to assume that the width will be 8 bytes?

drrtuy (commented):

primitives/linux-port/primitiveprocessor.h
mariadb-AlexanderPresniakov: Thanks, added CHAR

primitives/linux-port/column.cpp
mariadb-AlexanderPresniakov: Sorry, my bad, it was a TODO for myself (in some places in code null value is calculated with sign in mind, in others without it). Removed the comment

primitives/linux-port/primitiveprocessor.h
mariadb-AlexanderPresniakov: Thanks, I removed the width argument

primitives/linux-port/primitiveprocessor.h
```diff
@@ -482,16 +484,18 @@ T getNullValue(uint8_t type)
...
 
-    case execplan::CalpontSystemCatalog::CHAR:
-    case execplan::CalpontSystemCatalog::VARCHAR:
     case execplan::CalpontSystemCatalog::DATE:
     case execplan::CalpontSystemCatalog::DATETIME:
     case execplan::CalpontSystemCatalog::TIMESTAMP:
     case execplan::CalpontSystemCatalog::TIME:
     case execplan::CalpontSystemCatalog::VARBINARY:
     case execplan::CalpontSystemCatalog::BLOB:
+    case execplan::CalpontSystemCatalog::CHAR:
```
mariadb-AlexanderPresniakov, line 493: @mariadb-SergeyZefirov according to TypeHandlerChar::getNullValueForType CHARs are stored as dicts when their width is > 8, so we would have to handle them in getNullValue<128>(uint8_t type). But @drrtuy says that this case is impossible.

mysql-test/columnstore/bugfixes/mcol-4499.test
```diff
@@ -0,0 +1,27 @@
...
+DROP TABLE IF EXISTS twidechar;
+--enable_warnings
+
+CREATE TABLE twidevarchar (col VARCHAR(4)) ENGINE=ColumnStore;
+INSERT INTO twidevarchar VALUES (NULL),('a');
+SELECT col FROM twidevarchar WHERE col LIKE NULL;
+SELECT col FROM twidevarchar WHERE col NOT LIKE NULL;
+DROP TABLE twidevarchar;
+
+CREATE TABLE twidechar (col CHAR(5)) ENGINE=ColumnStore;
```
mariadb-SergeyZefirov: CHAR(5) is in dict only for specific collations. Please use wider column.

mysql-test/columnstore/bugfixes/mcol-4499-float-null.test
```diff
@@ -0,0 +1,19 @@
...
+CREATE DATABASE `doubles_and_nulls`;
+USE `doubles_and_nulls`;
+
+DROP TABLE IF EXISTS qatablefloat;
+--enable_warnings
+
+CREATE TABLE qatablefloat (col float) engine=columnstore;
+INSERT INTO qatablefloat VALUES (null);
+INSERT INTO qatablefloat VALUES (null);
+INSERT INTO qatablefloat VALUES (null);
```
drrtuy, line 14: I bet you want to insert couple non NULL values.

drrtuy (approved):
untested ACK
This patch calls for a rebase though.

mariadb-SergeyZefirov (changes requested):
Regading CHAR column width. Please use latin1 charset and latin1_bin collation. With these the width of character  columns corresponds to the with in bytes.

mysql-test/columnstore/bugfixes/mcol-4499.test
mariadb-AlexanderPresniakov: I've added test cases for char<2-9> and varchar<2-9>, both using ascii

mariadb-LeonidFedorov (approved):

PR #3252, opened 2024-07-25
chore(codestyle): MCOL-5405: replace windows CRLF with virtious linux one

mariadb-LeonidFedorov:
To see the exact diff invoke 
```git show --color --word-diff-regex=. MCOL-4266-windows-crlf```


mariadb-SergeyZefirov (approved):
I see no flaws, thus I approve!

PR #3253, opened 2024-07-26
MCOL-5587: Fix columnstore.cnf.

mariadb-AlanMologorsky:
fix(client): Fix columnstore.cnf file

This fix changes option file to apply '--quick' option only for 'mariadb' and 'mysql' clients instead of all MariaDB clients. Otherwise 'mysqladmin' uses this option, but it doesn't exist. As a result broken CI multinode MTR stage.

### DIFF
--- a/dbcon/mysql/columnstore.cnf
+++ b/dbcon/mysql/columnstore.cnf
@@ -1,4 +1,4 @@
-[client]
+[mariadb-client]
 quick
 
 [mysqld]


mariadb-AlanMologorsky:
@mariadb-SergeyZefirov Could you pls give me more details about which CLI clients should be affected by `--quick` option?

mariadb-SergeyZefirov:
> @mariadb-SergeyZefirov Could you pls give me more details about which CLI clients should be affected by `--quick` option?

I can not, I am just not an expert in that.

mariadb-SergeyZefirov (approved):

mariadb-LeonidFedorov (approved):
Superb! We need it in stable-23.10 for sure

mariadb-LeonidFedorov (approved):

PR #3257, opened 2024-08-01
MCOL-5776: support functions not in group by

mariadb-SergeyZefirov:
An attempt to make our support of SELECT GROUP BY to be closer to server's.

### DIFF
--- a/dbcon/execplan/aggregatecolumn.cpp
+++ b/dbcon/execplan/aggregatecolumn.cpp
@@ -46,6 +46,7 @@ using namespace joblist;
 
 namespace execplan
 {
+
 void getAggCols(execplan::ParseTree* n, void* obj)
 {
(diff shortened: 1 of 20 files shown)


mariadb-SergeyZefirov:
What does this patch do:
  - it prevents wrapping into aggregates values under GROUP BY keys, changes in ha_mcs_execplan.cpp are related to that;
  - it also disabled lowering of HAVING filter into WHERE if there are no aggregates (joblistfactory.cpp).

Other changes are related to build troubles (CMakefile.txt) and not much else.

There will be a separate PR for regression suite.

Right now it is still draft and is not rebased for me to have history of changes.

mariadb-SergeyZefirov:
https://github.com/mariadb-corporation/mariadb-columnstore-regression-test/pull/374 - relevant PR for regression suite.

drrtuy (approved):
untested ACK

PR #3261, opened 2024-08-05
MCOL-5779: use encoding for correct check of ALTER TABLE CHANGE COLUMN

mariadb-LeonidFedorov:


drrtuy (approved):
untested ACK

dbcon/ddlpackageproc/altertableprocessor.cpp
```diff
@@ -88,6 +90,14 @@ struct extentInfo
 
 namespace
 {
+
+bool checkTextTypesLengthAreUqual(const CalpontSystemCatalog::ColType& colType, const ColumnType& newType)
+{
+  auto* newTypeCharset = get_charset(newType.fCharsetNum, MYF(MY_WME));
+  auto newTypeBytesInChar = newTypeCharset ? newTypeCharset->mbmaxlen : 1;
+  return colType.colWidth * newTypeBytesInChar == newType.fLength * colType.getCharset()->mbmaxlen;
```
mariadb-SergeyZefirov: If the intention to check for equality of max text len of old and new types, then why do we multiply old width by new mbMaxLen and new width by (apparently) old mbMaxLen?

This need a comment on why is that.

mariadb-SergeyZefirov (approved):
I generally agree with what is done, but crucial piece of code needs a comment. ;)

mariadb-SergeyZefirov (approved):
I completely agree.

PR #3266, opened 2024-08-09
MCOL-5719: Move ownership mechanism to KV storage.

denis0x0D:
Move ownership mechanism to KV storage.


storage-manager/src/KVStorageInitializer.cpp
```diff
@@ -0,0 +1,74 @@
...
+   MA 02110-1301, USA. */
+#include "Config.h"
+#include "SMLogging.h"
+#include "KVStorageInitializer.h"
+
+namespace storagemanager
+{
+static std::shared_ptr<FDBCS::FDBDataBase> fdbDataBaseInstance;
+static std::unique_ptr<FDBCS::FDBNetwork> fdbNetworkInstance;
+static std::mutex kvStorageLock;
```
mariadb-LeonidFedorov, line 25: I'd better avoid static mutexes, we have a plenty of code, that starts before main() and we could have undefined statics initialization effects. Please consider explicit initialization inside of class object

storage-manager/src/Ownership.cpp
```diff
@@ -19,11 +19,13 @@
 #include "Config.h"
 #include "Cache.h"
 #include "Synchronizer.h"
+#include "KVStorageInitializer.h"
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <unistd.h>
 #include <fcntl.h>
 #include <boost/filesystem.hpp>
+#include "../fdb_wrapper_cpp/include/fdbcs.hpp"
```
mariadb-LeonidFedorov: having utils as the common engine path, we can write includes as <> without relative paths

storage-manager/src/KVStorageInitializer.h
```diff
@@ -0,0 +1,36 @@
...
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software
+   Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+   MA 02110-1301, USA. */
+
+#pragma once
+#include <memory>
+#include "../fdb_wrapper_cpp/include/fdbcs.hpp"
```
mariadb-LeonidFedorov: same on include paths

storage-manager/src/KVStorageInitializer.cpp
denis0x0D: @mariadb-LeonidFedorov  The static here means that particular global is not visible outside the translation unit. The idea behind this globals is to implement a singleton pattern, unfortunately we cannot initialize fdb network more than one time, but without a global variable you cannot say is it already init or not. I'm not sure which problem with static initialization you are talking inititialization order fiasco is not applicable here. If you know how to share information across threads without global, please let me know.

mariadb-LeonidFedorov (approved):

storage-manager/src/KVStorageInitializer.cpp
```diff
@@ -0,0 +1,74 @@
...
+std::shared_ptr<FDBCS::FDBDataBase> KVStorageInitializer::getStorageInstance()
+{
+  if (fdbDataBaseInstance)
+    return fdbDataBaseInstance;
+
+  std::unique_lock<std::mutex> lock(kvStorageLock);
+  if (fdbDataBaseInstance)
+    return fdbDataBaseInstance;
+
+  auto* config = Config::get();
```
drrtuy: I assume these two can be outside the crit section starts at line 32.

storage-manager/src/KVStorageInitializer.cpp
```diff
@@ -0,0 +1,74 @@
...
+  auto* logger = SMLogging::get();
+
+  if (!FDBCS::setAPIVersion())
+  {
```
drrtuy: There are 4 times you use the same code here, mb. it is worth to put it into an inline func?

storage-manager/src/KVStorageInitializer.h
```diff
@@ -0,0 +1,38 @@
...
+
+namespace storagemanager
+{
+// Represnts a `Key/Value` storage initializer based on Singleton pattern.
+// Initializes all needed `Key/Value` storage machinery once and return pointer
+// to the `Key/Value` storage.
+class KVStorageInitializer
+{
+ public:
+  static std::shared_ptr<FDBCS::FDBDataBase> getStorageInstance();
```
drrtuy, line 30: Are you sure that we need shared_ptr? The lifecycle of FDBDataBase equals with the lifecycle of SM, the runtime at all times has at least one instance, so mb it is worth to exchange ptr?
@mariadb-LeonidFedorov WDYT?

storage-manager/src/Ownership.cpp
```diff
@@ -124,28 +138,24 @@ bf::path Ownership::get(const bf::path& p, bool getOwnership)
...
-
 void Ownership::touchFlushing(const bf::path& prefix, volatile bool* doneFlushing) const
 {
   while (!*doneFlushing)
   {
-    TOUCH(prefix, "FLUSHING");
+    {
+      auto kvStorage = KVStorageInitializer::getStorageInstance();
+      auto tnx = kvStorage->createTransaction();
+      const std::string flushingKey = KVPrefixes[static_cast<uint32_t>(KVPrefixId::SM_OWNERSHIP)] +
```
drrtuy: TBH it is very hard to read which idx is used in the end.
I suggest to put idx-es cast outside to improve readability. Moreover I suggest to replace the key creation pattern  with an inline func that leverages ostringstream with preallocated buffer. WDYT @denis0x0D ?

storage-manager/src/Ownership.cpp
```diff
@@ -171,10 +181,23 @@ void Ownership::releaseOwnership(const bf::path& p, bool isDtor)
 
   if (isDtor)
   {
-    // This is a quick release.  If this is being destroyed, then it is through the graceful
-    // shutdown mechanism, which will flush data separately.
-    DELETE(p, "OWNED");
-    DELETE(p, "FLUSHING");
+    {
```
drrtuy: The pattern of this scope looks similar to another one both used later in this func and in `_takeOwnership`. Can we replace the boilerplate code with a func?

storage-manager/src/Ownership.cpp
```diff
@@ -124,28 +138,24 @@ bf::path Ownership::get(const bf::path& p, bool getOwnership)
...
+      auto tnx = kvStorage->createTransaction();
+      const std::string flushingKey = KVPrefixes[static_cast<uint32_t>(KVPrefixId::SM_OWNERSHIP)] +
+                                      prefix.string() +
+                                      ownerShipStates[static_cast<uint32_t>(OwnershipStateId::FLUSHING)];
+      tnx->set(flushingKey, "");
+      if (!tnx->commit())
+      {
+        const char* msg = "Ownership: commit `touchFlushing` failed ";
+        logger->log(LOG_CRIT, msg);
+        throw runtime_error(msg);
```
drrtuy, line 169: I see that previously DELETE doesn't throw and now it throws. Did you test this control path? SM might get SIGABRT that it doesn't handle that effectively crashes SM.

storage-manager/src/KVPrefixes.cpp
```diff
@@ -0,0 +1,24 @@
...
+   along with this program; if not, write to the Free Software
+   Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+   MA 02110-1301, USA. */
+
+#include "KVPrefixes.hpp"
+
+namespace storagemanager
+{
+// FDB recommends keep the key size up to 32 bytes.
+const char* KVPrefixes[2] = {/*OWNERSHIP*/ "SM_O_", /*META*/ "SM_M_"};
```
drrtuy, line 23: Later in the patch these prefixes(and suffixes) are always implicitly casted to string. I am curious did you try constexpr const string instead?

drrtuy (changes requested):

storage-manager/src/KVPrefixes.cpp
denis0x0D: UPD: I see that `constexpr` creates a trivial destrcutor for container, I guees it could be updated to constexpr string, will take a look

storage-manager/src/KVStorageInitializer.cpp
denis0x0D: make sense

storage-manager/src/KVStorageInitializer.cpp
denis0x0D: not clear to me, why do we need for each process, which tries to take a lock request a config and logger?

storage-manager/src/KVStorageInitializer.h
denis0x0D: This shared ptr comes here because of wrapper api, I can imagine a use case where FDBDataBase instance is not a static and also could be sent as a parameter to a function and move semantics is not applicable there. Having a just a row pointer could lead a memory leak, since user could forget to free that memory.

storage-manager/src/Ownership.cpp
denis0x0D: make sense, it could be placed in a outline function.

storage-manager/src/Ownership.cpp
denis0x0D: I see only one commit operation  which is not handled at line 304, good catch, thanks.
BTW - each transaction could have some `get/set/remove` operations in a row, and should end with `commit` operation. Only `commit` could fail, but not `get/set/remove` according to C API FDB.
Handling exception is responsibility of Ownership user. It is used in `IOCoordinator` which implements a `read/write/open` operation for the given filename. IOC itself used in Tasks - ReadTask, OpenTask and so on.
Each call of that functions a covered by try catch block:
```
   try
    {
      err = ioc->read(cmd->filename, &resp->payload[resp->returnCode], cmd->offset + resp->returnCode,
                      cmd->count - resp->returnCode);
    }
    catch (exception& e)
    {
      logger->log(LOG_ERR, "ReadTask: caught '%s'", e.what());
      errno = EIO;
      err = -1;
    }
```

storage-manager/src/Ownership.cpp
denis0x0D: make sense.

storage-manager/src/KVStorageInitializer.h
mariadb-LeonidFedorov: IMHO Double check pattern with lock could be replaced with internal static here having
```
static FDBCS::FDBDataBase& getStorageInstance()
{
    static InstanceWrapper instanceWrapper(....);
    return instanceWrapper.instance();
}
```
where InstanceWrapper ctor initilize everything inside it's object

denis0x0D: @mariadb-LeonidFedorov this approach looks cleaner, but put initialization in constructor is not recommended as well https://google.github.io/styleguide/cppguide.html#Doing_Work_in_Constructors

mariadb-LeonidFedorov: I believe this rule exists because Google doesn't use exceptions, but we do

storage-manager/src/KVStorageInitializer.cpp
drrtuy: B/c config and logger have their own global mutexes.

storage-manager/src/KVStorageInitializer.h
drrtuy: Holy Google! 
They are wise, we aren't in this case.

denis0x0D:
@drrtuy updated on review, please take a look

drrtuy (approved):
untested ACK

PR #3286, opened 2024-08-20
chore: UM join is multi-threaded now [develop-23.02]

drrtuy:


mariadb-SergeyZefirov (approved):
As I understand, this patch does not change algorithms in any significant way. I do not see here anything that can affect behavior in a negative way.

Thus, I approve this changeset.

mariadb-LeonidFedorov (approved):

mariadb-AlexeyAntipovsky (approved):

PR #3289, opened 2024-08-24
fix(rowgroup): RGData now uses uint64_t describing its buffer size[develop-23.02]

drrtuy:
This fact causes implicit heap overflow

### DIFF
--- a/utils/rowgroup/rowgroup.cpp
+++ b/utils/rowgroup/rowgroup.cpp
@@ -305,42 +305,23 @@ void UserDataStore::deserialize(ByteStream& bs)
 
 RGData::RGData(const RowGroup& rg, uint32_t rowCount)
 {
-  // cout << "rgdata++ = " << __sync_add_and_fetch(&rgDataCount, 1) << endl;
-  rowData.reset(new uint8_t[rg.getDataSize(rowCount)]);
+  RGDataSizeType s = rg.getDataSize(rowCount);
+  rowData.reset(new uint8_t[s]);
 
   if (rg.usesStringTable() && rowCount > 0)
     strings.reset(new StringStore());
 
   userDataStore.reset();
-
-
-#ifdef VALGRIND
-  /* In a PM-join, we can serialize entire tables; not every value has been
-   * filled in yet.  Need to look into that.  Valgrind complains that
-   * those bytes are uninitialized, this suppresses that error.
-   */
-  memset(rowData.get(), 0, rg.getDataSize(rowCount));  // XXXPAT: make valgrind happy temporarily
-#endif
-  memset(rowData.get(), 0, rg.getDataSize(rowCount));  // XXXPAT: make valgrind happy temporarily
 }
 
 RGData::RGData(const RowGroup& rg)
 {
-  // cout << "rgdata++ = " << __sync_add_and_fetch(&rgDataCount, 1) << endl;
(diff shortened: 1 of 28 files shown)


utils/rowgroup/rowgroup.cpp
```diff
@@ -1229,9 +1229,9 @@ uint32_t RowGroup::getDataSize() const
   return headerSize + (getRowCount() * offsets[columnCount]);
 }
```
kemm, line 1230: I think we should change `getDataSize()` too.
```c++
uin64_t RowGroup::getDataSize() const
{
  return getDataSize(getRowCount());
}
```

mariadb-AlexeyAntipovsky, line 1204: I think we should change `getDataSize()` too.
```c++
uin64_t RowGroup::getDataSize() const
{
  return getDataSize(getRowCount());
}
```

mariadb-SergeyZefirov (approved):
I completely agree with the changes.

mariadb-LeonidFedorov (approved):

utils/rowgroup/rowgroup.cpp
mariadb-AlexeyAntipovsky: Anyway, IMO there should be `static_cast<uint64_t>(offset_field)` in every `getDataSize`-like function to prevent int overflow when multiplying two uint32_t values.

utils/rowgroup/rowgroup.h
```diff
@@ -1560,11 +1561,11 @@ class RowGroup : public messageqcpp::Serializeable
   uint32_t sTableThreshold = 20;
   std::shared_ptr<bool[]> forceInline;
 
-  static const uint32_t headerSize = 18;
-  static const uint32_t rowCountOffset = 0;
-  static const uint32_t baseRidOffset = 4;
-  static const uint32_t statusOffset = 12;
-  static const uint32_t dbRootOffset = 14;
+  static const uint64_t headerSize = 18;
```
mariadb-AlexeyAntipovsky, line 1564: The `getHeaderSize()` signature should also be fixed (https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/3289/files#diff-f3dffeb8ad50140337201f2ed7ecd27d3ecc86aa2a9b714cdeedfb82dc454edcR1466)

mariadb-AlexeyAntipovsky (changes requested):

utils/rowgroup/rowgroup.cpp
drrtuy: Only `getDataSize` this makes sense. B/c the other two has 8192 as a number of rows so there must be 65K columns in the rowgroup to actually overflow.

mariadb-SergeyZefirov: We have a hard limit of 4096 columns per table. It will requre query like ```SELECT * FROM t1 JOIN t2 ON t1a = t2.b JOIN ...``` to join more than 16 tables  with max columns (or more than 32 - most values in row will occupy 8 bytes), to have an overflowiing condition.

But, we may hit another hard limit on row size on the MDB's side (online documentation has it at 64K).

What I'd suggest is to have a comment describing this situation, here or elsewhere.

mariadb-AlexeyAntipovsky (approved):

mariadb-AlexeyAntipovsky (approved):
lgtm

PR #3295, opened 2024-08-28
MCOL-5779: comment on alter table [stable 23.10]

mariadb-LeonidFedorov:
use encoding for correct check of ALTER TABLE CHANGE COLUMN

### DIFF
--- a/dbcon/ddlpackageproc/altertableprocessor.cpp
+++ b/dbcon/ddlpackageproc/altertableprocessor.cpp
@@ -45,6 +45,8 @@ using namespace ddlpackage;
 #include "messagelog.h"
 using namespace logging;
 
+#include "mariadb_my_sys.h"
+
 #include "we_messages.h"
 #include "we_ddlcommandclient.h"
 #include "we_ddlcommon.h"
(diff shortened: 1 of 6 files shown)


dbcon/ddlpackageproc/altertableprocessor.cpp
```diff
@@ -88,6 +90,14 @@ struct extentInfo
 
 namespace
 {
+
+bool checkTextTypesLengthAreUqual(const CalpontSystemCatalog::ColType& colType, const ColumnType& newType)
```
drrtuy: Uqual - is this a typo?

dbcon/execplan/calpontsystemcatalog.h
```diff
@@ -224,7 +224,7 @@ class CalpontSystemCatalog : public datatypes::SystemCatalog
     bool autoincrement = 0;  // set to true if  SYSCOLUMN autoincrement is �y�
     uint64_t nextvalue = 0;  // next autoincrement value
     uint32_t charsetNumber = default_charset_info->number;
-    const CHARSET_INFO* cs = nullptr;
+    const mutable CHARSET_INFO* cs = nullptr;
```
drrtuy, line 227: Are you sure we need a trick like this to make `getCharset` const? We actually changes the state of ColType instance

drrtuy (changes requested):

dbcon/execplan/calpontsystemcatalog.h
mariadb-LeonidFedorov: Yes, sure, getCharset initilizes this field if empty, I don't want to change the semantics

dbcon/ddlpackageproc/altertableprocessor.cpp
mariadb-LeonidFedorov: This is Albanian accent, I'll fix

drrtuy (approved):
untested ACK

PR #3299, opened 2024-09-02
[DUPLICATE STABLE-23.10] MCOL-5695: Add FoundationDB package as a dependency for CMAPI.

mariadb-AlanMologorsky:


mariadb-AlanMologorsky:
Discussed with @mariadb-LeonidFedorov and prefer to make another solution for all branches except develop.

PR #3304, opened 2024-09-05
MDEV-33091 rename PCRE_INCLUDES to use PKG_CHECK_MODULES naming, PCRE_INCLUDE_DIR

grooverdan:
…

The 10.6 server changed the name of the include.

ref: https://github.com/mariaDB/server/commit/3a33ae86010051a1022c392f2936b5fc9f3ab19e

Noticed thanks to @LinuxJedi

I also backported the same commit to 10.5 - https://github.com/MariaDB/server/pull/3484 however since its less maintained, appended a workaround mapping for columnstore there.


drrtuy:
@grooverdan Thx for the patch. 
This change is actual for all versions of servers, I suppose?

grooverdan:
yes.

mariadb-LeonidFedorov (approved):

PR #3305, opened 2024-09-05
chore(install): MCOL-4886 remove chown to reduce upgrade time for existing installations.

drrtuy:


mariadb-LeonidFedorov (dismissed):

mariadb-LeonidFedorov:
looks the change breaks package functionality
all smokes fails

PR #3320, opened 2024-09-12
fix(PrimProc): MCOL-5651 Add a workaround to avoid choosing an incorr…

mariadb-AlexeyAntipovsky:
…ect TupleHashJoinStep as a joiner


mariadb-LeonidFedorov (approved):

drrtuy (approved):
untested ACK

PR #3322, opened 2024-09-16
feat(joblist): better dot graphs that represents joblist steps execution tree. This also introduces a simplest version of EXPLAIN-like representation.

drrtuy:


mariadb-SergeyZefirov (approved):
This is very valuable for diagnostics and MCS development. Also, it does not interfere with any functionality- and/or performance-related parts of the MCS.

Thus, I approve wholeheartedly.

PR #3331, opened 2024-10-01
fix(PrimProc): MCOL-5651 Add a workaround to avoid choosing an incorrect TupleHashJoinStep as a joiner [stable-23.10]

mariadb-AlexeyAntipovsky:


drrtuy (approved):
utACK

PR #3332, opened 2024-10-03
MCOL-5802 Add support for blob insertion into FDB.

denis0x0D:


storage-manager/src/KVStorageInitializer.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+  fdbNetworkInstance = std::make_unique<FDBCS::FDBNetwork>();
+  if (!fdbNetworkInstance->setUpAndRunNetwork())
+  {
+    const char* msg = "Ownership: FDB setUpAndRunNetwork failed.";
+    logger->log(LOG_CRIT, msg);
+    throw std::runtime_error(msg);
+  }
+
+  std::string clusterFilePath = config->getValue("ObjectStorage", "fdb_cluster_file_path");
```
drrtuy, line 54: We are using modified camel case, so plz rename the param name `fdb_cluster_file_path`.

utils/fdb_wrapper_cpp/src/fdbcs.cpp
```diff
@@ -19,10 +19,20 @@
...
+#include <boost/uuid/uuid_io.hpp>
+#include <boost/uuid/random_generator.hpp>
+#include <boost/lexical_cast.hpp>
+#include <chrono>
+#include "fdbcs.hpp"
 
 namespace FDBCS
 {
+
+#define RETURN_ON_ERROR(retcode) \
```
drrtuy, line 32: Can we do the same without C PP macro?

denis0x0D: I'm not sure how to make a function on c++ which return only on specific case (error) and not return in other case. Can you please show an example

denis0x0D:
This one moved to https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/3351
since it's hard to rebase it on top

PR #3333, opened 2024-10-04
MCOL-5791: memory leaks in plugin ASAN based

mariadb-SergeyZefirov:


dbcon/mysql/ha_mcs_impl.cpp
```diff
@@ -702,6 +703,10 @@ vector<string> getOnUpdateTimestampColumns(string& schema, string& tableName, in
       // exemgrClient->shutdown();
       // delete exemgrClient;
       // exemgrClient = 0;
+      if (rowGroup)
+      {
+        delete rowGroup;
+      }
```
mariadb-LeonidFedorov: you can use delete on nullptr according the standard, so the check is unnecessary, but after you delete the memory it's better to null the pointer, so it should look like
 ```
delete rowGroup;
rowGroup = nullptr;
```

File: writeengine/bulk/cpimport.cpp [hunk 2]
```diff
@@ -188,6 +188,7 @@ void printUsage()
   exit(EXIT_SUCCESS);
 }
 
+volatile bool finished = false;
```
mariadb-LeonidFedorov: seems to be unused

mariadb-LeonidFedorov (approved):

PR #3334, opened 2024-10-08
MCOL-5720 Move meta to FDB

denis0x0D:


denis0x0D:
This one moved to separate PR.

PR #3335, opened 2024-10-08
fix(plugin): MCOL-5703 fix server crash on 'UNION ALL VALUES' queries

mariadb-AlexeyAntipovsky:
Such queries are not supported by the ColumnStore anyway.


drrtuy (approved):
untested ACK

PR #3343, opened 2024-11-05
MCOL 5724 Move journal to FDB.

denis0x0D:


utils/fdb_wrapper_cpp/test/test_fdb_api.cpp
```diff
@@ -16,10 +16,16 @@
    MA 02110-1301, USA. */
 
 #include <vector>
-#include "../include/fdbcs.hpp"
+#include "fdbcs.hpp"
```
drrtuy, line 19: Why don't you put this into the global `tests` directory?

utils/fdb_wrapper_cpp/include/fdbcs.hpp
```diff
@@ -104,5 +110,76 @@ class DataBaseCreator
...
+// This class represetns a machinery to handle a data `blob`.
+class BlobHandler
+{
+ public:
+  BlobHandler(std::shared_ptr<KeyGenerator> keyGen, uint32_t blockSizeInBytes = 100000)
+   : keyGen_(keyGen), blockSizeInBytes_(blockSizeInBytes)
+  {
+    // Block size in 100KB shows the best performance.
+    keySizeInBytes_ = keyGen_->getKeySize();
+    assert(keySizeInBytes_);
```
drrtuy, line 143: The asserts are necessary here?
They will cause SIGABRT if they are  not caught properly thus SM will crash.

drrtuy: Is there other way to stop an execution with an error w/o using assert?

drrtuy: AFAIK NDEBUG is set in the context of SM.

utils/fdb_wrapper_cpp/include/fdbcs.hpp
```diff
@@ -104,5 +110,76 @@ class DataBaseCreator
...
+  // Removes a `blob` by the given `key`, on error returns false.
+  bool removeBlob(std::shared_ptr<FDBCS::FDBDataBase> database, const ByteArray& key);
+
+ private:
+  size_t insertData(Block& block, const std::string& blob, const size_t offset);
+  void insertKey(Block& block, const std::string& value);
+  std::pair<bool, Keys> getKeysFromBlock(const Block& block);
+  Keys generateKeys(const uint32_t num);
+  bool isDataBlock(const Block& block);
+  bool commitKeys(std::shared_ptr<FDBCS::FDBDataBase> database, std::unordered_map<Key, Block>& map,
```
drrtuy, line 164: Name the map type plz.

utils/fdb_wrapper_cpp/include/fdbcs.hpp
```diff
@@ -104,5 +110,76 @@ class DataBaseCreator
...
+                  const Keys& keys);
+  bool commitKey(std::shared_ptr<FDBCS::FDBDataBase> database, const Key& key, const ByteArray& value);
+  bool removeKeys(std::shared_ptr<FDBCS::FDBDataBase> database, const Keys& keys);
+  std::unordered_map<uint32_t, uint32_t> computeNumKeysForEachTreeLevel(const int32_t treeLen,
```
drrtuy, line 168: Name the map type plz. I have no idea what is this map for.

utils/fdb_wrapper_cpp/include/fdbcs.hpp
```diff
@@ -104,5 +110,76 @@ class DataBaseCreator
...
+  inline float log(const uint32_t base, const uint32_t value);
+
+  std::shared_ptr<KeyGenerator> keyGen_;
+  uint32_t blockSizeInBytes_;
+  uint32_t keySizeInBytes_;
+  uint32_t numKeysInBlock_;
+  uint32_t dataBlockSizeInBytes_;
+  // FIXME: Doc says that 10MB is limit, currently taking in account `key` size and `value` size, but 10MB
+  // limit returns error on transaction.
+  const uint32_t maxTnxSize_{8192000};
```
drrtuy, line 179: Do these three const contributes to space used by every instance of BlobHandler?
If they do you better move these constants outside. The strings will become char* unfortunately.

drrtuy: I mean I wish them to be strings by choosing b/w mem consumption and extra string ctor call I personally prefer extra call. WDYT @denis0x0D ?

utils/fdb_wrapper_cpp/src/fdbcs.cpp
```diff
@@ -111,6 +121,13 @@ bool Transaction::commit() const
       std::cerr << "fdb_future_block_until_ready error, code: " << (int)err << std::endl;
       return false;
     }
+    err = fdb_future_get_error(future);
```
drrtuy, line 124: Could you ellaborate on this sequence you added? A programmer is obliged to always read an error?

utils/fdb_wrapper_cpp/src/fdbcs.cpp
```diff
@@ -206,6 +223,312 @@ std::shared_ptr<FDBDataBase> DataBaseCreator::createDataBase(const std::string c
...
+{
+  const size_t endOfBlock = std::min(offset + dataBlockSizeInBytes_, blob.size());
+  auto& dataBlock = block.second;
+  if (!block.first)
+  {
+    dataBlock.reserve(blockSizeInBytes_);
+    dataBlock.insert(dataBlock.end(), dataBlockIdentifier.begin(), dataBlockIdentifier.end());
+    block.first += dataBlockIdentifier.size();
+  }
+  dataBlock.insert(dataBlock.begin() + block.first, blob.begin() + offset, blob.begin() + endOfBlock);
```
drrtuy, line 273: Don't you need to increase block.first by the leng of the data inserted?

utils/fdb_wrapper_cpp/src/fdbcs.cpp
```diff
@@ -206,6 +223,312 @@ std::shared_ptr<FDBDataBase> DataBaseCreator::createDataBase(const std::string c
...
+    keys.push_back(std::move(key));
+    offset += keySizeInBytes_;
+  }
+
+  return {true, keys};
+}
+
+bool BlobHandler::isDataBlock(const Block& block)
+{
+  return block.second.compare(0, keyBlockIdentifier.size(), keyBlockIdentifier) != 0;
```
drrtuy, line 402: Can you ellaborate on the semanitcs of this expression @denis0x0D?

utils/fdb_wrapper_cpp/src/fdbcs.cpp
```diff
@@ -206,6 +223,312 @@ std::shared_ptr<FDBDataBase> DataBaseCreator::createDataBase(const std::string c
...
+    }
+
+    if (dataBlockReached)
+      break;
+
+    Keys nextKeys;
+    for (const auto& block : blocks)
+    {
+      auto keysPair = getKeysFromBlock(block);
+      if (!keysPair.first)
```
drrtuy, line 441: Why don't you check for `isDataBlock` is true iterating blocks on this level of a tree?
For example you do check this in the previous loop over `currentKeys`.

drrtuy: Did I get it right that the key blocks can't be mixed with data blocks?

storage-manager/src/PrefixCache.cpp
```diff
@@ -593,13 +595,17 @@ int PrefixCache::ifExistsThenDelete(const string& key)
...
-  bool journalExists = bf::exists(journalPath);
+  auto kvStorage = KVStorageInitializer::getStorageInstance();
+  auto tnx = kvStorage->createTransaction();
+  auto resultPair = tnx->get(journalName);
+
+  bool journalExists = resultPair.first;
   // assert(objectExists == bf::exists(cachedPath));
 
-  size_t objectSize = (objectExists ? bf::file_size(cachedPath) : 0);
+  //size_t objectSize = (objectExists ? bf::file_size(cachedPath) : 0);
```
drrtuy, line 605: remove if you don't need this.

storage-manager/src/Replicator.cpp
```diff
@@ -121,9 +124,17 @@ void Replicator::printKPIs() const
...
+  auto keyGen = std::make_shared<FDBCS::BoostUIDKeyGenerator>();
+  FDBCS::BlobHandler blobWriter(keyGen);
+  const std::string dataValue((char*)data);
+  if (!blobWriter.writeBlob(kvStorage, objectFilename, dataValue))
+  {
+    return -1;
+  }
 
+  int fd, err;
   OPEN(objectFilename.c_str(), O_WRONLY | O_CREAT);
```
drrtuy, line 138: Do you need those file-based operations that take place after this line?

storage-manager/src/Replicator.cpp
```diff
@@ -297,164 +309,61 @@ int Replicator::addJournalEntry(const boost::filesystem::path& filename, const u
...
-                    strerror_r(errno, errbuf, 80));
-      errno = l_errno;
-      if (err < 0)
-        return err;
-      else
-        return 0;
-    }
-    l_errno = errno;
-    return err;
+    mpLogger->log(LOG_CRIT, "Cannot remove journal blob.");
```
drrtuy, line 342: Can not

storage-manager/src/Replicator.cpp
```diff
@@ -215,54 +234,48 @@ ssize_t Replicator::_write(int fd, const void* data, size_t length)
...
-  size_t count = 0;
-  int version = 1;
-  int l_errno;
-  char errbuf[80];
-  bool bHeaderChanged = false;
-  string headerRollback = "";
-  string journalFilename = msJournalPath + "/" + filename.string() + ".journal";
+  const int version = 1;
+  const auto journalName = getJournalName(msJournalPath + "/" + filename.string() + ".journal");
+  const auto journalSizeName = getJournalName(msJournalPath + "/" + filename.string() + "_size" + ".journal");
```
drrtuy, line 244: Do you actually need "_size" + ".journal" suffix here?

storage-manager/src/Replicator.cpp
```diff
@@ -215,54 +234,48 @@ ssize_t Replicator::_write(int fd, const void* data, size_t length)
...
   uint64_t offlen[] = {(uint64_t)offset, length};
-  size_t count = 0;
-  int version = 1;
-  int l_errno;
-  char errbuf[80];
-  bool bHeaderChanged = false;
-  string headerRollback = "";
-  string journalFilename = msJournalPath + "/" + filename.string() + ".journal";
+  const int version = 1;
+  const auto journalName = getJournalName(msJournalPath + "/" + filename.string() + ".journal");
```
drrtuy, line 243: Do you actually need ".journal" suffix here?

storage-manager/src/MetadataFile.cpp
```diff
@@ -98,6 +105,7 @@ MetadataFile::MetadataConfig::MetadataConfig()
 
   try
   {
+    cout << "create dir for meta " << msMetadataPath.string() << endl;
```
drrtuy, line 108: I didn't get why do you need to create a meta dir at this point?

storage-manager/src/MetadataFile.cpp
```diff
@@ -215,16 +233,52 @@ void MetadataFile::printKPIs()
...
+  {
+    SMLogging::get()->log(LOG_CRIT, "Metadatafile::stat() failed with error: %s", ex.what());
+    return -1;
+  }
 
-  out->st_size = getLength();
   return 0;
 }
 
+int MetadataFile::stat(struct stat* out)
```
drrtuy, line 264: What is this stat is used for @@denis0x0D?

storage-manager/src/MetadataFile.cpp
```diff
@@ -319,10 +373,19 @@ metadataObject MetadataFile::addMetadataObject(const boost::filesystem::path& fi
...
-    boost::filesystem::create_directories(mFilename.parent_path());
+  {
+    auto kvStorage = KVStorageInitializer::getStorageInstance();
+    auto keyGen = std::make_shared<FDBCS::BoostUIDKeyGenerator>();
+    FDBCS::BlobHandler blobWriter(keyGen);
+    stringstream stream;
+    write_json(stream, *jsontree);
+    if (!blobWriter.writeBlob(kvStorage, mFilename.string(), stream.str()))
+    {
+      SMLogging::get()->log(LOG_CRIT, "Metadatafile: cannot commit tnx set().");
```
drrtuy, line 384: What is `txn set()` in the context? Is there a way to alter the text so that the error nature is clearer?

storage-manager/src/IOCoordinator.cpp
```diff
@@ -211,29 +215,23 @@ ssize_t IOCoordinator::read(const char* _filename, uint8_t* data, off_t offset,
...
-      int l_errno = errno;
-      fileLock.unlock();
-      cache->doneReading(firstDir, keys);
-      logger->log(LOG_CRIT, "IOCoordinator::read(): Got an unexpected error opening %s, error was '%s'",
-                  jFilename.c_str(), strerror_r(l_errno, buf, 80));
-      errno = l_errno;
-      return -1;
-    }
+    // else
+    // Journal already merged with object file.
```
drrtuy, line 230: Remove if you don't need.

storage-manager/src/IOCoordinator.cpp
```diff
@@ -857,7 +855,14 @@ int IOCoordinator::_truncate(const bf::path& bfpath, size_t newSize, ScopedFileL
     if (result & 0x1)
       replicator->remove(cachePath / firstDir / objects[i].key);
     if (result & 0x2)
-      replicator->remove(journalPath / firstDir / (objects[i].key + ".journal"));
+    {
+      const auto journalName =
```
drrtuy, line 859: You have identical blocks here and at the line 913. Make these a function.

storage-manager/src/IOCoordinator.cpp
```diff
@@ -1021,37 +1030,18 @@ int IOCoordinator::copyFile(const char* _filename1, const char* _filename2)
...
-
-  vector<pair<string, size_t> > newJournalEntries;
+  vector<pair<string, size_t>> newJournalEntries;
   ScopedReadLock lock(this, filename1);
   ScopedWriteLock lock2(this, filename2);
   MetadataFile meta1(metaFile1, MetadataFile::no_create_t(), false);
   MetadataFile meta2(metaFile2, MetadataFile::no_create_t(), false);
   vector<metadataObject> objects = meta1.metadataRead(0, meta1.getLength());
+
+  if (!meta1.exists())
```
drrtuy, line 1040: Why do you still need meta file?

storage-manager/src/IOCoordinator.cpp
```diff
@@ -1290,204 +1286,83 @@ std::shared_ptr<uint8_t[]> IOCoordinator::mergeJournal(const char* object, const
...
+  auto resultPair = blobReader.readBlob(kvStorage, journalPath);
+  if (!resultPair.first)
     return -1;
-  ScopedCloser s(journalFD);
 
+  const std::string& journalData = resultPair.second;
+  size_t journalOffset = 0;
   // grab the journal header and make sure the version is 1
-  std::shared_ptr<char[]> headertxt = seekToEndOfHeader1(journalFD, &l_bytesRead);
+  std::shared_ptr<char[]> headertxt = seekToEndOfHeader1_(journalData, &l_bytesRead);
```
drrtuy, line 1318: I am missing why do you need to return shared_ptr with a copy of a header out of `seekToEndOfHeader1_`. This looks wasteful to me. You need an offset to copy the offset bytes into stringistream, right @denis0x0D?

utils/fdb_wrapper_cpp/include/fdbcs.hpp
denis0x0D: The assert is using here is to prevent situation when we can choose a key and datablock size which could cause undefined behavior - division by zero and so on. Those params are not configurable during runtime, so it's only a compile time options and only for debug build.

utils/fdb_wrapper_cpp/include/fdbcs.hpp
denis0x0D: During the stage when we build a tree, we periodically commit the parts of the tree when the size of the keys and values of that tree, exceeded the transaction size limit, this is fdb internal restriction. The strings ctro/dctor is not affected this.

utils/fdb_wrapper_cpp/src/fdbcs.cpp
denis0x0D: If the future has an error, we need to handle it/

utils/fdb_wrapper_cpp/src/fdbcs.cpp
denis0x0D: We have to identify the data block and the key block, every block has special header starting from the beginning of the block, so this function returns true when we processing the data block and false otherwise.

utils/fdb_wrapper_cpp/src/fdbcs.cpp
denis0x0D: Not really, the semantics of this function is following - 
1) we have a block to insert with fixed size.
2) we have a blob with fixed size, could be greater than block 
3) we have offset in blob from where we have to start to insert data.

a)If the size of the remaining data in blob is greater than block size, we insert the block until the end.
b)If the size of the remaining data in blob is less than block size, that means that is a last block, because blob is in the end, and we do not need this block for insertion anymore.
Conclusion: so we do not need to increment `block.first` for data block, this is different comparing to the key block.

storage-manager/src/PrefixCache.cpp
denis0x0D: Sure, I will do in in patch related to `journal to meta`

storage-manager/src/Replicator.cpp
denis0x0D: Yes, this file operations for the cache files, of files loaded from S3

utils/fdb_wrapper_cpp/src/fdbcs.cpp
denis0x0D: Data blocks are placed as a leafes of that tree, it means we have to reach the lowest level of the tree. 
Iterating using BFS, if we reach first data blocks, it means all blocks on that level represents data.

drrtuy (changes requested):

storage-manager/src/MetadataFile.cpp
denis0x0D: This function emulates a stat, default stat returns stat info struct, but since we do not have files anymore, we need to emulate it somehow

denis0x0D:
This one is closed because of the stand alone PRS for each MCOL

PR #3350, opened 2024-11-11
fix(DEC):MCOL-5805,5808 to resolve UM-only node crash inside DEC when there is no local PP to send the local requests to.

drrtuy:
Need to revert 5602 patch b/c it introduces an error, namely the number of messages to ack in sendAcks msg are not set. I will re-create 5602.

### DIFF
--- a/dbcon/joblist/distributedenginecomm.cpp
+++ b/dbcon/joblist/distributedenginecomm.cpp
@@ -181,6 +181,17 @@ struct QueueShutdown
 
 namespace joblist
 {
+
+bool isDefaultLocalConnectionId(const uint32_t connectionId)
+{
+  return defaultLocalConnectionId() == connectionId;
+}
+
+bool needToSendToLocalPM(const bool isExeMgr, const uint32_t connectionId)
+{
+  return isExeMgr && !isDefaultLocalConnectionId(connectionId);
+}
+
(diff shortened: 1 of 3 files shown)


mariadb-AlexeyAntipovsky (approved):
lgtm

dbcon/joblist/distributedenginecomm.cpp
```diff
@@ -417,14 +430,14 @@ void DistributedEngineComm::Listen(boost::shared_ptr<MessageQueueClient> client,
...
-    ++map_tok->second->unackedWork[0];
+    (void)atomicops::atomicInc(&map_tok->second->unackedWork[0]);
     map_tok->second->queue.push(sbs);
   }
   lk.unlock();
 
   if (fIsExeMgr)
   {
-    const uint32_t originalPMCount = pmCount;
+    decltype(pmCount) originalPMCount = pmCount;
```
mariadb-DenisKhalikov, line 440: Why not just `const auto`?

mariadb-DenisKhalikov (approved):

dbcon/joblist/distributedenginecomm.cpp
drrtuy: Originally that was written before we have C++17/20.

PR #3351, opened 2024-11-19
MCOL-5802 Add support for blob insertion into FDB.

denis0x0D:


drrtuy (approved):
untested ACK

utils/fdb_wrapper_cpp/include/fdbcs.hpp
```diff
@@ -21,7 +21,13 @@
...
 #include <thread>
 #include <memory>
-
+#include <vector>
+#include <unordered_map>
+#include <boost/uuid/uuid.hpp>
+#include <boost/uuid/uuid_io.hpp>
+#include <boost/uuid/random_generator.hpp>
+#include <boost/lexical_cast.hpp>
+#
```
mariadb-AlexeyAntipovsky: Huh?

utils/fdb_wrapper_cpp/src/fdbcs.cpp
```diff
@@ -19,10 +19,20 @@
...
+#include <chrono>
+#include "fdbcs.hpp"
 
 namespace FDBCS
 {
+
+#define RETURN_ON_ERROR(retcode) \
+  if (!retcode)                  \
+    return retcode;
+
```
mariadb-AlexeyAntipovsky: I like this variant because it is less error-prone:
```suggestion
#define RETURN_ON_ERROR(retcode) do {\
    typeof(retcode) rc = (retcode);  \
    if (!rc) {                       \
      return (rc);                   \
    }                                \
  } while (false)
```

For example, with this macro, we will not do the `commitKeys` twice if an error occurs on line 350.

mariadb-AlexeyAntipovsky (changes requested):

utils/fdb_wrapper_cpp/src/fdbcs.cpp
denis0x0D: Thanks, fixed

mariadb-AlexeyAntipovsky (approved):

PR #3356, opened 2024-11-29
MCOL-5844: pass ubsan checks

mariadb-SergeyZefirov:
### fix(ubsan): MCOL-5844 - iron out UBSAN reports

The most important fix here is the fix of possible buffer overrun in
DATEFORMAT() function. A "%W" format, repeated enough times, would
overflow the 256-bytes buffer for result. Now we use ostringstream to
construct result and we are safe.

Changes in date/time projection functions made me fix difference between us and server behavior. The new, better behavior is reflected in changes
in tests' results.

Also, there was incorrect logic in TRUNCATE() and ROUND() functions in  computing the decimal "shift."


mariadb-LeonidFedorov (approved):

utils/loggingcpp/message.cpp
```diff
@@ -127,6 +127,30 @@ void Message::Args::add(uint64_t u64)
   fArgs.push_back(u64);
 }
 
+void Message::Args::add(int128_t i128)
```
drrtuy, line 132: I sppose you can re-use [this routine.](https://github.com/mariadb-corporation/mariadb-columnstore-engine/blob/develop/datatypes/mcs_int128.cpp#L77)

drrtuy (commented):
I suggest to re-use TSInt128::toString() in logging.

PR #3360, opened 2024-12-02
fix(ubsan): MCOL-5844 - iron out UBSAN reports

mariadb-SergeyZefirov:
The most important fix here is the fix of possible buffer overrun in DATEFORMAT() function. A "%W" format, repeated enough times, would overflow the 256-bytes buffer for result. Now we use ostringstream to construct result and we are safe.

Changes in date/time projection functions made me fix difference between us and server behavior. The new, better behavior is reflected in changes in tests' results.

Also, there was incorrect logic in TRUNCATE() and ROUND() functions in computing the decimal "shift."


mariadb-SergeyZefirov:
This is same as https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/3356 but for stable and properly squashed and rebased.

mariadb-LeonidFedorov (approved):

PR #3364, opened 2024-12-03
Feat/em join algo parallel exec 23.10

drrtuy:
This increases granularity of workload in terms of Primitive Jobs sent by EM to PP. Previously the min unit was a workload unit worth 1 extent now one can have lesser, fractional units.


mariadb-AlexeyAntipovsky (approved):
lgtm

PR #3370, opened 2024-12-13
fix(MTR tests): Fixes autopilot MTR tests (#3368)

mariadb-SergeyZefirov:
Fixes in UBSAN related commit introduced more server-compatible behavior that differ fom our old behavior. Thus, old tests broke and their results had to be changed. This is what this patch does.


drrtuy (approved):
untested ACK

PR #3371, opened 2024-12-17
fix(group by, having): MCOL-5776: GROUP BY/HAVING closer to server's

mariadb-SergeyZefirov:
This patch introduces an internal aggregate operator SELECT_SOME that is automatically added to columns that are not in GROUP BY. It "computes" some plausible value of the column (actually, last one passed).

Along the way it fixes incorrect handling of HAVING being transferred into WHERE, window function handling and a bit of other inconsistencies.


dbcon/joblist/tuplehavingstep.cpp
```diff
@@ -125,9 +127,11 @@ void TupleHavingStep::expressionFilter(const ParseTree* filter, JobInfo& jobInfo
   ExpressionStep::expressionFilter(filter, jobInfo);
 
   // extract simple columns from parse tree
+#if 01
```
drrtuy, line 130: What is this for?

drrtuy (approved):
untested ACK

PR #3373, opened 2024-12-23
[FDB BlobHandler] Add writeOrUpdate, update `read` and `remove`.

denis0x0D:
This patch updates `read` and `remove` to take in account the size of the `keys` and `values` for one FDB transaction.


drrtuy (approved):
untested ACK

PR #3378, opened 2024-12-27
MCOL-5133: Stage1 to stand alone cli tool.

mariadb-AlanMologorsky:
- [x] [add] cluster api client class
- [x] [fix] cli cluster_app using cluster api client
- [x] [add] ClusterAction Enum
- [x] [add] toggle_cluster_state function to reduce code duplication

### DIFF
--- a/cmapi/cmapi_server/constants.py
+++ b/cmapi/cmapi_server/constants.py
@@ -82,3 +82,10 @@ MCS_INSTALL_BIN = '/usr/bin'
 IFLAG = os.path.join(MCS_ETC_PATH, 'container-initialized')
 LIBJEMALLOC_DEFAULT_PATH = os.path.join(MCS_DATA_PATH, 'libjemalloc.so.2')
 MCS_LOG_PATH = '/var/log/mariadb/columnstore'
+
+
+# client constants
+CMAPI_PORT = 8640  #TODO: use it in all places
+CURRENT_NODE_CMAPI_URL = f'https://localhost:{CMAPI_PORT}'
+REQUEST_TIMEOUT: float = 30.0
+TRANSACTION_TIMEOUT: float = 300.0  # 5 minutes
--- a/cmapi/cmapi_server/managers/transaction.py
+++ b/cmapi/cmapi_server/managers/transaction.py
@@ -6,7 +6,7 @@ from signal import (
 )
 from typing import Optional, Type
 
-from cmapi_server.constants import DEFAULT_MCS_CONF_PATH
+from cmapi_server.constants import DEFAULT_MCS_CONF_PATH, TRANSACTION_TIMEOUT
 from cmapi_server.exceptions import CMAPIBasicError
 from cmapi_server.helpers import (
     get_id, commit_transaction, rollback_transaction, start_transaction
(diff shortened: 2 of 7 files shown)


cmapi/mcs_cluster_tool/cluster_app.py
```diff
@@ -157,29 +158,31 @@ def stop(
         # TODO: investigate more on how changing the hardcoded timeout
         #       could affect put_config (helpers.py broadcast_config) operation
         timeout = 0
-    _ = ClusterHandler.shutdown(logger=logger, in_transaction=True)
+
+    resp = client.shutdown_cluster({'in_transaction': True})
```
drrtuy, line 163: How much time does the shutdown_cluster wait till it returns?

drrtuy: Does it take user-specified graceful timeouts into account or this logic is expressed in a different manner?

cmapi/cmapi_server/managers/transaction.py
```diff
@@ -17,7 +17,7 @@ class TransactionManager(ContextDecorator):
     """Context manager and decorator to put any code inside CMAPI transaction.
 
     :param timeout: time in sec after transaction will be autocommitted,
-                    defaults to 300.0
+                    defaults to 300.0 (TRANSACTION_TIMEOUT)
```
drrtuy, line 20:
```suggestion
                    defaults to TRANSACTION_TIMEOUT
```

cmapi/cmapi_server/constants.py
```diff
@@ -82,3 +82,8 @@ class ProgInfo(NamedTuple):
 IFLAG = os.path.join(MCS_ETC_PATH, 'container-initialized')
 LIBJEMALLOC_DEFAULT_PATH = os.path.join(MCS_DATA_PATH, 'libjemalloc.so.2')
 MCS_LOG_PATH = '/var/log/mariadb/columnstore'
+
+
+# client constants
+REQUEST_TIMEOUT: float = 330.0  # 5,5 minutes
```
drrtuy: We need to experiment but to me these values are method-specific, e.g. `shutdown_cluster` should have it equal to graceful_timeout, `start_cluster` can have it as 330 secs,  but `set_mode`, `get_status`, should have lower values, so if it fails, it fails faster(30 secs should be enough). `load_s3` on contrary must have huge value here.

drrtuy (changes requested):
Plz address the comments

cmapi/mcs_cluster_tool/cluster_app.py
mariadb-AlanMologorsky: Fixed after discussing. Added timeouts.

cmapi/cmapi_server/managers/transaction.py
mariadb-AlanMologorsky: Fixed in another request.

cmapi/cmapi_server/constants.py
mariadb-AlanMologorsky: Fixed, added timeouts.

drrtuy (approved):
untested ACK

PR #3381, opened 2025-01-14
MCOL-5881 set/getThreadName use FreeBSD API

grooverdan:
Taken from FreeBSD ports, this uses the FreeBSD
APIs rather than the Linux specific prctl to change and retreive the thread names.

### DIFF
--- a/utils/common/threadnaming.cpp
+++ b/utils/common/threadnaming.cpp
@@ -22,13 +22,21 @@ namespace utils
 {
 void setThreadName(const char* threadName)
 {
+#ifdef __FreeBSD__
+  pthread_set_name_np(pthread_self(), threadName);
+#else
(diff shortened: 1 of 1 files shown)


drrtuy (approved):
untested ACK

drrtuy:
@grooverdan Which server releases do you want to propagate this?
Could you make PRs against the engine's `stable-23.10` and `develop`?

grooverdan:
> @grooverdan Which server releases do you want to propagate this?

I did here because of the FreeBSD patch against the 10.6 server. Remember you saying the develop-5 was important maintenance only, so assuming the below two are the other maintained branches.

> Could you make PRs against the engine's `stable-23.10` and `develop`?

done.

Thank you.

PR #3384, opened 2025-01-14
MCOL-5881 set/getThreadName use FreeBSD API

grooverdan:
Taken from FreeBSD ports, this uses the FreeBSD
APIs rather than the Linux specific prctl to change and retreive the thread names.


mariadb-LeonidFedorov:
LGTM, will be merged to this branch after the release 23.10.3 delivered

drrtuy (approved):
untested ACK

PR #3386, opened 2025-01-23
Columnstore 5.6.8 2

drrtuy:


vuvova:
Is it just a cherry-pick of my old commits? Then it's ok, approved. Did you make any changes and want me to take a look?

drrtuy:
These are some commits that are missing in the main branch you are going to use for 10.5, so yes, we have to merge these.

PR #3388, opened 2025-01-24
MCOL-5133: Stage2 stand alone cli tool

mariadb-AlanMologorsky:
- [x] [fix] client class methods
- [x] [fix] Transaction management at endpoint handling level
- [x] [fix] add_node and set_mode methods to use TransactionManager

### DIFF
--- a/cmapi/cmapi_server/handlers/cluster.py
+++ b/cmapi/cmapi_server/handlers/cluster.py
@@ -108,18 +108,12 @@ class ClusterHandler():
         return response
 
     @staticmethod
-    def start(
-        config: str = DEFAULT_MCS_CONF_PATH, in_transaction: bool = False
-    ) -> dict:
+    def start(config: str = DEFAULT_MCS_CONF_PATH) -> dict:
         """Method to start MCS Cluster.
 
         :param config: columnstore xml config file path,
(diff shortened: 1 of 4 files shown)


cmapi/cmapi_server/handlers/cluster.py
```diff
@@ -160,29 +146,19 @@ def shutdown(
             'Cluster shutdown command called. Shutting down the cluster.'
         )
         operation_start_time = str(datetime.now())
-        if not in_transaction:
-            with TransactionManager():
-                toggle_cluster_state(ClusterAction.STOP, config)
-        else:
-            toggle_cluster_state(ClusterAction.STOP, config)
-
+        toggle_cluster_state(ClusterAction.STOP, config)
```
drrtuy, line 149: I am missing why don't we want TransactionManager here?

drrtuy (commented):
Plz comment on the question

cmapi/cmapi_server/handlers/cluster.py
mariadb-AlanMologorsky: Discussed. 
Tramsaction Manager used in a top level calls.

drrtuy (approved):
untested ACK

PR #3397, opened 2025-01-31
MCOL-5133: Stage3 stand alone cli tool.

mariadb-AlanMologorsky:
MAJOR: Some logic inside node remove changed significantly using active 
    nodes list from Columnstore.xml to broadcast config after remove.

 - [x] [fix] remove node logic to use Transaction manager
 - [x] [fix] cluster set api key call using totp on a top level cli call
 - [x] [add] missed docstrings

### DIFF
--- a/cmapi/cmapi_server/controllers/api_clients.py
+++ b/cmapi/cmapi_server/controllers/api_clients.py
@@ -1,7 +1,13 @@
 import requests
 from typing import Any, Dict, Optional, Union
+
+import pyotp
+
 from cmapi_server.controllers.dispatcher import _version
-from cmapi_server.constants import CURRENT_NODE_CMAPI_URL
+from cmapi_server.constants import (
+    CMAPI_CONF_PATH, CURRENT_NODE_CMAPI_URL, SECRET_KEY,
+)
+from cmapi_server.helpers import get_config_parser, get_current_key
 
 
 class ClusterControllerClient:
@@ -10,6 +16,10 @@ class ClusterControllerClient:
             request_timeout: Optional[float] = None
     ):
         """Initialize the ClusterControllerClient with the base URL.
+
+        WARNING: This class only handles the API requests, it does not
+        handle the transaction management. So it should be started
+        at level above using TransactionManager (decorator or context manager).
 
         :param base_url: The base URL for the API endpoints,
                          defaults to CURRENT_NODE_CMAPI_URL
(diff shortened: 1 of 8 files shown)


cmapi/cmapi_server/handlers/cluster.py
```diff
@@ -300,76 +221,45 @@ def remove_node(
...
             )
         except Exception as err:
-            rollback_transaction(
-                transaction_id, nodes=txn_nodes, cs_config_filename=config
-            )
             raise CMAPIBasicError('Error while removing node.') from err
 
         response['node_id'] = node
-        if len(txn_nodes) > 0:
+        active_nodes = get_active_nodes(config)
```
drrtuy, line 240: Could you elaborate on the reasons for this change @mariadb-AlanMologorsky ?

cmapi/cmapi_server/controllers/api_clients.py
```diff
@@ -0,0 +1,129 @@
...
+        :return: The response from the API.
+        """
+        url = f'{self.base_url}/cmapi/{_version}/cluster/{endpoint}'
+        try:
+            response = requests.request(
+                method, url, json=data, timeout=self.request_timeout
+            )
+            response.raise_for_status()
+            return response.json()
+        # TODO: different handler for timeout exception?
```
drrtuy: Good idea.
Could you point me where do we release the lock if something went south during execution of a command on a cluster ?

cmapi/cmapi_server/handlers/cluster.py
mariadb-AlanMologorsky: 1. Now TransactionManager class used and this value is out of inner function scope.
2. While I investigate how to make it visible I made a valuable conclusion that all the time it was wrong logic:
      Checking number of nodes which had successful start of transaction after removing one of them is invalid by default. At least there are corner case with removing last one node. I used active nodes from updated MCS  configuration instead.

cmapi/cmapi_server/controllers/api_clients.py
mariadb-AlanMologorsky: Api client itself just calling remote endpoint, thansaction mechanics never used here. Otherwise it's used on a top level. For eg. [here](https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/3397/files#diff-b6b767592660fba535cbc6c2668074f5e05457be70ecde8abcb3c0ff0c4d6a22R170)

drrtuy (approved):
untested ACK

PR #3398, opened 2025-02-04
Mcol 5886 support foreign table partitions

mariadb-SergeyZefirov:


dbcon/joblist/joblist.cpp
```diff
@@ -672,7 +672,7 @@ void JobList::graph(uint32_t sessionID)
     else if (typeid(*(qsi->get())) == typeid(CrossEngineStep))
     {
       tcn.schema = qsi->get()->schema();
-      tcn.table = qsi->get()->alias();
+      tcn.table = qsi->get()->alias(); // XXX is it correct? why not table name?
```
drrtuy: Alias are more specific than tables. People tend to know what do  their aliases mean in the context of a query, even in case of CEJ.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -313,13 +315,59 @@ void convertOuterJoinToInnerJoin(List<TABLE_LIST>* join_list, TableOnExprList& t
...
+  if (table->part_info)
+  {
+    List_iterator<partition_element> part_el_it(table->part_info->partitions);
+
+    partition_element* pe;
+
+    while ((pe = part_el_it++)) // this is how server does it.
+    {
+      std::string pn(pe->partition_name);
+      result.fPartNames.push_back(pn);
```
drrtuy, line 331: emplace_back?

drrtuy:
```suggestion
      result.fPartNames.emplace_back(pe->partition_name);
```

dbcon/execplan/calpontsystemcatalog.cpp
```diff
@@ -6319,5 +6328,47 @@ bool ctListSort(const CalpontSystemCatalog::ColType& a, const CalpontSystemCatal
...
+    if (a.fPartNames[i] < b.fPartNames[i])
+    {
+      return true;
+    }
+    if (a.fPartNames[i] > b.fPartNames[i])
+    {
+      return false;
+    }
+  }
+  if (l > a.fPartNames.size())
```
drrtuy: How this and line 6350 can be true? `l` is always <= both `fPartNames.size()` IMHO. Am I missing something here?

drrtuy (commented):
I would be delighted to see the reply on `bool operator <(const Partitions& a, const Partitions& b)` code.

dbcon/execplan/calpontsystemcatalog.cpp
mariadb-SergeyZefirov: Yes, it should be `i`.

mariadb-SergeyZefirov: No, `i` is equal to `l` here, the comparisons should be opposite - not "is `l` greater than `some.size()`," but "is `l` less than `some.size()`".

dbcon/mysql/ha_mcs_execplan.cpp
mariadb-SergeyZefirov: Thanks!

dbcon/execplan/calpontsystemcatalog.cpp
mariadb-SergeyZefirov: Fixed.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -313,13 +315,58 @@ void convertOuterJoinToInnerJoin(List<TABLE_LIST>* join_list, TableOnExprList& t
...
+  if (table->partition_names)
+  {
+    List_iterator<String> part_name_it(*(table->partition_names));
+
+    String* n;
+
+    while ((n = part_name_it++)) // this is how server does it.
+    {
+      std::string pn(n->ptr(), n->length());
+      result.fPartNames.push_back(pn);
```
drrtuy, line 348: You can use emplace_back here as well. Just call a different string ctor.

drrtuy, line 348:
```suggestion
      result.fPartNames.emplace_back({n->ptr(), n->length()});
```

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -313,13 +315,58 @@ void convertOuterJoinToInnerJoin(List<TABLE_LIST>* join_list, TableOnExprList& t
...
+    }
+  }
+  return result;
+}
+
+
+CalpontSystemCatalog::TableAliasName makeTableAliasName_(TABLE_LIST* table)
```
drrtuy, line 355: `makeTableAliasName_` is a peculiar choice TBH. Why do you need a prefix? 
If we adding `TableAliasName::partitions` attribute can we add another ctor that uses `getPartitions` internally?

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -313,13 +315,58 @@ void convertOuterJoinToInnerJoin(List<TABLE_LIST>* join_list, TableOnExprList& t
...
 {
   return make_aliasview(
       (table->db.length ? table->db.str : ""), (table->table_name.length ? table->table_name.str : ""),
       (table->alias.length ? table->alias.str : ""), getViewName(table), true, lower_case_table_names);
 }
 
+CalpontSystemCatalog::TableAliasName makeTableAliasName(TABLE_LIST* table)
+{
+  CalpontSystemCatalog::TableAliasName result = makeTableAliasName_(table);
+  result.partitions = getPartitions(table);
```
drrtuy, line 365: Can you explicitly apply move semantics on the rvalue here just in case?

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -2658,6 +2706,7 @@ SimpleColumn* buildSimpleColFromDerivedTable(gp_walk_info& gwi, Item_field* ifp)
           }
           sc->resultType(cols[j]->resultType());
           sc->hasAggregate(cols[j]->hasAggregate());
+	  // XXX partitions???
```
drrtuy, line 2709: Do we need to copy partitions here also?

drrtuy:
@mariadb-SergeyZefirov any updates on this?

drrtuy (changes requested):
Could you port this into stable-23.10 @mariadb-SergeyZefirov ?

dbcon/mysql/ha_mcs_execplan.cpp
mariadb-SergeyZefirov: ```If we adding TableAliasName::partitions attribute can we add another ctor that uses getPartitions internally?```

What do you mean by that?

mariadb-SergeyZefirov:
> Could you port this into stable-23.10

I will do.

PR #3407, opened 2025-02-14
Migration for Boost 1.85

arkamar:
Gentoo users cannot compile MariaDB with columnstore because the oldest available Boost version in Gentoo is 1.85, see https://bugs.gentoo.org/932735. Migration seems to be straightforward, however, I noticed PR #3394, therefore, I am not sure, if this change is feasible for others.

I have set the base to `develop-6` because I did the migration for Mariadb 10.6.21 and 10.11.11, both versions compile and pass tests. The change seems to apply to `develop` cleanly as well, but I didn't test it properly yet.

---

Boost 1.85 removed some deprecated code in filesystem module which is still used in columnstore:

- The boost/filesystem/convenience.hpp was removed but columnstore does not use any functionality from that file except indirect includes. Therefore this include is removed or replaced with more general boost/filesystem.hpp. The convenience.hpp header file was deprecated in filesystem V3 introduced in Boost 1.46.0.
- `normalize` method was removed and users are suggested to replace it with `lexically_normal` method, which was introduced in Boost 1.60.0.
- The `copy_option` was removed in favor of `copy_options` (note the trailing 's'), but enum values were renamed. Namely, `fail_if_exists` is replaced with `none` and `overwrite_if_exists` is replaced with `overwrite_existing`. The `copy_options` was introduced in Boost 1.74.0

### DIFF
--- a/oamapps/columnstoreSupport/columnstoreSupport.cpp
+++ b/oamapps/columnstoreSupport/columnstoreSupport.cpp
@@ -30,6 +30,7 @@
 #include <readline.h>
 #include <boost/filesystem.hpp>
 
+#include "boost_copy_options_compat.hpp"
 #include "mcsconfig.h"
 #include "liboamcpp.h"
 #include "configcpp.h"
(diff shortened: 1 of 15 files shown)


drrtuy:
Thx for the contribution @arkamar. 
Pardon my ignorance which MDB versions do you now have in Gentoo? The actual Columnstore that comes with MariaDB > 11.1 uses a local version of boost(currently at 1.84 and defined [here](https://github.com/mariadb-corporation/mariadb-columnstore-engine/blob/stable-23.10/cmake/boost.cmake#L40)).
Given that Columnstore depends on system boost I assume this must be 10.6 but I want to double check this fact.  Could you confirm the MariaDB version we are talking about?

arkamar:
We are talking about MariaDB 10.6 and 10.11 with respect to this PR.

I will apply this patch locally. My main intention with this PR is to increase awareness that the Boost 1.85 dropped some deprecated code. Maybe other downstreams might find this PR useful as well.

drrtuy:
@arkamar Could I ask you to put preprocessor `ifdef` around the changes, so that your patch doesn't affect those who are limited by older boost versions available in their distros?

drrtuy (changes requested):
Plz add preprocessor `ifdef` to take into account those, who use older boost versions.

arkamar:
I have added preprocessor guards. It looked a bit awkward to do it around all `copy_option` places, therefore, I created backward compatibility layer in separate header file. New form is always preferred with respect to a boost version, where it was introduced. I tried to build MariaDB 10.6.21 with this patch and with Boost 1.53.0, 1.73.0 and 1.87.0, everything builds fine and tests pass.

The commit message is updated with respect to recent changes.

arkamar:
It was quite an effort to install old Boost 1.53.0 to recent Gentoo :)

drrtuy:
Thank you for the work on this!
I can imagine how problematic it was to get 1.53 installed on Gentoo :).

drrtuy (approved):
untested ACK

mariadb-LeonidFedorov (approved):

arkamar:
Thanks for fast merge!

@drrtuy Will this change be applied to MariaDB server repo automatically or should I do something more?

PR #3413, opened 2025-02-21
fix(MCOL-4942): No-table-SELECT now can return empty set

mariadb-SergeyZefirov:
The query like "SELECT 1 WHERE 1=0" was returning a row despite unsatisfiable condition in WHERE. Now it returns an empty set.


mariadb-LeonidFedorov (approved):

dbcon/joblist/tupleconstantstep.cpp
```diff
@@ -585,7 +587,9 @@ void TupleConstantStep::formatMiniStats()
 }
 
 // class TupleConstantOnlyStep
-TupleConstantOnlyStep::TupleConstantOnlyStep(const JobInfo& jobInfo) : TupleConstantStep(jobInfo)
+TupleConstantOnlyStep::TupleConstantOnlyStep(const JobInfo& jobInfo)
+  : TupleConstantStep(jobInfo)
+  , fEmptySet(jobInfo.constantFalse)
```
drrtuy, line 592: I would call this impossible condition b/c there are a variety of reasons when empty set can delivered.

mysql-test/columnstore/bugfixes/MCOL-4942-constant-select-empty-set.test
```diff
@@ -0,0 +1,40 @@
...
+	UNION ALL
+
+	SELECT ID
+
+	FROM
+
+	(
+
+		SELECT NULL ID WHERE 1111=2222
+
```
drrtuy, line 35: Extra empty lines.

drrtuy (commented):
Plz address my comments.
untestedACK

dbcon/joblist/tupleconstantstep.cpp
mariadb-SergeyZefirov: I do not get it. Can you explain?

mysql-test/columnstore/bugfixes/MCOL-4942-constant-select-empty-set.test
mariadb-SergeyZefirov: I have moved reproduction query into the test as it is. That's why there are extra empty lines.

dbcon/joblist/tupleconstantstep.cpp
drrtuy: The naming emptySet first:
- renames jobInfo.constantFalse so that we loose connection b/w jobInfo.constantFalse and new field
- emptySet doesn't tell me why exactly do we get an empty set.
What if we preserve the naming using TupleConstantOnlyStep::constantFalse ?

mariadb-SergeyZefirov: This flag governs decision to output no data.

While JobInfo contains a fact that condition is constant false in an appropriately named field, this flag contains a decision to output empty set result and as such is appropriately named.

PR #3415, opened 2025-02-24
fix(MCOL-4942): No-table-SELECT now can return empty set

mariadb-SergeyZefirov:
The query like "SELECT 1 WHERE 1=0" was returning a row despite unsatisfiable condition in WHERE. Now it returns an empty set.


drrtuy (approved):
untested ACK

PR #3417, opened 2025-02-25
Allow packagers to use system Thrift library

arkamar:
The commit introduces a new CMake configuration option, `WITH_THRIFT`, which accepts "auto", "system", or "bundled". In "auto" mode (the default), the build system attempts to use the system-installed Thrift library and falls back to the bundled version if the system library is not available. Setting `WITH_THRIFT` to "system" enforces the use of the system Thrift, causing the configuration to fail if it isn't found, while "bundled" forces the use of the bundled version.

The change mainly useful for downstream maintainers as it gives them flexibility over dependency management.

Downstream-issue: https://bugs.gentoo.org/949680

---

I would ideally like to have this backported to Mariadb 11.4 as a packager of MariaDB in Gentoo, which means `stable-23.10` branch in this repo.

### DIFF
--- a/cmake/thrift.cmake
+++ b/cmake/thrift.cmake
@@ -1,3 +1,19 @@
+set(WITH_THRIFT "auto" CACHE STRING
+  "Which Thrift to use (possible values are 'bundled', 'system', or 'auto')")
+
+if(WITH_THRIFT STREQUAL "system" OR WITH_THRIFT STREQUAL "auto")
+  FIND_PACKAGE(Thrift)
+
+  if (Thrift_FOUND)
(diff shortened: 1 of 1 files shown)


drrtuy:
Hey @arkamar!
Thx for the contribution. 
Could you make a clone of this for `stable-23.10` also?

mariadb-LeonidFedorov (approved):
Thank you for your contribution!
We didn't test it out of scope of our pipelines, but it works for our build conveyer. Feel free to make the duplicate for stable-23.10 branch

arkamar:
Thanks for merge! I have opened backporting PR #3418.

PR #3419, opened 2025-02-28
fix(MCOL-5842): Fix JSON_OBJECT's handling of empty strings

mariadb-SergeyZefirov:
JSON_OBJECT() (and probably some other JSON functions) now properly handle empty strings in their arguments - JSON_OBJECT used to return NULL, now it returns empty string.


drrtuy:
We should do this for stable-23.10 @mariadb-SergeyZefirov as well.

drrtuy (approved):
untested ACK

PR #3420, opened 2025-03-03
fix(MCOL-5396): Fix possible infinite loop in plugin--PrimProc commun…

mariadb-SergeyZefirov:
…ication

If you manage to shut down PrimProc just before plugin is trying to send Calpont Select Execution Plan to PrimProc, you now get a nice error message about PrimProc being down instead of endless logs of failed reconnection attempts.


mariadb-SergeyZefirov:
There is no test because timing to reproduce is too tight and reproduction needs possible performance-interfering code changes.

mariadb-LeonidFedorov (approved):

PR #3421, opened 2025-03-03
fix(MCOL-5396): Fix possible infinite loop in plugin--PrimProc commun…

mariadb-SergeyZefirov:
…ication

If you manage to shut down PrimProc just before plugin is trying to send Calpont Select Execution Plan to PrimProc, you now get a nice error message about PrimProc being down instead of endless logs of failed reconnection attempts.

### DIFF
--- a/dbcon/mysql/ha_mcs_impl.cpp
+++ b/dbcon/mysql/ha_mcs_impl.cpp
@@ -19,6 +19,8 @@
 #define PREFER_MY_CONFIG_H
 #include <my_config.h>
 #include <unistd.h>
+#include <chrono>
+#include <thread>
 #include <string>
 #include <iostream>
 #include <stack>
@@ -129,6 +131,7 @@ using namespace funcexp;
 #include "statistics.h"
 #include "ha_mcs_logging.h"
 #include "ha_subquery.h"
+
 
 namespace cal_impl_if
 {
@@ -2352,8 +2355,18 @@ int ha_mcs::impl_rnd_init(TABLE* table, const std::vector<COND*>& condStack)
     ByteStream msg;
     ByteStream emsgBs;
 
+    int ntries = 10;
+
+    // XXX: MCOL-5396: unable to reach this code.
     while (true)
     {
+      string emsg;
+      if (ntries < 0)
+      {
+        emsg = "Lost connection to ExeMgr. Please contact your administrator";
+        setError(thd, ER_INTERNAL_ERROR, emsg);
+        return ER_INTERNAL_ERROR;
+      }
       try
       {
         ByteStream::quadbyte qb = 4;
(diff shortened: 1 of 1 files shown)


dbcon/mysql/ha_mcs_impl.cpp
```diff
@@ -2437,6 +2449,10 @@ int ha_mcs::impl_rnd_init(TABLE* table, const std::vector<COND*>& condStack)
 
         ti.conn_hndl = hndl;
 
+	using namespace std::chrono_literals;
```
drrtuy: We should either use this one time in the code or specify the full path using the constant.

dbcon/mysql/ha_mcs_impl.cpp
```diff
@@ -4943,8 +4973,23 @@ int ha_mcs_impl_pushdown_init(mcs_handler_info* handler_info, TABLE* table, bool
...
     ByteStream emsgBs;
 
+    int ntries = 10;
+
+    // The issue is MCOL-5396.
+    // The delay below is used to trigger old infinite loop condition and
+    // prove that mitigation works. While it looks like unused code, it is
+    // important enough to have it just in case.
+    // using namespace std::chrono_literals;
+    // std::this_thread::sleep_for(10000ms);
```
mariadb-LeonidFedorov: please remove unused code

mariadb-SergeyZefirov: The code is not unused - it is part of comment on how to reproduce the condition. Thus, I would like to have it there.

dbcon/mysql/ha_mcs_impl.cpp
mariadb-SergeyZefirov: Ah, I get it.

dbcon/mysql/ha_mcs_impl.cpp
drrtuy: You already put the note that you need sleep and then you add 2 extra lines: one with the code itself and another to explain this codeline is needed. Isn't it good enough to mention sleep and avoid the specific sleep syntax. What if the sleep syntax will change overtime for example?

mariadb-SergeyZefirov: I do my best to lower the workload of the (future) colleagues. In this particular case I relieve them from looking into documentation, this saves a minute or two, yet it saves.

The purpose is same or close to commented code next to FIXME in ha_mcs_execplan.cpp. To have code ready to be used yet disabled.

As for change of syntax, this is standard from C++11 and most probably set in stone.

mariadb-LeonidFedorov (approved):

mariadb-LeonidFedorov (changes requested):
build is failed

mariadb-LeonidFedorov (approved):

PR #3425, opened 2025-03-07
MCOL-5852: GROUP_CONCAT memory accounting and disk based version

mariadb-AlexeyAntipovsky:


mariadb-AlexeyAntipovsky:
I'll add some tests a bit later

utils/messageqcpp/inetstreamsocket.cpp
```diff
@@ -575,30 +534,19 @@ void InetStreamSocket::do_write(const ByteStream& msg, uint32_t whichMagic, Stat
...
     return;
 
-  const auto& longStrings = msg.getLongStrings();
   /* buf.fCurOutPtr points to the data to send; ByteStream guarantees that there
      are at least 12 bytes before that for the magic & length fields */
   realBuf = (uint32_t*)msg.buf();
   realBuf -= 3;
   realBuf[0] = magic;
   realBuf[1] = msglen;
-  realBuf[2] = longStrings.size();
```
drrtuy, line 585: I suppose you can remove this.

dbcon/joblist/groupconcat.cpp
```diff
@@ -716,40 +776,156 @@ const string GroupConcator::toString() const
...
+    bs >> s;
+    fConstCols.emplace_back(f, s);
+  }
+  bs >> fCurrentLength;
+  bs >> fGroupConcatLen;
+  bs >> fConstantLen;
+  bs >> fTimeZone;
+}
+
+class OrderByRow
```
drrtuy: Name overlap with the class that is used by ORDER BY, so you should name it differently IMHO. 
After this patch we can move both this sorting structs together with idborderby* into a specialized dir for a healty housekeeping.

drrtuy (commented):
We need a call on this. I am missing how did you replace LongStrings functionality removed from BS.

mariadb-SergeyZefirov:
Regarding ``uint64_t StringStore::storeString(const uint8_t* data, uint32_t len)``

If you change it to ``uint64_t StringStore::storeString(uint64_t oldOffset, const uint8_t* data, uint32_t len)`` then you may not add an additional data and replace old data if oldOffset is not equal to UINT64_MAX-1.

This will mitigate problem with TEXT columns and MIN/MAX/SOME aggregates - right now they accumulate data in StringStore, after this changes they won't.

I also have not completely understand changes related to disk aggregation. I should ask you about that during call.

mariadb-SergeyZefirov (commented):
I salute you in bringing the ``JSON_ARRAY_AGG`` into ``GROUP_CONCAT``. This is GREAT!

I also have generally approve the changes.

Yet, I would like to discuss changes to ``storeString`` - is it possible to  incorporate them into your changeset?

dbcon/joblist/groupconcat.cpp
mariadb-AlexeyAntipovsky: Good point, fixed

utils/messageqcpp/inetstreamsocket.cpp
mariadb-AlexeyAntipovsky: Then there will be a section of uninitialized memory, which valgrind will criticize me for. 8))

mariadb-AlexeyAntipovsky:
@mariadb-SergeyZefirov  There is still a lot of work to be done here to improve disk aggregation, so I will refactor StringStorage as part of this work.

PR #3432, opened 2025-03-12
MCOL-5861 cmapi read only nodes

mariadb-AlexanderPresniakov:
https://jira.mariadb.org/browse/MCOL-5861
Read-only nodes don't start WriteEngineServer and do not own dbroots. But as read-only nodes must have RO access to all the data in the cluster, all dbroots in the cluster are assigned all the dbroots.
Read-only nodes are written into ReadOnlyNodes part of the config.
On every change that could add or remove dbroots update_dbroots_of_readonly_nodes is called to assign all existing dbroots to every read-only node.

MCSProgs was added to avoid possible typos in service names (because there is no single naming convention between services and it is very easy to make an error)
Also added NodeConfig.modify_config context manager that reads config from specified path, provides its root and on (only successful exit) writes config to the output path

### DIFF
--- a/cmapi/cmapi_server/constants.py
+++ b/cmapi/cmapi_server/constants.py
@@ -77,17 +77,17 @@ class ProgInfo(NamedTuple):
 # on top level of process handling
 # mcs-storagemanager starts conditionally inside mcs-loadbrm, but should be
 # stopped using cmapi
-ALL_MCS_PROGS: dict[str, ProgInfo] = {
+ALL_MCS_PROGS: dict[MCSProgs, ProgInfo] = {
     # workernode starts on primary and non primary node with 1 or 2 added
     # to subcommand (DBRM_Worker1 - on primary, DBRM_Worker2 - non primary)
-    MCSProgs.STORAGE_MANAGER.value: ProgInfo(15, 'mcs-storagemanager', '', False, 1),
-    MCSProgs.WORKER_NODE.value: ProgInfo(13, 'mcs-workernode', 'DBRM_Worker{}', False, 1),
-    MCSProgs.CONTROLLER_NODE.value: ProgInfo(11, 'mcs-controllernode', 'fg', True),
-    MCSProgs.PRIM_PROC.value: ProgInfo(5, 'mcs-primproc', '', False, 1),
-    MCSProgs.EXE_MGR.value: ProgInfo(9, 'mcs-exemgr', '', False, 1),
-    MCSProgs.WRITE_ENGINE_SERVER.value: ProgInfo(7, 'mcs-writeengineserver', '', False, 3),
-    MCSProgs.DML_PROC.value: ProgInfo(3, 'mcs-dmlproc', '', False),
-    MCSProgs.DDL_PROC.value: ProgInfo(1, 'mcs-ddlproc', '', False),
+    MCSProgs.STORAGE_MANAGER: ProgInfo(15, 'mcs-storagemanager', '', False, 1),
+    MCSProgs.WORKER_NODE: ProgInfo(13, 'mcs-workernode', 'DBRM_Worker{}', False, 1),
+    MCSProgs.CONTROLLER_NODE: ProgInfo(11, 'mcs-controllernode', 'fg', True),
+    MCSProgs.PRIM_PROC: ProgInfo(5, 'mcs-primproc', '', False, 1),
+    MCSProgs.EXE_MGR: ProgInfo(9, 'mcs-exemgr', '', False, 1),
+    MCSProgs.WRITE_ENGINE_SERVER: ProgInfo(7, 'mcs-writeengineserver', '', False, 3),
+    MCSProgs.DML_PROC: ProgInfo(3, 'mcs-dmlproc', '', False),
+    MCSProgs.DDL_PROC: ProgInfo(1, 'mcs-ddlproc', '', False),
 }
 
 # constants for docker container dispatcher
--- a/cmapi/cmapi_server/constants.py
+++ b/cmapi/cmapi_server/constants.py
@@ -4,6 +4,7 @@ TODO: move main constant paths here and replace in files in next releases.
 """
 import os
 from typing import NamedTuple
+from enum import Enum
 
 
 # default MARIADB ColumnStore config path
@@ -53,6 +54,16 @@ CMAPI_SINGLE_NODE_XML = os.path.join(
     CMAPI_INSTALL_PATH, 'cmapi_server/SingleNode.xml'
 )
 
+class MCSProgs(Enum):
+    STORAGE_MANAGER = 'StorageManager'
+    WORKER_NODE = 'workernode'
+    CONTROLLER_NODE = 'controllernode'
+    PRIM_PROC = 'PrimProc'
+    EXE_MGR = 'ExeMgr'
+    WRITE_ENGINE_SERVER = 'WriteEngineServer'
+    DML_PROC = 'DMLProc'
+    DDL_PROC = 'DDLProc'
+
 # constants for dispatchers
 class ProgInfo(NamedTuple):
     """NamedTuple for some additional info about handling mcs processes."""
@@ -66,17 +77,17 @@ class ProgInfo(NamedTuple):
 # on top level of process handling
 # mcs-storagemanager starts conditionally inside mcs-loadbrm, but should be
 # stopped using cmapi
-ALL_MCS_PROGS = {
+ALL_MCS_PROGS: dict[str, ProgInfo] = {
     # workernode starts on primary and non primary node with 1 or 2 added
     # to subcommand (DBRM_Worker1 - on primary, DBRM_Worker2 - non primary)
-    'StorageManager': ProgInfo(15, 'mcs-storagemanager', '', False, 1),
-    'workernode': ProgInfo(13, 'mcs-workernode', 'DBRM_Worker{}', False, 1),
-    'controllernode': ProgInfo(11, 'mcs-controllernode', 'fg', True),
-    'PrimProc': ProgInfo(5, 'mcs-primproc', '', False, 1),
-    'ExeMgr': ProgInfo(9, 'mcs-exemgr', '', False, 1),
-    'WriteEngineServer': ProgInfo(7, 'mcs-writeengineserver', '', False, 3),
-    'DMLProc': ProgInfo(3, 'mcs-dmlproc', '', False),
-    'DDLProc': ProgInfo(1, 'mcs-ddlproc', '', False),
+    MCSProgs.STORAGE_MANAGER.value: ProgInfo(15, 'mcs-storagemanager', '', False, 1),
+    MCSProgs.WORKER_NODE.value: ProgInfo(13, 'mcs-workernode', 'DBRM_Worker{}', False, 1),
+    MCSProgs.CONTROLLER_NODE.value: ProgInfo(11, 'mcs-controllernode', 'fg', True),
+    MCSProgs.PRIM_PROC.value: ProgInfo(5, 'mcs-primproc', '', False, 1),
+    MCSProgs.EXE_MGR.value: ProgInfo(9, 'mcs-exemgr', '', False, 1),
+    MCSProgs.WRITE_ENGINE_SERVER.value: ProgInfo(7, 'mcs-writeengineserver', '', False, 3),
+    MCSProgs.DML_PROC.value: ProgInfo(3, 'mcs-dmlproc', '', False),
+    MCSProgs.DDL_PROC.value: ProgInfo(1, 'mcs-ddlproc', '', False),
 }
 
 # constants for docker container dispatcher
--- a/cmapi/cmapi_server/failover_agent.py
+++ b/cmapi/cmapi_server/failover_agent.py
@@ -95,7 +95,7 @@ class FailoverAgent(AgentBase):
         try:
             # TODO: remove test_mode condition and add mock for testing
             if not test_mode:
-                MCSProcessManager.stop_node(is_primary=nc.is_primary_node())
+                MCSProcessManager.stop_node(is_primary=nc.is_primary_node(), is_read_only=nc.is_read_only())
             logger.info(
                 'FA.enterStandbyMode(): successfully stopped node.'
             )
--- a/cmapi/cmapi_server/failover_agent.py
+++ b/cmapi/cmapi_server/failover_agent.py
@@ -95,9 +95,7 @@ class FailoverAgent(AgentBase):
         try:
             # TODO: remove test_mode condition and add mock for testing
             if not test_mode:
-                MCSProcessManager.stop_node(
-                    is_primary=nc.is_primary_node(),
-                )
+                MCSProcessManager.stop_node(is_primary=nc.is_primary_node())
             logger.info(
                 'FA.enterStandbyMode(): successfully stopped node.'
             )
--- a/cmapi/cmapi_server/failover_agent.py
+++ b/cmapi/cmapi_server/failover_agent.py
@@ -95,7 +95,10 @@ class FailoverAgent(AgentBase):
         try:
             # TODO: remove test_mode condition and add mock for testing
             if not test_mode:
-                MCSProcessManager.stop_node(is_primary=nc.is_primary_node(), is_read_only=nc.is_read_only())
+                MCSProcessManager.stop_node(
+                    is_primary=nc.is_primary_node(),
+                    is_read_only=nc.is_read_only(),
+                )
             logger.info(
                 'FA.enterStandbyMode(): successfully stopped node.'
             )
--- a/cmapi/cmapi_server/failover_agent.py
+++ b/cmapi/cmapi_server/failover_agent.py
@@ -97,7 +97,6 @@ class FailoverAgent(AgentBase):
             if not test_mode:
                 MCSProcessManager.stop_node(
                     is_primary=nc.is_primary_node(),
-                    is_read_only=nc.is_read_only(),
                 )
             logger.info(
                 'FA.enterStandbyMode(): successfully stopped node.'
--- a/cmapi/cmapi_server/handlers/cluster.py
+++ b/cmapi/cmapi_server/handlers/cluster.py
@@ -163,8 +163,8 @@ class ClusterHandler():
         """
         logger: logging.Logger = logging.getLogger('cmapi_server')
         logger.debug(
-            'Cluster add node command called. '
-            f'Adding node {node} in {"read-only" if read_only else "read-write"} mode.'
+            f'Cluster add node command called. Adding node {node} in '
+            f'{"read-only" if read_only else "read-write"} mode.'
         )
 
         response = {'timestamp': str(datetime.now())}
@@ -182,7 +182,9 @@ class ClusterHandler():
                         output_config_filename=config
                     )
                 else:
-                    logger.debug(f"Node {node} is read-only, skipping dbroot addition")
+                    logger.debug(
+                        f'Node {node} is read-only, skipping dbroot addition'
+                    )
 
         except Exception as err:
             raise CMAPIBasicError('Error while adding node.') from err
--- a/cmapi/cmapi_server/handlers/cluster.py
+++ b/cmapi/cmapi_server/handlers/cluster.py
@@ -15,7 +15,7 @@ from cmapi_server.helpers import (
     get_current_key, get_version, update_revision_and_manager,
 )
 from cmapi_server.node_manipulation import (
-    add_node, add_dbroot, remove_node, switch_node_maintenance,
+    add_node, add_dbroot, remove_node, switch_node_maintenance, update_dbroots_of_readonly_nodes,
 )
 from mcs_node_control.models.misc import get_dbrm_master
 from mcs_node_control.models.node_config import NodeConfig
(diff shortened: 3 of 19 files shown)


mariadb-AlanMologorsky:
Name branch as a feature/MCOL-.....
And the same about unified commits pls.

Short description is better than just JIRA link. Even summary from commit\s message.

cmapi/cmapi_server/failover_agent.py
```diff
@@ -95,7 +95,7 @@ def enterStandbyMode(self, test_mode = False):
         try:
             # TODO: remove test_mode condition and add mock for testing
             if not test_mode:
-                MCSProcessManager.stop_node(is_primary=nc.is_primary_node())
+                MCSProcessManager.stop_node(is_primary=nc.is_primary_node(), is_read_only=nc.is_read_only())
```
mariadb-AlanMologorsky: Line length less or equal than 79 symbols pls

cmapi/cmapi_server/handlers/cluster.py
```diff
@@ -216,7 +219,7 @@ def add_node(
         :return: result of adding node
         :rtype: dict
         """
-        logger.debug(f'Cluster add node command called. Adding node {node}.')
+        logger.debug('Cluster add node command called. Adding node %s in %s mode.', node, 'read-only' if read_only else 'read-write')
```
mariadb-AlanMologorsky: F-string is not the best style and have cons in pair with logging module, but anyway it's readable and compact and used more often in code.
And line length pls.

cmapi/cmapi_server/handlers/cluster.py
```diff
@@ -238,13 +241,18 @@ def add_node(
...
-                    host=node, input_config_filename=config,
-                    output_config_filename=config
-                )
+                if not read_only:  # Read-only nodes don't own dbroots
+                    add_dbroot(
+                        host=node, input_config_filename=config,
+                        output_config_filename=config
+                    )
+                else:
+                    logger.debug("Node %s is read-only, skipping dbroot addition", node)
```
mariadb-AlanMologorsky: I'd like f-string here.

cmapi/cmapi_server/helpers.py
```diff
@@ -513,6 +513,10 @@ def get_desired_nodes(config=DEFAULT_MCS_CONF_PATH):
     return [ node.text for node in nodes ]
 
 
+def get_read_only_nodes(root) -> list[str]:
```
mariadb-AlanMologorsky: docstring pls.

cmapi/cmapi_server/helpers.py
```diff
@@ -568,6 +573,12 @@ def get_dbroots(node, config=DEFAULT_MCS_CONF_PATH):
                 dbroots.append(
                     smc_node.find(f"./ModuleDBRootID{i}-{j}-3").text
                 )
+
+    if dbroots and nc.is_read_only():
+        logger = logging.getLogger("dbroots")
```
mariadb-AlanMologorsky: Single quotes pls.

cmapi/cmapi_server/helpers.py
```diff
@@ -568,6 +573,12 @@ def get_dbroots(node, config=DEFAULT_MCS_CONF_PATH):
...
+        logger.warning("Config contains dbroots %s for this read-only node, ignoring", dbroots)
```
mariadb-AlanMologorsky: Same about f-string.

cmapi/cmapi_server/managers/process.py
```diff
@@ -404,37 +405,46 @@ def is_node_processes_ok(
         return set(node_progs) == set(p['name'] for p in running_procs)
 
     @classmethod
-    def start_node(cls, is_primary: bool, use_sudo: bool = True):
+    def start_node(cls, is_primary: bool, use_sudo: bool = True, is_read_only: bool = False) -> None:
```
mariadb-AlanMologorsky: Line length pls.

cmapi/cmapi_server/managers/process.py
```diff
@@ -404,37 +405,46 @@ def is_node_processes_ok(
...
                 )
                 continue
             # TODO: additional error handling
-            if prog_name == 'controllernode':
+            if prog_name == MCSProgs.CONTROLLER_NODE.value:
                 cls._wait_for_workernodes()
-            if prog_name in ('DMLProc', 'DDLProc'):
+            if prog_name in (MCSProgs.DML_PROC.value, MCSProgs.DDL_PROC.value):
                 cls._wait_for_controllernode()
+            if is_read_only and prog_name == MCSProgs.WRITE_ENGINE_SERVER.value:
```
mariadb-AlanMologorsky: Line length pls.

cmapi/cmapi_server/managers/process.py
```diff
@@ -404,37 +405,46 @@ def is_node_processes_ok(
...
+                logging.debug('Node is in read-only mode, not starting WriteEngine')
```
mariadb-AlanMologorsky: Line length. 
Single quotes are welcome =)

cmapi/cmapi_server/managers/process.py
mariadb-AlanMologorsky: This method only could change logic of startup depends on dispatcher type and process name. This is the main idea. 
Anyway current condition indirectly changes process list during start node to in a new place.
It should be moved in `_get_sorted_progs` method or into constants - that's the only places to sort and store processes list. I'd prefer the first place.

cmapi/cmapi_server/managers/process.py
```diff
@@ -444,20 +454,25 @@ def stop_node(
...
+        :param is_read_only: if true, doesn't stop WriteEngine
+        :type is_read_only: bool, optional
         :raises CMAPIBasicError: immediately if one mcs process not stopped
         """
         # Every time try to stop all processes no matter primary it or slave,
         # so use full available list of processes. Otherwise, it could cause
         # undefined behaviour when primary gone and then recovers (failover
         # triggered 2 times).
         for prog_name in cls._get_sorted_progs(True, reverse=True):
+            if is_read_only and prog_name == MCSProgs.WRITE_ENGINE_SERVER.value:
```
mariadb-AlanMologorsky: Same for start_node. 
**With one major difference.**
It is an overhead to check all states.
Just stop all mcs services existed as a binary in system. It takes extra 1-2 condition time (that's extremely small in terms of whole node stop time) but removes all overhead from `stop_node` method.

In other words just keep this method as is - it will works as needed.

cmapi/cmapi_server/managers/process.py
```diff
@@ -444,20 +454,25 @@ def stop_node(
...
 
     @classmethod
-    def restart_node(cls, is_primary: bool, use_sudo: bool):
+    def restart_node(cls, is_primary: bool, use_sudo: bool, is_read_only: bool = False):
         """TODO: For next releases."""
         if cls.get_running_mcs_procs():
-            cls.stop_node(is_primary, use_sudo)
-        cls.start_node(is_primary, use_sudo)
+            cls.stop_node(is_primary, use_sudo, is_read_only)
+        cls.start_node(is_primary, use_sudo, is_read_only)
```
mariadb-AlanMologorsky, line 483: Based on comment above - no need to pass it into stop_node.

cmapi/cmapi_server/constants.py
```diff
@@ -51,6 +52,16 @@
     CMAPI_INSTALL_PATH, 'cmapi_server/SingleNode.xml'
 )
 
+class MCSProgs(Enum):
```
mariadb-AlanMologorsky, line 57: In this case I don't like the way we accessing service name:  `MCSProgs.<Serv_name>.value`.
With all respect to Enums `.value` notation is an ugly one in current case and IMO in some places readability is degraded.

Therefor we could think about more elegant way to declare and store services names, or, if no better ideas, rename it to `MCSProgsEnum` or `MCSProgNamesEnum`.

cmapi/cmapi_server/node_manipulation.py
```diff
@@ -359,12 +373,16 @@ def __remove_helper(parent_node, node):
...
 def _remove_node(root, node):
     '''
-    remove node from DesiredNodes, InactiveNodes, and ActiveNodes
+    remove node from DesiredNodes, InactiveNodes, ActiveNodes and (if present) ReadOnlyNodes
     '''
 
     for n in (root.find("./DesiredNodes"), root.find("./InactiveNodes"), root.find("./ActiveNodes")):
         __remove_helper(n, node)
 
+    read_only_nodes = root.find("./ReadOnlyNodes")
```
mariadb-AlanMologorsky: Single quotes pls.
Double quotes above only because this code never refactored in terms of styling.

cmapi/cmapi_server/node_manipulation.py
```diff
@@ -359,12 +373,16 @@ def __remove_helper(parent_node, node):
...
+    if read_only_nodes is not None:
```
mariadb-AlanMologorsky, line 403: Good for backward compatibility and xmls that don't has such a tag.
What do you think about adding this tag to the default xml in next releases?

cmapi/cmapi_server/node_manipulation.py
```diff
@@ -972,6 +990,19 @@ def _add_WES(root, pm_num, node):
     etree.SubElement(wes_node, "Port").text = "8630"
 
 
+def _add_read_only_node(root, node) -> None:
+    """Add node name to ReadOnlyNodes if it's not already there"""
```
mariadb-AlanMologorsky: upd docstring pls with params and param types.

cmapi/cmapi_server/node_manipulation.py
```diff
@@ -972,6 +990,19 @@ def _add_WES(root, pm_num, node):
...
+    read_only_nodes = root.find("./ReadOnlyNodes")
```
mariadb-AlanMologorsky: single quotes here and below pls

cmapi/mcs_node_control/models/node_config.py
```diff
@@ -566,4 +566,19 @@ def get_all_dbroots(self, root):
         for i in range(1, mod_count+1):
             for j in range(1, int(smc_node.find(f"./ModuleDBRootCount{i}-3").text) + 1):
                 dbroots.append(smc_node.find(f"./ModuleDBRootID{i}-{j}-3").text)
+
+        # TODO not sure about it
+        if dbroots and self.is_read_only(root):
+            module_logger.warning("Config contains dbroots %s for this read-only node, ignoring", dbroots)
```
mariadb-AlanMologorsky: f-string pls.

cmapi/mcs_node_control/models/node_config.py
```diff
@@ -566,4 +566,19 @@ def get_all_dbroots(self, root):
...
+            return []
+
         return dbroots
+
+    def is_read_only(self, root=None) -> bool:
+        """Checks if this node is in read-only mode"""
+        from cmapi_server.helpers import get_read_only_nodes # Avoid circular import
```
mariadb-AlanMologorsky: We should avoid this circular import "hacks". 
In my practice if any exist - it means that structure is wrong or function in a wrong place.
In the future such a one place produces several another etc.

To be honest some functions from helpers should be here. But not yet.
So my proposal is to move `get_read_only_nodes` function as a method here.

mariadb-AlanMologorsky (changes requested):

mariadb-AlanMologorsky:
And one thing about NEW readonly node naming - Seems should be discussed together with @drrtuy or @mariadb-LeonidFedorov or both. 
Because comparing to any "readonly" mention in previous code this naming could be confusing.

cmapi/cmapi_server/constants.py
mariadb-AlexanderPresniakov: I stored the value of the enum field as a dict key because I wanted to minimize the impact on the other code. But turns out there are only two other places where this dict is used, so i've switched dict to enum as a key

cmapi/cmapi_server/managers/process.py
mariadb-AlexanderPresniakov: I could move this logic to _get_sorted_progs, but there are two problems:
* to do it, we would have to pass read_only into _get_sorted_progs, including the call sites is_node_processes_ok and stop_node (so we'd have to calculate and pass the value from the config, which would complicate the logic and the method signatures)
* I need a log record explaining that WES was not started because it is a read-only node. But this is a non-destructive method, so it would look and act ugly

mariadb-AlanMologorsky: I understand the complication, but it's still not the place to modify list of processes. 
Previously I read the PR and it looks not a big deal to move it from here.
Ok, when I'll have a time I will return back to it and suggest the exact code changes.

cmapi/cmapi_server/logging_management.py
```diff
@@ -104,7 +104,8 @@ def enable_console_logging(logger: logging.Logger) -> None:
 def config_cmapi_server_logging():
     # add custom level TRACE only for develop purposes
     # could be activated using API endpoints or cli tool without relaunching
-    add_logging_level('TRACE', 5)
+    if not hasattr(logging, 'TRACE'):
+        add_logging_level('TRACE', 5)
```
mariadb-AlanMologorsky, line 108: JFYI I think to remove or rewrite this dev function.

cmapi/cmapi_server/managers/process.py
```diff
@@ -89,7 +98,8 @@ def _detect_processes(cls) -> None:
         if cls.mcs_progs:
             logging.warning('Mcs ProcessHandler already detected processes.')
 
-        for prog_name, prog_info in ALL_MCS_PROGS.items():
+        for prog, prog_info in ALL_MCS_PROGS.items():
+            prog_name = prog.value
```
mariadb-AlanMologorsky, line 102: JFYI It's what I mentioned before [here](https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/3432/files#r2019691224) about naming and access names. prog.value == prog_name. It's somehow confusing.

cmapi/cmapi_server/node_manipulation.py
```diff
@@ -17,12 +17,14 @@
 
 from cmapi_server import helpers
 from cmapi_server.constants import (
-    CMAPI_CONF_PATH, CMAPI_SINGLE_NODE_XML, DEFAULT_MCS_CONF_PATH, LOCALHOSTS,
+    CMAPI_CONF_PATH,
+    CMAPI_SINGLE_NODE_XML,
+    DEFAULT_MCS_CONF_PATH,
+    LOCALHOSTS,
```
mariadb-AlanMologorsky, line 23: Why do we need it on a separate lines?

cmapi/cmapi_server/node_manipulation.py
```diff
@@ -1138,3 +1182,75 @@ def _replace_localhost(root, node):
...
+    mod_count = int(smc_node.find("./ModuleCount3").text)
+    for i in range(1, mod_count + 1):
+        ip_addr = smc_node.find(f"./ModuleIPAddr{i}-1-3").text
+        module_num_to_addr[i] = ip_addr
+    return module_num_to_addr
+
+
+def update_dbroots_of_readonly_nodes(root: etree.Element) -> None:
+    """Read-only nodes do not have their own dbroots, but they must have all the dbroots of the other nodes
+    So this function sets list of dbroots of each read-only node to the list of all the dbroots in the cluster
```
mariadb-AlanMologorsky, line 1200: Seems to be vice versa. 
`sets list of all the dbroots in the cluster to the list of dbroots of each read-only node`

cmapi/cmapi_server/node_manipulation.py
```diff
@@ -1138,3 +1182,75 @@ def _replace_localhost(root, node):
...
+    dbroot_count_node.text = str(len(existing_dbroots))
+
+    # Write new dbroot IDs to the module mapping
+    for i, dbroot_id in enumerate(existing_dbroots, start=1):
+        dbroot_id_node = etree.SubElement(
+            sysconf_node, f"ModuleDBRootID{module_num}-{i}-3"
+        )
+        dbroot_id_node.text = str(dbroot_id)
+
+    logging.info("Added %d dbroots to read-only node %d: %s", len(existing_dbroots), module_num, sorted(existing_dbroots))
```
mariadb-AlanMologorsky, line 1242: Pls f-string notation.

cmapi/cmapi_server/node_manipulation.py
```diff
@@ -1138,3 +1182,75 @@ def _replace_localhost(root, node):
...
+        module_num_to_addr[i] = ip_addr
+    return module_num_to_addr
+
+
+def update_dbroots_of_readonly_nodes(root: etree.Element) -> None:
+    """Read-only nodes do not have their own dbroots, but they must have all the dbroots of the other nodes
+    So this function sets list of dbroots of each read-only node to the list of all the dbroots in the cluster
+    """
+    nc = NodeConfig()
+    pm_num_to_addr = get_pm_module_num_to_addr_map(root)
```
mariadb-AlanMologorsky, line 1203: When reading it looks like inside variable only one pm and corresponding addr. Should be something like `pm_nums_to_addrs` or `pm_num_to_addr_map`

cmapi/cmapi_server/node_manipulation.py
```diff
@@ -1138,3 +1182,75 @@ def _replace_localhost(root, node):
...
+    for ro_node in nc.get_read_only_nodes(root):
+        # Get PM num by IP address
+        this_ip_pm_num = None
```
mariadb-AlanMologorsky, line 1206: Confusing variable name. 
It's hard to understand what pm num stored here. This ip? Should has better naming to be more readable.

cmapi/cmapi_server/test/test_node_manip.py
```diff
@@ -1,18 +1,20 @@
...
 
 from lxml import etree
 
 from cmapi_server import node_manipulation
 from cmapi_server.constants import MCS_DATA_PATH
-from cmapi_server.test.unittest_global import (
-    tmp_mcs_config_filename, BaseNodeManipTestCase
-)
+from cmapi_server.node_manipulation import add_dbroots_of_other_nodes, remove_dbroots_of_node, update_dbroots_of_readonly_nodes
+from cmapi_server.test.unittest_global import BaseNodeManipTestCase, tmp_mcs_config_filename
```
mariadb-AlanMologorsky, line 11: Too long lines this and above.

cmapi/cmapi_server/test/test_node_manip.py
```diff
@@ -1,18 +1,20 @@
...
 from mcs_node_control.models.node_config import NodeConfig
 
-
 logging.basicConfig(level='DEBUG')
 
+SINGLE_NODE_XML = "./cmapi_server/SingleNode.xml"
```
mariadb-AlanMologorsky, line 16: JFYI I'm not sure that default Columnstore.xml is fully equal to SingleNode.xml

cmapi/cmapi_server/test/test_node_manip.py
```diff
@@ -21,12 +23,18 @@ def test_add_remove_node(self):
...
         hostaddr = socket.gethostbyname(socket.gethostname())
-        node_manipulation.add_node(
-            self.NEW_NODE_NAME, tmp_mcs_config_filename, self.tmp_files[0]
-        )
-        node_manipulation.add_node(
-            hostaddr, self.tmp_files[0], self.tmp_files[1]
-        )
+
+        with patch('cmapi_server.node_manipulation.update_dbroots_of_readonly_nodes') as mock_update_dbroots_of_readonly_nodes:
+            node_manipulation.add_node(
```
mariadb-AlanMologorsky, line 28: With this mock test name is not fully valid. Here we test add\remove node and  mocked function call once.
May be better to move it in separate test?

cmapi/cmapi_server/test/test_node_manip.py
```diff
@@ -40,10 +48,13 @@ def test_add_remove_node(self):
...
-        node_manipulation.remove_node(
-            self.NEW_NODE_NAME, self.tmp_files[1], self.tmp_files[2],
-            test_mode=True
-        )
+        with patch('cmapi_server.node_manipulation.update_dbroots_of_readonly_nodes') as mock_update_dbroots_of_readonly_nodes:
+            node_manipulation.remove_node(
+                self.NEW_NODE_NAME, self.tmp_files[1], self.tmp_files[2],
+                test_mode=True
+            )
+            mock_update_dbroots_of_readonly_nodes.assert_called_once()
```
mariadb-AlanMologorsky, line 56: It's not clear to me what we trying to test here several times. Several times test that we call the mocked function once?

cmapi/mcs_node_control/models/node_config.py
```diff
@@ -137,6 +138,26 @@ def write_config(self, tree, filename=DEFAULT_MCS_CONF_PATH):
             f.write(self.to_string(tree))
         replace(tmp_filename, filename)    # atomic replacement
 
+    @contextmanager
```
mariadb-AlanMologorsky, line 141: I'd like to move it in separate PR and integrate in other places tho.

cmapi/pyproject.toml
mariadb-AlanMologorsky, line 1: Why do we need this file?

mariadb-AlanMologorsky (changes requested):

drrtuy:
Will open a new PR from the new feature branch [MCOL-5964-5861-5863-um-only-node](https://github.com/mariadb-corporation/mariadb-columnstore-engine/tree/feat/MCOL-5964-5861-5863-um-only-node).

PR #3437, opened 2025-03-14
fix(TNS): fix for duplicate DISTINCT output

drrtuy:
The patch delivers CountingAllocator that is propagated into TNS and THJS to account for stdlib structs memory consumption, e.g. hashmaps and priority_queues. The allocator has been also propagated into some of the existing Allocators: STD, Fixed etc. ByteStream also got CountingAllocator support and now PP uses ByteStream with memory accounting processing PrimitiveJob.

### DIFF
--- a/dbcon/joblist/jobstep.cpp
+++ b/dbcon/joblist/jobstep.cpp
@@ -217,6 +217,12 @@ void JobStep::handleException(std::exception_ptr e, const int errorCode, const u
   {
     std::rethrow_exception(e);
   }
+  // Add it here for now to handle potential bad_alloc exceptions
+  catch (std::bad_alloc& exc)
+  {
+   std::cerr << methodName << " caught a bad_alloc exception. " << std::endl;
+    catchHandler(methodName + " caught " + exc.what(), errorCode, fErrorInfo, fSessionId);
+  }
   catch (const IDBExcept& iex)
   {
     std::cerr << methodName << " caught a internal exception. " << std::endl;
--- a/dbcon/joblist/resourcemanager.cpp
+++ b/dbcon/joblist/resourcemanager.cpp
@@ -347,23 +347,23 @@ bool ResourceManager::userPriorityEnabled() const
 // If both have space, return true.
 bool ResourceManager::getMemory(int64_t amount, boost::shared_ptr<int64_t>& sessionLimit, bool patience)
 {
-  bool ret1 = (totalUmMemLimit.fetch_sub(amount, std::memory_order_relaxed) >= 0);
+  bool ret1 = (atomicops::atomicSubRef(totalUmMemLimit, amount) >= 0);
   bool ret2 = sessionLimit ? (atomicops::atomicSub(sessionLimit.get(), amount) >= 0) : ret1;
 
   uint32_t retryCounter = 0, maxRetries = 20;  // 10s delay
 
   while (patience && !(ret1 && ret2) && retryCounter++ < maxRetries)
   {
-    totalUmMemLimit.fetch_add(amount, std::memory_order_relaxed);
+    atomicops::atomicAddRef(totalUmMemLimit, amount);
     sessionLimit ? atomicops::atomicAdd(sessionLimit.get(), amount) : 0;
     usleep(500000);
-    ret1 = (totalUmMemLimit.fetch_sub(amount, std::memory_order_relaxed) >= 0);
+    ret1 = (atomicops::atomicSubRef(totalUmMemLimit, amount) >= 0);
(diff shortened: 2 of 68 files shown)


mariadb-LeonidFedorov (approved):

mariadb-LeonidFedorov (approved):

mariadb-LeonidFedorov:
This code should be inspected a bit better

mariadb-LeonidFedorov (changes requested):
There is a lot of junior changes to be involved

dbcon/joblist/jobstep.cpp
```diff
@@ -217,6 +217,12 @@ void JobStep::handleException(std::exception_ptr e, const int errorCode, const u
   {
     std::rethrow_exception(e);
   }
+  // Add it here for now to handle potential bad_alloc exceptions
```
mariadb-LeonidFedorov, line 220: I'd like to see bad alloc test

dbcon/joblist/resourcemanager.cpp
```diff
@@ -346,44 +347,44 @@ bool ResourceManager::userPriorityEnabled() const
 // If both have space, return true.
 bool ResourceManager::getMemory(int64_t amount, boost::shared_ptr<int64_t>& sessionLimit, bool patience)
 {
-  bool ret1 = (atomicops::atomicSub(&totalUmMemLimit, amount) >= 0);
+  bool ret1 = (totalUmMemLimit.fetch_sub(amount, std::memory_order_relaxed) >= 0);
```
mariadb-LeonidFedorov: There's no reason to change the common behavior without updating the core tools

dbcon/joblist/tupleannexstep.cpp
```diff
@@ -1072,11 +1099,25 @@ void TupleAnnexStep::finalizeParallelOrderBy()
...
+  // Release memory before ctor
+  for (uint64_t id = 1; id <= fMaxThreads; id++)
+  {
+    fOrderByList[id]->returnAllRGDataMemory2RM();
+  }
+
+  StepTeleStats sts;
+  sts.query_uuid = fQueryUuid;
+  sts.step_uuid = fStepUuid;
+  sts.msg_type = StepTeleStats::ST_SUMMARY;
```
mariadb-LeonidFedorov, line 1112: A bit of SOLID should work here

dbcon/joblist/tuplehashjoin.cpp
```diff
@@ -373,14 +316,38 @@ void TupleHashJoinStep::startSmallRunners(uint index)
...
+  if (!allowDJS || isDML || (fSessionId & 0x80000000) || (tableOid() < 3000 && tableOid() >= 1000))
+  {
+    joinIsTooBig = true;
+    ostringstream oss;
+    oss << "(" << __LINE__ << ") " << logging::IDBErrorInfo::instance()->errorMsg(logging::ERR_JOIN_TOO_BIG);
+    fLogger->logMessage(logging::LOG_TYPE_INFO, oss.str());
+    errorMessage(oss.str());
+    status(logging::ERR_JOIN_TOO_BIG);
+    cout << "Join is too big, raise the UM join limit for now (small runner)" << endl;
+    abort();
```
mariadb-LeonidFedorov, line 342: Abort should be explained with the all team

mariadb-LeonidFedorov (changes requested):

mariadb-LeonidFedorov (approved):

dbcon/joblist/resourcemanager.cpp
drrtuy: I didn't pick this up TBH. Can you elaborate on this? What are the core tools?

dbcon/joblist/tuplehashjoin.cpp
drrtuy: Could you elaborate? The same code block had existed for a long time in the THJS. I just moved it into a separate function.

dbcon/joblist/tupleannexstep.cpp
drrtuy: I didn't get how SOLID is related to this but I will add a specialized ctor for the purpose as an external patch b/c the change isn't directly related to this patch.

dbcon/joblist/resourcemanager.cpp
drrtuy: Addressed with a separate commit

dbcon/joblist/tupleannexstep.cpp
drrtuy: Addressed with a separate PR

PR #3446, opened 2025-03-18
chore(build): update boost version from 1.84 to 1.88 

mariadb-LeonidFedorov:
- updated boost from **1.84** (December 13th, 2023) to **1.88** (April 10th, 2025)
- updated usage of `boost::filesystem` due API breaking change
- `boost::interprocess` patched ad-hoc (**sic!**), as `CtorArgN `from `boost/interprocess/detail/named_proxy.hpp` used for forwarding was inherited from base class with virtual d-tor before boost 1.85, and now it's standalone class, and we got  warning for class with virtual functions without virtual d-tor, that is hard to suppress locally, we just include the header


drrtuy (approved):
untestedACK

PR #3464, opened 2025-03-25
feat(): counting allocator for 23.10 with compilation fixes for boost  

drrtuy:


writeengine/shared/we_brm.cpp
```diff
@@ -555,13 +545,13 @@ int BRMWrapper::getTableLockInfo(uint64_t lockID, BRM::TableLockInfo* lockInfo,
 /* static */
 int BRMWrapper::getBrmRc(bool reset)
 {
-  if (m_ThreadDataPtr.get() == 0)
+  if (m_brmRc == 0)
```
mariadb-AlexeyAntipovsky, line 548: Afair, `thread_specific_ptr::get()` returns pointer to `T` — looks like this check is redundant.

PR #3469, opened 2025-03-27
MCOL-5992: fix NULLIF if first param is NULL

mariadb-LeonidFedorov:


mysql-test/columnstore/bugfixes/mcol-5992.test
```diff
@@ -0,0 +1,17 @@
...
+
+--disable_warnings
+DROP DATABASE IF EXISTS mcol5992;
+--enable_warnings
+
+CREATE DATABASE mcol5992;
+USE mcol5992;
+
+CREATE TABLE `CS_LEG` (`ID` bigint(20) DEFAULT NULL, `sess_count` smallint(6) DEFAULT NULL) ENGINE=Columnstore;
+CREATE TABLE `INNO_LEG` (`ID` bigint(20) DEFAULT NULL, `sess_count` smallint(6) DEFAULT NULL) ENGINE=InnoDB;
```
mariadb-AlexeyAntipovsky: This table is not used in the test, maybe remove it in order to reduce the CO₂ footprint? 8))

utils/funcexp/func_nullif.cpp
```diff
@@ -56,8 +56,17 @@ int64_t Func_nullif::getIntVal(rowgroup::Row& row, FunctionParm& parm, bool& isN
                                execplan::CalpontSystemCatalog::ColType& op_ct)
 {
   int64_t exp1 = parm[0]->data()->getIntVal(row, isNull);
+
+  // NULL equals anything is NULL, so we return NULL as a value of parm[0]
+  if (isNull)
+  {
+    isNull = true;
```
mariadb-AlexeyAntipovsky: (and below)
This line is a bit confusing IMO. Why is it necessary to to set `isNull` to `true` when `isNull` is `true` — I really spent several minutes trying to figure out what I had missed...

mariadb-AlexeyAntipovsky (approved):
lgtm

utils/funcexp/func_nullif.cpp
mariadb-LeonidFedorov: you are perfectly right :)

mysql-test/columnstore/bugfixes/mcol-5992.test
mariadb-LeonidFedorov: ++

PR #3477, opened 2025-04-01
MCOL-5852: GROUP_CONCAT memory accounting and disk based version [stable-23.10]

mariadb-AlexeyAntipovsky:


dbcon/joblist/groupconcat.cpp
```diff
@@ -758,60 +1023,196 @@ uint64_t GroupConcatOrderBy::getKeyLength() const
...
   // check if this is a distinct row
   if (fDistinct && fDistinctMap->find(row.getPointer()) != fDistinctMap->end())
     return;
 
   // this row is skipped if any concatenated column is null.
-  if (concatColIsNull(row))
+  if (!fIsJsonArrayAgg && concatColIsNull(row))
     return;
 
-  auto& orderByQueue = getQueue();
```
drrtuy, line 771: Could you elaborate on this? I think this and the corresponding changes down the code should be reversed.

dbcon/joblist/groupconcat.cpp
```diff
@@ -824,66 +1225,71 @@ void GroupConcatOrderBy::processRow(const rowgroup::Row& row)
...
+  go->fMemSize = 0;
+  uint32_t shift = fDataVec.size();
+  for (auto& rgdata : go->fDataVec)
+  {
+    fDataVec.emplace_back(std::move(rgdata));
+  }
+  go->fDataVec.clear();
 
-  auto& orderByQueue = getQueue();
-  auto& mergeQueue = go->getQueue();
```
drrtuy, line 843: We leak abstractions here for a mergeQueue. I would suggest to use API to retrieve queue from go.

dbcon/joblist/groupconcat.cpp
```diff
@@ -981,54 +1389,60 @@ const string GroupConcatOrderBy::toString() const
...
 }
 
-// GroupConcatNoOrder class implementation
-GroupConcatNoOrder::GroupConcatNoOrder()
- : fRowsPerRG(128), fErrorCode(ERR_AGGREGATION_TOO_BIG), fMemSize(0), fRm(NULL)
+uint64_t GroupConcatOrderBy::Hasher::operator()(const rowgroup::Row::Pointer& p) const
+{
+  Row& row = ts->row1;
+  row.setPointer(p);
+  return row.hash(colCount - 1);
```
drrtuy, line 1401: Rhetoric phrase.
This is what I like the most :)
Why do we hash using all columns `colCount - 1`. This assumes that RG contains only GB columns.

dbcon/joblist/groupconcat.cpp
```diff
@@ -981,54 +1389,60 @@ const string GroupConcatOrderBy::toString() const
...
+}
+
+uint64_t GroupConcatOrderBy::getCurrentRowIdx() const
+{
+  return ((fDataVec.size() - 1) << 16) + (fRowGroup.getRowCount() - 1);
+}
+
+uint64_t GroupConcatOrderBy::shiftGroupIdxBy(uint64_t idx, uint32_t shift)
 {
+  uint16_t rid = idx & 0xffff;
```
drrtuy: It this part the same as we have for GB?

dbcon/joblist/groupconcat.cpp
```diff
@@ -1038,59 +1452,61 @@ void GroupConcatNoOrder::processRow(const rowgroup::Row& row)
...
 {
   ostringstream oss;
   bool addSep = false;
 
-  fDataQueue.push(fData);
   size_t prevResultSize = 0;
 
   bool isNull = true;
-  while (fDataQueue.size() > 0)
+  bool addBrackets = true;
```
drrtuy, line 1500: If `addBrackets` is always true here than I didn't get why do we have it in the line 1505?

dbcon/joblist/groupconcat.cpp
```diff
@@ -1102,21 +1518,70 @@ uint8_t* GroupConcatNoOrder::getResultImpl(const string& sep)
     }
     size_t sizeDiff = oss.str().size() - prevResultSize;
     prevResultSize = oss.str().size();
-    if (!fRm->getMemory(sizeDiff, fSessionMemLimit))
```
drrtuy, line 1105: I am missing why we don't need account for memory consumption diff here?

dbcon/joblist/groupconcat.cpp
drrtuy: If yes, we should abstract this.

dbcon/joblist/groupconcat.cpp
mariadb-AlexeyAntipovsky: The `colCount` value here is the exact number of columns that will be concatenated (and `distinct` operates on them)

dbcon/joblist/groupconcat.cpp
mariadb-AlexeyAntipovsky: We set it to false after adding the opening bracket at line 1510.

dbcon/joblist/groupconcat.cpp
drrtuy: Discussed this offline and decided to give it a try. There will be no accounting for these allocations that potentially can be big but the life span is presumably short.

drrtuy (approved):
untested ACK

PR #3480, opened 2025-04-03
Fix(MCOL_4622): Implicit FLOAT->INT and DOUBLE->INT conversion is not like in InnoDB

AestheticAkhmad:
Fixed the implicit conversion of FLOAT to INT and DOUBLE to INT to behave according to InnoDB.


drrtuy (approved):
untested ACK

drrtuy:
Thank you for the contribution @AestheticAkhmad !

PR #3482, opened 2025-04-04
MCOL-5962 Fix rollback in mcs cli and passing errors during broadcasting config

mariadb-AlanMologorsky:
* fix(cmapi): ClusterControllerClient request error handling
* fix(cmapi): broadcast_new_config request error handling
* style(cmapi): logs

### DIFF
--- a/cmapi/cmapi_server/controllers/api_clients.py
+++ b/cmapi/cmapi_server/controllers/api_clients.py
@@ -1,12 +1,14 @@
-import requests
+import logging
 from typing import Any, Dict, Optional, Union
 
 import pyotp
+import requests
 
 from cmapi_server.controllers.dispatcher import _version
 from cmapi_server.constants import (
     CMAPI_CONF_PATH, CURRENT_NODE_CMAPI_URL, SECRET_KEY,
 )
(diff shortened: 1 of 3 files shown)


drrtuy (approved):
untestedACK

cmapi/cmapi_server/managers/process.py
```diff
@@ -430,7 +430,7 @@ def start_node(cls, is_primary: bool, use_sudo: bool = True):
                 cls._wait_for_controllernode()
             if not cls.start(prog_name, is_primary, use_sudo):
                 logging.error(f'Process "{prog_name}" not started properly.')
-                raise CMAPIBasicError(f'Error while starting "{prog_name}".')
+                raise CMAPIBasicError(f'Error during starting "{prog_name}".')
```
mariadb-AlexanderPresniakov: "During" is typically followed by a noun, not a gerund like "starting". So it would be either "during the start of", or "while starting", which is a more idiomatic way to refer to an action in progress.

mariadb-AlanMologorsky: Done

PR #3485, opened 2025-04-06
fix(MCOL-5386): Bitwise aggregation functions do not work with wide d…

AestheticAkhmad:
Fixed bitwise aggregation functions to work with wide decimals and behave like InnoDB.


drrtuy (approved):
untested ACK

PR #3501, opened 2025-04-15
chore(build): Ubuntu 20 fix build

mariadb-LeonidFedorov:
Build is broken since clang was turned off for Ubuntu 20.04, that has GCC 9 as vanilla compiler

groupconcat uses c++20 features that are missing for GCC 9.0

- std::ranges usage was replaced by closest boost alternative for GCC 9.0 in Ubuntu 20.04 that lacks this in stdlib
- std::span usage was replaced by boost::span for GCC 9.0 in Ubuntu 20.04 that lacks this in stdlib

### DIFF
--- a/dbcon/joblist/groupconcat.cpp
+++ b/dbcon/joblist/groupconcat.cpp
@@ -22,7 +22,7 @@
 #include <iostream>
 // #define NDEBUG
 #include <cassert>
-#include <ranges>
+
 #include <string>
 #include "windowfunction/idborderby.h"
 using namespace std;
@@ -467,9 +467,9 @@ void GroupConcator::initialize(const rowgroup::SP_GroupConcat& gcc)
   fRm = gcc->fRm;
   fSessionMemLimit = gcc->fSessionMemLimit;
 
-  for (const auto& str : views::keys(gcc->fConstCols))
+  for (const auto& str : gcc->fConstCols)
   {
-    fConstantLen += str.length();
+    fConstantLen += str.first.length();
   }
 }
 
@@ -714,7 +714,8 @@ void GroupConcator::outputRow(std::ostringstream& oss, const rowgroup::Row& row)
 
 bool GroupConcator::concatColIsNull(const rowgroup::Row& row)
 {
-  return ranges::any_of(fConcatColumns, [&](uint32_t idx) { return row.isNullValue(idx); });
+  return std::any_of(fConcatColumns.cbegin(), fConcatColumns.cend(),
+                     [&row](uint32_t idx) { return row.isNullValue(idx); });
 }
 
 int64_t GroupConcator::lengthEstimate(const rowgroup::Row& row)
@@ -987,6 +988,7 @@ void GroupConcatOrderBy::initialize(const rowgroup::SP_GroupConcat& gcc)
 
   fOrderByCond.resize(0);
 
+  fOrderByCond.reserve(gcc->fOrderCond.size());
   for (const auto& [idx, asc] : gcc->fOrderCond)
   {
     fOrderByCond.emplace_back(idx, asc);
@@ -994,9 +996,10 @@ void GroupConcatOrderBy::initialize(const rowgroup::SP_GroupConcat& gcc)
 
   fDistinct = gcc->fDistinct;
 
-  for (uint32_t x : views::values(gcc->fGroupCols))
+  fConcatColumns.reserve(fConcatColumns.size() + gcc->fGroupCols.size());
+  for (auto& x : gcc->fGroupCols)
   {
-    fConcatColumns.emplace_back(x);
+    fConcatColumns.emplace_back(x.second);
   }
 
   auto size = fRowGroup.getSizeWithStrings(fRowsPerRG);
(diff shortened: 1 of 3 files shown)


dbcon/joblist/groupconcat.cpp
```diff
@@ -714,7 +719,7 @@ void GroupConcator::outputRow(std::ostringstream& oss, const rowgroup::Row& row)
 
 bool GroupConcator::concatColIsNull(const rowgroup::Row& row)
 {
-  return ranges::any_of(fConcatColumns, [&](uint32_t idx) { return row.isNullValue(idx); });
+  return boost::algorithm::any_of(fConcatColumns, [&](uint32_t idx) { return row.isNullValue(idx); });
```
mariadb-AlexeyAntipovsky:
```suggestion
  return std::any_of(fConcatColumns.cbegin(), fConcatColumns.cend(), [&row](uint32_t idx) { return row.isNullValue(idx); });
```

dbcon/joblist/groupconcat.cpp
```diff
@@ -467,7 +470,9 @@ void GroupConcator::initialize(const rowgroup::SP_GroupConcat& gcc)
   fRm = gcc->fRm;
   fSessionMemLimit = gcc->fSessionMemLimit;
 
-  for (const auto& str : views::keys(gcc->fConstCols))
+  auto keys = gcc->fConstCols | boost::adaptors::transformed([](const auto& p) { return p.first; });
+
+  for (const auto& str : keys)
   {
     fConstantLen += str.length();
```
mariadb-AlexeyAntipovsky: I used `ranges` here and below because it looks a bit more readable than using `.first`/etc. But `boost::transform` here has the opposite effect, making the code harder to understand, IMO.

```suggestion
  for (const auto& val : gcc->fConstCols)
  {
    fConstantLen += val.first.length();
```

dbcon/joblist/groupconcat.cpp
```diff
@@ -994,7 +999,9 @@ void GroupConcatOrderBy::initialize(const rowgroup::SP_GroupConcat& gcc)
 
   fDistinct = gcc->fDistinct;
 
-  for (uint32_t x : views::values(gcc->fGroupCols))
+  auto values = gcc->fGroupCols | boost::adaptors::transformed([](const auto& p) { return p.second; });
+
+  for (uint32_t x : values)
   {
     fConcatColumns.emplace_back(x);
```
mariadb-AlexeyAntipovsky:
```suggestion
  for (const auto& val : gcc->fGroupCols)
  {
    fConcatColumns.emplace_back(val.second);
```

dbcon/joblist/groupconcat.cpp
```diff
@@ -1048,7 +1055,10 @@ void GroupConcatOrderBy::serialize(messageqcpp::ByteStream& bs) const
...
     sz = fDistinctMap->size();
     bs << sz;
-    for (const auto& idx : views::values(*fDistinctMap))
+
+    auto values = *fDistinctMap | boost::adaptors::transformed([](const auto& p) { return p.second; });
+
+    for (const auto& idx : values)
     {
       bs << idx;
     }
```
mariadb-AlexeyAntipovsky:
```suggestion
    for (const auto& [pointer, idx] : fDistinctMap)
    {
      bs << idx;
    }
```

drrtuy (approved):
untestedACK

dbcon/joblist/groupconcat.cpp
```diff
@@ -1435,7 +1445,9 @@ void GroupConcatNoOrder::initialize(const rowgroup::SP_GroupConcat& gcc)
   fRowGroup.setUseOnlyLongString(true);
   fRowsPerRG = 128;
 
-  for (uint32_t colIdx : views::values(gcc->fGroupCols))
+  auto values = gcc->fGroupCols | boost::adaptors::transformed([](const auto& p) { return p.second; });
+
+  for (uint32_t colIdx : values)
   {
     fConcatColumns.push_back(colIdx);
   }
```
mariadb-AlexeyAntipovsky:
```suggestion
  for (const auto& grpCol : gcc->fGroupCols)
  {
    fConcatColumns.push_back(grpCol.second);
  }
```

mariadb-AlexeyAntipovsky (changes requested):

dbcon/joblist/groupconcat.cpp
drrtuy: Funny fact, I've just suggested the same thing )

mariadb-LeonidFedorov: ++

dbcon/joblist/groupconcat.cpp
mariadb-LeonidFedorov: ++

dbcon/joblist/groupconcat.cpp
mariadb-LeonidFedorov: ++

dbcon/joblist/groupconcat.cpp
mariadb-LeonidFedorov: ++

dbcon/joblist/groupconcat.cpp
mariadb-LeonidFedorov: ++

mariadb-AlexeyAntipovsky (approved):
LGTM :rocket:

PR #3507, opened 2025-04-15
disable timestamp tests because of MCOL-5961

timofeyturenko:
disabling timestamp-related tests because they work only with one certain configuration of server timezone and session timezone


drrtuy (approved):
untested ACK

mariadb-LeonidFedorov (approved):

PR #3510, opened 2025-04-16
feat(MCOL-5886): support InnoDB's table partitions in cross-engine joins

mariadb-SergeyZefirov:
The purpose of this changeset is to obtain list of partitions from SELECT_LEX structure and pass it down to joblist and then to CrossEngineStep to pass to InnoDB.


drrtuy (approved):
untestedACK

drrtuy (changes requested):
MTR step fails the added test.

drrtuy (approved):
untestedACK

PR #3511, opened 2025-04-18
Fix thread arg errors (app instead of sequence)

mariadb-AlexanderPresniakov:
Third argument of Thread is a sequence. Without it we get this error:
`18/Apr/2025 07:30:50 [INFO] (cherrypy.error) {MainThread} ENGINE Bus STARTED
    self.run()
  File "/usr/share/columnstore/cmapi/python/lib/python3.9/threading.py", line 917, in run
    self._target(*self._args, **self._kwargs)
TypeError: __main__.worker() argument after * must be an iterable, not Application`

### DIFF
--- a/cmapi/cmapi_server/__main__.py
+++ b/cmapi/cmapi_server/__main__.py
@@ -36,7 +36,7 @@ def worker(app):
     repeatedly.
     """
     while True:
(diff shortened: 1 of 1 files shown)


mariadb-AlanMologorsky (approved):
Great finding!
LGTM

PR #3512, opened 2025-04-18
fix(PP,BPP): reduced the BPPMap lock scope to avoid holding lock whilst calling heavy BPPV dtor. It became heavy b/c of CountingAllocator.

drrtuy:
…


mariadb-AlexeyAntipovsky (approved):
lgtm

PR #3513, opened 2025-04-18
feat(TNS): change ORDER BY out-of-memory error message. Replaced bad_alloc with IDBExcept in TNS.

drrtuy:
…


mariadb-LeonidFedorov (changes requested):
this looks like wrong branch merging

mariadb-LeonidFedorov (approved):
Untested

PR #3522, opened 2025-04-25
fix(MCOL-5386): Bitwise aggregation functions do not work with wide decimals (updated previous PR)

AestheticAkhmad:
Updated the logic (as discussed in Zulip): bitwise operations work with wide decimals and result fits in int64_t.


drrtuy (changes requested):
Could you add similar test for decimal(18) column that is stored in 8 byte column internall?

AestheticAkhmad:
> Could you add similar test for decimal(18) column that is stored in 8 byte column internall?

Done. I described an issue (in Zulip chat: MCOL-5386) with the build after I pulled new commits. Could you please take a look?

drrtuy:
Thank you @AestheticAkhmad for the contribution!

drrtuy (approved):
untested ACK but we need to fix our regr before we merge.

PR #3525, opened 2025-04-29
feat(PrimProc): MCOL-5950 Improve disk-based aggregation finalization

mariadb-AlexeyAntipovsky:
Iterate over the rows in the plain vector of RGData instead of
    iterating over the hashmap. This reduces the complexity and speeds
    up finalization (by up to the twice in the certain cases)


utils/rowgroup/rowstorage.cpp
```diff
@@ -2267,94 +2305,62 @@ void RowAggStorage::finalize(std::function<void(Row&)> mergeFunc, Row& rowOut)
...
 
-        bool found = false;
-        auto& keyRow = fExtKeys ? fKeyRow : rowOut;
-        fKeysStorage->getRow(pos.idx, keyRow);
-        while (!found && prevInfo[pidx] == pinfo)
+        // TODO: store hashes in the RowGroupStorage?
+        uint64_t hash = hashRow(prevRow, fLastKeyCol);
+        auto [info, idx] = rowHashToIdx(hash);
+        nextWhileLess(info, idx);
+        uint64_t curidx = -1;
```
drrtuy: Mb worth turning this value into a named const?

drrtuy (commented):
Don't you want to reconsider `fStorage->dumpAll(false) in the end of finalize loop?` It looks like this eats up lots of needless IO.

mariadb-AlexeyAntipovsky:
> Don't you want to reconsider `fStorage->dumpAll(false) in the end of finalize loop?` It looks like this eats up lots of needless IO.

No, no, no, it's right where it should be (it's in an external loop, where we go through the generations into which we have to merge the data into - so at the end of each iteration we have to save the state of this generation before moving on to the next one). We could have boolean flags for the number of RowGroups, whether we changed this RowGroup, but `dumpAll` saves only those RowGroups that are in memory, which means one of the:
1. we merged something into some row in this rg
2. we got a row with the same hash, but not equal to the row in this rg.
IMO, case 2 is rare enough to noticeably affect performance.

utils/rowgroup/rowstorage.cpp
mariadb-AlexeyAntipovsky: :+1:

drrtuy (approved):
untested ACK

PR #3533, opened 2025-05-05
feat(plugin): MCOL-5998 unconditional innodb into MCS pushdown

drrtuy:


mariadb-SergeyZefirov (approved):
Appears to be a correct way to do that.

PR #3543, opened 2025-05-15
chore(build): refactor main CMakeLists.txt

mariadb-LeonidFedorov:
CMakeLIsts.txt refactored.
- INSTALL_LAYOUT condition removed
- main CMakeLists.txt was refactored and divided to smaller .cmake files
- WError only for bootstrap_mcs.sh (-DCOLUMNSTORE_MAINTAINER)
-  unconditional -fno-omit-frame-pointer, easier debugging and proper flamegraphs are expected
- cmake-format config added (as python code), but it's a config file 
- debian 11 was removed fom pipelines with it's unstable repos
- unittests are build with ASAN only for Ubuntu-24.04, due to strange faults with other distros 
- boost is build with the same compiler as the engine code, some supressions were added for one boost header
- stub for sanitizers pipeline was added to .drone.jsonnet


drrtuy (approved):
untested ACK

mariadb-AlexeyAntipovsky (approved):

PR #3568, opened 2025-05-30
feat/MCOL-6074: basic integrational tests for CMAPI

mariadb-AlexanderPresniakov:
Basic integrational testing for CMAPI using pytest + terraform + ansible
Real cluster is created (using columnstore-ansible-aws project), all the interactions happen on VMs, after the tests cluster is destroyed


cmapi/requirements-dev.txt
```diff
@@ -0,0 +1,3 @@
+# For integration tests
```
mariadb-AlanMologorsky, line 1: Please freeze indirect dependencies here.
Use requirements.txt as a reference.

cmapi/integration_tests/utils.py
```diff
@@ -0,0 +1,41 @@
+from contextlib import contextmanager
```
mariadb-AlanMologorsky: [styling]
Pls format imports in right order it will be helpful in the future work.

cmapi/integration_tests/utils.py
```diff
@@ -0,0 +1,41 @@
...
+import os
+
+
+@contextmanager
+def change_directory(new_dir):
+    """
+    Context manager for temporarily changing the current working directory.
+
+    Args:
```
mariadb-AlanMologorsky: To be honor all doctrings across CMAPI project are in sphinx style.
I'm not against Google style docstrings but this looks and reads inconsistent.
I'm not pushing to change it immediately, but at least make task to change current docstring styling to Google across all the CMAPI project or task to fix integration tests docstrings to sphinx style.

cmapi/integration_tests/conftest.py
```diff
@@ -0,0 +1,183 @@
...
+
+    cluster_already_exists = cluster_mgmt.cluster_exists(terraform_dir)
+
+    if create_cluster:
+        if cluster_already_exists:
+            print("Cluster already exists, destroying it first...")
+            cluster_mgmt.destroy_cluster(terraform_dir)
+        print("Creating the cluster...")
+        cluster_mgmt.create_cluster(terraform_dir)
+
```
mariadb-AlanMologorsky, line 75: Extra blank lines?

cmapi/integration_tests/conftest.py
```diff
@@ -0,0 +1,183 @@
...
+    Fixture to provide the cluster configuration.
+    """
+    with change_directory(terraform_dir):
+        # Read the cluster configuration from the terraform output
+        outputs = json.loads(subprocess.check_output(
+            ["terraform", "output", "-json"],
+            text=True,
+        ))
+
+        if outputs == {}:
```
mariadb-AlanMologorsky, line 95: `if outputs:`?

cmapi/integration_tests/conftest.py
```diff
@@ -0,0 +1,183 @@
...
+            pytest.fail("Terraform output is empty, looks like cluster is not yet created. Pass --create-cluster option to create it.")
+
+        key_file_path = Path(outputs["ssh_key_file"]["value"])
+        ssh_user = request.config.getoption("--ssh-user")
+
+        mcs_nodes = []
+        for host_descr in outputs["columnstore_nodes"]["value"]:
```
mariadb-AlanMologorsky, line 102: Is it safe here? Are 100% sure that this key exist? Get looks better in this case.

cmapi/integration_tests/integration_tests.yml
```diff
@@ -0,0 +1,23 @@
...
+- hosts: "primary,replicas"
+  become: yes
+  become_user: root
+  gather_facts: no
+
+  tasks:
+    # Install cs_package_manager
+    - name: "Download cs_package_manager"
+      get_url:
+        url: "https://raw.githubusercontent.com/mariadb-corporation/mariadb-columnstore-engine/refs/heads/cs_package_manager_v3.5/cmapi/scripts/cs_package_manager.sh"
```
mariadb-AlanMologorsky: We do have this script in columnstore repo.
As I know the newest version and all changes are there. Moreover we distribute it with CMAPI package.

cmapi/integration_tests/ssh.py
```diff
@@ -0,0 +1,120 @@
+import concurrent.futures
+from contextlib import contextmanager
+import json
+from pathlib import Path
+from typing import Any, Callable
+from fabric import Connection
+
+
+class HostConfig:
```
mariadb-AlanMologorsky: `@dataclass`?

cmapi/integration_tests/ssh.py
```diff
@@ -0,0 +1,120 @@
...
+        self.private_ip = private_ip
+        self.public_fqdn = public_fqdn
+        self.key_file_path = key_file_path
+        self.ssh_user = ssh_user
+
+    def __repr__(self) -> str:
+        return f"HostConfig(name={self.name}, private_ip={self.private_ip}, public_fqdn={self.public_fqdn}"
+
+    @contextmanager
+    def ssh_connection(self):
```
mariadb-AlanMologorsky, line 22: [minor]
Seems to be misleading naming of the class. <placeholder>Config can't do connections imo.

cmapi/integration_tests/ssh.py
```diff
@@ -0,0 +1,120 @@
...
+    def __init__(self, mcs_hosts: list[HostConfig], maxscale_hosts: list[HostConfig]) -> None:
+        self.mcs_hosts = mcs_hosts
+        self.maxscale_hosts = maxscale_hosts
+
+    def __repr__(self) -> str:
+        return f"ClusterConfig(mcs_hosts={self.mcs_hosts}"
+
+    @property
+    def primary(self) -> HostConfig:
+        return self.mcs_hosts[0]
```
mariadb-AlanMologorsky, line 68: Is it guartied that first in a list is primary all the time?

cmapi/integration_tests/ssh.py
```diff
@@ -0,0 +1,120 @@
...
+
+    with concurrent.futures.ThreadPoolExecutor(max_workers=max_workers) as executor:
+        # Start the function execution on each host
+        future_to_host = {executor.submit(func, host): host for host in hosts}
+
+        # Process results as they complete
+        for future in concurrent.futures.as_completed(future_to_host):
+            host = future_to_host[future]
+            try:
+                result = future.result()
```
mariadb-AlanMologorsky: May be having some timeout is better here?

cmapi/integration_tests/state.py
```diff
@@ -0,0 +1,24 @@
+from integration_tests.ssh import HostConfig
+
+def activate_single_node_config(host: HostConfig):
+    """
+    Stop MCS services on the host.
+    """
+    with host.ssh_connection() as conn:
+        print(f"{host.name}: Stopping MCS services...")
+        conn.sudo("systemctl stop mariadb mcs-workernode@1 mcs-workernode@2 mcs-writeengineserver mcs-storagemanager mcs-ddlproc mcs-dmlproc mcs-primproc mcs-controllernode", hide=True)
```
mariadb-AlanMologorsky, line 9: `mcs cluster stop` seems to be much much better and safer here.
Except `mariadb` service it should be stopped after and separately.
JFYI In next releases I can add cli command for it too cause I already has an implementation.

cmapi/integration_tests/test_add_node.py
```diff
@@ -0,0 +1,60 @@
+from integration_tests.ssh import ClusterConfig
+from integration_tests.utils import assert_dict_includes, drop_timestamp
+
+ALL_MCS_SERVICES = {"StorageManager", "workernode", "controllernode", "PrimProc", "WriteEngineServer", "DDLProc", "DMLProc"}
```
mariadb-AlanMologorsky, line 4: Better to get this from CMAPI code itself?

mariadb-AlanMologorsky (changes requested):

cmapi/integration_tests/conftest.py
mariadb-AlexanderPresniakov: No, in this case we know exactly what we are expecting (empty dictionary), general false value is not enough here

cmapi/integration_tests/conftest.py
mariadb-AlexanderPresniakov: Yes, we know it for sure, because we control contents of its output (in outputs.tf in columnstore-ansible-aws), and if it doesn't match -- it must fail

cmapi/integration_tests/ssh.py
mariadb-AlexanderPresniakov: No, it is not guaranteed, but we don't need such guarantees here, because this is just an alias to improve readability (compared to always calling .mcs_hosts[0], .mcs_hosts[1]). If you want to be super-specific, you can access .mcs_hosts[X] and call the variable accordingly

cmapi/integration_tests/state.py
mariadb-AlexanderPresniakov: Yes, but we are testing cmapi here, that's why i didn't want to use cmapi as a part of cmapi testing process. Like in unit tests, where some logic can be duplicated to avoid relying on the tested code

cmapi/integration_tests/test_add_node.py
mariadb-AlexanderPresniakov: Yes, but at the moment there is an ExeMgr service that this code would try to use

mariadb-AlanMologorsky (approved):
LGTM

PR #3570, opened 2025-06-01
chore(compilation): proper use of plugin_ref

mariadb-LeonidFedorov:
There is macro defined for proper use

```
#ifdef DBUG_OFF
typedef struct st_plugin_int *plugin_ref;
...
#define plugin_state(pi) ((pi)->state)
#else
typedef struct st_plugin_int **plugin_ref;
...
#define plugin_state(pi) ((pi)[0]->state)
#endif


drrtuy (approved):
untested ACK

PR #3574, opened 2025-06-03
MDEV-32854: json_depth

mariadb-RuchaDeodhar:
This patch is an extension for the https://jira.mariadb.org/browse/MDEV-32854 

Coumnstore uses some code related to json from the server. So this patch aims at keeping the json depth unchanged for columnstore.


mariadb-LeonidFedorov (changes requested):
The changeset is not compiling with the current 10.6 server

drrtuy:
> The changeset is not compiling with the current 10.6 server

As you might know the custom build feature doesn't work, so there is no way to check CI builds. I will check it manually with DEB and move the changes under preprocessor b/c the changes will become available with 12.1 AFAIK. @mariadb-RuchaDeodhar  correct me with the server's version, if I am wrong.

PR #3580, opened 2025-06-05
chore(bootstrap): bootstrap does not replace existing Columnstore.xml…

drrtuy:
… if the flag is set


mariadb-LeonidFedorov (approved):
untested

PR #3589, opened 2025-06-09
feat(optimizer): MCOL-5999/MCOL-6036 rule-based optimizer and initial parallel scan 4 CES

drrtuy:


dbcon/execplan/CMakeLists.txt
```diff
@@ -47,3 +47,11 @@ set(execplan_LIB_SRCS
 
 columnstore_library(execplan ${execplan_LIB_SRCS})
 columnstore_link(execplan messageqcpp ${NETSNMP_LIBRARIES} ${ENGINE_DT_LIB} pron loggingcpp)
+
+columnstore_executable(csep_rewrite csep_rewrite.cpp)
+columnstore_link(csep_rewrite
```
mariadb-LeonidFedorov: binary should be added to `debian/mariadb-plugin-columnstore.install` if you want it to be packaged or just use simple `add_executable`

drrtuy: That was a leftover that I deleted.

dbcon/mysql/rulebased_optimizer.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+// Apply iteratively until CSEP is converged by rule
+bool Rule::apply(execplan::CalpontSelectExecutionPlan& root) const
+{
+  bool changedThisRound = false;
+  bool hasBeenApplied = false;
+  do
+  {
+    changedThisRound = walk(root) && !applyOnlyOnce;
+    hasBeenApplied = changedThisRound;
```
mariadb-AlexeyAntipovsky: I'm a bit confused here - if `hasBeenApplied == changedThisRound` is `true` then another iteration will be executed. That is, only `false` can be returned. Is that supposed to be?

dbcon/mysql/rulebased_optimizer.cpp
```diff
@@ -0,0 +1,189 @@
...
+  for (auto& table : csep.derivedTableList())
+  {
+    auto* csepPtr = dynamic_cast<execplan::CalpontSelectExecutionPlan*>(table.get());
+    if (!csepPtr)
+    {
+      continue;
+    }
+
+    auto& csepLocal = *csepPtr;
+    rewrite |= walk(csepLocal);
```
mariadb-AlexeyAntipovsky, line 81: nit: Personally, I'm not comfortable using recursion in c++. Maybe rewrite it as an iterative traversal?

dbcon/mysql/rulebased_optimizer.cpp
drrtuy: Good catch.

dbcon/mysql/rulebased_optimizer.cpp
drrtuy: What is your concern with recursion for this case?
Recursion is easier to read but I agree that iterative traversal might be good here. If you don't mind I will leave this as an exercise b/c I need to deliver feature ASAP.

mariadb-AlexeyAntipovsky: Cause you never know when your stack might overflow (to be honest I don't think that it is significant concern here)

mariadb-AlexeyAntipovsky (approved):

mariadb-LeonidFedorov (approved):

drrtuy:
@mariadb-AlexeyAntipovsky @mariadb-LeonidFedorov check out the last commit plz.

mariadb-SergeyZefirov (approved):

mariadb-AlexeyAntipovsky (approved):

PR #3591, opened 2025-06-10
MDEV-32854: Make JSON_DEPTH_LIMIT configurable

mariadb-RuchaDeodhar:
This patch is the columnstore-part of the task.


drrtuy (changes requested):
Thx for the patch @mariadb-RuchaDeodhar .
There are two repetitive issues that you should fix:
1) remove extra boilerplate code that is copy-pasted in many places. Make it a common function and plz choose a meaningful name, so that a reader can understand what is the semantics of the code.
2) get rid of c-style casts replacing them with dynamic_cast + check if the cast was successful.

PR #3598, opened 2025-06-13
chore(ci): downgrade amazon/aws-cli image in publish step to avoid aws cli problems

mariadb-AlekseiBukhalov:


drrtuy (approved):
untested ack

PR #3603, opened 2025-06-17
MCOL-4833: verbose options for testS3Connection

mariadb-KristinaPavlova:
Add  verbose option for testS3Connection enabling (curl_verbose_option + ms3_debug)
[MCOL-4833](https://jira.mariadb.org/browse/MCOL-4833)

### DIFF
--- a/storage-manager/src/S3Storage.cpp
+++ b/storage-manager/src/S3Storage.cpp
@@ -112,7 +112,7 @@ S3Storage::ScopedConnection::~ScopedConnection()
   s3->returnConnection(conn);
 }
 
-S3Storage::S3Storage(bool skipRetry) : skipRetryableErrors(skipRetry)
+S3Storage::S3Storage(bool skipRetry, bool verbose) : skipRetryableErrors(skipRetry), verbose_enabled(verbose)
 {
   /*  Check creds from envvars
       Get necessary vars from config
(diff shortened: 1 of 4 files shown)


storage-manager/src/S3Storage.cpp
```diff
@@ -225,9 +225,9 @@ S3Storage::S3Storage(bool skipRetry) : skipRetryableErrors(skipRetry)
   }
 
   ms3_library_init();
-  if (libs3_debug == "enabled")
+  if (verbose_enabled || libs3_debug == "enabled")
   {
-    ms3_debug(1);
```
mariadb-LeonidFedorov, line 230: IMO that's wrong behaviour, you'll have ms3_debug turned off, if verbose is not enabled by command-line option but turned on in config file. 
Let's assume both option turn on verbose,
so it should be `ms3_debug(1);`

storage-manager/src/S3Storage.cpp
```diff
@@ -225,9 +225,9 @@ S3Storage::S3Storage(bool skipRetry) : skipRetryableErrors(skipRetry)
...
+    ms3_debug(verbose_enabled);
```
mariadb-LeonidFedorov: IMO that's wrong behaviour, you'll have ms3_debug turned off, if verbose is not enabled by command-line option but turned on in config file.
Let's assume both option turn on verbose,
so it should be ms3_debug(1);

mariadb-LeonidFedorov (approved):
untested ACK

PR #3613, opened 2025-06-20
MCOL-4240: Parsing support for ENUM with columnstore engine

mariadb-NedeljkoStefanovic:
It is done in the parser part. The current task is about what to add to the ProcessDDLStatement function. It is about how table with smallint is created, altered and about inserts and deletions.


dbcon/ddlpackage/serialize.cpp
```diff
@@ -1136,6 +1137,14 @@ int ColumnType::unserialize(ByteStream& bytestream)
   bytestream >> autoincrement;
   bytestream >> nextVal;
   bytestream >> charsetNum;
+  bytestream >> enumValuesNum;
+
```
mariadb-LeonidFedorov, line 1142: u can use 
`template <typename T>
void serializeInlineVector(ByteStream& bs, const std::vector<T>& v)
`
from `bytestream.h`

PR #3614, opened 2025-06-20
MCOL-5906: Fix error messages

mariadb-KristinaPavlova:
Subtask of [MCOL-5877](https://jira.mariadb.org/browse/MCOL-5877)
fixing erorr messages in 1) 2) 3) 6)  9) 10)


utils/joiner/joinpartition.cpp
```diff
@@ -821,7 +821,7 @@ uint64_t JoinPartition::writeByteStream(int which, ByteStream& bs)
     {
       fs.close();
       ostringstream os;
-      os << "Disk join could not write file " << filename << ": " << strerror(saveErrno) << endl;
+      os << "Disk join could not write to configured path check 'HashJoin' in columnstore.xml " << filename << ": " << strerror(saveErrno) << endl;
```
drrtuy: We should be able to retrieve the actual path from ResourceManager instance here. I suggest to look at RM ctor.

drrtuy: As per discussion, the filename path is actual and it is needed to retrieve the path from RM.

drrtuy: There are other small things: HashJoin doesn't have the temp path prefix but SystemConfig.SystemTempFileDir has it. I suggest to make C a capital letter in  `columnstore.xml`.

mariadb-LeonidFedorov (approved):

PR #3616, opened 2025-06-24
fix(cmapi): MCOL-5913 cmapi write local address 127.0.1.1 instead real ip address

mariadb-AlanMologorsky:
- [x] feat(cmapi): NetworkManager class for some ip hostname opoerations.
- [x] fix(cmapi): Use NetworkManager class to resolve ip and hostname in node_manipulation.add_node function
- [x] fix(cmapi): Minor docstring and formatting fixes

### DIFF
--- a/cmapi/cmapi_server/managers/network.py
+++ b/cmapi/cmapi_server/managers/network.py
@@ -1,0 +1,258 @@
+import errno
+import fcntl
+import logging
+import socket
+import struct
+from ipaddress import ip_address
+from typing import Optional
+
+try:
+    import psutil
+    _PSUTIL_AVAILABLE = True
+except ImportError:
+    psutil = None
+    _PSUTIL_AVAILABLE = False
+
+from cmapi_server.exceptions import CMAPIBasicError
+
+
+SIOCGIFADDR = 0x8915  # SIOCGIFADDR "socket ioctl get interface address"
+
+
+class NetworkManager:
+    @classmethod
+    def get_ip_version(cls, ip_addr: str) -> int:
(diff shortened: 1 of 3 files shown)


mariadb-LeonidFedorov (approved):
LGTM not tested

drrtuy (commented):
As per discussion in slack decided @mariadb-AlanMologorsky will look into `ipaddress` python lib to replace viable prefix check approach.

mariadb-AlanMologorsky:
@drrtuy rewrite and prettify my previous solution without ".local" usage.

cmapi/cmapi_server/managers/network.py
```diff
@@ -0,0 +1,215 @@
...
+                (
+                    'Unknown exception while getting IP address of an '
+                    f'{ifname!r} interface',
+                ),
+                exc_info=True
+            )
+        return ip_addr
+
+    @classmethod
+    def get_current_node_ips(
```
mariadb-AlexeyAntipovsky, line 137: IMO, using `psutil.net_if_addrs()` would be more understandable (and cross-platform)

mariadb-AlexeyAntipovsky (changes requested):

cmapi/cmapi_server/managers/network.py
mariadb-AlanMologorsky: Thx a lot! Good suggestion.
I'm planning to reuse this class in python files without psutil in future releases, so I added psutil support and keep fallback that uses batteries.

mariadb-AlexeyAntipovsky (approved):

PR #3618, opened 2025-06-24
MCOL-6058 fix mtrs on develop-23.02

mariadb-AlekseiBukhalov:
This patch improves handling of NULLs in textual fields in ColumnStore. Previously empty strings were considered NULLs and it could be a problem if data scheme allows for empty strings. It was also one of major reasons of behavior difference between ColumnStore and other engines in MariaDB family.

Also, this patch fixes some other bugs and incorrect behavior, for example, incorrect comparison for "column <= ''" which evaluates to constant True for all purposes before this patch.


mariadb-LeonidFedorov:
this should not be in develop-23.02

PR #3624, opened 2025-06-25
chore(MCOL-6018) Fix incorrect Field_decimal cast

mariadb-SergeyZefirov:
This is a fix of a problem found by UBSAN. MDB changed default type to represent a decimal result, C-style cast did not do proper type checking and this one-liner fixes that. Now we will have an assertion if type changes again.


mariadb-LeonidFedorov (changes requested):
It would not compile

mariadb-LeonidFedorov:
`<Field_decimal*<` is a simple mistype, please test your code before review request

mariadb-LeonidFedorov (approved):

PR #3629, opened 2025-06-27
chore(plugin): translator walks are now in separate units

drrtuy:
Tiny refactoring to break big unit into logical parts. JFYI nothing added/changed, just put symbols into separate units.


mariadb-SergeyZefirov (approved):
I agree with the changes.

mariadb-LeonidFedorov (approved):

PR #3630, opened 2025-06-27
MCOL-6071 drone refactor

mariadb-AlekseiBukhalov:


build/bootstrap_mcs.sh
```diff
@@ -62,6 +60,8 @@ optparse.define short=U long=ubsan desc="Build with UBSAN" variable=UBSAN defaul
 optparse.define short=v long=verbose desc="Verbose makefile commands" variable=MAKEFILE_VERBOSE default=false value=true
 optparse.define short=V long=add-branch-name-to-outdir desc="Add branch name to build output directory" variable=BRANCH_NAME_TO_OUTDIR default=false value=true
 optparse.define short=W long=without-core-dumps desc="Do not produce core dumps" variable=WITHOUT_COREDUMPS default=false value=true
+optparse.define short=s long=sccache desc="Build with sccache" variable=SCCACHE default=false value=true
```
mariadb-LeonidFedorov, line 63: just use $(arch), instead of extra option, the host knows it's architechture :)

build/bootstrap_mcs.sh
```diff
@@ -77,11 +77,49 @@ if [[ ! " ${DISTRO_OPTIONS[*]} " =~ " ${OS} " ]]; then
...
 fi
 
-pkg_format="deb"
 if [[ "$OS" == *"rocky"* ]]; then
     pkg_format="rpm"
+    CTEST_BIN_NAME=:"ctest3"
+else
+    pkg_format="deb"
+    CTEST_BIN_NAME="ctest"
 fi
```
mariadb-LeonidFedorov: please make this explicit function select_pkg_format, so that if we want to add Amazon Linux for example(which is rpm based), we know, what condition to change. Please check there is no similar checks anywhere else

build/bootstrap_mcs.sh
mariadb-AlekseiBukhalov: done

build/bootstrap_mcs.sh
mariadb-AlekseiBukhalov: done

mariadb-LeonidFedorov (approved):

PR #3632, opened 2025-06-30
MCOL-5964: no comm ro nodes

mariadb-SergeyZefirov:


writeengine/client/we_clients.cpp
```diff
@@ -224,6 +239,14 @@ void WEClients::Setup()
     snprintf(buff, sizeof(buff), "pm%u_WriteEngineServer", moduleID);
     string fServer(buff);
 
+    // Check if WES is configured for this module
+    if (!isWESConfigured(rm->getConfig(), fServer))
+    {
+      writeToLog(__FILE__, __LINE__, "Skipping WriteEngineServer client creation for " + fServer + " as the node is re
```
mariadb-AlexanderPresniakov: There is a typo: reAd-only

writeengine/client/we_clients.h
```diff
@@ -114,6 +114,18 @@ class WEClients
     return pmCount;
   }
 
+  uint32_t getRWConnections()
```
mariadb-AlexanderPresniakov: Just a style thing, but i would prefer "num" or "count" to be mentioned in the name, because it doesn't return connections

mariadb-SergeyZefirov:
This builds cleanly on Ubuntu 24.04.

mariadb-AlexanderPresniakov (approved):

drrtuy (approved):
untested ACK

PR #3641, opened 2025-07-07
MCOL-6071 drone refactor

mariadb-AlekseiBukhalov:


build/bootstrap_mcs.sh
```diff
@@ -586,7 +585,7 @@ run_unit_tests() {
     fi
 
     message "Running unittests"
-    cd $MARIA_BUILD_PATH
```
mariadb-LeonidFedorov: Looks that's wrong

mariadb-LeonidFedorov: You should run ctest from path, where CMakeFiles are generated, and it's $MARIA_BUILD_PATH

build/install_clang.sh
```diff
@@ -0,0 +1,32 @@
...
+fi
+
+change_ubuntu_mirror us
+
+message "Installing clang-${CLANG_VERSION}"
+
+if [[ "$CLANG_VERSION" == "20" ]]; then
+    retry_eval 5 apt-get clean && apt-get update && apt-get install -y wget curl lsb-release software-properties-common gnupg
+    wget https://apt.llvm.org/llvm.sh
+    bash llvm.sh 20
```
mariadb-LeonidFedorov: just pass the param instead of if else statements

build/bootstrap_mcs.sh
mariadb-AlekseiBukhalov: done

build/install_clang.sh
mariadb-AlekseiBukhalov: done

mariadb-LeonidFedorov (approved):

PR #3647, opened 2025-07-11
feat(MCOL-6082): Multiple readers of dbroots using OamCache logic

mariadb-SergeyZefirov:
This patch introduces centralized logic of selecting what dbroot is accessible in PrimProc on what node. The logic is in OamCache for time being and can be moved later.

### DIFF
--- a/dbcon/ddlpackageproc/droptableprocessor.cpp
+++ b/dbcon/ddlpackageproc/droptableprocessor.cpp
@@ -359,8 +359,7 @@ DropTableProcessor::DDLResult DropTableProcessor::processPackageInternal(ddlpack
       return result;
     }
 
-    boost::shared_ptr<std::map<int, int> > dbRootPMMap = oamcache->getDBRootToPMMap();
-    pmNum = (*dbRootPMMap)[dbRoot];
+    pmNum = oamcache->getOwnerPM(dbRoot);
 
     try
     {
@@ -464,7 +463,7 @@ DropTableProcessor::DDLResult DropTableProcessor::processPackageInternal(ddlpack
       return result;
     }
 
-    pmNum = (*dbRootPMMap)[dbRoot];
+    pmNum = oamcache->getOwnerPM(dbRoot);
 
     try
     {
@@ -1275,8 +1274,7 @@ TruncTableProcessor::DDLResult TruncTableProcessor::processPackageInternal(ddlpa
       bytestream << (uint32_t)colType.compressionType;
     }
 
-    boost::shared_ptr<std::map<int, int> > dbRootPMMap = oamcache->getDBRootToPMMap();
-    pmNum = (*dbRootPMMap)[useDBRoot];
+    pmNum = oamcache->getOwnerPM(useDBRoot);
 
     try
     {
--- a/dbcon/dmlpackageproc/dmlpackageprocessor.h
+++ b/dbcon/dmlpackageproc/dmlpackageprocessor.h
@@ -165,8 +165,6 @@ class DMLPackageProcessor
       std::cout << "Cannot make connection to WES" << std::endl;
     }
 
-    oam::OamCache* oamCache = oam::OamCache::makeOamCache();
-    fDbRootPMMap = oamCache->getDBRootToPMMap();
     fDbrm = aDbrm;
     fSessionID = sid;
     fExeMgr = new execplan::ClientRotator(fSessionID, "ExeMgr");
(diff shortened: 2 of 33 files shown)


dbcon/mysql/is_columnstore_extents.cpp
```diff
@@ -202,7 +202,7 @@ static int is_columnstore_extents_fill(THD* thd, TABLE_LIST* tables, COND* cond)
   TABLE* table = tables->table;
 
   BRM::DBRM::refreshShmWithLock();
-  std::unique_ptr<BRM::DBRM> emp(new BRM::DBRM());
+  BRM::DBRM* emp = new BRM::DBRM();
```
mariadb-AlexanderPresniakov: Why was the smart pointer replaced with a dumb one here? And there is no explicit delete of the emp

dbcon/mysql/is_columnstore_files.cpp
```diff
@@ -216,7 +217,7 @@ static int generate_result(BRM::OID_t oid, BRM::DBRM* emp, TABLE* table, THD* th
 static int is_columnstore_files_fill(THD* thd, TABLE_LIST* tables, COND* cond)
 {
   BRM::DBRM::refreshShmWithLock();
-  std::unique_ptr<BRM::DBRM> emp(new BRM::DBRM());
```
mariadb-AlexanderPresniakov: And the same thing here (the pointer IQ change)

mariadb-AlexanderPresniakov (commented):
Apart from two smart pointer conversions everything is great in my opinion
PS: also i think commented out code must be removed

dbcon/mysql/is_columnstore_files.cpp
mariadb-SergeyZefirov: Valid points, thanks!

drrtuy:
If the patch delivers read-only functionality we should clean up the commented code.

dbcon/dmlpackageproc/dmlpackageprocessor.h
```diff
@@ -489,6 +487,8 @@ class DMLPackageProcessor
                      uint32_t tableOid);
   int endTransaction(uint64_t uniqueId, BRM::TxnID txnID, bool success);
 
+  oam::OamCache* oamcache() { return oam::OamCache::makeOamCache(); }
```
mariadb-AlexeyAntipovsky, line 490: nit: `oamCache` looks like a more consistent naming, imo

oam/oamcpp/oamcache.cpp
```diff
@@ -73,7 +81,7 @@ void OamCache::checkReload()
   idbassert(txt != "");
   numDBRoots = config->fromText(txt);
 
-  dbRootPMMap.reset(new map<int, int>());
+  dbRootPMMap.reset(new map<int, set<int>>());
```
mariadb-AlexeyAntipovsky, line 84:
```suggestion
  dbRootPMMap.reset(new decltype(dbRootPMMap)::element_type());
```

mariadb-AlexeyAntipovsky (approved):

oam/oamcpp/oamcache.cpp
drrtuy: +1

drrtuy (approved):
untestedACK

dbcon/ddlpackageproc/droptableprocessor.cpp
```diff
@@ -425,8 +424,7 @@ DropTableProcessor::DDLResult DropTableProcessor::processPackageInternal(ddlpack
              << endl;
         Message::Args args;
         Message message(9);
-        args.add(fTxnid.id);
-        args.add(" Error in dropping table from systables.");
+        args.add("Error in dropping table from systables.");
```
drrtuy: Why do we remove `args.add(fTxnid.id)` ?

drrtuy (changes requested):
Plz address the comment.

dbcon/dmlpackageproc/dmlpackageprocessor.h
mariadb-SergeyZefirov: It is used as ```oamcache = ...``` throughout the code, thus, method has the same name.

oam/oamcpp/oamcache.cpp
mariadb-SergeyZefirov: I will remember this for future, but politely decline change for this concrete code, for historic (code style) reasons.

PR #3650, opened 2025-07-12
MCOL-6090

vuvova:
do not install anything unconditionally, before checking whether the target exists


mariadb-LeonidFedorov (approved):
Thank you!

PR #3653, opened 2025-07-14
MCOL-6098: Add vanilla rockylinux with gcc-8.5 to CI

mariadb-LeonidFedorov:
regular rockylinux 8 build will use plain vanilla gcc8.5, and run singlenode mtr
gcc-toolset will use gcc-toolset-11 and run multinode_mtr

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -16,33 +16,43 @@ local platforms_arm = {
 };
 
 local builddir = "verylongdirnameforverystrangecpackbehavior";
+
+local get_build_command(command) = "bash /mdb/" + builddir + "/storage/columnstore/columnstore/build/" + command + " ";
+
+local clang(version) = [get_build_command("install_clang_deb.sh " + version),
+                        get_build_command("update-clang-version.sh " + version + " 100"),
+                        get_build_command("install_libc++.sh " + version),
+                        "export CC=/usr/bin/clang",
+                        "export CXX=/usr/bin/clang++"
+                        ];
+
 local customEnvCommandsMap = {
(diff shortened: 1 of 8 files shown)


.drone.jsonnet
```diff
@@ -16,33 +16,43 @@ local platforms_arm = {
 };
 
 local builddir = "verylongdirnameforverystrangecpackbehavior";
+
+local get_build_command(command) = "bash /mdb/" + builddir + "/storage/columnstore/columnstore/build/" + command + " ";
+
+local clang(version) = [get_build_command("install_clang_deb.sh " + version),
+                        get_build_command("update-clang-version.sh " + version + " 100"),
+                        get_build_command("install_libc++.sh " + version),
```
mariadb-AlekseiBukhalov, line 24: maybe these 3 should be as one script?

mariadb-AlekseiBukhalov (approved):

writeengine/bulk/we_cmdargs.cpp
```diff
@@ -303,7 +303,7 @@ void WECmdArgs::parseCmdLineArgs(int argc, char** argv)
       fAllowMissingColumn = true;
     }
   }
-  if (vm.contains("max-errors"))
+  if (vm.count("max-errors"))
```
drrtuy, line 306: This looks like a part of another patch that handles multiple distros and needs older version of lohman @mariadb-LeonidFedorov ?

mariadb-LeonidFedorov: vm is a map, so contains is c++20 and count is the same for c++ olders

drrtuy (approved):
untested ACK

PR #3656, opened 2025-07-16
Added the support for pkgconf when finding system thrift

PavolSloboda:
as fedora uses pkgconf to help locate the thrift files during compilation instead of a Thrift.cmake file. I have added this logic in such a way that the existing logic should not be affected. Therefore enabling the use of system thrift even without pkgconf.


mariadb-LeonidFedorov (approved):

PR #3661, opened 2025-07-18
feat(cmapi): MCOL-6006: Disable failover when shared storage not detected

mariadb-AlanMologorsky:
feat(cmapi,failover): MCOL-6006 Disable failover when shared storage not detected

- Add SharedStorageMonitor thread to periodically verify shared storage:
  * Writes a temp file to the shared location and validates MD5 from all nodes.
  * Skips nodes with unstable recent heartbeats; retries once; defers decision if any node is unreachable.
  * Updates a cluster-wide stateful flag (shared_storage_on) only on conclusive checks.
- New CMAPI endpoints:
  * PUT /cmapi/{ver}/cluster/check-shared-storage — orchestrates cross-node checks.
  * GET /cmapi/{ver}/node/check-shared-file — validates a given file’s MD5 on a node.
  * PUT /cmapi/{ver}/node/stateful-config — fast path to distribute stateful config updates.
- Introduce in-memory stateful config (AppStatefulConfig) with versioned flags (term/seq) and shared_storage_on flag:
  * Broadcast via helpers.broadcast_stateful_config and enhanced broadcast_new_config.
  * Config PUT is now validated with Pydantic models; supports stateful-only updates and set_mode requests.
- Failover behavior:
  * NodeMonitor keeps failover inactive when shared_storage_on is false or cluster size < 3.
  * Rebalancing DBRoots becomes a no-op when shared storage is OFF (safety guard).
- mcl status improvements: per-node 'state' (online/offline), better timeouts and error reporting.
- Routing/wiring: add dispatcher routes for new endpoints; add ClusterModeEnum.
- Tests: cover shared-storage monitor (unreachable nodes, HB-based skipping), node manipulation with shared storage ON/OFF, and server/config flows.
- Dependencies: add pydantic; minor cleanups and logging.

### DIFF
--- a/cmapi/cmapi_server/controllers/api_clients.py
+++ b/cmapi/cmapi_server/controllers/api_clients.py
@@ -1,8 +1,9 @@
 import logging
-from typing import Any, Dict, Optional, Union
+from typing import Any, Dict, List, Optional, Union
 
 import pyotp
 import requests
+
 
 from cmapi_server.constants import (
     CMAPI_CONF_PATH, CURRENT_NODE_CMAPI_URL, SECRET_KEY, _version
@@ -31,19 +32,16 @@ class BaseClient:
 
         :param base_url: The base URL for the API endpoints,
                          defaults to CURRENT_NODE_CMAPI_URL
-        :type base_url: str, optional
-        :param request_timeout: request timeout, defaults to None
-        :type request_timeout: Optional[float], optional
         """
         self.base_url = base_url
         self.request_timeout = request_timeout
-        self.cmd_url = None
+        self.cmd_class = None
 
     def _request(
         self, method: str, endpoint: str,
         data: Optional[Dict[str, Any]] = None,
         throw_real_exp: bool = False
-    ) -> Union[Dict[str, Any], Dict[str, str]]:
+    ) -> Union[Dict[str, Any], List[Any]]:
         """Make a request to the API.
 
         :param method: The HTTP method to use.
@@ -51,7 +49,8 @@ class BaseClient:
         :param data: The data to send with the request.
         :return: The response from the API.
         """
-        url = f'{self.cmd_url}/{endpoint}'
+
+        url = f'{self.base_url}/cmapi/{_version}/{self.cmd_class}/{endpoint}'
         cmapi_cfg_parser = get_config_parser(CMAPI_CONF_PATH)
         key = get_current_key(cmapi_cfg_parser)
         headers = {'x-api-key': key}
@@ -59,10 +58,10 @@ class BaseClient:
             headers['Content-Type'] = 'application/json'
             data = {'in_transaction': True, **(data or {})}
         try:
-            response = requests.request(
+            response = get_traced_session().request(
                 method, url, headers=headers,
                 params=data if method == 'GET' else None,
-                json=data if method in ('PUT', 'POST') else None,
+                json=data if method in ('PUT', 'POST', 'DELETE') else None,
                 timeout=self.request_timeout, verify=False
             )
             response.raise_for_status()
@@ -76,11 +75,12 @@ class BaseClient:
             if throw_real_exp:
                 raise exc
             raise CMAPIBasicError(message)
-        # TODO: different handler for timeout exception?
-        except requests.exceptions.HTTPError as exc:
+        except requests.HTTPError as exc:
             resp = exc.response
+            request_url = exc.request.url if exc.request else url
+            status_code = resp.status_code if resp is not None else 'N/A'
             error_msg = str(exc)
-            if resp.status_code == 422:
+            if status_code == 422:
                 # in this case we think cmapi server returned some value but
                 # had error during running endpoint handler code
                 try:
@@ -89,19 +89,26 @@ class BaseClient:
                 except requests.exceptions.JSONDecodeError:
                     error_msg = resp.text
             message = (
-                f'API client got an exception in request to {exc.request.url} '
-                f'with code {resp.status_code if resp is not None else "NA"} '
-                f'and error: {error_msg}'
+                f'API client got an HTTPError exception in request to {request_url} with code '
+                f'{status_code} and error: {error_msg}'
             )
             logging.error(message)
             if throw_real_exp:
                 raise exc
             raise CMAPIBasicError(message)
+        except requests.exceptions.Timeout:
+            message = f'Request to {url} timed out after {self.request_timeout}'
+            logging.error(message)
+            if throw_real_exp:
+                raise exc
+            raise CMAPIBasicError(message)
         except requests.exceptions.RequestException as exc:
-            status_code = exc.response.status_code if exc.response else 'NA'
+            resp = exc.response
+            request_url = exc.request.url if exc.request else url
+            status_code = resp.status_code if exc.response is not None else 'N/A'
             message = (
                 'API client got an undefined error in request to '
-                f'{exc.request.url} with code {status_code!r} and '
+                f'{request_url} with code {status_code!r} and '
                 f'error: {str(exc)}'
             )
             logging.error(message)
@@ -120,7 +127,7 @@ class ClusterControllerClient(BaseClient):
             self, base_url: str = CURRENT_NODE_CMAPI_URL,
             request_timeout: Optional[float] = None
     ):
-        """Initialize the ClusterControllerClient with the base URL.
+        """Initialize the BaseClient with the base URL.
 
         :param base_url: The base URL for the API endpoints,
                          defaults to CURRENT_NODE_CMAPI_URL
@@ -129,7 +136,7 @@ class ClusterControllerClient(BaseClient):
         :type request_timeout: Optional[float], optional
         """
         super().__init__(base_url, request_timeout)
-        self.cmd_url = f'{self.base_url}/cmapi/{_version}/cluster'
+        self.cmd_class = 'cluster'
 
     def start_cluster(
             self, extra: Dict[str, Any] = dict()
@@ -307,6 +314,15 @@ class ClusterControllerClient(BaseClient):
             'PUT', 'upgrade-cmapi', {'version': version, **extra}
         )
 
+    def check_shared_storage(
+            self, extra: Dict[str, Any] = dict()
+    ) -> Union[Dict[str, Any], Dict[str, str]]:
+        """Check if shared storage working.
+
+        :return: The response from the API.
+        """
+        return self._request('PUT', 'check-shared-storage', extra)
+
 
 class NodeControllerClient(BaseClient):
     """Client for the NodeController API.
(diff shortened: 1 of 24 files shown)


cmapi/cmapi_server/controllers/api_clients.py
```diff
@@ -4,33 +4,114 @@
...
-from cmapi_server.controllers.dispatcher import _version
 from cmapi_server.constants import (
     CMAPI_CONF_PATH, CURRENT_NODE_CMAPI_URL, SECRET_KEY,
 )
 from cmapi_server.exceptions import CMAPIBasicError
 from cmapi_server.helpers import get_config_parser, get_current_key
 
 
-class ClusterControllerClient:
+_version = '0.4.0'
```
mariadb-AlexanderPresniakov: Maybe move it to constants.py?

cmapi/cmapi_server/controllers/api_clients.py
```diff
@@ -4,33 +4,114 @@
...
+            return response.json()
+        # TODO: different handler for timeout exception?
+        except requests.HTTPError as exc:
+            resp = exc.response
+            error_msg = str(exc)
+            if resp.status_code == 422:
+                # in this case we think cmapi server returned some value but
+                # had error during running endpoint handler code
+                try:
+                    resp_json = response.json()
```
mariadb-AlexanderPresniakov: Better use resp instead of response here

cmapi/cmapi_server/helpers.py
```diff
@@ -324,24 +319,32 @@ def broadcast_new_config(
...
-    with open(sm_config_filename) as f:
-        sm_config_text = f.read()
-
     headers = {'x-api-key': key}
-    body = {
-        'manager': root.find('./ClusterManager').text,
-        'revision': root.find('./ConfigRevision').text,
-        'timeout': 300,
-        'config': config_text,
-        'cs_config_filename': cs_config_filename,
```
mariadb-AlexanderPresniakov, line 350: mcs_config_filename

cmapi/cmapi_server/controllers/endpoints.py
```diff
@@ -345,17 +344,58 @@ def put_config(self):
         request_revision = request_body.get('revision', None)
         request_manager = request_body.get('manager', None)
         request_timeout = request_body.get('timeout', None)
+        only_stateful_config = request_body.get('only_stateful_config', False)
```
mariadb-AlexanderPresniakov: I propose to create a new endpoint for the stateful config. This endpoint became very complex and the stateful logic almost doesn't intersect with the XML config functionality. So it will be easier to read/keep in mind/use/test if it is a separate entity

cmapi/cmapi_server/controllers/endpoints.py
```diff
@@ -345,17 +344,58 @@ def put_config(self):
...
+        if only_stateful_config:
+            # if stateful config is provided, we do not need other params
+            mandatory = (request_timeout, request_stateful_config)
+        else:
+            mandatory = (request_revision, request_manager, request_timeout)
         if None in mandatory:
             raise_422_error(
                 module_logger, func_name, 'Mandatory attribute is missing.')
 
+        if request_stateful_config:
```
mariadb-AlexanderPresniakov: There are many parameters, some of them are optional and all of them are typed. Some parameters, like stateful config, are nested. I think it's time for us to start using Pydantic, which is almost a standard tool for such tasks, and it provides validation, readable diagnostics, optional\default fields, optional parameter descriptions and static typing, so IDE will highlight wrong names\values\incomplete parameter sets, etc.
It even provides discriminators -- one could determine which sets of fields is required in some request depending on some field value, in this case it would be only_stateful_config

cmapi/cmapi_server/managers/application.py
```diff
@@ -27,3 +30,87 @@ def get_version(cls) -> str:
...
+
+    Uses a versioned config with thread-safe updates to avoid stale writes.
+    Flags are stored as key-value pairs in a dictionary.
+    # TODO: Change version.term after primary changes.
+    """
+
+    _lock = threading.RLock()
+    _version = StateConfigVersion(term=0, seq=0)
+    # declare flags with default values
+    _flags: dict[str, Any] = {
```
mariadb-AlexanderPresniakov: I propose to use a typed container for flags too, because we know which fields are allowed and what are their types. A dataclass would do, but pydantic model would be even better, because we would get typing\validation\etc all the way up, and it would be impossible to make a typo in a key name, for example. Also it would be very helpful if in the future more complex nested fields would be introduced. Pydantic models can be converted to dict or to json or vice versa at any moment.

```
from pydantic import BaseModel, Field

class StateConfigVersionModel(BaseModel):
    term: int = Field(..., ge=0)
    seq: int = Field(..., ge=0)

class StatefulFlagsModel(BaseModel):
    shared_storage_on: bool = False

class StatefulConfigModel(BaseModel):
    version: StateConfigVersionModel
    flags: StatefulFlagsModel

In the endpoint:
payload = StatefulConfigModel.model_validate(cherrypy.request.json)
if payload.flags.shared_storage_on: ...
```

cmapi/cmapi_server/handlers/cluster.py
```diff
@@ -399,3 +406,68 @@ def set_log_level(
...
+        active_nodes = get_active_nodes()
+        all_responses: dict = dict()
+        sm_parser = configparser.ConfigParser()
+        sm_config_str = NodeConfig().get_current_sm_config()
+        sm_parser.read_string(sm_config_str)
+        storage_type = sm_parser.get(
+            'ObjectStorage', 'service', fallback='LocalStorage'
+        )
+        file_dir = '/var/lib/columnstore/data1'
+        if storage_type.lower == 's3':
```
mariadb-AlexanderPresniakov: Lower is a function

cmapi/cmapi_server/handlers/cluster.py
```diff
@@ -399,3 +406,68 @@ def set_log_level(
...
+            file_dir = '/var/lib/columnstore/storagemanager/metadata/data1'
+
+        with tempfile.NamedTemporaryFile(
+            mode='wb+', delete=True, dir=file_dir, prefix='mcs_test_shared_'
+        ) as temp_file:
+            file_data = rb'File to check shared storage working.'
+            temp_file.write(file_data)
+            tmp_file_md5 = hashlib.md5(file_data)
```
mariadb-AlexanderPresniakov: md5 returns a hash object, it is better to convert it to str with .hexdigest()

cmapi/cmapi_server/handlers/cluster.py
```diff
@@ -399,3 +406,68 @@ def set_log_level(
...
+                    base_url=f'https://{node}:{CMAPI_PORT}'
+                )
+                node_response = client.check_shared_file(
+                    file_path=tmp_file_path, check_sum=tmp_file_md5
+                )
+                logging.debug(f'Finished checking file on {node!r}')
+                all_responses[node] = node_response
+
+        nodes_success_responses = (
+            v['success'] for _, v in all_responses.items()
```
mariadb-AlexanderPresniakov: v['success'] for v in all_responses.values()?

mariadb-AlexanderPresniakov: Also this is dangerous, better convert the generator to list (because once you iterate this generator in all, you will not get anything in sum and later)

mariadb-AlexanderPresniakov (changes requested):

cmapi/cmapi_server/controllers/api_clients.py
mariadb-AlanMologorsky: Done.

cmapi/cmapi_server/controllers/api_clients.py
mariadb-AlanMologorsky: Nice catch!
Done.

cmapi/cmapi_server/helpers.py
mariadb-AlanMologorsky: No idea why this rename is needed.

cmapi/cmapi_server/controllers/endpoints.py
mariadb-AlanMologorsky: I implemented a new endpoint, but after some investigation I found that this change requires many more modifications in various places, which would significantly increase the PR size. Therefore, let's move it to a separate task.

JFYI, I’ve kept the new endpoint in this PR, marked for future releases.

cmapi/cmapi_server/controllers/endpoints.py
mariadb-AlanMologorsky: Implemented Models and some validation.

cmapi/cmapi_server/managers/application.py
mariadb-AlanMologorsky: Done.

cmapi/cmapi_server/handlers/cluster.py
mariadb-AlanMologorsky: Good catch! Thx. Done.

cmapi/cmapi_server/handlers/cluster.py
mariadb-AlanMologorsky: Done.

mariadb-AlanMologorsky: Very good catch.
Done.

cmapi/cmapi_server/handlers/cluster.py
mariadb-AlanMologorsky: Done.

cmapi/cmapi_server/helpers.py
mariadb-AlexanderPresniakov: In the receiving code it is mcs_config_filename, so everything will break without it

mariadb-AlanMologorsky: O_o Great catch from an old times. Fixed it but potentially it could break something if wrong filepath passed to an endpoint. Before it's everytime takes default path from constants. JFYI For now we never change mcs config filepath, so this is legacy argument, that never changes its value.

cmapi/cmapi_server/controllers/endpoints.py
mariadb-AlexanderPresniakov: Ok, then please add pydantic schema for this endpoint too (i think you'll need a discriminator feature to determine which set of arguments is required depending on one parameter)

mariadb-AlanMologorsky: Seems more investigation is needed on the discriminator feature.
I’m not sure it’s a good idea to add it in the current request (already 900+ lines added).
The time spent on the current solution is getting ridiculous. It seems better to test it and merge first, and then handle the additional tasks afterwards.

mariadb-AlexanderPresniakov (approved):
LGTM
(but the test needs fixing)

mariadb-LeonidFedorov (approved):
The current state of code looks good to me, the approval doesn't consider merge without fixes of tests, but the approach and the implementation look neat

mariadb-LeonidFedorov (approved):

cmapi/cmapi_server/controllers/endpoints.py
mariadb-AlanMologorsky: Done.

PR #3666, opened 2025-07-22
feat(optimizer): MCOL-5250 rewrite queries with DISTINCT

mariadb-AlexeyAntipovsky:
... as aggregated queries.
So query
```
SELECT DISTINCT <cols list>
FROM <from list>
WHERE <where clause>
GROUP BY <aggregate>
HAVING <having clause>
ORDER BY <orderby list>
LIMIT <limit>
```
will become
```
SELECT *
FROM
  (
    SELECT <cols list>
    FROM <from list>
    WHERE <where clause>
    GROUP BY <aggregate>
    HAVING <having clause>
  ) a
GROUP BY 1,2,3,...,N
ORDER BY <orderby list>
LIMIT limit
```

### DIFF
--- a/dbcon/execplan/calpontselectexecutionplan.cpp
+++ b/dbcon/execplan/calpontselectexecutionplan.cpp
@@ -1205,56 +1205,32 @@ SCSEP CalpontSelectExecutionPlan::clone()
   for (const auto& subPlan : fSubSelects)
   {
     auto* subCSEP = dynamic_cast<CalpontSelectExecutionPlan*>(subPlan.get());
-    if (subCSEP)
-    {
-      newPlan->fSubSelects.push_back(subCSEP->clone());
-    }
-    else
-    {
-      throw runtime_error("CalpontSelectExecutionPlan::clone() - subPlan is not a CalpontSelectExecutionPlan");
-    }
+    idbassert_s(subCSEP != nullptr, "subPlan is not a CalpontSelectExecutionPlan");
+    newPlan->fSubSelects.push_back(subCSEP->clone());
   }
 
   newPlan->fDerivedTableList.clear();
   for (const auto& drvTable: fDerivedTableList)
   {
     auto* drvCSEP = dynamic_cast<CalpontSelectExecutionPlan*>(drvTable.get());
-    if (drvCSEP)
-    {
-      newPlan->fDerivedTableList.push_back(drvCSEP->clone());
-    }
-    else
-    {
-      throw runtime_error("CalpontSelectExecutionPlan::clone() - drvTable is not a CalpontSelectExecutionPlan");
-    }
+    idbassert_s(drvCSEP != nullptr, "derivedTable is not a CalpontSelectExecutionPlan");
+    newPlan->fDerivedTableList.push_back(drvCSEP->clone());
   }
 
   newPlan->fUnionVec.clear();
   for (const auto& subPlan : fUnionVec)
   {
     auto* subCSEP = dynamic_cast<CalpontSelectExecutionPlan*>(subPlan.get());
-    if (subCSEP)
-    {
-      newPlan->fUnionVec.push_back(subCSEP->clone());
-    }
-    else
-    {
-      throw runtime_error("CalpontSelectExecutionPlan::clone() - subPlan is not a CalpontSelectExecutionPlan");
-    }
+    idbassert_s(subCSEP != nullptr, "unionVec is not a CalpontSelectExecutionPlan");
+    newPlan->fUnionVec.push_back(subCSEP->clone());
   }
 
   newPlan->fSelectSubList.clear();
   for (const auto& subPlan : fSelectSubList)
   {
     auto* subCSEP = dynamic_cast<CalpontSelectExecutionPlan*>(subPlan.get());
-    if (subCSEP)
-    {
-      newPlan->fSelectSubList.push_back(subCSEP->clone());
-    }
-    else
-    {
-      throw runtime_error("CalpontSelectExecutionPlan::clone() - subPlan is not a CalpontSelectExecutionPlan");
-    }
+    idbassert_s(subCSEP != nullptr, "subPlan is not a CalpontSelectExecutionPlan");
+    newPlan->fSelectSubList.push_back(subCSEP->clone());
   }
 
   newPlan->fSubSelectList.clear();
--- a/dbcon/execplan/calpontselectexecutionplan.cpp
+++ b/dbcon/execplan/calpontselectexecutionplan.cpp
@@ -1197,4 +1197,73 @@ execplan::SCSEP CalpontSelectExecutionPlan::cloneForTableWORecursiveSelectsGbObH
   return newPlan;
 }
 
+SCSEP CalpontSelectExecutionPlan::clone()
+{
+  auto newPlan = cloneWORecursiveSelects();
+
+  newPlan->fSelectSubList.clear();
+  for (const auto& subPlan : fSubSelects)
+  {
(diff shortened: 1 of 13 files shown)


dbcon/mysql/rbo_apply_rewrite_distinct.cpp
```diff
@@ -0,0 +1,181 @@
...
+  rcCloned->derivedTable(tableAlias);
+  rcCloned->derivedRefCol(rc.get());
+  rcCloned->resultType(rc->resultType());
+  rcCloned->operationType(rc->operationType());
+  rcCloned->colPosition(colPos);
+
+  if (const auto* rcsc = dynamic_cast<execplan::SimpleColumn*>(rc.get()); rcsc != nullptr)
+  {
+    rcCloned->columnName(rcsc->columnName());
+    rcCloned->timeZone(rcsc->timeZone());
```
drrtuy: Can't you blindly set tz?

dbcon/mysql/rbo_apply_rewrite_distinct.cpp
```diff
@@ -0,0 +1,181 @@
...
+#include "existsfilter.h"
+#include "functioncolumn.h"
+#include "logicoperator.h"
+
+namespace optimizer
+{
+
+bool matchRewriteDistinct(execplan::CalpontSelectExecutionPlan& csep)
+{
+  return csep.distinct() && csep.tableList().size() > 0;
```
drrtuy: Are you sure that we want to re-write any DISTINCT queries? Originally we discussed we use the rule for the category with DISTINCT + WFs.

dbcon/execplan/calpontselectexecutionplan.cpp
```diff
@@ -966,4 +966,73 @@ execplan::SCSEP CalpontSelectExecutionPlan::cloneWORecursiveSelects()
...
+  for (const auto& subPlan : fSubSelects)
+  {
+    auto* subCSEP = dynamic_cast<CalpontSelectExecutionPlan*>(subPlan.get());
+    if (subCSEP)
+    {
+      newPlan->fSubSelects.push_back(subCSEP->clone());
+    }
+    else
+    {
+      throw runtime_error("CalpontSelectExecutionPlan::clone() - subPlan is not a CalpontSelectExecutionPlan");
```
drrtuy: I think we need a specific exception class for the purpose.
JFYI I plan to add try..catch somewhere to restore from RBO failures, so the failed rules can be ignored.

drrtuy (commented):
Need some responses.

dbcon/mysql/rbo_apply_rewrite_distinct.cpp
```diff
@@ -0,0 +1,181 @@
...
+  // fill SimpleColumn data
+  rcCloned->schemaName("");
+  rcCloned->tableName(tableAlias);
+  rcCloned->oid(0);
+  rcCloned->tableAlias(tableAlias);
+  rcCloned->data("");
+
+  // fill ReturnedColumn data
+  rcCloned->charsetNumber(rc->charsetNumber());
+  rcCloned->alias(tableAlias + "." + rc->alias());
```
drrtuy: Both tableAlias and column alias should be in backtics as other columns are.

dbcon/mysql/rbo_apply_rewrite_distinct.cpp
mariadb-AlexeyAntipovsky: `ReturnColumn` type does not have a timeZone field. But I'm not sure if it is necessary.

dbcon/mysql/rbo_apply_rewrite_distinct.cpp
mariadb-AlexeyAntipovsky: Why not? At the moment, with this rewrite, we lost only optimization of the `SELECT DISTINCT ... LIMIT n` (without `ORDER BY`) query. But we can use an approximation for an early cut-off of the aggregation phase in this case.

dbcon/execplan/calpontselectexecutionplan.cpp
mariadb-AlexeyAntipovsky: Good point.

mariadb-SergeyZefirov:
You cannot push arbitrary HAVING predicates into subquery. The only predicates that can be pushed there should operate on the GROUP BY keys.

One example would be ``SELECT COUNT(x) FROM t ORDER BY y HAVING COUNT(x) > 10`` - if you push HAVING predicate into subquery, you will get implicit GROUP BY on an empty key and incorrect filtering.

drrtuy:
> You cannot push arbitrary HAVING predicates into subquery. The only predicates that can be pushed there should operate on the GROUP BY keys.
> 
> One example would be `SELECT COUNT(x) FROM t ORDER BY y HAVING COUNT(x) > 10` - if you push HAVING predicate into subquery, you will get implicit GROUP BY on an empty key and incorrect filtering.

Good point but there are couple details to consider.
- Rule won't re-write the query presented b/c there is no DISTINCT in it
- According with the code GB columns are also pushed into a new derived, so pushing HAVING is legit IMHO

dbcon/mysql/rbo_apply_rewrite_distinct.cpp
```diff
@@ -0,0 +1,199 @@
...
+  }
+  rc->incRefCount();
+
+  if (!rc->alias().empty())
+  {
+    colName = rc->alias();
+  }
+  if (colName.empty())
+  {
+    colName = "$col_" + std::to_string(colPos);
```
mariadb-LeonidFedorov: This columnname manipulation, and following should be encapsulated IMHO

mariadb-LeonidFedorov (changes requested):
Please address the comment on columnname manipulation, without proper helper (I do not insist on unittest) this looks like ad-hoc patching

mariadb-LeonidFedorov:
All the other code looks good to me!

mariadb-LeonidFedorov (approved):

dbcon/rbo/rbo_apply_rewrite_distinct.cpp
```diff
@@ -0,0 +1,181 @@
...
+    //This "order by" column does not belong to "group by" columns, so it should be an aggregated column
+    auto* aggCol = new execplan::AggregateColumn();
+    auto obcCloned = boost::shared_ptr<execplan::ReturnedColumn>(aggCol);
+
+    aggCol->asc(obc->asc());
+    aggCol->nullsFirst(obc->nullsFirst());
+    aggCol->aggOp(execplan::AggregateColumn::SELECT_SOME);
+    aggCol->aggParms().emplace_back(rcCloned);
+
+    //obcCloned->orderPos(colPos + orderByColPos);
```
drrtuy: Let's remove the commented line.

dbcon/rbo/rbo_apply_rewrite_distinct.cpp
```diff
@@ -0,0 +1,181 @@
...
+  }
+  else if (const auto* rcwc = dynamic_cast<execplan::WindowFunctionColumn*>(rc.get()))
+  {
+    rcCloned->timeZone(rcwc->timeZone());
+  }
+  rc->incRefCount();
+
+  auto colName = getSimpleColumnAlias(*rc, colPos);
+  rcCloned->columnName(colName);
+  rcCloned->alias("`" + tableAlias + "." + colName);
```
drrtuy: Do we need a closing backtick here?

drrtuy (approved):
small things to fix but in general untestedACK

PR #3668, opened 2025-07-24
feat/MCOL-5964 5861 5863  read-only node 

drrtuy:


cmapi/cmapi_server/node_manipulation.py
```diff
@@ -96,14 +97,23 @@ def add_node(
...
+                _add_WES(c_root, pm_num, node)
+            else:
+                logging.info('Node is read-only, skipping WES addition.')
+                _add_read_only_node(c_root, node)
+
             _add_DBRM_Worker(c_root, node)
             _add_Module_entries(c_root, node)
             _add_active_node(c_root, node)
             _add_node_to_ExeMgrs(c_root, node)
             if use_rebalance_dbroots:
```
drrtuy, line 120:
```suggestion
            if use_rebalance_dbroots && not read_only:
```

versioning/BRM/extentmap.cpp
```diff
@@ -1826,7 +1826,7 @@ void ExtentMap::save(const string& filename)
 
   if (!out)
   {
-    log_errno("ExtentMap::save(): can't open file " + filename);
+    log_errno("ExtentMap::save(): open");
```
drrtuy: @mariadb-SergeyZefirov We lost `filename` here.

versioning/BRM/extentmap.cpp
```diff
@@ -4541,7 +4541,7 @@ void ExtentMap::getDbRootHWMInfo(int OID, uint16_t pmNumber, EmDbRootHWMInfo_v&
            "There are no DBRoots for OID "
         << OID << " and PM " << pmNumber << endl;
     log(oss.str(), logging::LOG_TYPE_CRITICAL);
-    throw invalid_argument(oss.str());
+    // previously it was logic error with invalid_argument(oss.str()) exception.
```
drrtuy: Why do you remove this here @mariadb-SergeyZefirov ? Mb it is worth to uncomment and leave as it was.

cmapi/cmapi_server/node_manipulation.py
drrtuy: @mariadb-AlexanderPresniakov ^

mariadb-AlexanderPresniakov: @drrtuy if not read_only is on the next line

cmapi/cmapi_server/constants.py
```diff
@@ -60,6 +61,15 @@
     CMAPI_INSTALL_PATH, 'cmapi_server/SingleNode.xml'
 )
 
+class MCSProgs(Enum):
```
mariadb-AlexeyAntipovsky: nit: Why not to use `StrEnum` here? This will remove the need for using `.value` when the string value of an enum is required.

mariadb-AlanMologorsky: Pls, return back 'ExeMgr', otherwise it brakes backward compatibility.

cmapi/cmapi_server/constants.py
```diff
@@ -73,17 +83,16 @@ class ProgInfo(NamedTuple):
...
-    'controllernode': ProgInfo(11, 'mcs-controllernode', 'fg', True),
-    'PrimProc': ProgInfo(5, 'mcs-primproc', '', False, 1),
-    'ExeMgr': ProgInfo(9, 'mcs-exemgr', '', False, 1),
-    'WriteEngineServer': ProgInfo(7, 'mcs-writeengineserver', '', False, 3),
-    'DMLProc': ProgInfo(3, 'mcs-dmlproc', '', False),
-    'DDLProc': ProgInfo(1, 'mcs-ddlproc', '', False),
+    MCSProgs.STORAGE_MANAGER: ProgInfo(15, 'mcs-storagemanager', '', False, 1),
+    MCSProgs.WORKER_NODE: ProgInfo(13, 'mcs-workernode', 'DBRM_Worker{}', False, 1),
+    MCSProgs.CONTROLLER_NODE: ProgInfo(11, 'mcs-controllernode', 'fg', True),
+    MCSProgs.PRIM_PROC: ProgInfo(5, 'mcs-primproc', '', False, 1),
```
mariadb-AlanMologorsky, line 95: Here lost ExeMgr too.

cmapi/cmapi_server/handlers/cluster.py
```diff
@@ -158,20 +163,25 @@ def add_node(node: str, config: str = DEFAULT_MCS_CONF_PATH) -> dict:
         :rtype: dict
         """
         logger: logging.Logger = logging.getLogger('cmapi_server')
-        logger.debug(f'Cluster add node command called. Adding node {node}.')
+        logger.debug(
+            f'Cluster add node command called. Adding node {node} in '
+            f'{"read-only" if read_only else "read-write"} mode.'
```
mariadb-AlanMologorsky: OMG @drrtuy it's still confusing when reading.
We have to different states of nodes with same readonly naming.

cmapi/cmapi_server/managers/process.py
```diff
@@ -7,7 +7,8 @@
 import psutil
 
 from cmapi_server.exceptions import CMAPIBasicError
-from cmapi_server.constants import MCS_INSTALL_BIN, ALL_MCS_PROGS
+from cmapi_server.constants import MCS_INSTALL_BIN, ALL_MCS_PROGS, MCSProgs, ProgInfo
```
mariadb-AlanMologorsky: Pls use parentheses to move importing objects to the next line.

cmapi/cmapi_server/node_manipulation.py
```diff
@@ -529,6 +555,19 @@ def unassign_dbroot1(root):
         i += 1
 
 
+def _get_existing_db_roots(root: etree.Element) -> list[int]:
+    '''Get all the existing dbroot IDs from the config file'''
```
mariadb-AlanMologorsky, line 592: Pls use double quotes in docstring for style consistency.

cmapi/cmapi_server/node_manipulation.py
```diff
@@ -157,20 +167,32 @@ def remove_node(
 
         if len(active_nodes) > 1:
             pm_num = _remove_node_from_PMS(c_root, node)
-            _remove_WES(c_root, pm_num)
+
+            is_read_only = node in node_config.get_read_only_nodes(c_root)
```
mariadb-AlanMologorsky: At the same time we have new method `is_read_only` in `NodeConfig` class.
May be better to use it here?

cmapi/pyproject.toml
```diff
@@ -0,0 +1,22 @@
+[tool.ruff]
```
mariadb-AlanMologorsky, line 1: This file not related to PR should be added in different PR IMO.

mariadb-AlanMologorsky (changes requested):

cmapi/cmapi_server/constants.py
drrtuy: This is an explicit request made by me. With EM cmapi stuck trying to shut down EM. Moreover there is no CMAPI version that is compatible with MCS that has EM.

cmapi/cmapi_server/constants.py
drrtuy: ditto

cmapi/cmapi_server/handlers/cluster.py
drrtuy: Yes, we need to re-factor status output.

cmapi/pyproject.toml
drrtuy: +1

drrtuy:
untested ACK

oam/oamcpp/oamcache.h
```diff
@@ -42,9 +43,14 @@ class OamCache
...
-  dbRootPMMap_t getDBRootToConnectionMap();
-  PMDbrootsMap_t getPMToDbrootsMap();
+  int getClosestPM(int dbroot); // who can access dbroot's records for read requests - either owner or us.
+  int getClosestConnection(int dbroot); // connection index to owner's PM or ours PM - who can access dbRoot.
+  int getOwnerConnection(int dbroot); // connection index to owner's PM.
+  int getOwnerPM(int dbroot); // Owner's PM index.
+  std::vector<int> getPMDBRoots(int PM); // what DBRoots are owned by given PM.
+  std::vector<int> getAllDBRoots(); // get all DBRoots.
+  bool isAccessibleBy(int dbRoot, int pmId);
+  bool isOffline(int dbRoot); // not registered in map.
```
mariadb-AlexeyAntipovsky: Is it make sense to make this methods `const`?

oam/oamcpp/oamcache.cpp
```diff
@@ -237,4 +233,79 @@ string OamCache::getModuleName()
   return moduleName;
 }
 
+bool OamCache::isAccessibleBy(int dbRoot, int pmId)
+{
+  return (*dbRootPMMap)[dbRoot].count(pmId);
```
mariadb-AlexeyAntipovsky, line 238: (and below) Is it ok that elements can be implicitly created in the map if `dbRoot` is not there?

dbcon/joblist/primitivestep.h
```diff
@@ -1458,7 +1459,7 @@ class TupleBPS : public BatchPrimitive, public TupleDeliveryStep
   uint32_t blocksPerJob;
 
   /* Pseudo column filter processing.  Think about refactoring into a separate class. */
-  bool processPseudoColFilters(uint32_t extentIndex, boost::shared_ptr<std::map<int, int>> dbRootPMMap) const;
+  bool processPseudoColFilters(uint32_t extentIndex, oam::OamCache* oamCache) const;
```
mariadb-AlexeyAntipovsky, line 1462: nit: If we mark some `OamCache` methods as `const`, we could use `const oam::OamCache*` here...

dmlproc/dmlprocessor.cpp
```diff
@@ -1845,7 +1845,7 @@ void DMLProcessor::operator()()
 
 void RollbackTransactionProcessor::processBulkRollback(BRM::TableLockInfo lockInfo, BRM::DBRM* dbrm,
                                                        uint64_t uniqueId,
-                                                       OamCache::dbRootPMMap_t& dbRootPMMap,
+                                                       OamCache* oamcache,
```
mariadb-AlexeyAntipovsky, line 1850: nit: `const OamCache*`?

mariadb-AlanMologorsky (approved):
All needed fixes discussed, LGTM

cmapi/cmapi_server/constants.py
drrtuy: Good point for future improvement.

oam/oamcpp/oamcache.cpp
drrtuy: @mariadb-SergeyZefirov ^

mariadb-SergeyZefirov: It is OK.

First, this is refactored from old code that used the same logic, possibly inserting element if there was none.

Second, even if dbRoot is inserted, the nodes it is associated with is empty. It is unaccessible anyway.

Third, memory overhead is negligible.

dmlproc/dmlprocessor.cpp
mariadb-SergeyZefirov: Why?

We can reload, if we implement on-the-fly addition of nodes the OAMCache (or similar object) will not be const.

oam/oamcpp/oamcache.h
mariadb-SergeyZefirov: Would it require changes more than marking method(s) const?

drrtuy:
@mariadb-SergeyZefirov Build has been broken. Looks like namespace issue.

cmapi/cmapi_server/handlers/cluster.py
mariadb-AlanMologorsky: Not only status.
From a customer point of view we have command that sets mode to readonly and we have command that add readonly node. How should we explain that it's totally different things?

cmapi/pyproject.toml
mariadb-AlanMologorsky: Discussed, preffer to keep it in PR for linting PR code.

cmapi/cmapi_server/constants.py
mariadb-AlanMologorsky: @drrtuy @mariadb-AlexanderPresniakov let's make a task this idea will be lost instead.

oam/oamcpp/oamcache.cpp
mariadb-AlexeyAntipovsky: Now we have a single point of responsibility, so we can easily add verification that the requested dbroot exists. IMO, it's better to end the operation as early as possible with an error than to try to figure out where incorrect entries in the map came from.

I do not insist on making these changes as part of this pull request, but in general it would be good to have them.

dmlproc/dmlprocessor.cpp
mariadb-AlexeyAntipovsky: Do we really want to reload in the middle of an operation?

oam/oamcpp/oamcache.h
mariadb-AlexeyAntipovsky: It will require replacing `[]` with `.at()` in `const`-methods.

mariadb-AlexeyAntipovsky (approved):
I don't require that the proposed changes should be addressed in this PR, it looks good enough to me.

cmapi/cmapi_server/constants.py
mariadb-AlexanderPresniakov: Great idea, thanks, i totally forgot about (str, Enum) inheritance!

cmapi/cmapi_server/handlers/cluster.py
mariadb-AlexanderPresniakov: Turned out there exists a term "read replica" (in AWS RDS and Postgre), we can urgently replace "read-only" in this PR with "read-replica", before it's too late. What do you think about it? @drrtuy @mariadb-AlanMologorsky

cmapi/cmapi_server/node_manipulation.py
mariadb-AlexanderPresniakov: I checked it again and turns out this code is correct:
node_config.is_read_only() checks if this node (where this Cmapi instance is running) is read-only, and
node_config.get_read_only_nodes() lets you check if any node is read-only. In this case remove_node can receive any node to remove, so we have to continue using <node> in nc.get_read_only_nodes()

cmapi/cmapi_server/handlers/cluster.py
drrtuy: Agree. Change it plz as per yesterday's discussion @mariadb-AlexanderPresniakov.

mariadb-AlexanderPresniakov: @drrtuy done

mariadb-LeonidFedorov (approved):

PR #3670, opened 2025-07-25
MCOL-5974 chore: mcol-5480 test in bugfixes now works

mariadb-SergeyZefirov:
There are two parts of changeset, the most pertinent is in ha_mcs_ddl.cpp. The change there is to remove forsing TIMESTAMP columns to be NOT NULL if no constraints were given. This particular forcing made server to provide a NULL value bit that is not processed by our batch insert procedures.

E.g., for triple (TIMESTAMP,INT,STRING) server provided three isNull bits, let's say, 010, but we read only two first, but for INT and STRING. TIMESTAMP column was recorded as NOT NULL in Calpont System Catalog tables and thus is not processed and/or accounted for. Thus, for triple (2020-10-20 11:22:33, NULL 'asd') with 010 null bits, we read  (2020-10-20 11:22:33, 0 NULL because null mask is 001 for us. Zero is some random value we incorrectly read.

The second part of changes is due to ROB unable to compile under 11.4. I disabled the transformation.

I see a lot of MTR problems with stable-23.10 under 11.4-enterprise, though. My changes appear as not changing much in that regard.

### DIFF
--- a/dbcon/mysql/ha_mcs_ddl.cpp
+++ b/dbcon/mysql/ha_mcs_ddl.cpp
@@ -867,12 +867,14 @@ int ProcessDDLStatement(string& ddlStatement, string& schema, const string& /*ta
           return rc;
         }
 
+#if MYSQL_VERSION_ID < 110400
         // For TIMESTAMP, if no constraint is given, default to NOT NULL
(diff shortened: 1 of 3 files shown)


drrtuy (approved):
untested ACK

PR #3672, opened 2025-07-28
chore(ci): collect asan ubsan and libc++ build with mtr and regression status ignored

mariadb-LeonidFedorov:


build/prepare_test_container.sh
```diff
@@ -103,6 +105,15 @@ prepare_container() {
         execInnerDocker "$CONTAINER_NAME" '/setup-repo.sh'
     fi
 
+    # FIX THIS HACK
```
mariadb-AlekseiBukhalov, line 108: is it for later?

mariadb-AlekseiBukhalov (approved):

build/prepare_test_container.sh
mariadb-LeonidFedorov: I'll remove this comment, as we discussed this is normal way

drrtuy (approved):
untested ACK

PR #3675, opened 2025-07-29
chore(ci): run MTR as the Server Does

mariadb-LeonidFedorov:
This PR should fail on MTR with warnings on memory.
This is expected, there is the expected solution, that should be merged over


mariadb-LeonidFedorov:
@mariadb-AlekseiBukhalov please consider the second commit only

mariadb-AlekseiBukhalov (approved):

drrtuy (approved):
untested ACK

PR #3677, opened 2025-07-30
chore(tests): make local mtr runnner smarter

mariadb-LeonidFedorov:


tests/scripts/run_mtr.sh
```diff
@@ -0,0 +1,108 @@
...
+
+source $COLUMNSTORE_SOURCE_PATH/build/utils.sh
+
+
+optparse.define short=s long=suite desc="whole suite to run" variable=SUITE_NAME
+optparse.define short=t long=test_full_name desc="Testname with suite as like bugfixes.mcol-4899" variable=TEST_FULL_NAME default=""
+
+optparse.define short=f long=full desc="Run full MTR" variable=RUN_FULL default=false value=true
+optparse.define short=r long=record desc="Record the result" variable=RECORD default=false value=true
+optparse.define short=e long=extern desc="Run as --extern" variable=EXTERN default=true value=false
```
mariadb-AlekseiBukhalov: so if we pass --extern it actually will run without extern

maybe rename it to long=without-extern?

mariadb-AlekseiBukhalov (changes requested):

mariadb-AlekseiBukhalov (approved):

PR #3703, opened 2025-08-11
Fixed "raise None" that happened on signal interception

mariadb-AlexanderPresniakov:
Also fixed couple of (possibly) unbound variables


mariadb-AlanMologorsky (approved):
LGTM

PR #3706, opened 2025-08-14
MCOL-6121: aggregates for subqueries

mariadb-LeonidFedorov:


dbcon/mysql/columnstore.cnf
```diff
@@ -7,5 +7,5 @@ plugin-load-add=ha_columnstore.so
 
 collation_server = utf8_general_ci
 character_set_server = utf8
-
-loose-columnstore_use_import_for_batchinsert = ON
\ No newline at end of file
+columnstore_innodb_queries_use_mcs = on
```
drrtuy, line 10: We don't want this in the config by default. At least not uncommented.

mariadb-LeonidFedorov: I suggest to keep it until the feature branch is ready

dbcon/mysql/rulebased_optimizer.h
```diff
@@ -27,22 +27,28 @@
...
 
-class RBOptimizerContext {
-public:
+class RBOptimizerContext
+{
+ public:
   RBOptimizerContext() = delete;
-  RBOptimizerContext(cal_impl_if::gp_walk_info& walk_info) : gwi(walk_info) {}
-  // gwi lifetime should be longer than optimizer context. 
+  RBOptimizerContext(cal_impl_if::gp_walk_info& walk_info) : gwi(walk_info)
```
drrtuy, line 37: Do you have an idea how does you formater produce different repr for this file?

mariadb-LeonidFedorov: we have  .clang-format commited, so basically we should have the same formats, but I doubt your formatter is on, as far as in the `formatting` commit there are changes, that should be for any formatter
as example look at `makeUnionFromTable`, there was no ident in the function body

PR #3707, opened 2025-08-14
MCOL-6127: QA range filters misses NULLs

mariadb-LeonidFedorov:


drrtuy (approved):
untested ACK(after conflict resolution)

drrtuy (approved):
untested ACK

PR #3711, opened 2025-08-15
Update mcs_backup_manager to v3.17

mariadb-AllenHerrera:
3.16 added retry rsync logic for backups
3.17 improved dbrm backups with save_brm, write poll checks and write locks with optional flags to disable


drrtuy (approved):
untested ACK

PR #3717, opened 2025-08-20
MCOL-6147: on node removal, really remove it

mariadb-AlexanderPresniakov:
Even when asked `mcs cluster node remove --node <node>`, cmapi wasn't removing it. Instead it was moving it into DesiredNodes, and failover was adding it back, which is a very frustrating and non-intuitive behavior

### DIFF
--- a/cmapi/cmapi_server/node_manipulation.py
+++ b/cmapi/cmapi_server/node_manipulation.py
@@ -123,7 +123,7 @@ def add_node(
 def remove_node(
     node: str, input_config_filename: str = DEFAULT_MCS_CONF_PATH,
     output_config_filename: Optional[str] = None,
-    deactivate_only: bool = True,
+    deactivate_only: bool = False,
     use_rebalance_dbroots: bool = True, **kwargs
 ):
     """Remove node from a cluster.


mariadb-AlanMologorsky (changes requested):
This mechanism worked fine before, so I can't see the real cause of the issue. The fix in this PR addresses the symptom rather than the root cause. I highly recommend investigating further to identify the original reason for the issue.
JFYI, I’m not against actually removing the node, but it should be done with a clear understanding of the root cause.

mariadb-AlanMologorsky:
I found the issue.
https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/3397/files#diff-4703e5b8056f9b8c5e6561e4ef30b8015c5c48699ba37ceb45062dcb8e78e7c1L124

Please make request one liner, cause we don't need such an argument in API.

mariadb-AlexanderPresniakov:
> I found the issue. https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/3397/files#diff-4703e5b8056f9b8c5e6561e4ef30b8015c5c48699ba37ceb45062dcb8e78e7c1L124
> 
> Please make request one liner, cause we don't need such an argument in API.

@mariadb-AlanMologorsky Done!

PR #3720, opened 2025-08-21
fix(MCOL-6018): Prevent buffer overflow

mariadb-SergeyZefirov:
This patch prevents buffer overflow in date format function.


mariadb-SergeyZefirov:
It was redundant one.

PR #3722, opened 2025-08-21
chore(build): suppress warning from boost, dump boost build to file(/…

mariadb-LeonidFedorov:
…dev/null mostly) for columnstore maintaiers


drrtuy (approved):
untested ACK

PR #3728, opened 2025-08-22
MCOL-6127: fix(rbo): fix crash on simple filter for null handling

mariadb-LeonidFedorov:
and fixed loosing the last range, it was produced by (col1 >= Xm AND col1 <= Xn+1) OR col1 IS NULL structure, thaw was rewritten to (col1 >= Xn OR col1 IS NULL) AND (col1 <= Xn+1 OR col1 IS NULL)

he new structure avoids triggering line 3191 in 

doOR() from jlf_execplantojoblist.cpp
that sets okToCombine = false when it encounters AND operations inside OR trees. 

Then  falls back to 
doExpressionFilter()

Which 
creates minimal job steps (SubAdapterStep, AnnexStep) that don't push filters down to the foreign engine.

Now each OR only contains simple filters, which should allow proper job step creation instead of falling back to expression evaluation.

The logic is  equivalent:

Original: (A ∧ B) ∨ C
Transformed: (A ∨ C) ∧ (B ∨ C)

### DIFF
--- a/dbcon/execplan/simplefilter.cpp
+++ b/dbcon/execplan/simplefilter.cpp
@@ -294,11 +294,11 @@ const string SimpleFilter::toString(bool compact) const
   else
   {
     // For other operand types, show them compactly
-    output << endl << "LHS: " << fLhs->data();
+    output << endl << "LHS: " << fLhs->toString();
   }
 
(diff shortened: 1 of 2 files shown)


drrtuy (approved):
untested ACK

PR #3731, opened 2025-08-24
MCOL-4623: sec_to_time() does not support FLOAT/DOUBLE UNSIGNED

AestheticAkhmad:
Added support of sec_to_time() for all mentioned types:
CLOB
UFLOAT
UDOUBLE
TIME
LONGDOUBLE
TIMESTAMP - max TIME fallback
DATE - max TIME fallback
DATETIME - max TIME fallback


drrtuy (approved):
untestedACK

PR #3735, opened 2025-08-26
feat(joblist): MCOL-5756 MCOL-5222 ORDER BY on UNIONs in outer select

mariadb-AlexeyAntipovsky:


dbcon/mysql/ha_mcs_impl_if.h
```diff
@@ -434,13 +439,15 @@ const std::string bestTableName(const Item_field* ifp);
...
-                                              bool isRefItem = false);
+                                              bool isRefItem = false, bool isUnion = false);
 execplan::ReturnedColumn* buildFunctionColumn(Item_func* item, gp_walk_info& gwi, bool& nonSupport,
-                                              bool selectBetweenIn = false);
-execplan::ReturnedColumn* buildArithmeticColumn(Item_func* item, gp_walk_info& gwi, bool& nonSupport);
+                                              bool selectBetweenIn = false, bool isUnion = false);
+execplan::ReturnedColumn* buildArithmeticColumn(Item_func* item, gp_walk_info& gwi, bool& nonSupport,
+                                                bool isUnion);
 execplan::ConstantColumn* buildDecimalColumn(const Item* item, const std::string& str, gp_walk_info& gwi);
-execplan::SimpleColumn* buildSimpleColumn(Item_field* item, gp_walk_info& gwi);
```
mariadb-LeonidFedorov, line 442: Sorry, for it, but boolean params like this `bool isUnion` are Evil.
There is a good replacement with enum class to have this param named
and instead of  `buildSimpleColumn(ifp, gwi, false)` we can read 
`buildSimpleColumn(ifp, gwi, SimpleColumnOption::IS_UNION)`

dbcon/joblist/joblistfactory.cpp
```diff
@@ -1985,29 +1985,200 @@ void makeJobSteps(CalpontSelectExecutionPlan* csep, JobInfo& jobInfo, JobStepVec
...
-  CalpontSelectExecutionPlan::SelectList& selectVec = csep->unionVec();
+  CalpontSelectExecutionPlan::SelectList& unionVec = csep->unionVec();
   uint8_t distinctUnionNum = csep->distinctUnionNum();
-  RetColsVector unionRetCols = csep->returnedCols();
+  uint32_t unionRetColsCount = csep->returnedCols().size();
   JobStepVector unionFeeders;
 
-  for (CalpontSelectExecutionPlan::SelectList::iterator cit = selectVec.begin(); cit != selectVec.end();
-       cit++)
+  std::decay_t<decltype(csep->orderByCols())> expOrderByCols;
```
mariadb-LeonidFedorov: `std::remove_cvref_t` instead of `std::decay_t`?

dbcon/mysql/ha_mcs_execplan_walks.cpp
```diff
@@ -119,7 +119,7 @@ void gp_walk(const Item* item, void* arg)
       if (ifp)
       {
         // XXX: this looks awfuly wrong.
```
mariadb-LeonidFedorov: Does it look wrong still?

mariadb-LeonidFedorov:
I liked the code! I would ask to consider if it's possible to extract some helper functions to make it more readable

dbcon/mysql/ha_mcs_impl_if.h
drrtuy: Enum makes sense if there are multiple options IMHO. Any modern IDE will show you a variable name, so semantics is visible right away.

mariadb-LeonidFedorov: IDE will show, but this is an extra step. Readability of diff with function(a, b, false, false, true, true) is pathetic, and there is a very easy solution

dbcon/mysql/ha_mcs_execplan_walks.cpp
mariadb-AlexeyAntipovsky: IDK what Sergey means by this comment. Probably it is as awful as before... 8))

dbcon/execplan/calpontselectexecutionplan.h
```diff
@@ -55,6 +55,23 @@ enum RM_PARMS
   UMSMALLSIDEMEMORY,
 };
 
+// query type of select plan.
+// TODO: move it somewhere?
+#undef DELETE  // Windows defines this...
```
drrtuy: We are not going to see windows here AFAIK

dbcon/mysql/ha_mcs_impl_if.h
drrtuy: This is the example, no extra steps needed.
Снимок экрана от 2025-09-04 17-58-12

dbcon/joblist/tupleunion.cpp
```diff
@@ -124,20 +126,19 @@ namespace
 
   void normalizeIntToStringWithScale(const Row& in, Row* out, uint32_t i) 
   {
-    ostringstream os;
     double d = in.getIntField(i);
     d /= exp10(in.getScale(i));
-    os.precision(15);
-    os << d;
-    utils::NullString ns(os.str());
+    char buf[datatypes::INT128MAXPRECISION + 1];
```
drrtuy: Boilerplate that is cloned or almost cloned 4 times. IMHO should be replaced with a routine.

drrtuy (approved):
untestedACK except some boilerplate code

dbcon/joblist/tupleunion.cpp
mariadb-AlexeyAntipovsky: Good point, thanks!

PR #3736, opened 2025-08-26
Speed up columnstore CI packaging

mariadb-LeonidFedorov:
Debug packages for rocky were build extra 45 mins and never used. To have debug symbols, I just do not strip binaries on CI builds + boost now compiling only required components


mariadb-AlexanderPresniakov (approved):
Looks great to me

PR #3750, opened 2025-09-04
Tracing fixes

mariadb-AlexanderPresniakov:
Moved most of the trace logging into a separate JSONL log file json_trace.log. JSONL is a format where each line is a JSON experession. It was done because trace logging is mostly for the tools, so we don't clutter the main log anymore. The trace attributes like trace_id/span_id/parent_span_id are kept in the main log file.

Introduced a script that reads multiple jsonl trace log files (from different hosts of the cluster), merges them and generates a sequence diagram of their communication.

Trace logger (the one that logs minimal trace events like trace_begin\trace_end in the main log file) doesn't take trace_params (tid\sid\psid) in the end of the format line, because each its line already contains them.

Save small part of the response into the span, if the response was a JSON string. We'll use it later to show on SVG diagram (not shown right now, without interactive tooltips it would clutter the image)
Don't print span attrs into the main log (there are sentry headers, possibly part of the response and lots of other stuff).

An example of the diagram generated from the json logs:
imagen

### DIFF
--- a/cmapi/cmapi_server/cmapi_logger.conf
+++ b/cmapi/cmapi_server/cmapi_logger.conf
@@ -3,6 +3,9 @@
     "filters": {
         "add_ip_filter": {
             "()": "cmapi_server.logging_management.AddIpFilter"
+        },
+        "trace_params_filter": {
+            "()": "cmapi_server.logging_management.TraceParamsFilter"
         }
     },
     "formatters": {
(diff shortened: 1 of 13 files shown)


cmapi/cmapi_server/cmapi_logger.conf
```diff
@@ -52,6 +60,16 @@
             "maxBytes": 1024,
             "backupCount": 3,
             "encoding": "utf8"
+        },
+        "json_trace": {
+            "level": "DEBUG",
+            "class": "logging.handlers.RotatingFileHandler",
+            "formatter": "json_formatter",
+            "filename": "/var/log/mariadb/columnstore/json_trace.log",
```
mariadb-AlanMologorsky: NEAT
Better to name it with cmapi prefix or postfix.

mariadb-AlanMologorsky (approved):
LGTM

PR #3751, opened 2025-09-04
fix(PP,THJS): MCOL-6106 Fix race condition in TupleHashJoinStep

mariadb-AlexeyAntipovsky:


dbcon/joblist/diskjoinstep.cpp
```diff
@@ -444,6 +444,12 @@ void DiskJoinStep::joinFcn(const uint32_t threadID)
   smallNullMem[0].reset(new uint8_t[smallNullRow.getSize()]);
   smallNullRow.setData(rowgroup::Row::Pointer(smallNullMem[0].get()));
   smallNullRow.initToNull();
+  funcexp::FuncExpWrapper localFE2;
+  if (thjs->fe2)
+  {
+    // Make thread's own copy to prevent a possible race condition when evaluating expressions
+    localFE2 = *thjs->fe2;
```
drrtuy, line 451: Does this copy-asignment do deep-clone or it manipulates SRCP?

drrtuy (approved):
untestedACK

dbcon/joblist/diskjoinstep.cpp
mariadb-AlexeyAntipovsky: deep-clone

PR #3753, opened 2025-09-05
MCOL-5963: null handling in IF with datetime

mariadb-LeonidFedorov:
Fix:  IF function returninng a wrong value when the first expression is FALSE, the second expression is NULL, the third expression is a datetime, and the table engine is Columnstore.

### DIFF
--- a/utils/funcexp/func_if.cpp
+++ b/utils/funcexp/func_if.cpp
@@ -28,6 +28,7 @@ using namespace std;
 #include "functor_all.h"
 #include "functioncolumn.h"
 #include "predicateoperator.h"
+#include "constantcolumn.h"
 using namespace execplan;
 
 #include "rowgroup.h"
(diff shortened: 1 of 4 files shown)


utils/funcexp/func_if.cpp
```diff
@@ -124,9 +125,104 @@ CalpontSystemCatalog::ColType Func_if::operationType(FunctionParm& fp,
...
+  if (chooseTemporalIfOtherIsNull(rt2, rt1, isNullConst2, chosen))
+  {
+    resultType = chosen;
+    return chosen;
+  }
+
+  // Preserve temporal type when the opposite side is a constant (e.g. TRUE, numeric literal).
+  // This matches expected behavior in tests like IF(0=1, true, foo_datetime) and IF(0=1, 5, foo_datetime),
+  // where the result should be datetime, not coerced to DOUBLE.
+  auto isTemporal = [](CalpontSystemCatalog::ColDataType t) -> bool {
```
drrtuy: I would put this into mcs_datatype.h to be re-used.

utils/funcexp/func_if.cpp
```diff
@@ -124,9 +125,104 @@ CalpontSystemCatalog::ColType Func_if::operationType(FunctionParm& fp,
...
+    resultType = rt2;
+    return rt2;
+  }
+  if (isConst2 && isTemporal(rt1.colDataType))
+  {
+    resultType = rt1;
+    return rt1;
+  }
+
+  auto isStringType = [](CalpontSystemCatalog::ColDataType t) -> bool {
```
drrtuy: Check out datatypes::isCharType.

drrtuy (approved):
LGTM except checkers that must be defined in the common place to be re-used.

PR #3754, opened 2025-09-09
MCOL 5675 - fix error message for null-safe equal

mariadb-KristinaPavlova:


mariadb-LeonidFedorov (changes requested):
Please update relevant tests, that suffer from the behaviour changes, see the CI builds

drrtuy (approved):
untested ACK

PR #3756, opened 2025-09-09
fix(timeout): MCOL-5508 make node unresponse timeout configurable

mariadb-AlanMologorsky:
Read DBRM unresponsive timeout from Columnstore.xml DBRMUnresponsiveTimeout Tag instead of hardcoded 300 sec value.


mariadb-LeonidFedorov (approved):
LGTM

mariadb-AlexeyAntipovsky (approved):

PR #3765, opened 2025-09-11
Add forward/reverse DNS validation when adding nodes by hostname

mariadb-AlexanderPresniakov:
Sometimes there is no reverse DNS for the IP of the host, in this case IP isn't interchangeable with a host name.

In the case shown in [MCOL-6159](https://jira.mariadb.org/browse/MCOL-6159), CMAPI instance could not find matching hostname for its IP in the ActiveNodes and decided that it will not start MCS processes.

In this PR I added a check, that if node is added to the cluster by a host name, at least one of its IPs must resolve back to the passed hostname. If it's not true, we return an error and propose to add node by the IP.

Also the error i added was not visible in the API response, so i added couple of context managers to fix it and to minimize code duplication.

And i started converting logs from f-strings to %s syntax, because Sentry groups logs by the format strings, so if we use f-strings, it creates a separate error for each record. I love f-strings, but we have to do it.

### DIFF
--- a/cmapi/cmapi_server/node_manipulation.py
+++ b/cmapi/cmapi_server/node_manipulation.py
@@ -61,7 +61,6 @@ def switch_node_maintenance(
     node_config.write_config(config_root, filename=output_config_filename)
     # TODO: probably move publishing to cherrypy.engine failover channel here?
 
-
 def add_node(
     node: str, input_config_filename: str = DEFAULT_MCS_CONF_PATH,
     output_config_filename: Optional[str] = None,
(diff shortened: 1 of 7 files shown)


cmapi/cmapi_server/node_manipulation.py
```diff
@@ -96,6 +95,11 @@ def add_node(
     node_config = NodeConfig()
     c_root = node_config.get_current_config_root(input_config_filename)
 
+    # If a hostname (not IP) is provided, ensure fwd/rev DNS consistency.
+    # Skip validation for localhost aliases to preserve legacy single-node flows.
+    if not NetworkManager.is_ip(node) and node not in LOCALHOSTS:
```
mariadb-AlanMologorsky: What do you think about `not NetworkManager.is_only_loopback_hostname(node)`?

mariadb-AlanMologorsky (approved):
LGTM

cmapi/cmapi_server/node_manipulation.py
mariadb-AlexanderPresniakov: Thanks! I replaced the LOCALHOSTS check

PR #3766, opened 2025-09-11
chore(QA): Merge preparations

drrtuy:
Added MTR tests for TPC-H, docs and removed print outs.


mariadb-LeonidFedorov (approved):
Nice! Untested

PR #3768, opened 2025-09-12
feat(joblist): descend into SelectFilter even if sub is correlated.

drrtuy:
The solution for TPC-H Q20 with multiple partitioned tables involved in a query.


mariadb-AlexeyAntipovsky (approved):
lgtm

PR #3782, opened 2025-09-19
fix(shmem,brm,scripts): mcs-savebrm.py wrapper now cleans shmem locks before calling save_brm.

drrtuy:
This is to address atomicity of `systemctl stop mcs-workernode` operation. As of now CMAPI cleans shmem locks before it shuts WN daemon down that might result in DBRM inconsistencies.

### DIFF
--- a/oam/install_scripts/mcs-savebrm.py.in
+++ b/oam/install_scripts/mcs-savebrm.py.in
@@ -11,6 +11,7 @@ import struct
 import subprocess
 import sys
 import time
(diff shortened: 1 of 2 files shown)


mariadb-LeonidFedorov (approved):

mariadb-LeonidFedorov (changes requested):
please check the multinode tests

mariadb-LeonidFedorov (approved):

PR #3802, opened 2025-09-30
fix(cs_package_manager): RHEL install

mariadb-AlanMologorsky:
Do not download MariaDB-server for RHEL installation Fix yum repo by adding module_hotfixes = 1


mariadb-AllenHerrera (approved):
untested ack

PR #3806, opened 2025-10-02
 MCOL-5843: fix(extents): print corrent N/A for temporal extents

mariadb-LeonidFedorov:
we could see that
```
| Part#     Min                           Max                           Status
 0.0.3     32767-15-63                   32768-00-00                   Enabled |
```
here the the invalid values "32767-15-63" and "32768-00-00" come from `INT32_MAX` and `INT32_MIN`  being interpreted as dates.

### DIFF
--- a/datatypes/mcs_datatype.cpp
+++ b/datatypes/mcs_datatype.cpp
@@ -23,6 +23,7 @@
 #include <sstream>
 #include <cerrno>
 #include <cstring>
+#include <limits>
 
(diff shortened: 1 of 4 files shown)


mariadb-SergeyZefirov (commented):
It looks mergeworthy.

mariadb-KristinaPavlova (approved):

drrtuy (approved):
untested ACK

PR #3807, opened 2025-10-07
DNM feat(JSON data type): MCOL-6197: implement support for JSON type

mariadb-SergeyZefirov:
This patch does exactly this, it implements support for JSON in DDL.

Right now we use server's check for JSON validity on INSERT. We do not implement JSON validity check during updates, it is postponed for later work.

### DIFF
--- a/dbcon/ddlpackageproc/ddlpackageprocessor.cpp
+++ b/dbcon/ddlpackageproc/ddlpackageprocessor.cpp
@@ -195,9 +195,10 @@ execplan::CalpontSystemCatalog::ColDataType DDLPackageProcessor::convertDataType
 
     case ddlpackage::DDL_BLOB: colDataType = CalpontSystemCatalog::BLOB; break;
 
-    case ddlpackage::DDL_TEXT: colDataType = CalpontSystemCatalog::TEXT; break;
+    case ddlpackage::DDL_TEXT:
+    case ddlpackage::DDL_JSON: colDataType = CalpontSystemCatalog::TEXT; break;
 
-    default: throw runtime_error("Unsupported datatype!");
+    default: throw runtime_error("Unsupported DDL datatype!");
   }
 
   return colDataType;
@@ -227,6 +228,8 @@ std::string DDLPackageProcessor::buildTableConstraintName(const int oid, ddlpack
     case ddlpackage::DDL_CHECK: prefix = "ck_"; break;
 
     case ddlpackage::DDL_NOT_NULL: prefix = "nk_"; break;
+
+    case ddlpackage::DDL_VALIDATE_JSON: prefix = "jk_"; break;
 
     default: throw runtime_error("Unsupported constraint type!"); break;
   }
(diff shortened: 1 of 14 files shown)


dbcon/ddlpackage/ddlpkg.cpp
```diff
@@ -96,6 +96,9 @@ ColumnType::ColumnType(int type)
 
     default: fPrecision = 10; break;
   }
+  if (type == DDL_JSON)
```
mariadb-AlexeyAntipovsky: nit: Why the separate if-statement and not a switch-branch? Also if this is a stub for future code I'd like to see comment here...

dbcon/ddlpackageproc/ddlpackageprocessor.cpp
```diff
@@ -195,9 +195,10 @@ execplan::CalpontSystemCatalog::ColDataType DDLPackageProcessor::convertDataType
 
     case ddlpackage::DDL_BLOB: colDataType = CalpontSystemCatalog::BLOB; break;
 
-    case ddlpackage::DDL_TEXT: colDataType = CalpontSystemCatalog::TEXT; break;
+    case ddlpackage::DDL_TEXT:
+    case ddlpackage::DDL_JSON: colDataType = CalpontSystemCatalog::TEXT; break;
```
mariadb-AlexeyAntipovsky, line 199: nit: IMO, it's better to put code as a multi-line block in case of fallthrough cases.

writeengine/server/we_ddlcommon.h
```diff
@@ -178,7 +178,8 @@ inline int convertDataType(int dataType)
 
     case ddlpackage::DDL_BLOB: calpontDataType = execplan::CalpontSystemCatalog::BLOB; break;
 
-    case ddlpackage::DDL_TEXT: calpontDataType = execplan::CalpontSystemCatalog::TEXT; break;
+    case ddlpackage::DDL_TEXT:
+    case ddlpackage::DDL_JSON: calpontDataType = execplan::CalpontSystemCatalog::TEXT; break;
```
mariadb-AlexeyAntipovsky, line 182: nit: same here - IMO, it's better to put code as a multi-line block in case of fallthrough switch cases.

mariadb-AlexeyAntipovsky (approved):

writeengine/server/we_ddlcommon.h
mariadb-SergeyZefirov: I did what code around does, take a look. Otherwise I would have to change much more than single line of code.

dbcon/ddlpackageproc/ddlpackageprocessor.cpp
mariadb-SergeyZefirov: Again, I did what code around does. Otherwise I would have to change more than single line.

dbcon/ddlpackage/ddlpkg.cpp
mariadb-SergeyZefirov: Good catch, I deleted that. It was a stub for logic, but logic itself belongs to postprocessing function.

PR #3813, opened 2025-10-10
MCOL-6032: check that required directories are in the correct state

mariadb-AlexanderPresniakov:
When in shared FS storage mode, check that data1 and dbrm directories exist and have correct permissions when starting cmapi.


mariadb-AlanMologorsky:
LGTM

mariadb-AlanMologorsky (approved):

PR #3816, opened 2025-10-17
fix(cmapi): MCOL-6091 CMAPI gives DMLProc only 10 seconds for a greceful stop.

mariadb-AlanMologorsky:
Fix default timeout for stop node, stop dml proc, shutdown Controller, put_config handler etc. All places that could cause reducing dmlproc graceful stop timeout are fixed.

### DIFF
--- a/cmapi/cmapi_server/managers/process.py
+++ b/cmapi/cmapi_server/managers/process.py
@@ -3,11 +3,18 @@ from __future__ import annotations
 import logging
 import os.path
 import socket
+import time
 from time import sleep
 
 import psutil
 
-from cmapi_server.constants import ALL_MCS_PROGS, MCS_INSTALL_BIN, MCSProgs, ProgInfo
+from cmapi_server.constants import (
+    ALL_MCS_PROGS,
+    DMLPROC_SHUTDOWN_TIMEOUT,
+    MCS_INSTALL_BIN,
+    MCSProgs,
+    ProgInfo,
+)
 from cmapi_server.exceptions import CMAPIBasicError
 from cmapi_server.process_dispatchers.base import BaseDispatcher
 from cmapi_server.process_dispatchers.container import ContainerDispatcher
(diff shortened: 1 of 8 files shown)


drrtuy:
@mariadb-AlanMologorsky Could you replace literal value with a named constant or use one if it exists?

drrtuy (approved):
untestedACK

cmapi/cmapi_server/managers/process.py
```diff
@@ -238,32 +245,49 @@ def _wait_for_controllernode(cls) -> bool:
...
+                sleep_for = min(10, remaining)
+                logging.info(
+                    (
+                        f'Waiting for DMLProc to stop. Seconds left ~{remaining}. '
+                        f'Sleeping {sleep_for} seconds before next check.'
+                    )
+                )
+                next_log_in = LOG_INTERVAL
+
+            sleep_for = min(10, remaining)
```
mariadb-AlexanderPresniakov, line 282: We can put this line before if and do it unconditionally (as min is used)

cmapi/cmapi_server/managers/process.py
```diff
@@ -238,32 +245,49 @@ def _wait_for_controllernode(cls) -> bool:
...
+            sleep(sleep_for)
+            next_log_in -= sleep_for
+
+        logging.error(
+            'DMLProc did not stopped gracefully by DBRM command within '
```
mariadb-AlexanderPresniakov: *didn't stop

mariadb-AlexanderPresniakov (approved):
Two cosmetic fixes, otherwise LGTM

cmapi/cmapi_server/managers/process.py
mariadb-AlanMologorsky: Do You want just skip waiting messages in log?

cmapi/cmapi_server/managers/process.py
mariadb-AlanMologorsky: Done

PR #3824, opened 2025-10-23
boost::unrodered_flat_map for join

mariadb-LeonidFedorov:
No benchmarks yet. Any ideas like is it good idea to replace std::vector with small_vector or any other is appreciated


timofeyturenko:
performance comparison with stable-23.10 https://palantir.mariadb.net/report_comparisons/74/

dbcon/joblist/largehashjoin.h
```diff
@@ -107,13 +108,12 @@ class HashJoin
...
   // HjHasher<element_t> > hash_t;
-  typedef
-      typename std::tr1::unordered_multimap<typename element_t::second_type, typename element_t::first_type>
-          hash_t;
-  typedef typename std::tr1::unordered_multimap<typename element_t::second_type,
-                                                typename element_t::first_type>::iterator hashIter_t;
-  typedef typename std::tr1::unordered_multimap<typename element_t::second_type,
-                                                typename element_t::first_type>::value_type hashPair_t;
+  // Now using boost::unordered_flat_map with vector for better cache locality
+  typedef typename boost::unordered_flat_map<typename element_t::second_type,
```
mariadb-AlexeyAntipovsky, line 112: `using` instead of `typedef`?

utils/joiner/joiner.h
```diff
@@ -50,9 +51,8 @@ class Joiner
 {
  public:
   //		typedef std::tr1::unordered_multimap<uint64_t, uint64_t> hash_t;
-  typedef std::tr1::unordered_multimap<uint64_t, uint64_t, std::tr1::hash<uint64_t>, std::equal_to<uint64_t>,
-                                       utils::SimpleAllocator<std::pair<uint64_t const, uint64_t> > >
-      hash_t;
+  // Using boost::unordered_flat_map with vector for multimap behavior
+  typedef boost::unordered_flat_map<uint64_t, std::vector<uint64_t>> hash_t;
```
mariadb-AlexeyAntipovsky, line 55: The allocator type was lost.

utils/joiner/joiner.h
```diff
@@ -83,51 +93,50 @@ class Joiner
...
+    else if (!it->second.empty() && (it->second[0] & MSB))
       return true;
     else
     {
-      newMatches->push_back(joblist::ElementType(it->second | MSB, value));
-      range = h->equal_range(value);
-
-      for (; range.first != range.second; ++range.first)
-        range.first->second |= MSB;
+      // Add all values to newMatches and mark them
```
mariadb-AlexeyAntipovsky, line 134: Do we really need to add all values here? We added only one value before

utils/joiner/tuplejoiner.cpp
```diff
@@ -41,11 +41,10 @@ namespace joiner
 constexpr const size_t DEFAULT_BUCKET_COUNT = 10;
 
 template <typename HashTable>
-std::unique_ptr<HashTable> makeHashMap(size_t bucketCount, ResourceManager* resourceManager)
+std::unique_ptr<HashTable> makeHashMap(size_t /*bucketCount*/, ResourceManager* /*resourceManager*/)
 {
-  return std::unique_ptr<HashTable>(new HashTable(bucketCount, TupleJoiner::hasher(),
-                                                  typename HashTable::key_equal(),
-                                                  utils::STLPoolAllocator<typename HashTable::value_type>(resourceManager)));
+  // boost::unordered_flat_map doesn't need bucket_count or allocator in constructor
```
mariadb-AlexeyAntipovsky, line 46: `std::ordered_map` also doesn't need it, but we use custom allocator to track memory usage

mariadb-AlexeyAntipovsky (changes requested):

mariadb-AlexeyAntipovsky:
There is a LOT of code like
```c++
      auto it = destHashTbl->find(e.second);
      if (it != destHashTbl->end())
      {
        it->second.push_back(e.first);
      }
      else
      {
        std::vector<typename element_t::first_type> vec;
        vec.push_back(e.first);
        destHashTbl->emplace(e.second, std::move(vec));
      }
  ```
  It would be nice to isolate this boilerplate somewhere

utils/joiner/joiner.h
mariadb-SergeyZefirov: Look just below, the values were/are added there.

mariadb-SergeyZefirov (commented):
I consider PR as good modulo Alexei's request, where the most important is notice that allocator type has been lost.

PR #3827, opened 2025-10-23
fix(perf): add reserves for fill after creation containers pattern

mariadb-LeonidFedorov:
Added reserve() calls to prevent unnecessary  reallocations throughout the Columnstore codebase.

### DIFF
--- a/dbcon/ddlpackage/serialize.cpp
+++ b/dbcon/ddlpackage/serialize.cpp
@@ -125,6 +125,8 @@ int AlterTableStatement::unserialize(ByteStream& bytestream)
   // read alter action list
   quadbyte action_count;
   bytestream >> action_count;
+  // Reserve space for actions
+  fActions.reserve(action_count);
 
   for (unsigned int i = 0; i < action_count; i++)
   {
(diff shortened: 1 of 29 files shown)


timofeyturenko:
performance comparisin https://palantir.mariadb.net/report_comparisons/75/

dbcon/ddlpackage/serialize.cpp
```diff
@@ -125,6 +125,8 @@ int AlterTableStatement::unserialize(ByteStream& bytestream)
   // read alter action list
   quadbyte action_count;
   bytestream >> action_count;
+  // Reserve space for actions
```
mariadb-AlexeyAntipovsky, line 128: nit: I don't like comments like this, IMO they only pollute the code.

utils/rowgroup/rowstorage.cpp
```diff
@@ -1234,6 +1234,8 @@ class RowGroupStorage
   void getTmpFilePrefixes(std::vector<std::string>& prefixes) const
   {
     char buf[PATH_MAX];
+    // Reserve space for 2 prefixes
+    prefixes.reserve(prefixes.size() + 2);
```
mariadb-AlexeyAntipovsky, line 1238: IMO, this is meaningless reserve

mariadb-AlexeyAntipovsky (approved):

mariadb-SergeyZefirov (approved):
I agree with changes.

PR #3828, opened 2025-10-24
Configurable failover interval

mariadb-AlexanderPresniakov:
Added cmapi's own config section into Columnstore.xml to keep there parameters that must be shared by all cluster nodes.

Failover sampling interval is kept there.


mariadb-AlanMologorsky (approved):
LGTM 
But pls rebase branch from current stable.

PR #3831, opened 2025-10-27
Enable json input\output in endpoints by default

mariadb-AlexanderPresniakov:
And opt out of it by using disable_json decorator
There are not that many endpoints that don't consume/produce JSON, we decorate them with disable_json


mariadb-AlanMologorsky (approved):
LGTM but tests failed, can't approve at least before green tests.

mariadb-AlanMologorsky (approved):
Need to be rebased from stable-23.10

PR #3832, opened 2025-10-28
MCOL-6218 : null safe operator failing

mariadb-KristinaPavlova:
[MCOL-6218](https://jira.mariadb.org/browse/MCOL-6218)


drrtuy (approved):
untestedACK if the tests are passed.

PR #3844, opened 2025-11-06
Mcol 6210 count aux column

mariadb-SergeyZefirov:
THIS IS WORK IN PROGRESS AND OPENED TO GET SUGGESTIONS!!!


mariadb-SergeyZefirov:
The logic inside mcs_aux_count() implementation is taken out of next_row() function. But it does not work correctly - I've got an "internal error" message on execution.

If any of you, my dear colleagues, have a hint of what is incorrect there, let me know.

PR #3830, opened 2025-10-27
fix(correctness): replace volatiles with atomics

mariadb-LeonidFedorov:
## Benchmarks
https://palantir.mariadb.net/report_comparisons/76/

## Executive Summary

This refactoring addresses **critical thread-safety bugs** caused by misuse of the `volatile` keyword. The `volatile` keyword in C++ only prevents compiler optimizations—it does **not** provide atomicity or memory ordering guarantees required for thread-safe code. All instances where `volatile` was used for inter-thread communication have been replaced with proper `std::atomic` types.

### Impact Classification
- **🔴 CRITICAL:** 3 fixes preventing guaranteed production crashes/corruption
- **🔴 HIGH:** 3 fixes preventing likely production bugs  
- **🟡 MEDIUM:** 5 fixes preventing platform/timing-dependent bugs
- **🟢 LOW:** 4 fixes improving correctness in rare/non-critical paths

---

### 1. DistributedEngineComm::unackedWork Arrays 🔴🔴🔴
**Files:** `dbcon/joblist/distributedenginecomm.{h,cpp}`  
**Risk:** CRITICAL - Flow control corruption

**Problem:**
```cpp
// BEFORE: Arrays accessed by multiple threads without locks
boost::scoped_array<volatile uint32_t> unackedWork;
(void)atomicops::atomicInc(&unackedWork[i]);  // Race condition!
```

**Fix:**
```cpp
// AFTER: Proper atomic array
std::vector<std::atomic<uint32_t>> unackedWork;
unackedWork[i].fetch_add(1, std::memory_order_relaxed);
```

**Evidence of No Lock:**
- Code explicitly unlocks before access: `lk.unlock(); mqe->unackedWork[i].fetch_add(...)`
- Comment states: _"Other threads can be touching mqe->unackedWork at the same time"_

**Bug Symptoms:**
- Incorrect flow control → query deadlocks
- PM overwhelmed with data
- Message loss
- Occurs on **every multi-PM query**

---

### 2. DistributedEngineComm::pmCount 🔴🔴🔴
**Files:** `dbcon/joblist/distributedenginecomm.{h,cpp}`  
**Risk:** CRITICAL - Crashes during topology changes

**Problem:**
```cpp
// BEFORE: Manual memory barrier + volatile
volatile uint32_t pmCount;
atomicops::atomicMb();  // Manual barrier indicates known race!
pmCount = newPmCount;
```

**Fix:**
```cpp
// AFTER: Atomic with proper memory ordering
[...]

PR #3846, opened 2025-11-07
fix(correctness): replace volatiles with atomics [develop-23.02]

mariadb-LeonidFedorov:
backport of https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/3830

### DIFF
--- a/dbcon/joblist/distributedenginecomm.cpp
+++ b/dbcon/joblist/distributedenginecomm.cpp
@@ -206,7 +206,7 @@ void DistributedEngineComm::reset()
 }
 
 DistributedEngineComm::DistributedEngineComm(ResourceManager* rm, bool isExeMgr)
- : fRm(rm), pmCount(0), fIsExeMgr(isExeMgr)
+ : fRm(rm), pmCount{0}, fIsExeMgr(isExeMgr)
 {
   if (fIsExeMgr)
   {
@@ -356,9 +356,8 @@ int32_t DistributedEngineComm::Setup()
 
   fWlock.swap(newLocks);
   fPmConnections.swap(newClients);
-  // memory barrier to prevent the pmCount assignment migrating upward
-  atomicops::atomicMb();
-  pmCount = newPmCount;
+  // atomic store with release semantics ensures all previous writes are visible
+  pmCount.store(newPmCount, std::memory_order_release);
 
   newLocks.clear();
   newClients.clear();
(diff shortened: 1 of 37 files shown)


mariadb-AlexeyAntipovsky (approved):

dbcon/joblist/distributedenginecomm.cpp
```diff
@@ -448,9 +447,7 @@ void DistributedEngineComm::addQueue(uint32_t key, bool sendACKs)
   condition* cond = new condition();
   uint32_t firstPMInterleavedConnectionId =
       key % (fPmConnections.size() / pmCount) * fDECConnectionsPerQuery * pmCount % fPmConnections.size();
-  boost::shared_ptr<MQE> mqe(new MQE(pmCount, firstPMInterleavedConnectionId, flowControlEnableBytesThresh));
-
-  mqe->queue = StepMsgQueue(lock, cond);
```
mariadb-SergeyZefirov, line 453: This is not present in new code. Looks like an omission.

utils/libmarias3/libmarias3
mariadb-SergeyZefirov, line 1: Should libmarias3 be updated?

writeengine/shared/we_brm.cpp
```diff
@@ -297,22 +297,19 @@ int BRMWrapper::getFboOffset(const uint64_t lbid, int& oid, uint16_t& dbRoot, ui
 //------------------------------------------------------------------------------
 BRMWrapper* BRMWrapper::getInstance()
 {
-  if (m_instance == 0)
+  BRMWrapper* tmp = m_instance.load(std::memory_order_acquire);
```
mariadb-SergeyZefirov, line 300: I think that this should be under lock, otherwise I see a race condition despite use of atomics.

mariadb-SergeyZefirov (approved):
The changes are straightforward, structurally so.

My biggest concern is omission of ``queue`` field initialization in distributedenginecomm module.

A second one is libmarias4  changes.

I approve this PR nonetheless, but please comment here on why ``queue`` is not initialized and why libmarias4 is updated.

dbcon/joblist/distributedenginecomm.cpp
mariadb-LeonidFedorov: I was moved to ctor, line 1216

utils/libmarias3/libmarias3
mariadb-LeonidFedorov: That's wrong, thank you

writeengine/shared/we_brm.cpp
mariadb-LeonidFedorov: It's usual double checked pattern, or I'm missing something

mariadb-SergeyZefirov: I see, I was confused because locking is done after tmp load and nullptr check.

drrtuy (approved):
untestedACK
Too many changes together, so hard to evaluate the consequences IMHO. This doesn't blow up on basic functionality so green light from me.

PR #3852, opened 2025-11-10
Added fix for div and mod functions in columnstore - MCOL-4601

mayankadhikary-lab:
Description: The MOD and DIV operators in MariaDB ColumnStore were not behaving consistently with InnoDB (the standard MariaDB storage engine), leading to incorrect results, particularly with floating-point or DECIMAL types.

1. MOD Function (func_mod.cpp)
Analysis and Root Cause
The MOD operator was failing because the implementation in Func_mod::getDecimalVal only considered the first parameter (parm[0]) for the precision calculation and failed to use the second parameter (parm[1]) at all. This resulted in wrong precision and consequently, an incorrect remainder calculation.
Fix Summary
The fix ensures that:
1.	The second parameter (parm[1]) is correctly retrieved and used.
2.	The precision (scale) is determined by taking the maximum precision of both input parameters (d.scale and d1.scale) to guarantee the most accurate result.
3.	The calculation is performed using high-precision integer types (int128_t) to avoid floating-point loss.

Verification result:

image

2. DIV Function (func_div.cpp)
Analysis and Root Cause
The primary issue in the Func_div::getIntVal function was the reliance on the double data type. The function used getDoubleVal to retrieve operands, forcing any high-precision DECIMAL data into the lossy binary floating-point format. This resulted in an inexact quotient that, when cast to int64_t, produced an incorrect integer result due to the inherent loss of precision.
Fix Summary
To ensure mathematically exact results for integer division, the fix replaces all reliance on floating-point arithmetic with the native high-precision decimal type:
1.	Dual Decimal Retrieval: Operands are retrieved as IDB_Decimal objects (e.g., using a new getDecimalVal accessor), entirely bypassing the lossy double conversion.
2.	Scale Normalization: The scales of the dividend (dec1) and divisor (dec2) are normalized to a common value to ensure both are represented with the same precision before division.
3.	High-Precision Division: The division is performed using the IDB_Decimal arithmetic methods.
5.	Final Cast: The resulting high-precision decimal quotient is then cast directly to int64_t. Because the division was performed precisely, this final conversion correctly yields the intended integer result

Verification result:

image

### DIFF
--- a/utils/funcexp/func_div.cpp
+++ b/utils/funcexp/func_div.cpp
@@ -47,16 +47,23 @@ CalpontSystemCatalog::ColType Func_div::operationType(FunctionParm& /*fp*/,
 int64_t Func_div::getIntVal(rowgroup::Row& row, FunctionParm& parm, bool& isNull,
                             CalpontSystemCatalog::ColType& /*op_ct*/)
 {
-  double val1 = parm[0]->data()->getDoubleVal(row, isNull);
-  double val2 = parm[1]->data()->getDoubleVal(row, isNull);
+  IDB_Decimal val1 = parm[0]->data()->getDecimalVal(row, isNull);
+  IDB_Decimal val2 = parm[1]->data()->getDecimalVal(row, isNull);
 
-  if (val2 == 0 || val2 == NAN)
+  if (val2.value == 0)
   {
     isNull = true;
     return 0;
   }
   // MCOL-179 InnoDB doesn't round or convert to int before dividing.
-  return static_cast<int64_t>(val1 / val2);
+  int scale = std::max(val1.scale, val2.scale);
+
+  int128_t new_val1 = val1.value * pow(10, scale - val1.scale);
+  int128_t new_val2 = val2.value * pow(10, scale - val2.scale);
+
+  int128_t result = new_val1 / new_val2;
+  return static_cast<int64_t>(result);
+      
 
 #if 0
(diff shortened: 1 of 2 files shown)


mariadb-LeonidFedorov:
Could you please add the tests covering your fix columnstore/basic/t/mcs227_div_function.test
To run the test and to update the reference output columnstore/basic/r/mcs227_div_function.result, you can use
--record of following script

➜  sudo tests/scripts/run_mtr.sh --help
usage: tests/scripts/run_mtr.sh [OPTIONS]
OPTIONS:
        
	-s --suite:                  whole suite to run
	-t --test_full_name:         Testname with suite as like bugfixes.mcol-4899
	-f --full:                   Run full MTR [default:false]
	-r --record:                 Record the result [default:false]
	-e --no-extern:              Run with --extern [default:false]
	-c --color:                  run with unbufer to colorize output [default:false]
        -? --help  :  usage

mayankadhikary-lab:
Adding the detailed document, containing the description, analysis, code changes and verification result with test cases for both mod and div functions. 
[Issue - 4601.docx](https://github.com/user-attachments/files/23469948/Issue.-.4601.docx)

drrtuy:
> Adding the detailed document, containing the description, analysis, code changes and verification result with test cases for both mod and div functions. [Issue - 4601.docx](https://github.com/user-attachments/files/23469948/Issue.-.4601.docx)

Could you plz implement the test cases in a form of MTR tests?

PR #3859, opened 2025-11-12
fix(plugin): MCOL-6088 This ensures that `SELECT col AS alias ... GRO…

mariadb-LeonidFedorov:
# Fix for GROUP BY Ambiguous Column Reference Bug

## Problem Description

When a GROUP BY clause contains a column name that is ambiguous (i.e., it matches both a table column and a SELECT list alias pointing to a different column), ColumnStore was throwing an error:

```
ERROR 1178 (42000): The storage engine for the table doesn't support MCS-2016: Non supported item 'cat_id' on the GROUP BY list.
```

The expected behavior is to handle the ambiguity gracefully, similar to how standard MariaDB would handle it (with a warning if appropriate).

## Example Query

```sql
CREATE TABLE test2 (
  id INT UNSIGNED NOT NULL,
  GBP DOUBLE UNSIGNED NOT NULL,
  cat_id INT UNSIGNED NOT NULL,
  cat_id_level_0 INT UNSIGNED NOT NULL
) ENGINE=Columnstore;

-- This query has an ambiguous GROUP BY reference
-- 'cat_id' could refer to:
--   1. The table column 'cat_id'
--   2. The SELECT alias 'cat_id' (which is cat_id_level_0 AS cat_id)

SELECT cat_id_level_0 AS entity_id, SUM(GBP) AS spend, cat_id_level_0 AS cat_id
FROM test2
GROUP BY cat_id;
```

## Root Cause

In the `processGroupBy()` function in `ha_mcs_execplan.cpp`, when processing a `FIELD_ITEM` in the GROUP BY clause:

1. `buildSimpleColumn()` was called to create a column reference
2. If this failed (returned NULL) due to ambiguity or other issues, the code immediately set `nonSupportItem`, triggering the error
3. The code didn't check if there was a matching alias in the SELECT list that could resolve the reference

## Solution

Modified `processGroupBy()` function in `ha_mcs_execplan.cpp` to handle ambiguous column references more gracefully:

1. When `buildSimpleColumn()` succeeds, use the table column and check SELECT list by `sameColumn()` match
2. If table column found in SELECT list, also check for alias match as a fallback
3. When `buildSimpleColumn()` returns NULL or non-SimpleColumn, match by alias name in SELECT list


dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -5724,6 +5724,61 @@ int processHaving(SELECT_LEX& select_lex, gp_walk_info& gwi, SCSEP& csep,
...
+    if (agg_check->aggParms().empty())
+    {
+      // Aggregate with no parameters (like COUNT(*)) - can't use for GROUP BY
+      return nullptr;
+    }
+
+    ReturnedColumn* inner_col = agg_check->aggParms()[0].get();
+    AggregateColumn* inner_agg = dynamic_cast<AggregateColumn*>(inner_col);
+
+    if (inner_agg)
```
mariadb-SergeyZefirov: Can it be recursive? E.g., ``SUM(COUNT(group_by_col))``.

mariadb-SergeyZefirov (approved):
I agree with changes as the solve exactly what need to be solved and thus can be merged.

I have a question about nested aggregate unwrapping, but I do not see it as super important. If we differ in behavior from the server for single aggregate, we also may differ for nested aggregates.

mysql-test/columnstore/basic/t/mcol-4525.test
```diff
@@ -83,11 +83,10 @@ INSERT INTO t1 VALUES(10,'bye',10);
 INSERT INTO t1 VALUES(10,'sam',10);
 INSERT INTO t1 VALUES(10,'bob',10);
 SET columnstore_select_handler=ON;
---error 1178
-SELECT SUM(col1) AS col2 FROM t1 GROUP BY col2;
+SELECT SUM(col1) AS col2 FROM t1 GROUP BY col2 ORDER BY col2;
 SELECT col1 c FROM t1 ORDER BY AVG(col1);
 SET columnstore_select_handler=AUTO;
-SELECT SUM(col1) AS col2 FROM t1 GROUP BY col2;
+SELECT SUM(col1) AS col2 FROM t1 GROUP BY col2 ORDER BY col2;
```
drrtuy, line 89: I suggest to use sorted_result instead of OB if OB is not needed.

mysql-test/columnstore/basic/t/mcol-4525.test
```diff
@@ -83,11 +83,10 @@ INSERT INTO t1 VALUES(10,'bye',10);
 INSERT INTO t1 VALUES(10,'sam',10);
 INSERT INTO t1 VALUES(10,'bob',10);
 SET columnstore_select_handler=ON;
---error 1178
-SELECT SUM(col1) AS col2 FROM t1 GROUP BY col2;
+SELECT SUM(col1) AS col2 FROM t1 GROUP BY col2 ORDER BY col2;
```
drrtuy, line 86: I suggest to use sorted_result instead of OB if OB is not needed.

PR #3861, opened 2025-11-14
fix(PP,THJS): MCOL-6083 fix disk joins freezes

mariadb-AlexeyAntipovsky:


dbcon/joblist/resourcemanager.cpp
```diff
@@ -398,6 +398,14 @@ bool ResourceManager::getMemory(int64_t amount, bool patience)
   return ret1;
 }
 
+bool ResourceManager::getMemoryForce(int64_t amount, boost::shared_ptr<int64_t>& sessionLimit)
+{
+  bool ret1 = (atomicops::atomicSubRef(totalUmMemLimit, amount) >= 0);
+  bool ret2 = sessionLimit ? (atomicops::atomicSub(sessionLimit.get(), amount) >= 0) : ret1;
+
+  return (ret1 && ret2);
```
mariadb-LeonidFedorov, line 406: is it correct, that `getMemoryForce` doesn't restore counters on failure?

mariadb-AlexeyAntipovsky: That's the point. We always want to process incoming RowGroups regardless of whether there is enough memory (so as not to lose data). In order to avoid being killed by the OOM-killer (for example, in a parallel query), I want this memory to be accounted (it will be returned later, either after destroying the TupleJoiner or converting it to DiskJoin).

mariadb-LeonidFedorov (approved):

utils/common/countingallocator.h
```diff
@@ -43,6 +43,12 @@ const constexpr int64_t MemoryLimitLowerBound = 500 * 1024 * 1024;
 // Higher values demonstrate slower response to memory limit violations.
 const constexpr int64_t CheckPointStepSize = 100 * 1024;
 
+enum class AllocMode
+{
+  STRICT,  ///
```
mariadb-SergeyZefirov: Should here be comment? ;)

utils/rowgroup/rowaggregation.cpp
```diff
@@ -5144,7 +5144,7 @@ void GroupConcat::deserialize(messageqcpp::ByteStream& bs)
     bs >> f;
     uint8_t s;
     bs >> s;
-    fOrderCond.emplace_back(f, static_cast<bool>(s));
+    fOrderCols.emplace_back(f, static_cast<bool>(s));
```
mariadb-SergeyZefirov, line 5147: Is it correct? Why there was "condition" and now "columns"?

mariadb-SergeyZefirov (commented):
I generally agree with changes.

The question about comment is non-essential to functionality, but can help in future.

The question about changes in rowaggregation.cpp ("condition vs columns") should be addressed. though.. It appears as something out of scope for this PR.

utils/common/countingallocator.h
mariadb-AlexeyAntipovsky: Exactly! I forgot to push... Thanks!

utils/rowgroup/rowaggregation.cpp
mariadb-AlexeyAntipovsky: There was a typo here — as you can see, a couple of lines above we have reserved space for fOrderCols, and fOrderConds deserialized below.

PR #3871, opened 2025-11-28
Fix bug with the WriteEngineServer service disappearing from the services list.

mariadb-AlanMologorsky:
In rare cases, if the Columnstore.xml configuration is changed without restarting the CMAPI service, the WriteEngineServer service may disappear from the mcs_progs services list on the primary node.


mariadb-AlexanderPresniakov (approved):
LGTM

drrtuy (approved):
untestedACK

PR #3872, opened 2025-11-28
fix(query-acceleator): Fixes SUGSEGV in MCOL-6238

mariadb-SergeyZefirov:
Now all SimpleColumns in ConstantFilter are updated during Query Accelerator processing.


drrtuy (approved):
untestedACK but I would like to see some MTR tests enabled if this patch fixes any failures in TPC-H.

mariadb-LeonidFedorov (approved):

PR #3879, opened 2025-12-04
Relax permission checks in invariant checks (check only user's perms)

mariadb-AlexanderPresniakov:


mariadb-AlanMologorsky (approved):
LGTM

PR #3884, opened 2025-12-12
MCOL-6251: query accelerator subqueries

mariadb-SergeyZefirov:


dbcon/joblist/subquerytransformer.cpp
```diff
@@ -120,7 +120,7 @@ SJSTEP& SubQueryTransformer::makeSubQueryStep(execplan::CalpontSelectExecutionPl
 
 
   // Update v-table's alias.
-  fVtable.name("$sub");
+  fVtable.name("$sub"+fVtable.name()+"/"+fVtable.alias());
```
drrtuy: Are you sure about this? This might implicitly break lots of invariants in  compiler.

dbcon/execplan/simplecolumn.cpp
```diff
@@ -358,7 +316,7 @@ const string SimpleColumn::toString(bool compact) const
   output << "Column: " << data();
   datatypes::Charset cs(fResultType.charsetNumber);
   output << endl
-         << "Info: " << schemaName() << "." << tableName() << "." << columnName()
+         << "Info: " << schemaName() << "." << tableName() << "." << columnName() << ", table alias " << tableAlias()
```
drrtuy: Can you represent ", table alias " in a more compressed way, e.g. alias after tableName() in parenthesis? 
```suggestion
         << "Info: " << schemaName() << "." << tableName() 
         << "(" << tableAlias << ")" 
          << "."  << columnName() << ", table alias " << tableAlias()
```

drrtuy: When the code prints out a column(and this happens a lot when debugging) for every column there is an extra ` table alias ` line added. This is a cognitive waste and might make formatting even less readable b/c the column description line becomes even wider than originally.

dbcon/joblist/subquerytransformer.cpp
mariadb-SergeyZefirov: Yes, I am pretty sure.

The only places where "$sub" is checked are handful of places in joblist code and there comparisons are checked against substring.

```
    else if (ta.compare(0, 5, "$sub_") && ta.compare(0, 4, "$exp"))  // compare != 0
      name = ss + ta + "." + name;

...
   if (tn.compare(0, 4, "$sub") && tn.compare(0, 4, "$exp"))
    {
      if (!useAlias)
        name = ss + name;
    }
    else if (tn.compare(0, 4, "$sub") == 0)
    {
```

dbcon/execplan/simplecolumn.cpp
```diff
@@ -123,21 +123,21 @@ void getSimpleColsExtended(execplan::ParseTree* n, void* obj)
...
+  else if (f)
   {
-    selectFilter->setSimpleColumnListExtended();
-    list->insert(list->end(), selectFilter->simpleColumnListExtended().begin(),
-                 selectFilter->simpleColumnListExtended().end());
+    f->setSimpleColumnListExtended();
+    list->insert(list->end(), f->simpleColumnListExtended().begin(), f->simpleColumnListExtended().end());
   }
-  else if (cf)
+  else if (lo) // XXX: should it be default case?
```
mariadb-SergeyZefirov, line 131: @drrtuy This is the important variant I was talking about.

mariadb-LeonidFedorov (approved):

PR #3899, opened 2026-01-15
fix(cmapi): MCOL-6285 Invariant checks follows symlinks.

mariadb-AlanMologorsky:
Now invariant checks look at given directory and check if it is symlink. If yes then real path will be used.
Added option to turn off invariant checks
```
[application]
invariant_checks = false
```

### DIFF
--- a/cmapi/cmapi_server/invariant_checks.py
+++ b/cmapi/cmapi_server/invariant_checks.py
@@ -1,27 +1,78 @@
 import logging
 import os
-from typing import Optional, Tuple
+from typing import List, Optional, Tuple
 
 from mr_kot import Runner, Status, any_of, check, check_all, fact, parametrize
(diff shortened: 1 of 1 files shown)


cmapi/cmapi_server/invariant_checks.py
```diff
@@ -1,27 +1,78 @@
...
+    result: List[str] = []
+    current: str = base_path
+
+    # Resolve base path if it's a symlink
+    if os.path.islink(current):
+        current = os.path.realpath(current)
+        logger.debug('Base path %s is a symlink, resolved to %s', base_path, current)
+
+    result.append(current)
+
```
mariadb-AlexanderPresniakov, line 35: IMO it can be made a single loop (instead of separate processing for the base and for the subdirs), but it isn't crucial

mariadb-AlexanderPresniakov (approved):
LGTM

PR #3900, opened 2026-01-15
[DNM] MCOL-6285 Fix/invariant checks follows symlinks arm64

mariadb-AlanMologorsky:


mariadb-AlanMologorsky:
Needed only for arm builds.

PR #3906, opened 2026-01-19
MCOL-5983: Add COMPARE_BITMASK primitive filter for bitwise AND pattern optimization

mariadb-LeonidFedorov:
This PR adds support for pushing down bitwise AND filter patterns (col & mask) = mask as primitive filters, enabling significant performance improvements for Bloom filter-based queries.

Background
Bloom filters are commonly used in specialized databases (e.g., chemical/molecular databases) to accelerate substring searches. A typical query pattern is:

```
SELECT * FROM molecules 
WHERE smiles LIKE '%C(=O)C(O)%' 
  AND (signature & @bloom_mask) = @bloom_mask;
```
Previously, the bitwise AND expression was evaluated in FE1 (Function Expression Group 1) after all primitive filters, negating the benefit of the Bloom filter pre-filtering.

Pattern Recognition: `doSimpleFilter()` detects `FunctionColumn("&") = ConstantColumn` where both mask arguments match 
- Filter Creation: Creates a `pColStep` with `COMPARE_BITMASK` operator instead of falling back to `ExpressionFilter`
- SIMD Execution: The primitive processor uses `bwAnd() + cmpEq()` SIMD intrinsics for vectorized filtering
- Filter Reordering: `walkTree()` uses cost-based heuristics to execute cheap bitmask filters before expensive LIKE filters

### DIFF
--- a/dbcon/joblist/jlf_execplantojoblist.cpp
+++ b/dbcon/joblist/jlf_execplantojoblist.cpp
@@ -29,6 +29,7 @@
 #include <map>
 #include <climits>
 #include <cmath>
+#include "mcs_datatype.h"
 using namespace std;
 
 #include <boost/shared_ptr.hpp>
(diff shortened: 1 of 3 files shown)


mariadb-SergeyZefirov (approved):

mariadb-AlexeyAntipovsky (approved):
LGTM!

dbcon/joblist/jlf_execplantojoblist.cpp
```diff
@@ -1913,6 +1913,91 @@ const JobStepVector doSimpleFilter(SimpleFilter* sf, JobInfo& jobInfo)
     if (jsv.empty())
       throw runtime_error("Unhandled SimpleFilter");
   }
+  // Handle bitmask pattern: (col & mask) = mask
+  // This allows bitwise AND filters to be pushed down as primitive filters
+  else if (lhsType == FUNCTIONCOLUMN && opeq == *sop)
```
drrtuy: WTB an equvalent `mask & col` expression?

dbcon/joblist/jlf_execplantojoblist.cpp
```diff
@@ -1913,6 +1913,91 @@ const JobStepVector doSimpleFilter(SimpleFilter* sf, JobInfo& jobInfo)
...
+          CalpontSystemCatalog::ColType ct = sc->colType();
+
+          if (!sc->schemaName().empty() && sc->isColumnStore())
+          {
+            ct = jobInfo.csc->colType(sc->oid());
+            ct.charsetNumber = sc->colType().charsetNumber;
+          }
+
+          // Only for integer types
+          if (ct.colDataType == CalpontSystemCatalog::BIGINT ||
```
drrtuy: [There](https://github.com/mariadb-corporation/mariadb-columnstore-engine/blob/f3d554ed48f208582d5220e4ab702feb749a3fa9/datatypes/mcs_datatype.h#L413) is a routine to filter integer DTs.

drrtuy (changes requested):
I like the idea.
Plz comment on the question regarding equivalent expressions `mask & col` and use an existing routine to filter int types.

dbcon/joblist/jlf_execplantojoblist.cpp
```diff
@@ -3344,11 +3429,48 @@ namespace joblist
...
+        if (sf)
+        {
+          const SOP& op = sf->op();
+          // LIKE is expensive
+          if (op && (op->op() == OP_LIKE || op->op() == OP_NOTLIKE))
+            return 1000;
+          // Check if LHS is a bitmask function (col & mask) = mask
+          const FunctionColumn* fc = dynamic_cast<const FunctionColumn*>(sf->lhs());
+          if (fc && fc->functionName() == "&")
+            return 10;  // Bitmask is cheap
```
drrtuy: This prioritized op will surely strike back in big expressions. You are tuning for this particular use case. I suggest to check for you explicit case here, namely `col & const` to avoid unexpected side effects for other expressions that contains bitwise op. 
This statement is true only if this affects priority for FuncExp filtering.

mariadb-AlekseiBukhalov (approved):

dbcon/joblist/jlf_execplantojoblist.cpp
mariadb-LeonidFedorov: sure!

dbcon/joblist/jlf_execplantojoblist.cpp
mariadb-LeonidFedorov: Good notice!

dbcon/joblist/jlf_execplantojoblist.cpp
mariadb-LeonidFedorov: Agreed

primitives/linux-port/column.cpp
```diff
@@ -207,6 +214,13 @@ inline bool colCompare_(const T& val1, const T& val2, uint8_t COP, uint8_t rf)
 
     case COMPARE_NULLEQ: return val1 == val2 && rf == 0;
 
+    case COMPARE_BITMASK:
+      // Bitmask comparison only valid for integral types
+      if constexpr (std::is_integral_v<T>)
```
drrtuy, line 219: This should be unnecessary TBH. We check for the types at the dispatcher layer choosing b/w integral and vector types.

drrtuy (approved):
untested ACK

PR #3911, opened 2026-01-26
MCOL-6204: new crashtrace [develop-23.02]

mariadb-LeonidFedorov:
Problem: The workernode process crashes when a signal (SIGINT, SIGTERM, or SIGHUP) arrives during startup, specifically while the `SlaveComm` object is being constructed.

Root Cause: In `ServiceWorkerNode::Child()` signal handlers are registered before the comm object is constructed:


drrtuy (approved):
untested ACK

PR #3913, opened 2026-01-29
fix(ASAN): MCOL-5604 - fix ASAN diagnostics

mariadb-SergeyZefirov:
This simple patch removes use of type handler of column operations which can be incorrect.


drrtuy (approved):
untested ACK

PR #3920, opened 2026-02-06
MCOL-6293 - table to list partitions

mariadb-SergeyZefirov:


mariadb-SergeyZefirov:
This patch introduces INFORMATION_SCHEMA.COLUMNSTORE_PARTITIONS table to query partitions of tables and columns. It has TABLE_SCHEMA, TABLE_NAME, COLUMN_NAME and PARTITION_ID columns, all textual. PARTITION_ID encoding is compatible with CALSHOWPARTITIONS and CALDROPPARTITIONS functions.

If two columns share partition, only one column will be listed in the result set, but search for partitions of specific column will return correct list of partitions for that column.

drrtuy (approved):
untested ACK

PR #3921, opened 2026-02-06
MCOL-5983: Fix bitmask comparison optimization when using user-defined variables as the mask value.

mariadb-LeonidFedorov:
Queries using user variables in bitmask comparisons returned incorrect results:

```sql
SET @mask = 0x15;
SELECT * FROM t WHERE (sig & @mask) = @mask;  -- Returns ALL rows instead of filtered
```

While the same query with a literal worked correctly:

```sql
SELECT * FROM t WHERE (sig & 0x15) = 0x15;  -- Works correctly
```

### Root Cause

MariaDB stores hex user variables (e.g., `SET @mask = 0x15`) as binary strings with `charset=binary`. When [val_str()](cci:1://file:///home/ubuntu/proj/MariaDBEnterprise/sql/item_func.h:4184:2-4201:3) is called, it returns raw binary bytes (`\x15`) instead of a text representation (`"21"`).

The [ConstantColumn](cci:1://file:///home/ubuntu/proj/MariaDBEnterprise/storage/columnstore/columnstore/dbcon/execplan/constantcolumn.cpp:124:0-136:1) constructor parses this string using `atoll()`, which returns 0 for binary input, causing the mask value to be incorrectly set to 0.

### Solution

In [buildConstantColumnMaybeNullUsingValStr()](cci:1://file:///home/ubuntu/proj/MariaDBEnterprise/storage/columnstore/columnstore/dbcon/mysql/ha_mcs_execplan_helpers.h:56:0-56:97), detect user variables with binary charset and interpret the string as a little-endian integer:

```cpp
if (str && str->length() > 0 && str->length() <= 8 &&
    str->charset() == &my_charset_bin)
{
  // Interpret binary string as little-endian integer
  uint64_t val = 0;
  const unsigned char* p = (const unsigned char*)str->ptr();
  for (size_t i = 0; i < str->length(); i++)
    val |= ((uint64_t)p[i]) << (i * 8);
  ...
}
```

Using [charset() == &my_charset_bin](cci:1://file:///home/ubuntu/proj/MariaDBEnterprise/sql/item_func.h:3949:2-3949:53) is a reliable way to detect binary data, as opposed to scanning for non-printable characters.

### Changes

- **[dbcon/mysql/ha_mcs_execplan_helpers.cpp](cci:7://file:///home/ubuntu/proj/MariaDBEnterprise/storage/columnstore/columnstore/dbcon/mysql/ha_mcs_execplan_helpers.cpp:0:0-0:0)**: Detect binary charset user variables and convert to integer
- **[dbcon/mysql/ha_mcs_execplan_walks.cpp](cci:7://file:///home/ubuntu/proj/MariaDBEnterprise/storage/columnstore/columnstore/dbcon/mysql/ha_mcs_execplan_walks.cpp:0:0-0:0)**: Use [buildConstantColumnMaybeNullUsingValStr()](cci:1://file:///home/ubuntu/proj/MariaDBEnterprise/storage/columnstore/columnstore/dbcon/mysql/ha_mcs_execplan_helpers.h:56:0-56:97) which handles binary user variables
- **[dbcon/joblist/jlf_execplantojoblist.cpp](cci:7://file:///home/ubuntu/proj/MariaDBEnterprise/storage/columnstore/columnstore/dbcon/joblist/jlf_execplantojoblist.cpp:0:0-0:0)**: Minor refactor — parse mask value before creating `pColStep`

### Testing

```sql
CREATE TABLE t (id INT, sig BIGINT UNSIGNED) ENGINE=columnstore;
INSERT INTO t VALUES (1, 0x3F), (2, 0x15), (3, 0x2A), (4, 0x07), (5, 0x00);

-- Both queries now return the same result: id=1,2
SELECT * FROM t WHERE (sig & 0x15) = 0x15;

SET @mask = 0x15;
SELECT * FROM t WHERE (sig & @mask) = @mask;

-- String variables still work normally
SET @str = 'hello';
SELECT @str;  -- Returns 'hello'
```

### DIFF
--- a/dbcon/joblist/jlf_execplantojoblist.cpp
+++ b/dbcon/joblist/jlf_execplantojoblist.cpp
@@ -1950,7 +1950,6 @@ const JobStepVector doSimpleFilter(SimpleFilter* sf, JobInfo& jobInfo)
 
         if (maskStr == rhsStr)
         {
-          CalpontSystemCatalog::OID tbl_oid = tableOid(sc, jobInfo.csc);
           CalpontSystemCatalog::ColType ct = sc->colType();
 
           if (!sc->schemaName().empty() && sc->isColumnStore())
@@ -1962,6 +1961,18 @@ const JobStepVector doSimpleFilter(SimpleFilter* sf, JobInfo& jobInfo)
           // Only for integer types
           if (datatypes::isInteger(ct.colDataType))
           {
+            int64_t maskValue = 0;
+            try
+            {
+              maskValue = static_cast<int64_t>(std::stoull(maskStr));
+            }
+            catch (...)
+            {
+              jsv = doExpressionFilter(sf, jobInfo);
+              return jsv;
+            }
+
+            CalpontSystemCatalog::OID tbl_oid = tableOid(sc, jobInfo.csc);
             string alias(extractTableAlias(sc));
             string view(sc->viewName());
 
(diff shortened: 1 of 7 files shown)


mariadb-SergeyZefirov (approved):
Looks good to me.

drrtuy:
Close as per discussion in Jira.

PR #3925, opened 2026-02-11
Set lower_case_table_names=2 for MTR

timofeyturenko:


build/run_regression.sh
mariadb-AlekseiBukhalov: i guess it should be removed here

mariadb-AlekseiBukhalov (changes requested):

build/run_mtr.sh
mariadb-AlekseiBukhalov: please move this higher, maybe right after source $(optparse.build), so that we sure that DISTRO is passed and set_cnf_path will work correctly

mariadb-AlekseiBukhalov (changes requested):

mariadb-AlekseiBukhalov (approved):

PR #3935, opened 2026-02-27
MCOL-4270 Fix DROP COLUMN IF EXISTS fails if the column is already present

jaeheonshim:
Add IF EXISTS support to ALTER TABLE DROP COLUMN by updating the drop_column_def structure in ddl.y to include the opt_if_exists clause. MariaDB's server already handles the case where the column doesn't exist before ColumnStore sees the statement, so ColumnStore just needs to parse and execute the drop normally.

Removed error checking for DROP COLUMN IF EXISTS in ha_mcs_ddl.cpp since the parser will no longer fail with this clause.

Modified tests to expect correct outcome for both drop column variants.


drrtuy (approved):
tested ACK

drrtuy:
JFYI CI failures are caused by the recent 11.4, 11.8 rebase.

PR #3952, opened 2026-03-19
Feat/perf io opts

drrtuy:
### Summary

Reduce lock contention and redundant disk reads in the IO subsystem.

### Changes

1. **Partitioned FD cache + thread pool for async loads** (`iomanager.cpp`, `primitiveserver.cpp`)
   - Shard `fdcache` by OID to eliminate contention between IO threads.
   - Narrow critical sections — lock only during cache lookup/insert.
   - Replace per-request `boost::thread` creation with a static `ThreadPool`; bump `asyncMax` 20→32.

2. **Lock-free FileBlockRequestQueue** (`fileblockrequestqueue.h/.cpp`)
   - Replace `std::deque` + `boost::mutex` + `boost::condition` with `moodycamel::BlockingConcurrentQueue` (lock-free MPMC).
   - Vendor `concurrentqueue.h`, `blockingconcurrentqueue.h`, `lightweightsemaphore.h` into `utils/common/`.

3. **Chunk readahead** (`iomanager.cpp`)
   - After decompressing a chunk, cache all remaining blocks (not just the requested ones) into block cache via `bulkInsert()`.
   - Avoids re-reading and re-decompressing the same chunk on sequential scans.

### Risks

- `size()` on the queue is now approximate.
- Readahead increases block cache memory pressure (beneficial for scans, neutral otherwise).

### DIFF
--- a/primitives/blockcache/fileblockrequestqueue.cpp
+++ b/primitives/blockcache/fileblockrequestqueue.cpp
@@ -23,72 +23,37 @@
  *                                                                         *
  ***************************************************************************/
 
-#include <boost/thread.hpp>
-#include <boost/thread/condition.hpp>
-
 #include "fileblockrequestqueue.h"
-
-using namespace std;
 
 namespace dbbc
 {
-fileBlockRequestQueue::fileBlockRequestQueue() : queueSize(0), readersWaiting(0)
+fileBlockRequestQueue::fileBlockRequestQueue()
 {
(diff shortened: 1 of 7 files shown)


mariadb-LeonidFedorov:
I would suggest to use CMake FetchContent macro for 
https://github.com/cameron314/concurrentqueue
unstead of direct copying here

PR #3956, opened 2026-03-24
MCOL-6325: fix uninitialized shared_ptr usage

mariadb-AlexeyAntipovsky:


dbcon/joblist/tuple-bps.cpp
```diff
@@ -1550,7 +1550,7 @@ void TupleBPS::join()
 
 void TupleBPS::sendError(uint16_t status)
 {
-  SBS msgBpp;
+  SBS msgBpp = boost::make_shared<ByteStream>();
```
mariadb-LeonidFedorov, line 1553: Wow!
Why not just `ByteStream stream;` BTW?

mariadb-AlexeyAntipovsky: `fDec->write` below uses SBS as argument

mariadb-LeonidFedorov (approved):

mariadb-LeonidFedorov:
untested ACK

PR #3959, opened 2026-03-27
fix(EM rebuild): MCOL-6302, MCOL-6321 - restore EM in case of header overflow, with test

mariadb-SergeyZefirov:
This patch adds logic to read block offsets of a dictionary column from tokens in a pseudo-column. These block offsets then are used to restore LBIDs in extent map and set proper HWM for a proper LBID.

An accompanying test tests rebuild process and integrity of the restored data.


mariadb-LeonidFedorov (approved):
I like the work produced. The code is nice and clean

PR #3960, opened 2026-03-30
fix(CI): Run future suite with innodb_queries_use_mcs=ON

mariadb-AlekseiBukhalov:
## Problem
CI runs the `future` MTR suite with `--extern`, so `suite.opt` options (including `--columnstore_innodb_queries_use_mcs=on`) are never applied to the running server. The server starts with default `columnstore_innodb_queries_use_mcs=OFF` from `columnstore.cnf`. Since this is a `PLUGIN_VAR_READONLY` variable, it can't be changed at runtime. Without it, InnoDB queries are never routed to CS's select handler.

All 7 `future/rbo_parallel_ces*` and `query_accelerator*` tests fail.

## Fix
Run the `future` suite separately with `innodb_queries_use_mcs=ON` written to CI server config before server restart (same pattern as `lower_case_table_names=2`).

## Companion change
Regression-test repo branch `fix-tests`: updates `bug5315.sql.ref.log` — `Rows matched: 0` → `Rows matched: 2` (correct behavior after MCOL-3529).


mariadb-AlekseiBukhalov:
will be merged as a part of https://github.com/mariadb-corporation/mariadb-columnstore-engine/pull/3970

PR #3966, opened 2026-04-07
Infer reports

mariadb-LeonidFedorov:


mariadb-SergeyZefirov (approved):
These changes are all trivial and on point.

PR #3967, opened 2026-04-08
MCOL-6331: Fix server crash on JSON_TABLE with ColumnStore subquery

mariadb-LeonidFedorov:
isMCSTable() crashed dereferencing NULL db_plugin for virtual tables
like JSON_TABLE that have no storage engine plugin assigned.

Additionally, the old logic incorrectly returned true (is ColumnStore)
when plugin info was unavailable. Now returns false — unknown tables
are not ColumnStore tables.

Added MTR regression test bugfixes.MCOL-6300-json-table-null-db-plugin-crash.


mariadb-SergeyZefirov (approved):
I understand changes and agree with them.

mariadb-SergeyZefirov (approved):
I completely understand what is going on and I agree with changes.

PR #3978, opened 2026-04-17
MCOL-6328: fix array out of bound in rbo_groupby_wrap_columns.cpp

vivarum:
Should fix an out of bound array indexing in rbo_groupby_wrap_columns.cpp when a window function has either a partition list of size > 1 or an order by list of size > 1.


mariadb-LeonidFedorov:
Thank you for the patch @vivarum !

mariadb-LeonidFedorov (approved):

mysql-test/columnstore/bugfixes/MCOL-6328-fix-array-out-of-bound-for-window-func-groupby-rbo.test
mariadb-AlekseiBukhalov, line 1: I would suggest to extend the test a little
 
 Add the exact minimal repro from the ticket:
  SELECT AVG(b), STD(b) OVER (PARTITION BY abs(b) ORDER BY b, a
                              ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING)
  FROM t ORDER BY 1, 2;

Fill the table with more rows to reliably trigger the crash pre-fix (INSERT INTO t SELECT seq, seq % 150 FROM seq_1_to_1500;)
Add one query that combines multi-column PARTITION BY with multi-column ORDER BY in the same window spec, thats the PARTITION -> ORDER BY transition in the changed code, currently not covered

mariadb-AlekseiBukhalov (changes requested):

mysql-test/columnstore/bugfixes/MCOL-6328-fix-array-out-of-bound-for-window-func-groupby-rbo.test
mariadb-LeonidFedorov: Sounds good! @vivarum could you please do that, as the PR comes from your fork, we cannot add the commits over your code easily

vivarum: Yes, I'll do that shortly.

vivarum: @mariadb-AlekseiBukhalov I added more test cases according to the suggestions above. Please, let me know if there are any comments!

mariadb-AlekseiBukhalov (approved):

PR #3986, opened 2026-04-20
MCOL-4258: feat(sp)  Support stored functions in Columnstore SELECT projection (connector-side evaluation)

mariadb-LeonidFedorov:
Implement support for MariaDB stored functions (CREATE FUNCTION ... RETURN) in SELECT queries against Columnstore tables without falling back to the MariaDB server's default execution path.

Previously, any query with a stored function (Item_func::FUNC_SP) against a Columnstore table was rejected with error MCS-1001 / ERR_SP_FUNCTION_NOT_SUPPORT (@bug4383). This patch removes that restriction for the SELECT projection case and evaluates stored functions on the connector side.

== Architecture / Approach ==

Stored functions (Item_func_sp) require the MariaDB server context (THD, SP framework, etc.) for evaluation. This context is not available inside the remote ExeMgr/PrimProc processes where Columnstore normally evaluates FunctionColumn expressions. Therefore, a "connector-side post-processing" approach is used:

1. PLAN BUILDING (ha_mcs_execplan.cpp, getSelectPlan):
   - When a FUNC_SP item is encountered in the SELECT list, instead of raising an error, a ConstantColumn(0) placeholder is pushed into CSEP::returnedCols at the corresponding position. This reserves a slot in the ExeMgr output row.
   - The Item_func_sp pointer and its output field index are recorded in gp_walk_info::storedFuncCols.
   - After getSelectPlan() completes, cs_get_select_plan() transfers this metadata into ha_columnstore_select_handler::stored_func_cols.

2. ROW DELIVERY (ha_mcs_pushdown.cpp, next_row):
   - After fetchNextRow() fills table->record[0] from ExeMgr data, the connector performs post-processing for each row:

     a) SOURCE FIELD POPULATION: iterates select_lex->item_list to find Item_field items. For each one, copies the value from the corresponding tmp table field (filled by ExeMgr) into the source table field (t1->record[0]) that the Item_field points to. This is necessary because Item_func_sp's arguments are Item_field objects bound to source table fields, which are not populated in select_handler mode. Uses dbug_tmp_use_all_columns() to allow writing to source table fields in debug builds.

     b) SP EVALUATION: for each stored function column, calls the appropriate Item_func_sp->val_*() method (val_int, val_real, val_decimal, val_str depending on result_type()), then writes the result into the output field in table->record[0], replacing the placeholder value.

3. DATA STRUCTURES:
   - StoredFuncColumnInfo: holds Item_func_sp* and output_field_idx
   - ha_columnstore_select_handler::stored_func_cols: vector of above
   - gp_walk_info::storedFuncCols: temporary storage during plan building

== What works (columnstore_select_handler=ON) ==

- SP in SELECT projection with source columns present
- Multiple SPs in one query
- Multi-argument SPs (columns from same or different tables)
- Nested SPs: sf(sf(col))
- SP with expression arguments: sf(col + 1), sf(col * 2 + 1)
- SP with constant arguments: sf(10)
- NULL propagation
- All return types: INT, BIGINT, VARCHAR, DECIMAL, DATE, BOOLEAN
- INNER JOIN, LEFT JOIN, self-join
- LIMIT
- INSERT INTO ... SELECT, CREATE TABLE AS SELECT
- WHERE filter on non-SP columns

== Known limitations ==

- SP in WHERE/ORDER BY/GROUP BY/HAVING: these clauses go through Columnstore's filter/sort/group expression evaluation in ExeMgr, which cannot call Item_func_sp. Falls back in AUTO mode.
- DISTINCT with SP: ExeMgr cannot deduplicate on SP placeholder values
- Arithmetic on SP result in projection (e.g. sf(a) + 1): the outer '+' operator is processed by buildFunctionColumn which tries to push it to ExeMgr.
- Aggregate on SP result (e.g. MIN(sf(a))): same issue.
- SP-only projection without source column in SELECT list: the source table field is not populated because ExeMgr doesn't return the column.

### DIFF
--- a/dbcon/mysql/ha_mcs_execplan.cpp
+++ b/dbcon/mysql/ha_mcs_execplan.cpp
@@ -6717,13 +6717,23 @@ int processSelect(SELECT_LEX& select_lex, gp_walk_info& gwi, SCSEP& csep, vector
       {
         Item_func* ifp = static_cast<Item_func*>(item);
 
-        // @bug4383. error out non-support stored function
+        // Stored functions are evaluated on the connector side after ExeMgr
+        // returns rows. Push a ConstantColumn placeholder so ExeMgr reserves a
+        // slot in the output row, and record the Item_func_sp* for post-processing.
         if (ifp->functype() == Item_func::FUNC_SP)
         {
-          gwi.fatalParseError = true;
-          gwi.parseErrorText = IDBErrorInfo::instance()->errorMsg(ERR_SP_FUNCTION_NOT_SUPPORT);
-          setError(gwi.thd, ER_CHECK_NOT_IMPLEMENTED, gwi.parseErrorText, gwi);
-          return ER_CHECK_NOT_IMPLEMENTED;
+          uint32_t outputIdx = gwi.returnedCols.size();
+          ConstantColumn* cc = new ConstantColumn((int64_t)0, ConstantColumn::NUM);
+          cc->timeZone(gwi.timeZone);
+          cc->resultType(colType_MysqlToIDB(ifp));
+          if (ifp->name.length)
+            cc->alias(ifp->name.str);
+          else if (!itemAlias.empty())
+            cc->alias(itemAlias);
+          SRCP srcp(cc);
+          gwi.returnedCols.push_back(srcp);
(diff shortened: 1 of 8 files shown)


mariadb-SergeyZefirov (approved):
I agree with changes, despite Gemini's criticisms.

PR #3989, opened 2026-04-20
fix(MCOL-6340): fix CrossEngine credentials for full MTR suite on 11.4+

mariadb-AlekseiBukhalov:
Part of [MCOL-6340](https://jira.mariadb.org/browse/MCOL-6340) — make the full columnstore MTR suite pass on 11.4+ under `--extern`.

## What broke

`run_mtr.sh` wrote `loose-columnstore_innodb_queries_use_mcs=ON` to a cnf drop-in globally (read-only startup variable, can't be scoped per-suite without a restart; `--extern` forbids restart). That routed plain InnoDB tests through `CrossEngineStep`, which then failed because `cejuser` setup was copy-pasted across tests and had drifted.

The global QA path also exposed several pre-existing Select Handler / CES bugs that don't fire when QA is suite-scoped — see _Pre-existing bugs surfaced_ below.

## Fix

- **`b5dfe7c24`** — Shared `cross_engine.inc` / `drop_cross_engine.inc` (idempotent CREATE/DROP USER for `localhost` + `127.0.0.1`); 24 tests refactored. Drop runner-level writes of `lower_case_table_names=2` and `innodb_queries_use_mcs=ON` from `run_mtr.sh` — both already applied per-suite via `suite.opt`.
- **`a7d465876`** — Disable `MCOL-6321-test-em-rebuild` (crashes PrimitiveServer on 4MB+ longtext, [MCOL-6302](https://jira.mariadb.org/browse/MCOL-6302)).
- **`9e96be104`** — Stabilize flaky `mcol-3785` (missing ORDER BY), `mcol-5669` (UNION ordering + 1-ULP DECIMAL drift through CES), `mcol-5963` (CES emits `HH:MM:SS.0` vs InnoDB `HH:MM:SS`).
- **`f0a6f0607`** — Lowercase ~77 mixed-case identifiers across 119 `.test`/`.result`. `suite.opt` `--lower_case_table_names=2` entries kept as defense-in-depth.

## Pre-existing bugs surfaced

Filed while triaging failures the global QA enable produced. All pre-existing in the Select Handler / CES code paths; reproducible only with `columnstore_innodb_queries_use_mcs=ON`. With QA now scoped to the `future` suite, the affected tests pass again in this PR — bugs remain open for follow-up.

- [MCOL-6343](https://jira.mariadb.org/browse/MCOL-6343) — Select handler returns wrong representative byte for `GROUP BY` / `DISTINCT` on `CHAR(1) latin1` (counts correct, byte from utf8 source instead of latin1 row).
- [MCOL-6344](https://jira.mariadb.org/browse/MCOL-6344) — ExeMgr disconnect on `IN (SELECT … WHERE x IS NOT NULL OR x IS NULL)` (regression of MCOL-4642 — `NOT IN` form was fixed, `IN` form still broken in `InSub::handleFunc`).
- [MCOL-6345](https://jira.mariadb.org/browse/MCOL-6345) — `MCS-1000 'not joined'` on self-join via derived table over a default-engine table (CEJ classifier fails to equate the base table use with its appearance inside the derived sub-query).
- [MCOL-6302](https://jira.mariadb.org/browse/MCOL-6302) — Restart of `mcsRebuildEM`-rebuilt MCS crashes server. Workaround in this PR: disable `MCOL-6321-test-em-rebuild`.

## Validation

CI [#3489](https://ci.columnstore.mariadb.net/mariadb-corporation/mariadb-columnstore-engine/3489/) — full MTR on 10.6/11.4/11.8.


drrtuy (approved):
untested ACK

PR #3991, opened 2026-04-22
MCOL-6336: fix DATE_SUB/DATE_ADD with INTERVAL NULL crashes with BatchPrimitiveProcessor error instead of returning NULL

AestheticAkhmad:
Fixes: [MCOL-6336](https://jira.mariadb.org/browse/MCOL-6336)
Added a null check before date add processing.

### DIFF
--- a/utils/funcexp/func_date_add.cpp
+++ b/utils/funcexp/func_date_add.cpp
@@ -822,8 +822,15 @@ int64_t Func_date_add::getIntVal(rowgroup::Row& row, FunctionParm& parm, bool& i
   else
     funcType = static_cast<OpType>(parm[3]->data()->getIntVal(row, isNull));
 
-  uint64_t value =
-      helpers::dateAdd(val, parm[1]->data()->getStrVal(row, isNull).safeString(""), unit, dateType, funcType);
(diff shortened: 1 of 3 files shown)


utils/funcexp/func_date_add.cpp
```diff
@@ -822,8 +822,15 @@ int64_t Func_date_add::getIntVal(rowgroup::Row& row, FunctionParm& parm, bool& i
   else
     funcType = static_cast<OpType>(parm[3]->data()->getIntVal(row, isNull));
 
-  uint64_t value =
-      helpers::dateAdd(val, parm[1]->data()->getStrVal(row, isNull).safeString(""), unit, dateType, funcType);
+  const auto expr = parm[1]->data()->getStrVal(row, isNull).safeString("");
+
+  if (isNull)
+  {
+    isNull = true;
```
drrtuy: Do you think you need this explicit assignment here?

AestheticAkhmad: I think yes, as I noticed such assignment in multiple places in Columnstore, so I assumed it is kind of convention to do that. But I can remove it.

Edit: I removed this and rebased to stable.

drrtuy (approved):
untestedACK

PR #3997, opened 2026-04-30
MCOL-6332 Fix ORDER BY using numeric positional column reference and DESC order on non-grouped columns in aggregation queries

vivarum:
This PR addresses [MCOL-6332](https://jira.mariadb.org/browse/MCOL-6332).

Fixes:
- Fix error when using a reference id in an ORDER BY section and specifying a DESC sort order for a non-agg function column that is not under a group by list


mariadb-AlexeyAntipovsky (approved):

mariadb-SergeyZefirov (approved):
Simple fix, with test. What not to like?

PR #3999, opened 2026-05-05
MCOL-4627 fix SET int_col=decimal_col returns an error

AestheticAkhmad:
This issue [MCOL-4627](https://jira.mariadb.org/browse/MCOL-4627) was resolved by solving [MCOL-4628](https://jira.mariadb.org/browse/MCOL-4628). Thus, only testcases were added.

Edit: Gemini found an inconsistency bug, where Columnstore was rounding differently from InnoDB. Now it is fixed, and the testcases are adjusted.


mariadb-SergeyZefirov (approved):
I agree with changes.

mysql-test/columnstore/bugfixes/mcol_4627.test
```diff
@@ -0,0 +1,58 @@
...
+CREATE DATABASE mcol_4627;
+USE mcol_4627;
+
+--disable_warnings
+DROP TABLE IF EXISTS t1;
+--enable_warnings
+CREATE TABLE t1 (i INT, d DECIMAL(10,1)) ENGINE=ColumnStore;
+INSERT INTO t1 VALUES (0, 1.5);
+UPDATE t1 SET i=d;
+SELECT * FROM t1;
```
mariadb-AlexeyAntipovsky, line 15: I would suggest adding the "--sorted_result" flag before the select statements to avoid depending on the row order.

mariadb-AlexeyAntipovsky (approved):

PR #4001, opened 2026-05-07
ci(drone): ignore mtr/regression failures for ASan/UBSan nightly

mariadb-AlekseiBukhalov:
## Summary
- Sanitizer pipelines (ASan, UBSan) on `ubuntu:24.04` are currently non-functional and frequently fail at runtime.
- Their failures bubble up to the nightly Slack notification, making the whole nightly look red even when everything else is green.
- Reuse the existing `ignoreFailureStepList` mechanism (same `["regression", "mtr"]` pattern already used by the `libcpp` pipeline) so failures in mtr / regression tests of sanitizer builds no longer fail the pipeline.

## Test plan
- [ ] `jsonnet .drone.jsonnet` parses cleanly (verified locally).
- [ ] Confirm in next nightly that ASan/UBSan pipelines report success even when mtr/regression steps fail, while non-sanitizer pipelines still report failures normally.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -715,8 +715,9 @@ local AllPipelines =
     for triggeringEvent in events
     for server in servers[current_branch]
   ] +
+  // sanitizers are non-functional; ignore mtr/regression failures so nightly stays green
   [
-    Pipeline(b, platform, triggeringEvent, a, server, flag, "")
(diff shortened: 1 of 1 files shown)


mariadb-SergeyZefirov (approved):
I agree.

PR #4002, opened 2026-05-08
feat(mtr): MCOL-5430 Add a test for using window functions with GROUP BY

mariadb-AlexeyAntipovsky:


mariadb-SergeyZefirov (approved):
I agree.

PR #4006, opened 2026-05-12
fix(json_append): Fix JSON_APPEND crash under 12.3 CS

mariadb-SergeyZefirov:
This patch prevents redefinition of isleap macro to allow building under 12.3 CS and fixes heap buffer overrun in the JSON_APPEND function implementation.


mariadb-AlexeyAntipovsky (approved):

drrtuy (approved):
untested ACK

PR #4013, opened 2026-05-20
MCOL-4601 Fix UB for mod div overvlow case in decimals

mariadb-VasilyKozhukhovskiy:
This is a follow up fix for previous changes.

I didn't handle the case of overflow, since I didn't want to throw any errors or write warnings for now. But in some cases it can be a UB (which UBSan shows with overflow tests enabled).

So, this should fix UB by handling overflow and returning NULL for such cases.

It also adds returning NULL when the result value (i.e. int128) cannot fit in result type - int64/uint64, which is not UB but leads to a truncation of the result.


mariadb-SergeyZefirov (approved):
I agree.

mariadb-AlexeyAntipovsky:
Is it ok that the results differ from InnoDB? For example:
```
CREATE TABLE t8i (a DECIMAL(38,0)) engine=innodb;
SELECT a, a MOD 1.11 as r FROM t8i;
+----------------------------------------+------+
| a                                      | r    |
+----------------------------------------+------+
| 99999999999999999999999999999999999999 | 0.21 |
| 12345678901234567899999999999999999999 | 0.75 |
+----------------------------------------+------+
```

mariadb-VasilyKozhukhovskiy:
> Is it ok that the results differ from InnoDB? For example:
> 
> ```
> CREATE TABLE t8i (a DECIMAL(38,0)) engine=innodb;
> SELECT a, a MOD 1.11 as r FROM t8i;
> +----------------------------------------+------+
> | a                                      | r    |
> +----------------------------------------+------+
> | 99999999999999999999999999999999999999 | 0.21 |
> | 12345678901234567899999999999999999999 | 0.75 |
> +----------------------------------------+------+
> ```

ColumnStore uses int128 for Decimal type with max precision supported (which is 38 - wide Decimal). Int128 is not enough to fit the scaled value: 99999999999999999999999999999999999999*100 to perform the operation. So it's the case where the int128 overflow is handled and NULL is returned.

As I understand, InnoDB uses some custom arrays inside its implementation for Decimal type, so it can operate on Decimals with greater precision (>38).

mariadb-AlexeyAntipovsky (approved):

PR #4014, opened 2026-05-20
MCOL-4601 Fix exception in func_div.cpp when one operand is double and the other one is decimal

mariadb-VasilyKozhukhovskiy:
Fixes failed nightly build.


mariadb-SergeyZefirov (approved):
I agree with changes.

mariadb-AlexeyAntipovsky (approved):

PR #4017, opened 2026-05-25
fix(JSON): MCOL-6405 - JSON_SET/JSON_REPLACE work as in server

mariadb-SergeyZefirov:
This fix prevents appending command on empty intermediate JSON result and this simple fix makes Columnstore's behavior identical to the server's.

clean code; correct result


mariadb-VasilyKozhukhovskiy (approved):
Looks good to me.

PR #4022, opened 2026-06-01
MCOL-6221 Sec to time rounds values

mariadb-VasilyKozhukhovskiy:


utils/funcexp/func_sec_to_time.cpp
```diff
@@ -72,32 +132,20 @@ string Func_sec_to_time::getStrVal(rowgroup::Row& row, FunctionParm& parm, bool&
 
     case execplan::CalpontSystemCatalog::DOUBLE:
     case execplan::CalpontSystemCatalog::UDOUBLE:
-    {
-      datatypes::TDouble d(parm[0]->data()->getDoubleVal(row, isNull));
-      val = d.toMCSSInt64Round();
-      break;
-    }
+      return secToTimeWithUsec(parm[0]->data()->getDoubleVal(row, isNull));
```
mariadb-SergeyZefirov: Please note that you are changing call to general ``datatypes`` functionality to yours.

The ``toMCSSInt64Round`` function you replace here is also used in func_period_diff.cpp.

So, would it be better to have your new function moved to datatypes/ source code? This way it will be accessible to other parts of our code more easily, if the need come.

utils/funcexp/func_sec_to_time.cpp
```diff
@@ -41,6 +43,64 @@ using namespace execplan;
 using namespace logging;
 namespace funcexp
 {
+namespace
```
mariadb-SergeyZefirov: In my opinion, this code belongs to datatypes/ part of our source tree. The justification is an accessibility of the functionality.

mariadb-SergeyZefirov (approved):
I generally approve the changes.

It would be nice, but not strictly necessary, for new functionality to be in datatypes/ source code somewhere. For example, next to ``toMCSSInt64Round`` template function.

``Func_period_diff`` uses the same logic as in what you replaced here. If you think that moving your function to datatypes/ is too much, please, add some comments there in Func_period_diff about possible improvement path. Again, not strictly necessary, just a nice thing to do. ;)

utils/funcexp/func_sec_to_time.cpp
mariadb-VasilyKozhukhovskiy: Thanks for the good advice. Moved constants and helper functions to datatypes/mcs_decimal.h and to a new header datatypes/mcs_time.h where it seemed suitable.

PR #4023, opened 2026-06-05
fix(BRM): MCOL-6404 fix race cond while setting DBRM read-only

mariadb-AlexeyAntipovsky:
Improve ColumnStore read-only transition and cleanup handling

- Coordinate read-only startup with both active cpimport jobs and transactions
- Use a shared jobs mutex/condition variable for cpimport and transaction state changes
- Reject new cpimport jobs and transaction IDs while waiting to enter read-only mode
- Notify read-only waiters when cpimport jobs finish or transaction count reaches zero

### DIFF
--- a/versioning/BRM/masterdbrmnode.cpp
+++ b/versioning/BRM/masterdbrmnode.cpp
@@ -60,7 +60,7 @@
 #define THREAD_EXIT                                                                                    \
   {                                                                                                    \
     mutex.lock();                                                                                      \
-    for (vector<IOSocket*>::iterator _it = activeSessions.begin(); _it != activeSessions.end(); ++_it) \
+    for (auto _it = activeSessions.begin(); _it != activeSessions.end(); ++_it)                        \
       if (p->sock == *_it)                                                                             \
       {                                                                                                \
         activeSessions.erase(_it);                                                                     \
@@ -86,7 +86,7 @@ MasterDBRMNode::MasterDBRMNode()
 
   config = config::Config::makeConfig();
 
-  if (config == NULL)
+  if (config == nullptr)
     throw invalid_argument("MasterDBRMNode: Configuration error.");
 
   runners = 0;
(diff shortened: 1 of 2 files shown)


versioning/BRM/masterdbrmnode.cpp
```diff
@@ -1169,7 +1168,7 @@ void MasterDBRMNode::doForceClearAllCpimportJobs(messageqcpp::IOSocket* sock)
   std::cout << "doForceClearCpimprtJobs" << std::endl;
 #endif
   ByteStream reply;
-  std::unique_lock lk(cpimportMutex);
+  std::unique_lock lk(jobsMutex);
   sm.clearAllCpimportJobs();
```
mariadb-VasilyKozhukhovskiy, line 1172: Should not call to jobsCond.notify_one() be added here after clearing all jobs?

versioning/BRM/masterdbrmnode.cpp
```diff
@@ -1521,64 +1492,66 @@ void MasterDBRMNode::doFinishCpimportJob(ByteStream& msg, ThreadParams* p)
...
-    msg >> block;
-    msg >> isDDL;
-    txnid = sm.newTxnID(sessionID, (block != 0), (isDDL != 0));
-    reply << (uint8_t)ERR_OK;
-    reply << (uint32_t)txnid.id;
-    reply << (uint8_t)txnid.valid;
-#ifdef BRM_VERBOSE
-    cerr << "newTxnID returning id=" << txnid.id << " valid=" << txnid.valid << endl;
-#endif
+    reply << (uint8_t)ERR_READONLY;
```
mariadb-VasilyKozhukhovskiy, line 1523: Maybe need to keep a 6-byte response here, smth like:
reply << (uint8_t)ERR_READONLY << (uint32_t)0 << (uint8_t)0

To avoid getting "bad response" errors in logs here: https://github.com/mariadb-corporation/mariadb-columnstore-engine/blob/fix/MCOL-6404_startreadonly-race/versioning/BRM/dbrm.cpp#L3117? Or it doesn't matter much?

versioning/BRM/masterdbrmnode.cpp
mariadb-AlexeyAntipovsky: Good point. Done.

versioning/BRM/masterdbrmnode.cpp
mariadb-AlexeyAntipovsky: Actually it will be a [network error](https://github.com/mariadb-corporation/mariadb-columnstore-engine/blob/fix/MCOL-6404_startreadonly-race/versioning/BRM/dbrm.cpp#L3108) — which is the same as in `DBRM::newCpimportJob()` so I think it's ok.

mariadb-VasilyKozhukhovskiy (approved):

PR #4028, opened 2026-06-12
MCOL-6221 Fix stack-use-after-scope in TreeNode::getIntVal/getUintVal

mariadb-VasilyKozhukhovskiy:
Converter parsed a dangling pointer into the dead safeString() temporary. Caught by ASan in PrimProc via SEC_TO_TIME(varchar) in the MCOL-6221 mtr tests.


mariadb-SergeyZefirov (approved):
I agree.

PR #4031, opened 2026-06-16
MCOL-6432 Fix ASan-detected memory leaks on the ColumnStore DDL path

mariadb-VasilyKozhukhovskiy:


mariadb-AlexeyAntipovsky (approved):
lgtm

PR #4032, opened 2026-06-17
MCOL-6416 Refactoring and fixes for ci pipelines

mariadb-VasilyKozhukhovskiy:
- Split each sanitizer stage (heavy, nightly only) into two parallel stages to make them fit in their time quotas
- Fix failing regression test test500.sh on Debian 13 and Ubuntu 26.04 (see PR in https://github.com/mariadb-corporation/mariadb-columnstore-regression-test repo)
- Add more granular explicit timeouts per stage step
- Remove from ignored tests all tests except: test400.sh (regression test, fails sporadically), cmapi, upgrade

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -764,9 +764,9 @@ local AllPipelines =
     for triggeringEvent in events
     for server in servers[current_branch]
   ] +
-  // last argument is to ignore mtr and regression failures
+  // last argument (ignoreFailureStepList) ignores only test400
   [
-    Pipeline(b, platform, triggeringEvent, a, server, "", "", ["regression"])
+    Pipeline(b, platform, triggeringEvent, a, server, "", "", ["test400.sh"])
     for a in ["amd64"]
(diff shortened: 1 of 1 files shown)


.drone.jsonnet
```diff
@@ -168,7 +188,7 @@ local Pipeline(branch, platform, event, arch="amd64", server="10.6-enterprise",
     ],
   },
 
-  local regression_tests_base = if (event == "cron") then [
+  local regression_tests_base = if (event != "cron") then [
```
mariadb-SergeyZefirov: That seems excessive.

mariadb-SergeyZefirov: I see! This is how you test, I believe.

mariadb-SergeyZefirov (approved):
It is a good job. I commented on change in the logic that should not go into the main branch, otherwise I agree with changes.

mariadb-AlanMologorsky (approved):
LGTM

PR #4034, opened 2026-06-26
MCOL-6338: make rebuildEM better

mariadb-SergeyZefirov:
This particular patch introduces SAT-based backtracking search into LBID ranges restoration process (look for SATProblem in tools/rebuildEM/rebuildEM.h).

It uses Picosat as a SAT solver. Picosat is introduced into our source code as two source files (picosat.c and picosat,h).

The CNF for SAT is constructed from all guesses for all possible ranges that dictionary data record could occupy. Then SAT problem is solved and we restore LBID ranges from satisfying solution. Picosat solves such problems pretty quickly.

There is a reproduction (tools/rebuildEM/repro) that should be run manually for time being.


mariadb-AlexeyAntipovsky (approved):
Looks good. 👍  I think after a quick cleanup (indentation in a couple of places, and it makes sense to make the detailed logs wrapped with the `doVerbose()` flag), it's ready for testing.

PR #4039, opened 2026-07-03
fix(aggregation): MCOL-6441 Proper memory usage accounting

mariadb-AlexeyAntipovsky:


mariadb-VasilyKozhukhovskiy (approved):
Looks good to me.

PR #4041, opened 2026-07-03
fix(BRM): MCOL-6404 fix race cond while setting DBRM read-only (cherry-pick to 23.02)

mariadb-AlexeyAntipovsky:
Improve ColumnStore read-only transition and cleanup handling

- Coordinate read-only startup with both active cpimport jobs and transactions
- Use a shared jobs mutex/condition variable for cpimport and transaction state changes
- Reject new cpimport jobs and transaction IDs while waiting to enter read-only mode
- Notify read-only waiters when cpimport jobs finish or transaction count reaches zero


mariadb-VasilyKozhukhovskiy (approved):
I remember this fix! LGTM.

PR #4044, opened 2026-07-06
Remove extra 'yum update' in run_cmapi_test.sh that leads to sporadic failures in tests

mariadb-VasilyKozhukhovskiy:
The error occurs sometimes in container preparation step when installing colstore cmapi package in yum-based distributives.

"yum update" here seems to have been a mistranslation of the deb branch's apt-get update — the catalog refresh it was meant to mirror is something dnf does automatically on install, the full system upgrade it actually performs already runs in prepare_test_container.sh with the --nobest safety flag, and repeating it without that flag is what crashes on rockylinux:9 whenever the base image lags the mirror's glibc.

Example of a failed step caused by this problem:
https://ci.columnstore.mariadb.net/mariadb-corporation/mariadb-columnstore-engine/3952/3/14
https://ci.columnstore.mariadb.net/mariadb-corporation/mariadb-columnstore-engine/3955/13/13

### DIFF
--- a/build/run_cmapi_test.sh
+++ b/build/run_cmapi_test.sh
@@ -28,7 +28,18 @@ prepare_environment() {
   if [[ "$PKG_FORMAT" == "deb" ]]; then
     execInnerDocker $CONTAINER_NAME "apt-get clean && apt-get update -y && apt-get install -y mariadb-columnstore-cmapi"
   else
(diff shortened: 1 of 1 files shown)


timofeyturenko (approved):

mariadb-AlanMologorsky (approved):
LGTM

PR #4056, opened 2026-07-15
fix(BRM): MCOL-6434 - BRM access race fix

mariadb-SergeyZefirov:
In heavy cpimport/query processing loads the access races were observed. This particular patch attempts to prevent that by using much more tested upgradable mutext implementation from boost::interprocess.


mariadb-VasilyKozhukhovskiy (approved):
Looks good to me.

PR #4058, opened 2026-07-21
Mcol 6405 fix json functions on big data

mariadb-SergeyZefirov:


utils/funcexp/func_json_array_append.cpp
```diff
@@ -34,30 +34,31 @@ std::string Func_json_array_append::getStrVal(rowgroup::Row& row, FunctionParm&
...
   utils::NullString tmpJS(js);
+  Func_json_multipath_state state(fp, 1, 2);;
   for (size_t i = 1, j = 0; i < fp.size(); i += 2, j++)
   {
     const char* rawJS = tmpJS.str();
     const size_t jsLen = tmpJS.length();
-    JSONPath& path = paths[j];
+    JSONPath& path = state.paths[j];
 
     if (!path.parsed && parseJSPath(path, row, fp[i], false))
```
mariadb-VasilyKozhukhovskiy, line 44: Looks like these flags (i.e. path.parsed / isModeParsed / arg2Parsed) will be always returning false now when the state is a local variable?

Maybe they should be removed now and some comment be added about adding caching for such things where possible in future (for better performance on big load)?

mariadb-VasilyKozhukhovskiy (commented):
Looks like CI is red and not all steps have passed (perhaps need to just re-run?).

utils/funcexp/func_json_array_append.cpp
mariadb-SergeyZefirov: These flags are parsed inside get___Val function.

mariadb-SergeyZefirov:
> Looks like CI is red and not all steps have passed (perhaps need to just re-run?).

This is somewhat intentional.

Enterprise Server 12.3 does not build with Debian derivatives, that is why many builds are red. It also breaks regression tests, Stas allowed for that, we need 12.3 in our CI so that we will not miss anything, including, but not limited, 12.3 successful build under Debian. ;)

utils/funcexp/func_json_array_append.cpp
mariadb-VasilyKozhukhovskiy: I mean that now the state is a local variable and it's created anew on each call, so looks like not much use of using these flags now (they do not carry across calls), as I understand.

That's my only pick, but probably you know better:) Apart from that LGTM.

mariadb-VasilyKozhukhovskiy:
> > Looks like CI is red and not all steps have passed (perhaps need to just re-run?).
> 
> This is somewhat intentional.
> 
> Enterprise Server 12.3 does not build with Debian derivatives, that is why many builds are red. It also breaks regression tests, Stas allowed for that, we need 12.3 in our CI so that we will not miss anything, including, but not limited, 12.3 successful build under Debian. ;)

I didn't understand that, sorry. I just see that the build was killed and didn't finish. Now I know the reason.

mariadb-VasilyKozhukhovskiy (approved):

PR #4059, opened 2026-07-21
Remove hard deps on the certain server version

timofeyturenko:


timofeyturenko:
testing showed that Columnstore works only with server that was compiled together with it

PR #4072, opened 2026-07-27
MCOL-6492: Disable pagers for systemd service probes

RazvanLiviuVarzaru:
Package scripts use systemctl cat to check whether MariaDB and ColumnStore service units exist. systemctl can start less for these queries, making the result depend on TERM and pager behavior. If the pager fails, systemctl can return a nonzero status and make the scripts select a non-systemd path.

Pass --no-pager to every service probe. This option is supported by the systemd versions used by RHEL 7 and newer.


drrtuy (approved):
untested ACK

timofeyturenko (approved):

PR #4079, opened 2026-07-31
fix(PrimProc): MCOL-5283 Add data conversion for string data types

mariadb-AlexeyAntipovsky:
Make the behavior identical to InnoDB:
* Support for `CHAR()`, `VARCHAR` and `BLOB` columns in math functions and arithmetic
* Support above data types along with DATE/TIME/DATETIME/TIMESTAMP for aggregation functions SUM and AVG.


utils/rowgroup/rowaggregation.cpp
```diff
@@ -1449,13 +1450,36 @@ void RowAggregation::doSum(const Row& rowIn, int64_t colIn, int64_t colOut, int
...
+      wideValue = strtoll128(str.c_str(), sat, nullptr);
+      break;
+    }
+
+    case execplan::CalpontSystemCatalog::TIMESTAMP:
+    {
+      auto str = DataConvert::timestampToString1(rowIn.getUintField(colIn), fTimeZone);
+      isWideDataType = true;
+      bool sat = false;
+      wideValue = strtoll128(str.c_str(), sat, nullptr);
```
mariadb-SergeyZefirov, line 1477: There are two cases of copy-pasted code. Factoring them out would be nice.

Also it would be nice to have test for aggregation ``SUM(char_column)`` with invalid double values. Like "wot a number".  refactored code may help in that case. ;)

But it is not mandatory, just trying to glue what may fall here.

mariadb-SergeyZefirov (approved):
Most of code changes here is to support what previously was not supported. I completely agree with that. Thus, I approve.

I left a comment about missing corner case test, it was slightly premature, as we test with "not a number" string in the test added.  It would be nice to have some indication that results are in line with InnoDB or MyISAM. A comment would be OK.

utils/rowgroup/rowaggregation.cpp
mariadb-AlexeyAntipovsky: :+1: 

> Also it would be nice to have test for aggregation SUM(char_column) with invalid double values

The last 4 queries in the test do exactly that: sum and avg across all columns, including non-numeric data (one query each for ColumnStore and InnoDB, so we'll see if the server behavior changes)

PR #4087, opened 2026-08-13
fix(funcexp): MCOL-6475 fix SYSDATE return value

mariadb-AlexeyAntipovsky:


mariadb-SergeyZefirov (approved):
I agree with changes.

PR #4094, opened 2026-08-21
Feature and test

mariadb-SergeyZefirov:


dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -2326,6 +2326,11 @@ uint32_t setAggOp(AggregateColumn* ac, Item_sum* isp)
   Item_sum::Sumfunctype agg_type = isp->sum_func();
   uint32_t rc = 0;
 
+#if MYSQL_VERSION_ID >= 130100
+  if (isp->filter_expr)
+    return ER_CHECK_NOT_IMPLEMENTED;
+#endif
```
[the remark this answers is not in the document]
mariadb-SergeyZefirov: This is blatantly false, there is no such MDEV-11924.  Doing suggested change would result in build failure.

dbcon/mysql/ha_mcs_execplan.cpp
```diff
@@ -2326,6 +2326,11 @@ uint32_t setAggOp(AggregateColumn* ac, Item_sum* isp)
   Item_sum::Sumfunctype agg_type = isp->sum_func();
   uint32_t rc = 0;
 
+#if MYSQL_VERSION_ID >= 130100
+  if (isp->filter_expr)
```
mariadb-VasilyKozhukhovskiy, line 2330: It doesn't compile for me, filter_expr is a private member in the server. Probably has_filter() member function should be used instead.

Note, I had to add -DWITH_UNITTESTS=NO option to avoid build errors for tests.

mysql-test/columnstore/basic/r/MCOL-6534-aggregate-filter.result
```diff
@@ -0,0 +1,6 @@
+DROP DATABASE IF EXISTS MCOL_6534;
```
mariadb-VasilyKozhukhovskiy, line 1: I cannot find the .test file for this .result file, does it exist?

mariadb-VasilyKozhukhovskiy (changes requested):
Some more notes in addition to the two inline comments: 
1. It looks like the check is not done for window functions. Instead of the error, columnstore returns incorrect result:
```
CREATE TABLE t(x integer) ENGINE=COLUMNSTORE;
INSERT INTO t VALUES (1),(2),(3),(4),(5);
SELECT x, SUM(x) FILTER (WHERE x > 3) OVER () AS s FROM t ORDER BY x;
```
2. The MTR test will fail on server versions prior to 13.1 that do not support FILTER clause. Claude recommends to add a new file "disable_before_13.1.inc":
```
if (`SELECT (sys.version_major(), sys.version_minor()) < (13, 1)`)
{
    skip FILTER clause requires MariaDB 13.1+;
} 
```
and --source it from the test.
3. The current error text looks a bit not too informative:
`The storage engine for the table doesn't support Non supported aggregate type on the select clause`
Maybe it would be better to print something like this?
`FILTER clause of aggregate functions is not supported by Columnstore`

PR #4101, opened 2026-08-28
fix(json): MCOL-6563 - fix crashes with CS 13.1

mariadb-SergeyZefirov:
This oneliner introduces enough initialization of JSON engine so that crashes do not happen.


mariadb-VasilyKozhukhovskiy (approved):
Looks good to me.

timofeyturenko (approved):

PR #4112, opened 2026-09-16
ci(woodpecker): port .drone.jsonnet to .woodpecker.jsonnet with a PR-only matrix

mariadb-LeonidFedorov:
Woodpecker 3.x port of the Drone pipeline, generated from `.woodpecker.jsonnet` by `scripts/ci/generate_woodpecker.sh` — the same generator orthus uses (jsonnet → one `.woodpecker/<name>.yml` per workflow, `--check` staleness gate wired in as the `check-jsonnet-ymls-sync` step, `woodpecker-cli lint`).

Draft: opened to see how the build actually behaves on Woodpecker.

## What changes

`.drone.jsonnet` expands every matrix entry for **both** events (`events = ["pull_request", "cron"]`), so a pull request pays for the full nightly matrix. Here the matrix is split by trigger:

| | contents |
|---|---|
| `prMatrix` | `ubuntu:26.04` + `rockylinux:9` on `11.8-enterprise`, test groups `smoke` + `cmapi` + `mtr`. This is everything a PR runs. |
| `nightlyMatrix` | **empty by design** — nightly is not generated yet. |

The nightly machinery is already in place: `event: 'cron'` selects the long regression list and the `latest` publish, `buildKey`/`envKey` give the sanitizer and compiler variants (ASan/UBSan/libc++/gcc-toolset/clang-20), `testGroups` gives `upgrade` and `regression`. Enabling nightly = listing entries in `nightlyMatrix` and regenerating.

Generated per-PR workflow, 18 steps:

```
submodules → check-jsonnet-ymls-sync → clone-mdb → build ∥ cmapi-build
  → createrepo → pkg → publish-cmapi-build, publish-pkg
  → smoke / cmapi-test / mtr  (each + its log step + publish-<log>)
```

`build/createrepo.sh` and `build/report_test_stage.sh` hardcoded the Drone workspace mount; they now resolve `${CI_WORKSPACE:-/drone/src}` and work under both CIs. `.drone.jsonnet` is untouched — the two systems can run side by side.

## Not ported yet

Needed before `nightlyMatrix` is filled in: the `dockerfile → dockerhub → multi-node MTR` chain (rockylinux:8 + gcc-toolset only), the Infer/SAST pipeline, and the Slack notification pipeline that fans in from every nightly workflow. All three are flagged as TODO in `.woodpecker.jsonnet`.

## Testing

- `scripts/ci/generate_woodpecker.sh` generates and lints both workflows clean; `--check` reports them in sync.
- Dependency graph of every generated workflow has no dangling or forward `depends_on`.
- A filled-in nightly matrix compiles cleanly out-of-tree (12 pipelines, no dangling deps) — the `upgrade` and `regression` step groups are ported but **not exercised** by any generated workflow yet.


mariadb-VasilyKozhukhovskiy (approved):
Looks good to me as a starting point. This change would bring back regularly running checks for each submitted PR and then a more complete test matrix could be prepared to bring up the full nightly runs using the new CI.

mariadb-AlexeyAntipovsky (approved):
Looks great. Thanks, Leonid!

PR #4114, opened 2026-09-22
feat(cmapi, ci, cbom): MCOL-6570 generate sbom for cmapi.

mariadb-AlanMologorsky:


mariadb-StanislavUsov (approved):
Approved with a condition to verify the output completeness against actual dependencies

PR #4116, opened 2026-09-23
MCOL-6496 — Versioned copy-on-write shared memory for the BRM

mariadb-AlexeyAntipovsky:
Make BRM (ExtentMap/CopyLocks/VSS/VBBM) versioned:
* writer copies the published area, changes the copy, and publishes it
* writers synced by a simple mutex placed within shared memory
* readers take no lock at all
* remove Undoable class - it isn't needed anymore, all changes reside in a separated shmem segment, that can be simply dropped
* MasterSegmentTable is useless either, as there are no rwlocks now, and we have separate metadata for each entity


mariadb-SergeyZefirov (approved):
I read the code for stable-23.10 amd approved it.

